PR #1, opened 2024-05-01
Add `op-node-derivers` design doc

protolambda:
**Description**

Design doc for refactor of the Driver/Derivation in op-node, and introduction of "Deriver" to make the core chain validation processes of the rollup-node more extensible. This supports new projects such as interop by providing cleaner abstractions for major new functionality.

### DIFF
--- a/op-node-derivers.md
+++ b/op-node-derivers.md
@@ -517,6 +517,17 @@ The derivation is somewhat statefully connected to the engine however:
 We need a first step that addresses this, so the derivation process that generates the block attributes
 is shareable with the new driver design.
 
+In summary, the changes would look like:
+- Remove the Engine-queue stage.
+- As derivation pipeline, return a "tentative" block-attributes (not the safe-block yet),
+  along with information whether or not this is the last entry of a span-batch.
+- Modify the derivation pipeline such that the success-status of the last attributes is passed on:
+  if failed, we proceed to pull more attributes on the same block-height.
+- Remember engine sync-status in the EngineController.
+- Handle consolidation vs. force-processing of attribute as a separate object, called with the attributes from the pipeline.
+- Move safe-blocks memory and finalization as a separate object, called after applying attributes to the engine.
+- Handle queued unsafe blocks (the payloads queue) as a separate object, called before the derivation is used.
+
 #### 2) Incremental `OnEvent` adoption
 
 We can incrementally change the Driver state-loop sub-processes to fit the `OnEvent` signature,
--- a/op-node-derivers.md
+++ b/op-node-derivers.md
@@ -1,0 +1,554 @@
+# `op-node-derivers`
+
+#### Metadata
+
+Authors: @protolambda.
+Created: May 1, 2024.
+
+# Purpose
+
+<!-- *This section is also sometimes called “Motivations” or “Goals”.*
+ *It is fine to remove this section from the final document,
+  but understanding the purpose of the doc when writing is very helpful.* -->
+
+The core component of native cross-L2 interop is
+[message dependency-validation](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/messaging.md).
+This is a new core chain-validation process, adjacent to derivation,
+that the rollup node implementation will have to support.
+
+The purpose of this design doc is to support the integration of new rollup-node processes like this,
+by improving the abstractions, encapsulating complexity, and decoupling chain-state from derivation-state better.
+
+# Summary
+
+<!-- *Most (if not all) documents should have a summary.
+While the length will likely be proportional to the length of the full document,
+the summary should be as succinct as possible.* -->
+
+To support "cross-unsafe" block validation of Interop:
+- Slim down and simplify the derivation, by decoupling the `EngineQueue`.
+- Encapsulate the driver sub-processes, by splitting the state-loop into smaller composable "derivers".
+
+
+# Problem Statement + Context
+
+<!-- *Describe the specific problem that the document is seeking to address as well as information
+needed to understand the problem and design space.
+If more information is needed on the costs of the problem, this is a good place to that information.* -->
+
+The op-node splits the "when" (driver) and the "how" (derivation).
+This separation of scheduling and processing is key for fault-proving,
+where there is no "when", and where the "how" needs to be synchronous.
+
+The problem is that neither driver or derivation is extensible,
+and with every feature that is forced into it, the total becomes more fragile and complicated.
+
+## Tech-debt
+
+The "tech-debt" topic applies to the op-node, not directly to the rollup-node spec.
+However, this is an area that is under-specified,
+and as de-facto reference implementation the op-node should not be encumbered with tech-debt that makes it unreadable.
+
+Things like the unsafe-block processing and sequencing work were fitted into safe-block processing code-path:
+1) This overloads the responsibilities of the derivation-pipeline, in particular the "engine-queue".
+2) This overloads the responsibilities of the driver state-loop, monolithically and synchronously executing everything.
+
+This is a remnant of the first Bedrock iteration, which synchronously processed L1 deposit events, and nothing else.
+
+This worked reasonably well up to now, since it was mostly centered around a single new resource: the "unsafe" block.
+
+The OP-Stack is growing a lot: new types of resources,
+new customizations, and new validation rules are being introduced.
+This growth is something we need to layer on more solid abstractions.
+
+#### Driver tech-debt
+
+The `Driver` "schedules" the work.
+But currently really it only does some prioritization and is otherwise one monolithic synchronous loop.
+This needs to be addressed if we want to make it extensible with new asynchronous validation processes,
+rather than overloading the existing ones.
+
+Places where scheduling can improve:
+- Sequencing error handling has hot-loop edge-cases.
+- Derivation steps have tight time-out needs due to locking up unrelated resources.
+- Slow unsafe-block processing stalls the consolidation process for safe-head progression.
+- We also interleave sequencing and derivation steps, whereas these can be parallel.
+- We lock up all execution when reading the chain-state due to consistency needs.
+- There is no ability to attach a new validation process to the driver,
+  making engine-control sharing unnecessarily difficult.
+
+Stylistically there is more to improve too: the `Driver` is a *struct of 29 properties*, all managed in one place.
+
+#### Derivation tech-debt
+
+The `DerivationPipeline` is overloaded with more than just L1 to L2 data transformation:
+- The safe-block changes that happen are cached for later finality signals to work.
+  (More state that derivation has to deal with.) 
+- A plasma "input fetcher" may override fetching/traversal/finality as special case.
+  (Special hooks, rather than substituting what a rollup does.)
+- The EngineQueue derivation stage hosts a payloads queue that is not related to deriving from L1.
+  (No prioritization control as caller.)
+- The EngineQueue wraps an `EngineController`, rather than producing outputs.
+  (Derivation-caller has no control of what/when things are processed. Context timeouts make little sense.)
+- The derivation is started / reset with a "find sync start" algorithm.
+  (Taking a long time to run, not replaceable, and quite fragile.)
+
+And the above bloats the code-path that is fault-proven;
+the unsafe-block and finalization parts are dead code in the fault-proof context.
+
+### Interop
+
+With interop, we are adding an entirely new block safety type:
+["cross-unsafe"](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/verifier.md#cross-unsafe-inputs).
+
+This requires the op-node to:
+- asynchronously validate "unsafe" blocks to become "cross-unsafe".
+- block "safe" progression until blocks are "cross-unsafe" first (applicable to L1-derived block attributes also).
+- maintain a view of external L2 chains, and reorg local L2 chain-state when remote L2 chain-state is invalidated.
+
+### Alt-DA
+
+To support alt-DA and other customizations,
+there is a desire to extend the driver and derivation behavior with a bundled interface.
+
+Adding or modifying functionality should not require many sparse changes.
+Rather, we want a more plugin-like interface, that allows something to fit in and override core functionality,
+while reusing default functionality where desired, and still being structured to be fault-provable.
+
+Extending the stack should not require duplication of the state-transition 
+modifications between the proof system and regular node.
+
+Alt-DA also changes the meaning of "inputs";
+DA commmitments alone are not sufficient to finalize data as we would with rollup data,
+as the underlying data is not finalzied when the L1-registration of the commitment is.
+With plasma, a challenge period has to expire first.
+In other words, alt-DA changes the finality system.
+
+# Alternatives Considered
+
+<!-- *List out a short summary of each possible solution that was considered.
+Comparing the effort of each solution* --> 
+
+### Status-quo
+
+To integrate interop in the op-node as-is, we would have to:
+- Modify the L2 forkchoice state to include cross-unsafe.
+- Modify the payload-consolidation to have a cross-unsafe pre-requisite.
+- Modify the payload-force-processing to not instantly produce a safe block (that would bypass cross-unsafe).
+- Modify the various forkchoice-updates of EngineController to be consistent with the cross-unsafe head.
+- Modify the engine-queue/controller to allow reorgs through a different path than the current block-attributes path.
+- Modify the find-sync-start loop to add a block tag of cross-unsafe.
+- Modify the driver to accomodate for a special-purpose process that maintains a view of the cross-L2 safety,
+  while explicitly synchronizing the engine-control access in the main state-loop.
+
+This is high-effort (medium in size, but hard to test), and accumulates a lot more tech-debt,
+since it makes everything less extensible and adds complexity without simplifying any existing code.
+
+### Many `RepeatCond` and some shared locks
+
+See design doc [here (private)](https://github.com/ethereum-optimism/protocol-quest/issues/202).
+The gist is that it centers around conditions/effects,
+and picks the Go standard-lib solution for conditions that share resources, even though not a popular choice.
+
+From a specs perspective, this direction is promising, as it is relatively easy to match with a spec
+that describes these conditions and effects without the scheduling part.
+The effects change the chain-state after all, and should be better described in the rollup-node specs.
+
+This models each driver sub-process as a `RepeatCond` (managed condition/effect loop on `sync.Cond`).
+Conditionals can share locks, to avoid conflicting usage of resources.
+E.g. unsafe-block-processing and sequencing both modify the tip of the chain,
+independent of production of L2 attributes from L1 data.
+An illustrative draft of this can be found here:
+[monorepo PR 10051](https://github.com/ethereum-optimism/optimism/pull/10051);
+
+Pros:
+- The conditions/effect functions are easy to call manually, to perform sequences in non-flaky ways during testing.
+- It's easy to match the implementation with the specification rules.
+- Easily extensible/compositional. Adding `RepeatCond`, swapping conditions/effects, combining them, is easy.
+
+Cons:
+- The draft does not cleanly support prioritization of effects within a single resource.
+- It introduces more locks and `sync.Cond`, and doesn't follow the preference of communicating state.
+
+This is medium-effort, clears a decent amount of current tech-debt, but also introduces some new tech-debt.
+The pros are not exclusive to this solution, and the cons are.
+
+### Small state-loops
+
+This builds on the `RepeatCond` condition/effect idea, in particular the categorization by shared resource.
+There's an `unsafe`, `cross-unsafe`, `safe`, `finalized` and `currentL1` resource. 
+And each proceeds 1 at a time and can be signaled a reorg.
+
+When a condition is being checked, it implies a prestate for the effect, if the lock is not released in-between.
+By hosting the conditions/effects of the same resource together, it is easy to prioritize between them,
+and to keep a lock during execution of a condition/effect:
+the other conditions/effects would not be applicable while the primary resource itself is taken.
+
+The challenging thing here is to not copy-paste the issues of the current larger monolithic state-loop
+into these grouped smaller state-loops.
+
+Into this direction, we need to consider that:
+- When a resource is idle, it should be readable by many.
+- The loop itself should not inline key prioritization/scheduling logic, it makes it more difficult to test.
+
+The state-loops approach is not dynamically extensible: events are manually merged,
+prioritized, and categories of events are limited.
+
+This is medium-effort, but not all that much more testable than the current approach, and retains some tech-debt.
+
+### Derivers as block providers
+
+Another way to look at it, from @axelKingsley, is to think of it as a blocks-first system with "providers" and a "core".
+The "providers" produce blocks, and the "core" prioritizes and processes them, along with safety updates.
+
+There are some issues we identified:
+- Not all state changes are in fact blocks.
+- Creating blocks involves the engine (a contentious resource).
+- The "core" should not replicate the existing monolithic state-loop and synchronous execution issues.
+
+But this does start to introduce a plugin-like abstraction.
+A provider may do many things (alt-DA deriver) or just one thing (rollup safe-blocks deriver).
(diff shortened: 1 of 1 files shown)


op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+### Potential for test-vectors
+
+Defining standards for derivers, and making events encodeable as JSON,
+could be a solid basis for multi-client test-vectors.
+A test-runner would simply iterate through a list of encoded events, applied to a deriver,
+on top of some pre-state, and then assert some post-state properties.
+
+### Alt-clients: Rust / Java
+
+Typed events and event-switches are very natural in Rust and Java, and even more common than in Go.
```
clabby, line 481: Can confirm, this is a nice pattern for us :)

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+
+### Migration path
+
+This one of the more involved refactors, that we cannot execute all at once in op-node.
+This is a proposal for a 3-phase migration:
+
+#### 1) Derivation/Engine decoupling
+
+The derivation pipeline currently is coupled too tightly to the Engine.
+In particular, the `EngineQueue` should be removed,
```
clabby, line 492: 👍

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+- Not all state changes are in fact blocks.
+- Creating blocks involves the engine (a contentious resource).
+- The "core" should not replicate the existing monolithic state-loop and synchronous execution issues.
+
+But this does start to introduce a plugin-like abstraction.
+A provider may do many things (alt-DA deriver) or just one thing (rollup safe-blocks deriver).
+Ideally we find a way such that the derivation, safety-progression and scheduling logic
+is modular such that we can just combine a few providers to build the driver.
+
+We believe we do not need runtime plugins as binaries:
```
tynes, line 214: Ideally reth's builder api could be used as an inspiration for enabling a no-fork op-node with custom plugins

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+worker, ok := driver.workers[event.Resource]
+if !ok {
+    worker = ... // spawn new worker
+    driver.workers[event.Resource] = worker // keep it around
+}
+worker.OnEvent(ctx, event)
+```
+
+#### Instrumentation
+
```
tynes, line 387: This will be particularly helpful in interop

tynes (commented):
This is very thorough and well thought through

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+<!-- *An overview of what could go wrong. Also any open questions that need more work to resolve.* -->
+
+## Introduction of concurrency bugs
+
+When running derivers on different threads, there is state that may be shared between them, and needs to be safe.
+Deadlocks are especially important to avoid.
+Derivers should not acquire one resource before acquiring the other one when other derivers also do so.
+`ctx`-timeouts might unblock this, but the stall of locking up like this is already bad.
+Instead, derivers should design such that they communicate contentious work to a shared type of `Resource` worker.
```
axelKingsley, line 532: The `Resource` worker is essentially a router/prioritizer here? This helps explain my earlier confusion about the relationship between Resources and Workers a bit. 

With regards to resources: Without something in the design to prevent this, it's inevitable that we'd eventually have some concurrency error(s) over time.

I suppose the root deriver would mostly be calling to high level router/prioritizer workers, and the true implementation `Deriver` workers are leaf nodes of the structure. Then managing contentious resources is just a matter of router implementation.

Are there any cases where work has multiple contentious Resources that aren't managed by just a single `Resource worker`?

axelKingsley (commented):
In general, I am overwhelmingly in favour of this design. The proposed event-driven model helps us tease apart systems both in implementation and in spec.

My comments are varied, but in particular I have a few thoughts+questions about handling concurrency, and the intended application of some of these patterns. My gut feeling is that these Events are Messages with intended recipients, and that message-passing concurrency lines up well with a channel based design. Would love to know how this thinking does/doesn't work here.

Overall, this is an exciting direction for the Derivation stack, and I bet we'll see a lot of unexpected benefits with the additional composability.

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+
+To encourage composability, and reduce code, events should be documented and small in scope.
+I.e. create an event for a deriver to turn attributes in an unsafe block,
+and follow that up with events to try to promote the block safety, suggest it for finality, and persisting it to the safe head DB.
+rather than tightly coupling the four acts: (1) block processing, (2) updating of the safe-head, (3) remembering the finality data all at once, (4) persisting the safe head change.
+
+
+#### Threads and parallel derivers
+
+Stacks can define a single `OnEvent` handler, and by using `Resource`, still execute on parallel threads, if desired.
```
ajsutton, line 372: I'm unclear on how the `OnEvent` implementation is controlled.  In the examples above it looks like events can be targeted to specific derivers (`root` vs `engine` vs `pipeline`) but this suggests that they always call the "stacks" `OnEvent` and it routes based on the `Resource`.

How the stack is created is a really key part of this because fault proofs vs op-node is going to want to control how messages are passed to switch between sync and async, but different stacks will want to plug in different derivers. It would be good to avoid having to have a interop+derive, interop+fp, altda+derive, altda+fp kind of stacks.  So I think we want to decouple how message passing happens from where the message is handled (which this part implies but doesn't seem to fit with some of the stuff above).

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+type Engine struct {
+  control EngineController
+}
+
+func (e *Engine) OnEvent(ctx context.Context, event *Event) {
+    switch x := event.Data.(type) {
+        case *NewPayload:
+            select {
+              // events can include ways to communicate the result, if necessary, across threads
+              case x.Result <- e.onNewPayload(x.Envelope):
```
ajsutton, line 308: This may break the ability for fault proofs to run everything synchronously - particularly if that channel might fill up and block.

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+type Pipeline struct {
+    root Deriver
+}
+
+func (p *Pipeline) OnEvent(ctx context.Context, event *Event) {
+    switch x := event.Data.(type) {
+    case *L1Change:
+        // generate block attributes and give them to the unsafe-block processor to extend the chain
+        // (or forward to next deriver if already known attributes)
+        p.root.OnEvent(ctx, &Event{Resource: UnsafeL2, Data: AttributesEvent{attributes}})
```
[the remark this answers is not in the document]
ajsutton: It would definitely be very useful to have a clear set of threading expectations around event handling.  For example does `OnEvent` have to be handled synchronously or could it be async? Is that deriver dependent?

If async is allowed and/or expected, then a deriver that isn't thread safe could just be wrapped in a deriver that executes all events on a single thread (either by channels or with a lock).

ajsutton (commented):
I think I generally like this idea, though I'm worried that `Resource` is being overloaded for destinations, threading and locks but that we're also suggesting some "combined" resources (like `engine-queue` which is actually working with finalized, safe and unsafe heads at once).  Having separate Resource instances for the single and combined locks seems like it has a lot of potential for race conditions unless something knows that the combined resources are an aggregate of the others and can acquire locks appropriately.

There's also a big open question in my mind for how the derivation pipeline/stack actually gets built with all the right derivers and the right threading details between them (eg to force it all to be single-threaded for fault proofs).

In terms of threading, personally I actually like the idea of events always being asynchronous since it's the most flexible for the receiving implementation and callers are never caught by surprise that something was async since it always is. For fault proofs where we want "synchronous" execution, that would mean outputting the event into a buffer and the driver loop would just keep pulling events from the buffer and forwarding them along (it would have to store the original event and the target).

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+
+Things like the unsafe-block processing and sequencing work were fitted into safe-block processing code-path:
+1) This overloads the responsibilities of the derivation-pipeline, in particular the "engine-queue".
+2) This overloads the responsibilities of the driver state-loop, monolithically and synchronously executing everything.
+
+This is a remnant of the first Bedrock iteration, which synchronously processed L1 deposit events, and nothing else.
+
+This worked reasonably well up to now, since it was mostly centered around a single new resource: the "unsafe" block.
+
+The OP-Stack is growing a lot: new types of resources,
```
trianglesphere, line 60: Unless we modify the execution client, I think there are three resources we have to take a lock over
- Unsafe Head
- Safe Head
- Finalized Head

Block building is a bit weird b/c we should serialize those calls with the head updates, it can track it's own validity (i.e. is the payload still the tip payload & building on the same chain).

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+while reusing default functionality where desired, and still being structured to be fault-provable.
+
+Extending the stack should not require duplication of the state-transition 
+modifications between the proof system and regular node.
+
+Alt-DA also changes the meaning of "inputs";
+DA commmitments alone are not sufficient to finalize data as we would with rollup data,
+as the underlying data is not finalzied when the L1-registration of the commitment is.
+With plasma, a challenge period has to expire first.
+In other words, alt-DA changes the finality system.
```
trianglesphere, line 125: It really becomes an issues in alt-DA because we wan to use data before it becomes safe.
Spec'd here: https://github.com/ethereum-optimism/design-docs/blob/8d66ce7bd1ec09f9d4f6ebb7765844b03440ed36/plasma-mode.md#finalization

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+- Measure the execution-time per event type.
+- Track utilization per resource.
+- Trace/debug log the results and timeouts.
+
+This can help identify bottlenecks, does not have to be rewritten with each op-stack mod/extension,
+and fits it all in just a few reusable/aggregate metrics charts.
+
+### Spec changes
+
+In the specs we enshrine:
```
trianglesphere, line 399: This document is important for describing how the op-node works & how to modify it, but I don't think we should enshrine implementation in the spec.

mslipper:
I'm curious what the plan is to roll this out. Is there a way we can roll this refactor out incrementally in order to reduce review burden and risk?

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+- (new, maybe) Sync-status change (if not implied by unsafe L2 head change)
+
+Condition/effect pairs:
+- Unsafe block addition
+- Unsafe block sync-trigger
+- Unsafe block processing
+- Sequencing
+- Payload attributes processing
+
+#### Cross-unsafe head (interop)
```
sebastianst, line 431: Could you elaborate on the example of the cross-unsafe head how this new head stage could be added in a modular way without touching the resource implementations of the unsafe and safe head? IIRC the safe head would be dependent on the cross-unsafe in that it would be a new dependent intermediary head stage.

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+In particular, the `EngineQueue` should be removed,
+to allow the attributes-generation and unsafe-block processing to be truly independent processes.
+
+The derivation is somewhat statefully connected to the engine however:
+- Block-attributes result in later promotion to a "safe" head
+- Not all attributes are equal: those generated from span-batches result in unsafe blocks, until the last in the span is processed.
+- The find-sync-start routine is coupled to the `EngineQueue` `Reset()`
+
+We need a first step that addresses this, so the derivation process that generates the block attributes
+is shareable with the new driver design.
```
sebastianst, line 501: I like the scope of step 1) Could you elaborate on some details of how this could be done towards the new Deriver design?

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+
+#### 2) Incremental `OnEvent` adoption
+
+We can incrementally change the Driver state-loop sub-processes to fit the `OnEvent` signature,
+without changing the synchronous behavior of the state loop.
+This way the node behaves the same, and there is no risk of new concurrency bugs.
```
sebastianst, line 507: So in this phase, we'd implement the derivers without worker routines?

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+
+To get the parallelism, the `Driver`, which itself can be the absolute root `Deriver`,
+can create a new go routine for each new `Resource` type, when first seen.
+```go
+worker, ok := driver.workers[event.Resource]
+if !ok {
+    worker = ... // spawn new worker
+    driver.workers[event.Resource] = worker // keep it around
+}
+worker.OnEvent(ctx, event)
```
sebastianst, line 383: So each resource has exactly one worker? What's the relationship between resource workers and derivers? Where are the derivers registered?

sebastianst (commented):
Thanks for this next iteration of the proposal, learned a lot about the existing driver design and where we want to get to!

I would love to understand better how we achieve modularity with the Deriver design. How would new derivers be wired into an existing system. Would there be some kind of registration mechanism, or would it happen by augmenting existing derivers via embeddings?

I also don't quite understand yet how everything would be _driven_ in this new paradigm. What would the main routine do, and would there be exactly one worker routine per Resource?

sebastianst:
> I'm curious what the plan is to roll this out. Is there a way we can roll this refactor out incrementally in order to reduce review burden and risk?

I believe @protolambda presents a nice rollout plan in three phases towards the end of the proposal, where the first removes the EngineQueue, the second introduces the new pattern into the Deriver without changing underlying concurrency, and the third would then incrementally add concurrency.

tynes:
> > I'm curious what the plan is to roll this out. Is there a way we can roll this refactor out incrementally in order to reduce review burden and risk?
> 
> @mslipper I believe @protolambda presents a nice rollout plan in three phases towards the end of the proposal, where the first removes the EngineQueue, the second introduces the new pattern into the Deriver without changing underlying concurrency, and the third would then incrementally add concurrency.

A POC implementation also hid the new logic behind a flag and kept the old logic such that its possible to incrementally roll out the software to production, ie we run some nodes that opt into the new code and then after being sure that its safe we can eventually remove the old code

op-node-derivers.md
protolambda: Interop is already adding one (cross-unsafe), and previous designs of plasma was adding one (subjective-safe). And I think of the L1 traversal as another form of block-label resource too. For this reason I don't want to enshrine unsafe/safe/finalized as the only ones that matter.

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+1) This overloads the responsibilities of the derivation-pipeline, in particular the "engine-queue".
+2) This overloads the responsibilities of the driver state-loop, monolithically and synchronously executing everything.
+
+This is a remnant of the first Bedrock iteration, which synchronously processed L1 deposit events, and nothing else.
+
+This worked reasonably well up to now, since it was mostly centered around a single new resource: the "unsafe" block.
+
+The OP-Stack is growing a lot: new types of resources,
+new customizations, and new validation rules are being introduced.
+This growth is something we need to layer on more solid abstractions.
```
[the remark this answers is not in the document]
protolambda: Will add, thanks

op-node-derivers.md
protolambda: Will add the reference to the specs, thanks.

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+This is medium-effort, but not all that much more testable than the current approach, and retains some tech-debt.
+
+### Derivers as block providers
+
+Another way to look at it, from @axelKingsley, is to think of it as a blocks-first system with "providers" and a "core".
+The "providers" produce blocks, and the "core" prioritizes and processes them, along with safety updates.
+
+There are some issues we identified:
+- Not all state changes are in fact blocks.
+- Creating blocks involves the engine (a contentious resource).
```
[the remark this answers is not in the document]
protolambda: Right, will clarify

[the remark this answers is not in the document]
protolambda: Technically it's not a "block" until processed by the engine however, as block-inputs do not have EVM-outputs like a state-root.

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+The stack only has to prove a small subset of objectively provable onchain data however,
+when the need arises for this generalization, with its complexity, it can be revisited.
+
+# Proposed Solution
+
+<!-- *A high level overview of the proposed solution. When there are multiple alternatives
+there should be an explanation of why one solution was picked over other solutions.
+As a rule of thumb, including code snippets (except for defining an external API) is likely too low level.* -->
+
+## Desired properties
```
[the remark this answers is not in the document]
protolambda: Will add it to the list, thanks

op-node-derivers.md
```diff
@@ -0,0 +1,554 @@
...
+Note that things can be entirely synchronous, if the caller ignores the resource preference, and just runs one thing at a time. This can support fault-proofs, reducing the custom code (no fault-proof specific driver logic, just a synchronous loop of calling the deriver of choice).
+
+To get the parallelism, the `Driver`, which itself can be the absolute root `Deriver`,
+can create a new go routine for each new `Resource` type, when first seen.
+```go
+worker, ok := driver.workers[event.Resource]
+if !ok {
+    worker = ... // spawn new worker
+    driver.workers[event.Resource] = worker // keep it around
+}
```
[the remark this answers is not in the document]
protolambda: There can be multiple consumers of an event here; a deriver might choose to forward the same event to multiple sub-systems.

op-node-derivers.md
```diff
@@ -0,0 +1,573 @@
...
+- Modify the driver to accomodate for a special-purpose process that maintains a view of the cross-L2 safety,
+  while explicitly synchronizing the engine-control access in the main state-loop.
+
+This is high-effort (medium in size, but hard to test), and accumulates a lot more tech-debt,
+since it makes everything less extensible and adds complexity without simplifying any existing code.
+
+### Many `RepeatCond` and some shared locks
+
+See design doc [here (private)](https://github.com/ethereum-optimism/protocol-quest/issues/202).
+The gist is that it centers around conditions/effects,
```
bitwiseguy, line 156: Is "it" referring to the "existing rollup driver implementation"? Might be good to use the more descriptive term here

op-node-derivers.md
```diff
@@ -0,0 +1,573 @@
...
+and picks the Go standard-lib solution for conditions that share resources, even though not a popular choice.
```
bitwiseguy, line 157: Why is the choice to use the Go standard-lib solution for shared resources unpopular? What group of people is it unpopular with?

protolambda: It's a part of the standard-lib that is only used in 2 places in geth, and generally discouraged as something that can be done better with other patterns. I think `sync.Cond` with the right abstractions is powerful, but it is rare. And I am more happy with the other event based solution, since it requires less locking and is more of a compositional interface that uses communication rather than shared state.

op-node-derivers.md
bitwiseguy: Nvm, I see "it" is referring to the proposed alternative design described in detail within the linked issue

op-node-derivers.md
protolambda: Good points, will combine into the other resource/event related discussion.

op-node-derivers.md
protolambda: Two main reusable building blocks:
- The safe-head handling can respond to events that describe when something can be promoted promoted.
- The unsafe-head handling can take responsibility over processing block-attributes into new blocks, and emit events that signal the unsafe-head change with the context of the block-attributes causing that change.

We can compose the above with one of two options:
- Regular: Translate the unsafe-block-from-attributes signal to a promote-to-safe signal
- Interop: Accept unsafe-block-from-attributes as trigger to check cross-unsafe status.

So depending on regular/interop mode, we can just translate one event in the other, to trigger the right behavior. Or in other words, if we don't tie together the exact derivers events, we can compose them by translating the events any way we like.

op-node-derivers.md
protolambda: Will expand the doc here :+1:

op-node-derivers.md
protolambda: Essentially yes. Just leave the event processing synchronous, as things are now. But organized such that we can start running event processing in parallel when we wish so in phase 3.

protolambda:
Alignment on:
- We address the `EngineQueue` situation first.
- We can take more inspiration from the reth builder API.
- Instrumentation feature of the deriver is good to have.
- We want an event-based API for derivers

Aware of:
- Need to keep fault-proofs synchronous. Code that we lift out of the `derive` package into the `driver` package,
  unless not used by Fault-Proofs, should be either synchronous or have a simple synchronous alternative.

Contention on:
- (Josh/Axel) How we adapt the specs. See [here](https://github.com/ethereum-optimism/design-docs/pull/1/files#r1591728581) Two viewpoints:
  - Do not enshrine implementation in the spec, or anything like this structure.
  - Do not enshrine implementation, but do enshrine the abstract state-changes,
    such that we have an overview of these in the specs.

Confusion on: *the exact usage of `Resource` / `OnEvent`*:
- Axel:
  - Should we call `Resource` an `Inbox`?
    - A: A resource is not the final receiver of the events, but rather ideal executing thread.
  - Should we call event `Message`?
    - A: Not every event has a prescribed destination, so no.
  - > The Resource worker is essentially a router/prioritizer here? This helps explain my earlier confusion about the relationship between Resources and Workers a bit.
    - A: It "routes" in the sense that it runs the deriver stack. It's parallel to other workers, who can focus on modifying other resources.
  - > Are there any cases where work has multiple contentious Resources that aren't managed by just a single Resource worker?
    - A: We can implement them, but ideally we create a new resource, to communicate state, rather than sharing state to communicate.
- Adrian:
  - Stacks:
    - > It would be good to avoid having to have a interop+derive, interop+fp, altda+derive, altda+fp kind of stacks. So I think we want to decouple how message passing happens from where the message is handled (which this part implies but doesn't seem to fit with some of the stuff above).
    - A: Composability is meant to make these different stacks very lean / easy to create.
  - Event routing:
    - > I'm unclear on how the OnEvent implementation is controlled.
    - `Resource` is meant to signal which *thread* runs the `OnEvent` call.
      There is still only one instance of the deriver stack.
  - Concurrency:
    - > It would definitely be very useful to have a clear set of threading expectations around event handling.
        > For example does OnEvent have to be handled synchronously or could it be async? Is that deriver dependent?
    - A: agreed, something that we can improve. The driver is meant to handle anything concurrent.
      The derivation is meant to handle anything synchronous, and there should be very minimal work to do
      between the derivation and the engine, such that almost none of the driver functionality has to be fault-proven.
- Seb:
  - So each resource has exactly one worker?
    - A: yes
  - What's the relationship between resource workers and derivers
    - A: worker executes, deriver maintains state and defines what happesn to an event.
  - Where are the derivers registered?
    - A: They are not "registered", but rather composed into a single root deriver.
- (Seb/Axel) Do we want a "decoupled pub/sub architecture"?
  - Would there be some registration process at the top level, and then on an incoming event a loop would go over all registrants and call OnEvent?
  - A: No, composition of event-handlers can serve the same. The `OnEvent` should maybe be renamed,
    it's meant as a standard interface, such that derivers are all compatible and easy to wrap.
- "If we don't have registration, how would we achieve the decoupled modularity?"
  - A: Components can be composed, such that they receive the events that are meant for them.
    Components do not have to dynamically register themselves during runtime.
    The `Deriver` interface allows any component to be composed as a target for events to be emitted into.

Proposal:
- Don't use channels in derivers, keep them fault-provable.
- Introduce an interface for emitting events, attached to the root driver (not any deriver).
  This can then target the right worker-thread for the given resource, if async, or just execute synchronously.
  ```go
  type Emitter interface {
    EmitEvent(event *Event, dest Deriver)
  }
  ```
- Rename `Resource` to something that better describes "thread concerned with single L1 or L2 block-tag label".

op-node-derivers.md
protolambda: Answered in aggregate resource/event comment below.

op-node-derivers.md
protolambda: Answered in aggregate resource/event comment below.

op-node-derivers.md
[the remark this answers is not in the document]
protolambda: Tried to answer in aggregate resource/event comment below. Let's discuss there, this discussion is getting a bit sparse but it's all related.

PR #2, opened 2024-05-02
Add eth_sendRawTransactionConditional Design Doc

hamdiallam:
**Description**

Design doc for supporting for `eth_sendRawTranscationConditional` that's needed for Bundlers in the 4337 ecosystem, especially as shared 4337 mempools begin to launch.


hamdiallam:
Waiting for @jakim929 review before sharing more widely

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+# Problem Statement + Context
+
+Bundlers aggregate UserOps from an external mempool into a single transaction sent to the enshrined 4337 `Entrypoint` contract. It is the Bundler's responsibility to ensure every UserOp in the transaction will not revert when its validation step is executed. In between tx submission and inclusion, there is a chance network state changes causes the transaction to revert, even after local bundler validation.
+
+With private Bundler mempools, the likelihood of these reverts are small. However shared 4337 mempools are coming into [production](https://medium.com/etherspot/decentralized-future-erc-4337-shared-mempool-launches-on-ethereum-b6c860072f41), launched on Ethereum and some L2 testnets, decentralizing 4337 infrastructure. This is possible with no change on Ethereum through special block builders like Flashbots which support atomic transaction inclusion. An endpoint like `eth_sendRawTransactionConditional` is required to launch a shared mempools on L2 as there's only a single block builder, the sequencer, and the likelihood of reverts with a shared mempool would be too high of a cost for bundlers to operate.
```
tynes, line 18: I think its important to be explicit around the problem of "multiple bundlers include the same user op". This is the precise reason why the public 4337 mempool needs a situation like this

tynes: Flashbots does not solve this problem

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+# Alternatives Considered
+
+## Continue with private 4337 mempools
+
+4337 account abstraction is live on optimism. Dapps likely utilize the bundler endpoints that are come with a vendor-specific SDK -- Alchemy, Pimilico, Thirdweb, etc. As 4337 infrastructure becomes more permissionless, we will likely later have to play catch-up to ensure the OP-Stack remains compatible while other L2 offerings have already moved towards supporting `eth_sendRawTransactionConditional`.
+
+## Verticalize the OP-Stack
+
```
tynes, line 28: Related to this paragraph, there is also a solution where the sequencer runs its own 4337 mempool software that somebody else maintains, ie not make it part of the op stack itself but instead part of our infrastructure. Are there any benefits to this sort of architecture?

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+With private Bundler mempools, the likelihood of these reverts are small. However shared 4337 mempools are coming into [production](https://medium.com/etherspot/decentralized-future-erc-4337-shared-mempool-launches-on-ethereum-b6c860072f41), launched on Ethereum and some L2 testnets, decentralizing 4337 infrastructure. This is possible with no change on Ethereum through special block builders like Flashbots which support atomic transaction inclusion. An endpoint like `eth_sendRawTransactionConditional` is required to launch a shared mempools on L2 as there's only a single block builder, the sequencer, and the likelihood of reverts with a shared mempool would be too high of a cost for bundlers to operate.
+
+
+# Alternatives Considered
+
+## Continue with private 4337 mempools
+
```
tynes, line 24: I don't follow how continuing to use private 4337 mempools is an alternative to implementing the RPC. Its not up to us to determine where the orderflow ends up. Some apps may not want their orderflow to go to a public mempool. I think the alternative is "do nothing"

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+## Verticalize the OP-Stack
+
+Rather than externalize the 4337 mempool, the op-stack could natively offer a UserOp mempool alongside the regular mempool, avoiding conflicts as the sequencer can ensure the bundled UserOps do not conflict with the latest chain state when sealing a new block. However, this adds additional complexity to the stack, where the proposer-builder seperation in 4337 nicely keeps account abstraction out of protocol. A chain operator can still choose to verticalize in proposed solution solution by allowlisting only its own bundler through `eth_sendRawTransactionConditional`, achieving the same outcome.
+
+
+# Proposed Solution
+
+1. Implement `eth_sendRawTransactionConditional` as described in the [spec](https://notes.ethereum.org/@yoav/SkaX2lS9j). A draft implementation for this [exists](https://github.com/ethereum/go-ethereum/compare/master...tynes:go-ethereum:eip4337)today. The conditional is applied on submission and tx inclusion.
```
tynes, line 34: Can you define submission and tx inclusion? This isn't precise enough to really know. The invariants are checked on entrance to the mempool and also during block building

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+2. Introduce additional validation.
+    * **Exclusive 4337 Entrypoint Support**: We assert `tx.to() == entrypoint_contract_address` to ensure that this endpoint isn't used for other usecases. Making it easy to rollback if deprecated in the future due to native AA or a better solution to this problem.
```
tynes, line 36: There are multiple entrypoint contract versions deployed at different addresses

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+    * **Runtime Shutoff**: The endpoint can be shutoff by the sequencer to address issues. As long a single bundler supports a fallback to `eth_sendRawTransaction`, 4337 liveness should remain OK.
+    * **Global Rate Limit**: Rate limit how many conditional txs can be buffered at any given time in the mempool.
```
tynes, line 38: "in the mempool" - this rate limiting should not be happening in the node software, the rate limiting should happen before the node software, ie a proxy

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+    * **Authentication**: Requests must be authenticated, flashbot-style, with a self-identified keypair. This enables the development of two authorization policy modules.
```
tynes, line 39: Please define the semantics exactly rather just saying just like flashbots

tynes, line 39: Can we do product research to make sure that the flashbots style auth scheme can work well? Also maybe flashbots will share their reputation code with us

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+        1. _Allowlist_: The sequencer can verticalize their chain with a bundler sidecar or have specific partners allowed to send conditional entrypoint txs.
+        2. _Local Rate Limiting_: The authenticated caller has flashbots-style [reputation](https://docs.flashbots.net/flashbots-auction/advanced/reputation) computed based on the success rate of conditional txs, enabling permisionless bundler participation
+
+The endpoint MUST be authenticated but the authorization policies may not be needed to permissionlessly launch this feature if a global rate limit with a runtime shut off is sufficient. A recommendation is to have the allowlist policy implemented with the keys of known bundlers registered, such that it is ready to enable. As this endpoint is used in production, additional policies like the local rate limits can be iterated on and implemented on a as-needed basis.
```
tynes, line 43: I would recommend on making the initial scope very obvious and then have a section for future iterations. It could be framed like "initial scope, these things can go wrong, if x goes wrong this is solution, if y goes wrong this is the solution..."

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+# Proposed Solution
+
+1. Implement `eth_sendRawTransactionConditional` as described in the [spec](https://notes.ethereum.org/@yoav/SkaX2lS9j). A draft implementation for this [exists](https://github.com/ethereum/go-ethereum/compare/master...tynes:go-ethereum:eip4337)today. The conditional is applied on submission and tx inclusion.
+2. Introduce additional validation.
+    * **Exclusive 4337 Entrypoint Support**: We assert `tx.to() == entrypoint_contract_address` to ensure that this endpoint isn't used for other usecases. Making it easy to rollback if deprecated in the future due to native AA or a better solution to this problem.
+    * **Runtime Shutoff**: The endpoint can be shutoff by the sequencer to address issues. As long a single bundler supports a fallback to `eth_sendRawTransaction`, 4337 liveness should remain OK.
```
tynes, line 37: Can you add some notes on the cloud infrastructure diff required to enable runtime shutoff and global rate limit?

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+
+We can implement this endpoint in op-geth with an additional layer of external validation to ensure this endpoint is safe from DoS attacks, does not supersede `eth_sendRawTransaction`, and can be deprecated if native account abstraction efforts in the future absolves the need for this endpoint.
+
+**Note:** As a result of [EIP-3074](https://eips.ethereum.org/EIPS/eip-3074), the interop project is also exploring external validation as a mechanism to preventing DoS attack. A low-stakes scoped version of this for `eth_sendRawTransactionConditional` is a good opportunity for learning.
+
+
+# Problem Statement + Context
+
+Bundlers aggregate UserOps from an external mempool into a single transaction sent to the enshrined 4337 `Entrypoint` contract. It is the Bundler's responsibility to ensure every UserOp in the transaction will not revert when its validation step is executed. In between tx submission and inclusion, there is a chance network state changes causes the transaction to revert, even after local bundler validation.
```
tynes, line 16: Can you add links to 4337 eip for terms like user op and entrypoint?

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+# Summary
+
+By providing an auxilliary transaction submission mechanism, [eth_sendRawTransactionConditional](https://notes.ethereum.org/@yoav/SkaX2lS9j), we can enable Bundlers to submit Entrypoint transactions with stronger guarantees, avoiding costly inadvertent reverts.
+
+We can implement this endpoint in op-geth with an additional layer of external validation to ensure this endpoint is safe from DoS attacks, does not supersede `eth_sendRawTransaction`, and can be deprecated if native account abstraction efforts in the future absolves the need for this endpoint.
+
+**Note:** As a result of [EIP-3074](https://eips.ethereum.org/EIPS/eip-3074), the interop project is also exploring external validation as a mechanism to preventing DoS attack. A low-stakes scoped version of this for `eth_sendRawTransactionConditional` is a good opportunity for learning.
+
+
+# Problem Statement + Context
```
tynes, line 14: Can you add links to dune dashboards that show 4337 usage between chains? It will help to build a narrative as to why this is important

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
+# Purpose
+
+To fully unlock ERC-4337 on the OP-Stack.
+
+# Summary
+
+By providing an auxilliary transaction submission mechanism, [eth_sendRawTransactionConditional](https://notes.ethereum.org/@yoav/SkaX2lS9j), we can enable Bundlers to submit Entrypoint transactions with stronger guarantees, avoiding costly inadvertent reverts.
```
tynes, line 7: Can you describe the semantics of the rpc in a tldr fashion?

ecopod/sendRawTransactionConditional.md
axelKingsley: +1, I'd like to understand this better too

ecopod/sendRawTransactionConditional.md
axelKingsley, line 34: In a PGA, transactions are reprioritized and re-executed rapidly until the 2s sealing time is up.

Are conditionals based only on information from already-built blocks, or are they also able to access state that was modified by an earlier tx in the same block? I assume it's the former, otherwise there are some strange behaviors when filtering on re-execution. But wanted to make sure my understanding is correct.

ecopod/sendRawTransactionConditional.md
hamdiallam: sounds good, i'll elaborate on this. And I believe flashbots does solve this problem through `eth_sendBundle`? Multiple bundlers can submit txs to flashbots which are omitted if they revert

ecopod/sendRawTransactionConditional.md
hamdiallam: So this is included as apart of the proposed solution. By having `eth_sendRawTransactionConditional` be authenticated, the sequencer can choose to only allow their own bundler in the allowlist

> A chain operator can still choose to verticalize in proposed solution solution by allowlisting only its own bundler through `eth_sendRawTransactionConditional`, achieving the same outcome.

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
...
+        1. _Allowlist_: The sequencer can verticalize their chain with a bundler sidecar or have specific partners allowed to send conditional entrypoint txs.
+        2. _Local Rate Limiting_: The authenticated caller has flashbots-style [reputation](https://docs.flashbots.net/flashbots-auction/advanced/reputation) computed based on the success rate of conditional txs, enabling permisionless bundler participation
+
+The endpoint MUST be authenticated but the authorization policies may not be needed to permissionlessly launch this feature if a global rate limit with a runtime shut off is sufficient. A recommendation is to have the allowlist policy implemented with the keys of known bundlers registered, such that it is ready to enable. As this endpoint is used in production, additional policies like the local rate limits can be iterated on and implemented on a as-needed basis.
+
+The restricted usage & rate limits prevents this endpoint from superseding `eth_sendRawTransaction`, only suitable for honest bundlers, and also easy to wind back if needed later on.
+
+# Risks & Uncertainties
```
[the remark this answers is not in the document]
tynes: This adds state lookups + simple invariant checking as part of the block builder. If you send a tx thru this api then the block builder pinky promises to only include the tx if the invariants are met. It could be good to talk about risk of equivocation. Right now if the block builder goes back on its promise, the only downside is reputational risk. There is no economic downside

ecopod/sendRawTransactionConditional.md
tynes: For the invariant to be held true, they must be evaluated against the pending state, ie the partially built block

ecopod/sendRawTransactionConditional.md
tynes: `eth_sendBundle` has the ability to express "do not include my tx if it reverts" (or it may just be implicit via that endpoint) but my current understanding is that the top level 4337 bundle will not revert if there is a single reverting userop. that particular userop will just revert and the others will be fine. This would be good to dig into to double check but seems like a huge dos vector if a single userop could cause the entire bundle to revert. If the top level tx (4337 bundle) doesn't revert but only its subcalls revert (userops) then `eth_sendBundle` does not solve the problem

ecopod/sendRawTransactionConditional.md
hamdiallam: "Do nothing" is likely a better paraphrasing. It's not up to use to determine orderflow but we could still rely on & even push for usage of the existing Bundler stack when when public mempools are live. Relying on a partner like Alchemy to provide vendor-specific AA entrypoints.

ecopod/sendRawTransactionConditional.md
tynes: For things that may cause denial of service, we want to be able to solve the problem with horizontal scalability. This leads me to believe a good longer term solution is a more generic "superchain backend" that can help with these invariants. I think `revm` + `EthersDB` (think foundry forking) is a really good way to horizontally scale here. Its also possible to build this on geth using a `StateDB` implementation that can fetch state from RPC

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,53 @@
+# Purpose
+
+To fully unlock ERC-4337 on the OP-Stack.
+
+# Summary
+
+By providing an auxilliary transaction submission mechanism, [eth_sendRawTransactionConditional](https://notes.ethereum.org/@yoav/SkaX2lS9j), we can enable Bundlers to submit Entrypoint transactions with stronger guarantees, avoiding costly inadvertent reverts.
+
+We can implement this endpoint in op-geth with an additional layer of external validation to ensure this endpoint is safe from DoS attacks, does not supersede `eth_sendRawTransaction`, and can be deprecated if native account abstraction efforts in the future absolves the need for this endpoint.
```
tynes, line 9: I recommend adding a link to the following document which is the current proposed AA roadmap: https://notes.ethereum.org/@yoav/AA-roadmap-May-2024

ecopod/sendRawTransactionConditional.md
protolambda, line 34: Gave this geth diff a quick review, and since it's over a year old I believe it needs a significant rebase.
- The tx-pool in geth was split into sub tx-pools. (one for type 1/2/3 txs, one for blob txs). Need to move some validation.
- The block-building code of `commitTransactions` may have changed a little. The surrounding `miner` code has been changing more, as the "pending" block is on-demand and the PoW/Clique era block builder is on its way out.

And the usage of `ValidateTxOptions` can be optimized, and should be rate-limited to avoid DoS:
- In the tx-pool it should measure the state-access cost, and remove txs that have lots of cost and idle around in the tx pool for too long.
- In the block-building code the cheaper validation checks like the nonce check should apply first.

And we definitely need some way of applying back-pressure to specific offchain actors with frequently invalid 4337 txs (the "local rate limiting" as described below). This has not been implemented in the linked geth diff, but I imagine there may be newer diffs in the ecosystem that do have something like it.

protolambda (commented):
Generally in favor of changes to op-geth to support 4337, but please structure the code as an isolated package as much as possible, and document/categorize the diff by updating the `fork.yaml`.

ecopod/sendRawTransactionConditional.md
tynes: @hamdiallam Can you add hyperlinks to the implementations by the other teams? [Arbitrum starts here](https://github.com/OffchainLabs/go-ethereum/blob/da4c975e354648c7be814ab9667b42f1c19cdc0f/arbitrum/conditionaltx.go#L25) and [Polygon starts here](https://github.com/maticnetwork/bor/blob/b8ad00095a9e3e508517d802c5358a5ce3e81ed3/internal/ethapi/bor_api.go#L70)

I haven't checked if other L2s support it

tynes: It would be good to compare precisely what these 2 implementations do to see the different tradeoffs that they made

ecopod/sendRawTransactionConditional.md
hamdiallam: Added a note in (2) about the scope of these rules

ecopod/sendRawTransactionConditional.md
axelKingsley: If invariants can come from partially built blocks, that creates some emergent behavior that doesn't sound great for UX:

Consider an invariant which is false on the first execution of an in-progress block. The transaction would get filtered out. But then if the invariant turns to *true* on recommit (because reprioritization changed the inclusion or outcome of the condition), the transaction has already been filtered out, and therefore still won't be included. Customers would see that the published block does not include their transaction, but *does* satisfy the invariant. In this way, invariants are less like simple `if` statements, and more like `never false` statements -- if the invariant filters the Tx even once, it wont' be included.

This is besides the extra compute pressure we'd be getting from checking these invariants on every recommit, which I want to be careful about.

ecopod/sendRawTransactionConditional.md
hamdiallam: > my current understanding is that the top level 4337 bundle will not revert if there is a single reverting userop.

It depends what part of the UserOp is reverting. If the validation phase of the UserOp reverts, the entire transaction does reverts, even if only a single UserOp fails

The execution phase of the UserOp is allowed to revert without concern to the top level tx

ecopod/sendRawTransactionConditional.md
hamdiallam: yea that's a good point, I added a section in `Optimized State Lookups` as areas for improvement in the proposed solution

ecopod/sendRawTransactionConditional.md
hamdiallam: added

ecopod/sendRawTransactionConditional.md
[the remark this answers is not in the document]
hamdiallam: I think it's good to highlight the compute requirement as a risk that could be a reason to rollback this endpoint

ecopod/sendRawTransactionConditional.md
hamdiallam: Sounds good i'll add the link to these.

Makes sense that the PR would require a bit of a refresh. Mainly included it as a reference as some of the details might be more appropriate in a specs PR, wdyt?

ecopod/sendRawTransactionConditional.md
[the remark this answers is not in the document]
hamdiallam: added a section on this

ecopod/sendRawTransactionConditional.md
hamdiallam: Agreed a shared codebase here would be immensely valuable. If we end up needing to go down this route, we can open up a communications channel with them

I don't anticipate us needing to implement local rate limiting in production

ecopod/sendRawTransactionConditional.md
hamdiallam: invariants are not applied on partially built blocks. Only the latest unsafe head. There is a chance that two conflicting bundler txs pass the mempool as they would be valid at the time of submission.

Based on the priority fee, one of them would get included and the other would then be dropped when the block is actually built.

+1 on the compute pressure, I added that as a risk. We could be smart in validation and reject the second competing bundler transactions that's detected to have a conflicting conditional, even when valid against the unsafe head but I think that's be overengineering to start and also introduces new behavior relative to the priority fee

hamdiallam: +1 on the extra compute pressure. There's an added section in the risks for this.

> The transaction would get filtered out. But then if the invariant turns to true on recommit (because reprioritization changed the inclusion or outcome of the condition)

This is a good point, the number of filtered out txs in the block building phase is something we need to keep metrics on. Would be interesting to see the metrics on arb which has this endpoint publicly available. With the usage being relatively low, my thinking is we can observe this behavior in testnet and be a bit reactionary with the safeguards present to rollback/shutoff the endpoint.

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,88 @@
...
+
+**Risk 3: Generalized External Validation.** Validation policies should be DRY'd between interop, eth_sendRawTransactionConditional, and any future use cases. These policies that are implementated should work well between these usecases as this approach is adopted and scales. The tech-debt here can grow quickly if each solution has it's own methods of preventing DoS and validation, especially operationally.
+
+**Risk 4: Excessive Compute/Operational Requirements**. This endpoint is a feature provided out of protocol by the block builder -- the sequencer. With failed conditional transactions, the sequencer is not compensated with charged gas like when processing a reverted transaction. There's also the added overhead of managing new services that must mitigate DoS and increases the surface area where manual intervention will be required. The wasted compute on failed conditionals or inability to effectively mitigate DoS may be a reason to rollback this feature.
```
axelKingsley, line 88: Thanks for including this section :)

Just want to note that even when the transaction is included (conditional passes), the compute to check the invariant is still unpaid. And in fact, compute pressure from a passing invariant is higher than compute pressure from a failed one (because passing invariants are checked on every recommit, while a failed one would be dropped and not recomputed)

ecopod/sendRawTransactionConditional.md
tynes: Got it, thanks! Could you add this to the doc so other people don't have to ask the same q?

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,88 @@
...
+
+    * **Only 4337 Entrypoint Contract Support**: `tx.to() == entrypoint_contract_address`
+
+        The rationale is to make it easier to rollback if deprecated in the future due to native account abstraction or better solutions. Otherwise new uses case might create unwanted dependencies on this endpoint.
+
+        There does exist different [versions](https://github.com/eth-infinitism/account-abstraction/releases) of the Entrypoint contract as the 4337 spec is iterated on. We'll need to stay up to date with these version as a part of op-stack [preinstalls](https://docs.optimism.io/builders/chain-operators/features/preinstalls) and pass through calls to all the supported versions of `EntryPoint`.
+
+    * **Authentication**: Allowlist Policy
+
+        Requests to this endpoint MUST be authenticated with a secp256K1 keypair, similar to [flashbots authentication](https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint#authentication). The [EIP-191](https://eips.ethereum.org/EIPS/eip-191) hash of the json-rpc payload must be signed and included in the the `X-Optimism-Signature` header of the request in a `<public key adddress>:<signature>` format.
```
tynes, line 53: Are we sure that we need this for the first release? Can we use a dumb IP rate limit to reduce scope? This could be like 90% effective for 10% of the work

tynes: Also making our own header fragments the tooling that is for flashbots vs optimism, i can def understand why would would want to modify it but really there should be a more generic header name that everybody uses. Its a lot of work to coordinate that

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,88 @@
...
+
+If high rates of rejected transactions are observed, additional validation rules & improvements can be iterated on.
+
+* **Local Rate Limiting**
+
+    Adopt flashbots-style [reputation](https://docs.flashbots.net/flashbots-auction/advanced/reputation) score that's computed based on the success rate of conditional txs. Designed to filter out inefficient callers and promote honest ones.
+
+* **Optimized State Lookups**
+
+    Verifying the conditional associated with the transaction requires state access, for which the sequencer is not compensated when the transaction is rejected. To minimize resource usage, these checks could also be applied by the validating proxy prior to the sequencer mempool. This could then leverage an optimized state store through solutions like [revm](https://github.com/bluealloy/revm) to reject failed conditionals both cheaper and faster.
```
tynes, line 78: Can you elaborate on this? What do you mean by optimized? Is this `revm` + `EthersDB`?

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,88 @@
...
+
+
+# Proposed Solution
+
+1. Implement `eth_sendRawTransactionConditional` with support for the conditionals described in the [spec](https://notes.ethereum.org/@yoav/SkaX2lS9j), for which a draft implementation [exists](https://github.com/ethereum/go-ethereum/compare/master...tynes:go-ethereum:eip4337) but requires a refresh. The conditional attached to the transaction is checked against the latest unsafe head the prior to mempool submisison and re-checked when included in the block being built.
+
+    * There exists implementations for [Arbitrum](https://github.com/OffchainLabs/go-ethereum/blob/da4c975e354648c7be814ab9667b42f1c19cdc0f/arbitrum/conditionaltx.go#L25) and [Polygon](https://github.com/maticnetwork/bor/blob/b8ad00095a9e3e508517d802c5358a5ce3e81ed3/internal/ethapi/bor_api.go#L70) conforming to the [spec](https://notes.ethereum.org/@yoav/SkaX2lS9j). On Polygon, the API is authenticated under the` bor` namespace but public on Arbitrum under the `eth` namespace.
```
tynes, line 41: I would be curious if they do anything at the mempool layer. The current implementation attempts to prune the invalid txs on every new block, including the conditional ones, so that adds a constant sized amount of work relative to blocktime that should be noted in this doc

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,88 @@
...
+        There does exist different [versions](https://github.com/eth-infinitism/account-abstraction/releases) of the Entrypoint contract as the 4337 spec is iterated on. We'll need to stay up to date with these version as a part of op-stack [preinstalls](https://docs.optimism.io/builders/chain-operators/features/preinstalls) and pass through calls to all the supported versions of `EntryPoint`.
+
+    * **Authentication**: Allowlist Policy
+
+        Requests to this endpoint MUST be authenticated with a secp256K1 keypair, similar to [flashbots authentication](https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint#authentication). The [EIP-191](https://eips.ethereum.org/EIPS/eip-191) hash of the json-rpc payload must be signed and included in the the `X-Optimism-Signature` header of the request in a `<public key adddress>:<signature>` format.
+
+        With the public key of the caller, we can implement an allowlist policy module, allowing the chain operator to verticalize by running their own bundler or delegating to partners.
+
+    * **Runtime Shutoff**
+
```
tynes, line 58: I would propose monitoring that automates the shutoff over a certain point + ip based rate limits and drop the signature style thing for now. Adding the signature thing would need to loop in product so that we can talk to the bundlers and make sure they would use it

ecopod/sendRawTransactionConditional.md
tynes: I think its fine to include a link to our legacy implementation

ecopod/sendRawTransactionConditional.md
```diff
@@ -0,0 +1,88 @@
...
+1. Implement `eth_sendRawTransactionConditional` with support for the conditionals described in the [spec](https://notes.ethereum.org/@yoav/SkaX2lS9j), for which a draft implementation [exists](https://github.com/ethereum/go-ethereum/compare/master...tynes:go-ethereum:eip4337) but requires a refresh. The conditional attached to the transaction is checked against the latest unsafe head the prior to mempool submisison and re-checked when included in the block being built.
+
+    * There exists implementations for [Arbitrum](https://github.com/OffchainLabs/go-ethereum/blob/da4c975e354648c7be814ab9667b42f1c19cdc0f/arbitrum/conditionaltx.go#L25) and [Polygon](https://github.com/maticnetwork/bor/blob/b8ad00095a9e3e508517d802c5358a5ce3e81ed3/internal/ethapi/bor_api.go#L70) conforming to the [spec](https://notes.ethereum.org/@yoav/SkaX2lS9j). On Polygon, the API is authenticated under the` bor` namespace but public on Arbitrum under the `eth` namespace.
+
+2. Implement external validation rules. These rules are not enshrined in the node software's implementation of `eth_sendRawTransactionConditional`, instead scoped a layer higher, i.e a proxy, and horizontally scalable.
```
tynes, line 43: I think the proxy that does auth should be standalone and only do auth

mslipper:
We're good to go on this. Follow-up actions:

- [ ] Confirm authentication scheme with block builders
- [ ] Describe what metrics/run books we'll need to shut off the endpoint in the event of a DoS, and who will run the pager
- [ ] Enforce some kind of minimum fee at the proxy layer in order to improve chances of transaction inclusion
- [ ] Include metrics around transaction inclusion
- [ ] Describe how bundles can fall back to the regular `eth_sendRawTransaction` endpoint in the event of global shutdown
- [ ] Include a diagram of how this would look in the cloud


We'll ecentually need a design doc for how we do the out-of-sequencer checks, but this doesn't have to happen now.

PR #3, opened 2024-05-02
Add Plasma Mode Design Doc

trianglesphere:
**Description**

This adds the design doc for plasma mode.

**Metadata**

- Fixes https://github.com/ethereum-optimism/client-pod/issues/830

### DIFF
--- a/plasma-mode.md
+++ b/plasma-mode.md
@@ -101,8 +101,8 @@ The validity is checked by a third party smart contract which understands the co
 DA Layers which have ZK bridges from their DA Chain to Ethereum can write a SC which emits the following event.
 
 ```solidity
-    event MarkedInvalid(bytes indexed challengedCommitment, uint256 indexed challengedBlockNumber)
-    event MarkedValid(bytes indexed challengedCommitment, uint256 indexed challengedBlockNumber)
+    event Invalid(bytes challengedCommitment, uint256 challengedBlockNumber)
+    event Valid(bytes challengedCommitment, uint256 challengedBlockNumber)
 ```
 
 The flow is that after a commitment is posted to L1, it can be challenged within `challenge_window` blocks.
--- a/plasma-mode.md
+++ b/plasma-mode.md
@@ -73,6 +73,11 @@ These events then get adapted into a golang interface that the derivation pipeli
 
 ### Data Availability Challenge Smart Contract Specification
 
+We propose two different DAC SC event specifications.
+V1 is implemented for the Centralized DA system and can be extended to arbitrary DA layers as long as the DA layer is able to validate data against the commitment on L2.
+V2 is not yet implemented, but is designed to work with an instant finality DA bridge. Several DA layers have a header chain relay from the DA layer to L1 which
+can be used to instantly challenge a commitment on L1.
+
 #### V1: Optimisic Bridge
 The DAC SC will emit the following event
 ```solidity
--- a/plasma-mode.md
+++ b/plasma-mode.md
@@ -17,13 +17,13 @@ We have the following requirements for Plasma Mode:
 Additionally, we take the following development-process requirement:
 - The op-node and specification repo does not need to be modified for each new DA Layer
 
-The largest challenge to this developement process requirement is that DA Providers have different ways of committing to data and semantics of challenges. We need to create a design that generically supports a broad set of DA Providers without heavy modification or forking of the OP Stack.
+The largest challenge to this development process requirement is that DA Providers have different ways of committing to data and semantics of challenges. We need to create a design that generically supports a broad set of DA Providers without heavy modification or forking of the OP Stack.
 
 ## High Level Overview of Plasma Mode
 
 Instead of posting data directly to L1, the batcher gives the data to a DA server which posts the data to a DA layer.
-The DA server returns a commiment to the data to the batcher & the batcher posts that commitment to L1.
-The op-node performs derivation with the modification that when it finds a commiment on L1, it asks its DA server to get the data associated with the commitment.
+The DA server returns a commitment to the data to the batcher & the batcher posts that commitment to L1.
+The op-node performs derivation with the modification that when it finds a commitment on L1, it asks its DA server to get the data associated with the commitment.
 Then the rest of the derivation pipeline functions as normal.
 Finalization is modified because a L2 block can only be finalized when the commitment that was used to derive it can no longer be challenged.
 Commitment finalization takes significantly longer in Plasma Mode because of the challenge period.
@@ -48,7 +48,7 @@ The other option to avoid compiling in the DA Layer specific logic is to use [go
 
 ## Generic Commitment
 
-The generic commtiment is [specified here](https://github.com/ethereum-optimism/specs/blob/main/specs/experimental/plasma.md#input-commitment-submission).
+The generic commitment is [specified here](https://github.com/ethereum-optimism/specs/blob/main/specs/experimental/plasma.md#input-commitment-submission).
 The relevant part is copied to this document:
 
 The da-service commitment is as follows: `da_layer_byte ++ payload`. The DA layer byte must be initially restricted to the range `[0, 127)`. This specification will not apportion DA layer bytes, but different DA layers should coordinate to ensure that the DA layer bytes do not conflict. DA Layers can do so in [this discussion](https://github.com/ethereum-optimism/specs/discussions/135). The payload is a bytestring which is up to the DA layer to specify. The DA server should be able to parse the payload, find the data on the DA layer, and verify that the data returned from the DA layer matches what was committed to in the payload.
@@ -88,7 +88,7 @@ event ChallengeStatusChanged(bytes indexed challengedCommitment, uint256 indexed
 
 The flow is that after a commitment is posted to L1, it can be challenged within `challenge_window` blocks.
 If it is challenged an event with the `Active` status is emitted.
-If the challenge has not been resolved within `resolve_window` blocks, the challenge is presumed to suceed and the commitment is invalid.
+If the challenge has not been resolved within `resolve_window` blocks, the challenge is presumed to succeed and the commitment is invalid.
 If the challenge is resolved within `resolve_window`, an event with the `Resolved` status is emitted and the commitment is valid.
 If a commitment is never challenged within `challenge_window` blocks, it is valid.
 
@@ -122,10 +122,10 @@ const (
 func CommitmentStatus(comm GenericCommitment, L1Source eth.BlockID) (CommStatus, error)
 ```
 
-Commitment Status Defintions:
+Commitment Status Definitions:
 - `OptimisticOK` means that the commitment is assumed to be valid, but could be challenged & invalidated later
 - `ConfirmedValid` means that the commitment cannot be invalidated on the given L1 chain
-- `ConfirmedInvalid` means that the commitment has been invalited & will not become valid on the given L1 chain
+- `ConfirmedInvalid` means that the commitment has been invalidated & will not become valid on the given L1 chain
 
 We adapt the SC events into this golang interface to support multiple different bridge types without having to modify derivation.
 In addition, this confirmation model is more closely aligned to how the derivation pipeline works with commitment validity.
@@ -181,7 +181,7 @@ This is proposal helps inform how we should implement finality tracking; however
 ## DA Server / Challenge Code Quality Integration
 
 External parties will be responsible for both the DA server and DA Challenge Contract.
-This means that code will not follow the standard developement process of the op-stack.
+This means that code will not follow the standard development process of the op-stack.
 
 ## Superchain Registry Integration
 
--- a/plasma-mode.md
+++ b/plasma-mode.md
@@ -115,7 +115,7 @@ The op-node uses the following interface inside the derivation pipeline.
 ```go
 type CommStatus int
 const (
-    OptimisticOK     = iota
+    AssumedValid     = iota
     ConfirmedValid
     ConfirmedInvalid
 )
@@ -123,7 +123,7 @@ func CommitmentStatus(comm GenericCommitment, L1Source eth.BlockID) (CommStatus,
 ```
 
 Commitment Status Definitions:
-- `OptimisticOK` means that the commitment is assumed to be valid, but could be challenged & invalidated later
+- `AssumedValid` means that the commitment is assumed to be valid, but could be challenged & invalidated later
 - `ConfirmedValid` means that the commitment cannot be invalidated on the given L1 chain
 - `ConfirmedInvalid` means that the commitment has been invalidated & will not become valid on the given L1 chain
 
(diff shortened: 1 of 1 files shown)


plasma-mode.md
```diff
@@ -0,0 +1,203 @@
...
+
+The Data Availability Challenge ensures that nodes are able to fetch the data corresponding to the commitment posted on L1.
+If the data is not available, nodes should not attempt to use the commitment.
+If the data is only made available to a small portion of nodes, it should either be made available to all nodes by posting the data to L1 or the commitment should be rejected.
+Ensuring that nodes have a consistent view of the data is important to ensure that nodes derive the same chain and the fault proofs function correctly.
+
+Each DA Layer will have to implement logic on L1 to validate their commitments. This logic will emit a specific set of events that the op-node will read.
+These events then get adapted into a golang interface that the derivation pipeline uses to manage L2 reorgs & finality.
+
+### Data Availability Challenge Smart Contract Specification
```
clabby, line 74: Can we describe how this contract will be integrated into the dispute protocol? I see that the current `DataAvailabilityChallenge` contract has not yet been integrated, which should be the case before launch.

We touched on this back in November for the Lattice team, and they seem to have never done it (cc @tynes, who was leading this discussion at the time). The dispute protocol is meant to be generic enough to allow for disputes to exist over varying types of information. If the abstractions aren't suitable for the DA challenge, we need to improve them. 

If we do not bring this contract into the dispute protocol, we risk having several different methods for disputing information, where we could have had standardized interfaces and behavior.

trianglesphere: So far we haven't be planning on integrating it into the dispute protocol. One of the things about integrating it into the dispute protocol is that the op-node relies on this SC for derivation while it doesn't rely on the dispute protocol for derivation.

plasma-mode.md
```diff
@@ -0,0 +1,203 @@
...
+The DA server returns a commiment to the data to the batcher & the batcher posts that commitment to L1.
+The op-node performs derivation with the modification that when it finds a commiment on L1, it asks its DA server to get the data associated with the commitment.
+Then the rest of the derivation pipeline functions as normal.
+Finalization is modified because a L2 block can only be finalized when the commitment that was used to derive it can no longer be challenged.
+Commitment finalization takes significantly longer in Plasma Mode because of the challenge period.
+Data from commitments is also used prior to knowing if the commitment is valid or not.
+
+# Proposed Solution
+
+We propose a generic commitment type which can be interpreted by a DA server to fetch & validate the data. Challenge commitments is done in per DA layer smart contract where the smart contract emits events about the status of the challenge. The op-node then interprets those events to handle commitment validity and finality.
```
tynes, line 34: Does this mean that every integration will need to implement their own challenge contract?

plasma-mode.md
```diff
@@ -0,0 +1,203 @@
...
+    Resolved,      // Emitted when a challenge is resolved. This means the data has been made available on L1.
+    Expired        // Never emitted
+}
+
+event ChallengeStatusChanged(bytes indexed challengedCommitment, uint256 indexed challengedBlockNumber, ChallengeStatus status);
+```
+
+The flow is that after a commitment is posted to L1, it can be challenged within `challenge_window` blocks.
+If it is challenged an event with the `Active` status is emitted.
+If the challenge has not been resolved within `resolve_window` blocks, the challenge is presumed to suceed and the commitment is invalid.
```
tynes, line 91: typo `suceed` -> `succeed`

plasma-mode.md
```diff
@@ -0,0 +1,203 @@
...
+
+
+#### V2: ZK Bridge
+
+DA Layers which have ZK bridges from their DA Chain to Ethereum can write a SC which emits the following event.
+
+```solidity
+    event MarkedInvalid(bytes indexed challengedCommitment, uint256 indexed challengedBlockNumber)
+    event MarkedValid(bytes indexed challengedCommitment, uint256 indexed challengedBlockNumber)
+```
```
tynes, line 106: If you make a dynamic length type `indexed` then it will be the hash of the data in the event when the length is more than 32. I recommend making the commitment not indexed if we want the data to be legible to `op-node`

tynes: Also small nit on the naming, just wondering why `Valid` and `Invalid` aren't better names for the events :)

plasma-mode.md
```diff
@@ -0,0 +1,203 @@
...
+If it is challenged an event with the `Active` status is emitted.
+If the challenge has not been resolved within `resolve_window` blocks, the challenge is presumed to suceed and the commitment is invalid.
+If the challenge is resolved within `resolve_window`, an event with the `Resolved` status is emitted and the commitment is valid.
+If a commitment is never challenged within `challenge_window` blocks, it is valid.
+
+This works essentially as an optimistic bridge. The commitment is assumed to be valid until someone claims that it is not.
+The validity is checked by a third party smart contract which understands the commitment format.
+
+
+#### V2: ZK Bridge
```
tynes, line 99: Can you give a more thorough example into how this will work with blobstream?

plasma-mode.md
```diff
@@ -0,0 +1,203 @@
...
+If the data is not available, nodes should not attempt to use the commitment.
+If the data is only made available to a small portion of nodes, it should either be made available to all nodes by posting the data to L1 or the commitment should be rejected.
+Ensuring that nodes have a consistent view of the data is important to ensure that nodes derive the same chain and the fault proofs function correctly.
+
+Each DA Layer will have to implement logic on L1 to validate their commitments. This logic will emit a specific set of events that the op-node will read.
+These events then get adapted into a golang interface that the derivation pipeline uses to manage L2 reorgs & finality.
+
+### Data Availability Challenge Smart Contract Specification
+
+#### V1: Optimisic Bridge
```
tynes, line 76: What is the relationship between v1 and v2?

protolambda (commented):
For efficient data-challenges, and later fault-proof integration, I believe we need a standardized notion of two types of commitments:
- `DataCommitment`: commitment to the data on the non-L1 DA. Can simply be a hash of it the data. Independent of the L1.
- `ClaimCommitment`: commitment to the *inclusion* of the `DataCommitment` on the L1. I.e. a hash of the L1 inclusion point (blockhash, blocknumber, timestamp, perhaps also event log-index) combined with the `DataCommitment`

The above enables the challenge-contract to localize data: if the same data is posted twice, it will not have the same commitment. If the data gets challenged, you challenge the `ClaimCommitment` (ideally by providing the inclusion info and `DataCommitment`, such that the `ClaimCommitment` is computed onchain, and the challenge is legible and easy to index).
It is important that the commitments are unique, since "data availability" properties at the time of inclusion, within the challenge window, are what need to be enforced. Without it, the fault-proof ends up being tightly coupled to the data challenge results, making the implementation of a pre-image oracle unnecessarily difficult (FP may ask: for `DataCommitment` and inclusion info, was it available, and the challenge contract wouldn't know how to verify the inclusion if all its challenges were `DataCommitment` instead of `ClaimCommitment`).

The rollup node / DA service can then reconstruct the `ClaimCommitment` from the *raw* included commitment data in the L1, to determine the safety. I.e. in the happy path there is no need for any EVM execution: commitments can be posted just as raw data. Only the challenge contract, when challenged in the sad-path, needs EVM execution.

Currently the `GET /get/<hex_encoded_commitment>` does not specify how L1 inclusion data is communicated to the plasma DA service, making the plasma DA layer service to implement this kind of optimized plasma challenge protocol to determine the safety.

The `ClaimCommitment` could also be a fixed-size versioned standard op-stack type, while the `DataCommitment` can be more free-form.

While technically possible to engineer around not having this separation, enhancing the data-retrieval API with the inclusion info, and recommending the data-commitment / claim-commitment structure, makes the Plasma protocol better: there is no duplicate-challenges foot-gun, and it is efficient by default.

plasma-mode.md
```diff
@@ -0,0 +1,203 @@
...
+
+Commitment Status Defintions:
+- `OptimisticOK` means that the commitment is assumed to be valid, but could be challenged & invalidated later
+- `ConfirmedValid` means that the commitment cannot be invalidated on the given L1 chain
+- `ConfirmedInvalid` means that the commitment has been invalited & will not become valid on the given L1 chain
+
+We adapt the SC events into this golang interface to support multiple different bridge types without having to modify derivation.
+In addition, this confirmation model is more closely aligned to how the derivation pipeline works with commitment validity.
+
+When integrating with derivation, safe blocks are derived from commitments that are OptimisiticOK.
```
axelKingsley, line 133: This + Proto's doc on Derivation architecture works nicely -- when blocks become OptimisticOK, the DA Deriver can emit the SafeBlock event. When they are ConfirmedInvalid, it emits a Reorg event, and so on. Super cool.

plasma-mode.md
```diff
@@ -0,0 +1,203 @@
...
+The benefit of L1 execution for `ConfirmedValid` is that we will be able to significantly reduce the finalization delay for DA Layers that have timely ZK bridges.
+
+## Finalization
+
+At a high level, finalization is process of marking L2 blocks as not being able to be be reorged.
+If a commitment is marked as invalid, we have to reorg the L2 chain, therfore finality in Plasma Mode is about tracking when commitments finalize.
+
+To implement finality, we track commitments (even skipped or invalid commitments in order). When the derivation pipeline has advanced to the point where
+the commitment status changes from `OptimisticOK` to `ConfirmedValid` or `ConfirmedInvalid`, we record the L1 block in which that occurred.
+We then wait for that L1 block to be finalized. Once that L1 block has been finalized, the commitment status can never change.
```
axelKingsley, line 171: And to confirm, OptimisticOK turns into Confirmed when:
* In Optimistic, when the challenge window expires
* In ZK, when a proof is supplied

axelKingsley (commented):
Familiar with this doc so

plasma-mode.md
trianglesphere: Yes this is correct. The simplest contract that DA layers would have to implement is validating a commitment against data (i.e. validate the KZG or Merkle Hash or Merkle Mountain Range, etc) on L1 & then hook that into the current DAC smart contract.

plasma-mode.md
trianglesphere: I didn't know that about the indexed. Also updated the name. They are more clear.

plasma-mode.md
trianglesphere: I haven't really scoped this out yet. I'd assume that the Celestia DAC would reach out to the blobstream contract and call https://github.com/succinctlabs/blobstreamx/blob/aac0842f17056e5343f66de7df44020c1637e8b7/contracts/src/BlobstreamX.sol#L252-L281

cc @tuxcanfly will this interface work with blobstream

plasma-mode.md
trianglesphere: In ZK it could also be if the challenge period expires. Blobstream is half ZK where you could immediately get `Valid/Invalid` but don't want to call those contracts unless there is a dispute

plasma-mode.md
```diff
@@ -0,0 +1,203 @@
...
+The other option to avoid compiling in the DA Layer specific logic is to use [go plugins](https://pkg.go.dev/plugin). We reject this solution because it requires that the DA fetching logic be implementing in Go and it add places restrictions on how the resulting binary may be used.
+
+## Generic Commitment
+
+The generic commtiment is [specified here](https://github.com/ethereum-optimism/specs/blob/main/specs/experimental/plasma.md#input-commitment-submission).
+The relevant part is copied to this document:
+
+The da-service commitment is as follows: `da_layer_byte ++ payload`. The DA layer byte must be initially restricted to the range `[0, 127)`. This specification will not apportion DA layer bytes, but different DA layers should coordinate to ensure that the DA layer bytes do not conflict. DA Layers can do so in [this discussion](https://github.com/ethereum-optimism/specs/discussions/135). The payload is a bytestring which is up to the DA layer to specify. The DA server should be able to parse the payload, find the data on the DA layer, and verify that the data returned from the DA layer matches what was committed to in the payload.
```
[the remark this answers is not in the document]
trianglesphere: DA Layers often want more than 32 bytes. Usually they put in block height / hash + a cryptographic commitment to the data. Celestia is 8 bytes for the block number + 32 bytes for crypto commitment.

trianglesphere:
> * DataCommitment: commitment to the data on the non-L1 DA. Can simply be a hash of it the data. Independent of the L1.
> * ClaimCommitment: commitment to the inclusion of the DataCommitment on the L1. I.e. a hash of the L1 inclusion point (blockhash, blocknumber, timestamp, perhaps also event log-index) combined with the DataCommitment

This document describes the generic `DataCommitment`.  The `ClaimCommitment` is currently the L1 block number, though maybe we also need log index to handle multiple commitments inside a single L1 block.

We actually do track `ClaimCommitment` internally in the op-node & it's part of the challenge contract interface.

plasma-mode.md
tynes: The main complaint around using the dispute protocol vs the singleton contract is that the dispute protocol requires watching many contracts. It would need to be very easy to find all open disputes (represented by diff contracts) in a stateless way. The challenger prob does this so its likely possible, I had trouble getting buy in from Lattice at the time due to this unknown.

I do think it makes sense to develop tooling around a dispute protocol, especially as we move plasma to be fault provable, we will need to weigh this tech debt against other projects in the future and maybe one day we can unify

plasma-mode.md
tynes: Do you mind clarifying this directly in the document?

plasma-mode.md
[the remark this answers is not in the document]
alfonso-op: @protolambda what is the benefit of a fixed-size commitment in this case? What does it buy us?

@trianglesphere if we were to go with a fixed-size commitment, while 32 bytes is too small for current use cases, couldn't we set a reasonable upper-bound that would be (e.g., 64 bytes)?

plasma-mode.md
```diff
@@ -0,0 +1,203 @@
...
+- There is a DA Challenge which validates that the data is available
+- Optionally, the DA challenge process makes the data available on Ethereum
+- The op-node is able to mark L2 as finalized when commitments used to derive the L2 chain are no longer able to marked as invalid
+
+Additionally, we take the following development-process requirement:
+- The op-node and specification repo does not need to be modified for each new DA Layer
+
+The largest challenge to this development process requirement is that DA Providers have different ways of committing to data and semantics of challenges. We need to create a design that generically supports a broad set of DA Providers without heavy modification or forking of the OP Stack.
+
+## High Level Overview of Plasma Mode
```
alfonso-op, line 22: This and the next section would benefit from a few diagrams to illustrate the main flows, namely posting data to the DA Layer, performing derivation for a commitment and handling challenges.

plasma-mode.md
```diff
@@ -0,0 +1,203 @@
...
+If the challenge succeeds, the `MarkedInvalid` event is emitted. The `MarkedInvalid` event should not be emitted unless the data is not available for all nodes.
+If the challenge fails, the `MarkedValid` event is emitted. If the `MarkedValid` event is emitted, the data MUST be available for all nodes.
+
+### Derivation Integration
+
+The op-node uses the following interface inside the derivation pipeline.
+```go
+type CommStatus int
+const (
+    OptimisticOK     = iota
```
alfonso-op, line 118: Nit: Could consider renaming to "AssumedValid," "OptimisticAssumedValid" or similar, which may be a bit more self-documenting than OptimisticOk and be more consistent with the other statuses.

plasma-mode.md
tuxcanfly: Yes, given a commitment and it's inclusion proof blobstream can resolve the challenge by returning whether the attested data is available on celestia.

However, we will need the inclusion proof to be submitted to L1.

plasma-mode.md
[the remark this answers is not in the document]
protolambda: A 32-byte identifier would be easier to work with in the EVM, and bound what the op-node has to deal with. The pre-image of it can still be arbitrarily large if we want, and that's what is posted to the L1 inbox I imagine.

plasma-mode.md
trianglesphere: Two step tracking: watch the factory which emits an a contract address as an event which then emits a status

PR #7, opened 2024-05-13
feat: txpool service

tynes:
**Description**

Adds a design doc for the txpool-service, which can
be used to horizontally scale denial of service protection
in the world of interop and beyond.

### DIFF
--- a/txpool-service.md
+++ b/txpool-service.md
@@ -1,0 +1,154 @@
+# Purpose
+
+The purpose of this document is to illustrate a way to be safe from denial of service attacks in the context of interop
+in the mempool. This document is not meant to illustrate a way to be safe from denial of service attacks in all cases
+but this should cover the majority.
+
+This document also can apply to [sendRawTransactionConditional](https://github.com/ethereum-optimism/design-docs/pull/2)
+or [Flashbots protect](https://protect.flashbots.net/about) revert protection types of designs.
+
+# Summary
+
+A new horizontally scalable service called `op-txpool` is introduced that enables denial of service protection
+in the context of interop.
+
+# Problem Statement + Context
+
+Internet services can be susceptible to spam or denial of service attacks when they are permissionless
+to participate in. This is not a new problem, SMTP (Simple Mail Transfer Protocol) has been fighting against
+spam for a long time to the point where it is [very difficult](https://news.ycombinator.com/item?id=9798800)
+to run your own SMTP server at home.
+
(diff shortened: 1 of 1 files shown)


txpool-service.md
```diff
@@ -0,0 +1,154 @@
...
+Interop adds a new requirement such that execution is required to know if a transaction entering
+the mempool can be considered valid. This is where `op-txpool` can be very useful - it can execute
+transactions to look for executing messages and then resolve the initiating message from a remote chain.
+This architecture removes the need for the EL client to be aware of the remote chain RPCs.
+It can run optionally, meaning that full nodes don't need to run it. The first iteration of `op-txpool`
+doesn't need to do any of the checking that is done in the mempool already but in the future it can
+also act as a first level of checking before the transactions reach full nodes.
+
+## Implementation
+
```
tynes, line 109: Should add this this needs rate limiting on the number of RPC requests it sends out and also can run with `proxyd` to load balance remote requests against a consistent backend

tynes:
This needs an architecture diagram, but the tldr is:
- Stateless edge executor services that caches data and does execution of transactions
- Backend database nodes optimized for storing and serving data, optimized for consistency, runs no mempool
- The executor runs a forking evm on top of the backend database, starts with RPC over HTTP, will need to optimize to a binary based transport in the future
- Can horizontally scale the database serving and the execution and optimize them independently
  - can push commonly used data from the sequencer to these edge nodes
- The executor can also act as a proxy for `eth_sendRawTransaction` in the future

hamdiallam:
for the design-doc of `sendRawTransactionConditional`, should we replace the proxy described with `op-txpool`?  We can iterate on a single rpc server and consolidate early or we can start separately if it gets us to our goals faster

tynes:
# New Proposal

Based on convo with @axelKingsley and @protolambda 

Instead of adding a completely new service to act as a transaction ingress, simply add execution and an `op-supervisor` client to a full node's mempool. This feature can be opt in, meaning that not all nodes need to do this, but nodes that sit in front of the sequencing node will need to have this feature on to filter out invalid transactions before forwarding them to the sequencer. Ultimately, the sequencer should also double check executing messages when building the block itself due to the lack of strong consistency with the check at the periphery and the downside of including a an invalid executing message resulting in an invalid block.

This cuts scope and simplifies the architecture. We shouldn't attempt to optimize here yet.

tynes:
Closing this in favor of the solution defined in https://github.com/ethereum-optimism/optimism/issues/10956

PR #8, opened 2024-05-14
feat: solidity feature development

tynes:
**Description**

Outlines a proposal on how to do solidity smart contract
development in a way that reduces the need for feature
branches. We are testing this out in interop and it would
be nice to have the org on board with this as a pattern.

### DIFF
--- a/smart-contract-feature-development.md
+++ b/smart-contract-feature-development.md
@@ -1,0 +1,94 @@
+# Purpose
+
+The purpose of this document is to provide guidance on how develop smart contracts without the need for feature branches.
+
+# Summary
+
+Feature branches result in a lot of effort spent keeping the feature branch up to date. We want to have modular code
+that enables us to easily compose features together such that we can get code mainlined early and toggle it off when
+running in production.
+
+# Problem Statement + Context
+
+We want to move away from a world with long lived feature branches.
+To do so, we need to have some flexibility when it comes to composition of features.
+This is easy in off-chain code as the cost is minimal for toggling off certain features but still having them in the source code.
+This is not the case for Solidity. This document exists to outline how to think about composition of features for Solidity code.
+When the feature is ready to be shipped, the feature specific contract should be merged into the base contract.
+
+# Alternatives Considered
+
+## Copy and Paste Implementation
+
+This was the solution for the fault proofs project, which worked well until the custom gas token feature was introduced. Now the contracts
+are in a state such that the custom gas token feature needs to be ported to the `OptimismPortal2` before the `OptimismPortal` and the
+`OptimismPortal2` can be merged. This may be a good solution for security critical code as we do not rely on the correctness
+of solidity inheritance and all of the logic lives in a single contract, making it easier to audit.
+
+# Proposed Solution
+
+Below includes general guidelines for how to develop solidity feature contracts that are meant to be merged into mainline at
+some point in the future.
+
+### Composition through Inheritance
+
+Solidity allows contacts to [inherit](https://docs.soliditylang.org/en/v0.8.17/contracts.html#inheritance) other contracts.
+The functionality of the contract can then be extended or overridden.
+For in development features that modify core contracts already running in production,
+a new contract should be created that inherits the core contract and all new functionality should be added to the new contract.
+The deploy script should be able to parse some sort of config and optionally deploy the feature enabled version of the contract.
+
+### Handling Inheritance
+
+Inherit the contract that needs new functionality. It is very easy to add net new functions. If certain functions need to be modified,
+then the original contract may need to have the `virtual` keyword added in places. The `virtual` keyword should only be added in
+places where it must be added and nowhere in addition.
+
+### Handling Semver
+
+The parent contract should update its `version` getter to be a real function. The child contract should implement a `version` getter like so:
+
+```solidity
+function version() public override returns (string) {
+  return string.concat(super.version(), "+feature-name");
+}
+```
+
+The `feature-name` should be some identifier that uniquely identifies the feature. It could also be a hardfork name if the release is tied to a hard fork.
+
+### Handling Storage
+
+Handling storage becomes complex and there is no solution that handles all cases well. We generally want to ensure that the contracts
+are forwards compatible, so if `solc` is managing storage, the base contract should have spacers when we know that a feature contract
+is scheduled to be merged soon.
+
+Another solution is to use unstructured storage, meaning not letting `solc` manage storage and instead using custom storage slots.
+Assuming that digests and unique preimages are used for the storage slot keys, then this can work well for ensuring there are no
+storage slot collisions. This works well if there is no need for `solc` to manage the storage layout.
+
+If there are multiple feature contracts based on the same base contract being worked on in parallel, then there is complexity around
+layout management and it is up to the teams working on features to coordinate with each other if they risk mangling storage layouts.
+Using unstructured storage greatly reduces the coordination overhead here.
+
+More attention should be paid when the contracts are going to be deployed to a production network. It is less important to have the
+storage layouts 100% perfect when the contracts are going to be used on a transient network like a devnet. It is up to the developers
+to be mindful of their own upgrade paths as projects may have varying requirements.
+
+Given enough time, there will be storage layout issues. In the worst case, they can be fixed at upgrade time using the `StorageSetter`
+contract, this pattern is already used in production for `Initializable`.
+
+### Deploy Script
+
+The deploy script should be able to parse the deploy config in a standard way where the same config option enables the feature across the
+entire codebase. This means that a single config key/value pair should be able to have predeploys set in the L2 genesis, have the feature
+enabled contracts deployed to L1 and also tell op-node/op-geth to use that particular feature.
+
+### Merging Contracts
+
+The feature specific contract should be merged back into the base contract when the scheduling allows. This means that the code is nearing
+ready to be deployed to production and there is nothing immediately before that is going to be deployed and released.
+
+# Risks & Uncertainties
+
+- Longer compile times due to more contracts in the `contracts-bedrock` package
+- Coordination around the "lock" on the contracts package adds a lot of planning overhead. We need a better strategy for handling scheduling of features.


smart-contract-feature-development.md
```diff
@@ -0,0 +1,94 @@
...
+are in a state such that the custom gas token feature needs to be ported to the `OptimismPortal2` before the `OptimismPortal` and the
+`OptimismPortal2` can be merged. This may be a good solution for security critical code as we do not rely on the correctness
+of solidity inheritance and all of the logic lives in a single contract, making it easier to audit.
+
+# Proposed Solution
+
+Below includes general guidelines for how to develop solidity feature contracts that are meant to be merged into mainline at
+some point in the future.
+
+### Composition through Inheritance
```
maurelian, line 33: have we considered composition through solidity libraries? I haven't thought deeply about this but given that they are stateless, they may be easier to reason about, and reduce the complexity of storage concerns.

tynes: This is actually how the custom gas token feature was implemented, see [here](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/libraries/GasPayingToken.sol). That pattern made it easy to share functionality between multiple contracts and only write the storage related code once. I do like this pattern and there are likely opportunities to use it in the future. It isn't immediately obvious if we want to enshrine this pattern everywhere, it would become part of the style guide to make storage happen through libraries that expose getters to storage

maurelian: Cool. I think we agree it's a useful tool for certain jobs.

smart-contract-feature-development.md
```diff
@@ -0,0 +1,94 @@
...
+to be mindful of their own upgrade paths as projects may have varying requirements.
+
+Given enough time, there will be storage layout issues. In the worst case, they can be fixed at upgrade time using the `StorageSetter`
+contract, this pattern is already used in production for `Initializable`.
+
+### Deploy Script
+
+The deploy script should be able to parse the deploy config in a standard way where the same config option enables the feature across the
+entire codebase. This means that a single config key/value pair should be able to have predeploys set in the L2 genesis, have the feature
+enabled contracts deployed to L1 and also tell op-node/op-geth to use that particular feature.
```
maurelian, line 84: Applying this to the proposed solution of 'Composition through Inheritance', I take this to mean that Deploy Scripts should deploy a different contract depending on the key/value (ie. `OptimismPortal2` rather than `OptimismPortal`. Correct?

smart-contract-feature-development.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Merging Contracts
```
maurelian, line 86: I am a bit confused about what is meant by merging contracts. Is composition through inheritance only meant to be a temporary solution during the development phase, and then contracts will eventually be consolidated into one using feature flags to define behaviour?

smart-contract-feature-development.md
```diff
@@ -0,0 +1,94 @@
...
+
+The feature specific contract should be merged back into the base contract when the scheduling allows. This means that the code is nearing
+ready to be deployed to production and there is nothing immediately before that is going to be deployed and released.
```
ajsutton, line 89: I think we consistently see that requiring strict ordering of changes becomes really problematic because there are so often unexpected things.  Fault Proofs for example was "done" when an RC was taken but then further changes were required.  Other changes had already been merged in though so we had to cherry-pick commits which introduces risks of missing some important commits.

For me, the really key property we need is to keep develop releasable at all times (our tests and release process tries to disprove the theory that it's ok to release by finding issues). I think this inheritance approach adds a useful tool to our toolbox so we can get closer to that goal, but it seems like it doesn't entirely solve the problem if we are still needing to manage this sequencing. That said, if we only do this merge once a feature is confirmed ready to ship then it's fine because anything else that needs to ship is safe to deploy the merged contracts.

smart-contract-feature-development.md
```diff
@@ -0,0 +1,94 @@
...
+are forwards compatible, so if `solc` is managing storage, the base contract should have spacers when we know that a feature contract
+is scheduled to be merged soon.
+
+Another solution is to use unstructured storage, meaning not letting `solc` manage storage and instead using custom storage slots.
+Assuming that digests and unique preimages are used for the storage slot keys, then this can work well for ensuring there are no
+storage slot collisions. This works well if there is no need for `solc` to manage the storage layout.
+
+If there are multiple feature contracts based on the same base contract being worked on in parallel, then there is complexity around
+layout management and it is up to the teams working on features to coordinate with each other if they risk mangling storage layouts.
+Using unstructured storage greatly reduces the coordination overhead here.
```
mds1, line 71: Another idea: When writing your feature contract, always add storage variables to the base contract, and everything else (constants, events, initializers, core logic, etc) is added to the feature contract. This way, when two feature contracts are in being worked on in parallel, the rule is "only append to the base contract storage", which should reduce the coordination required and reduce the chance of storage layout issues

smart-contract-feature-development.md
```diff
@@ -0,0 +1,94 @@
...
+
+```solidity
+function version() public override returns (string) {
+  return string.concat(super.version(), "+feature-name");
+}
+```
+
+The `feature-name` should be some identifier that uniquely identifies the feature. It could also be a hardfork name if the release is tied to a hard fork.
+
+### Handling Storage
```
mds1, line 59: One final idea is giving each feature contract a dedicated storage contract, instead of using the existing contract. I don't love this idea because it's a bit clunky, but it would work and keep all storage isolated, so wanted to mention it for completeness. Example:

```solidity
contract OptimismPortal {
  // --- snip ---
}

contract OptimismPortalFeatA is OptimismPortal {
  OptimismPortalFeatAStorage featAStorage; 

  function initialize() external initializer {
    // Need to avoid redeployment on future initialization calls, one way is use
    // create2/create3 with a constant salt, and only deploy if there is no code at
    // the expected address.
    featAStorage = new OptimismPortalFeatAStorage();
  }

  function setX(uint256 x) external {
    featAStorage.setX(x);
  }

  function getX() external returns (uint256) {
    return featAStorage.x();
  }
}

contract OptimismPortalFeatAStorage {
  address public immutable parent;
  uint256 public x;

  constructor() {
    parent = msg.sender;
  }

  function setX(uint256 _x) external {
    require(msg.sender == parent, "Unauthorized");
    x = _x;  
  }
}
```

smart-contract-feature-development.md
mds1, line 84: Can we flesh this out a bit more? Is this referring only to feature flags that are smart contract specific, or, is this a generic new feature flag proposal, even if the feature only affects e.g. op-node and not contracts?

smart-contract-feature-development.md
tynes: I actually like this solution a lot for the predeploys, at least for chain specific storage. In practice the `initialize` wouldn't need to deploy a contract, it could just use a known predeploy.

I don't love this for the L1 contracts, but it does create an abstraction that works for storage

smart-contract-feature-development.md
```diff
@@ -0,0 +1,94 @@
...
+
+Handling storage becomes complex and there is no solution that handles all cases well. We generally want to ensure that the contracts
+are forwards compatible, so if `solc` is managing storage, the base contract should have spacers when we know that a feature contract
+is scheduled to be merged soon.
+
+Another solution is to use unstructured storage, meaning not letting `solc` manage storage and instead using custom storage slots.
+Assuming that digests and unique preimages are used for the storage slot keys, then this can work well for ensuring there are no
+storage slot collisions. This works well if there is no need for `solc` to manage the storage layout.
```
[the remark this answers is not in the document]
tynes: Discussion in `ethereum/solidity` on unstructured storage - https://github.com/ethereum/solidity/issues/597

tynes:
I think that @ajsutton hit the main point really well that this solution will not solve everything. I believe that it can solve the immediate needs of interop, we already have contracts written using this style and its been pretty painless. For what its worth, we don't add much extra storage and mostly add new functions or modify existing functions slightly

smart-contract-feature-development.md
tynes: This really depends on the feature. Some features only impact the contracts while others may impact the client code. Interop impacts both while custom gas token does not

smart-contract-feature-development.md
tynes: I do like this idea altho it adds risk around code that may be shipped to prod being changed due to changes in experimental features. If the experimental feature changes, then it would result in changes in storage layout in the base contract. I suppose that could be ok, if its always empty

smart-contract-feature-development.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Composition through Inheritance
+
+Solidity allows contacts to [inherit](https://docs.soliditylang.org/en/v0.8.17/contracts.html#inheritance) other contracts.
+The functionality of the contract can then be extended or overridden.
+For in development features that modify core contracts already running in production,
+a new contract should be created that inherits the core contract and all new functionality should be added to the new contract.
+The deploy script should be able to parse some sort of config and optionally deploy the feature enabled version of the contract.
+
+### Handling Inheritance
```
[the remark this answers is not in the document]
tynes: For fault proofs, it would have been possible to follow this approach by inheriting the contract and adding new functions. Functions that were removed in the new portal would have to be overridden and revert. You cannot subtract interfaces under this approach, which isn't ideal. I think that if you need to subtract interfaces, then its better dealt with composability of smaller interfaces, which is in the family of the approach mentioned. This would look like composing `OptimismPortal` from 2 contracts, one is all deposits and one is all withdrawals, then for `OptimismPortal2` implement a new withdrawal contract and compose it with the same deposit contract.

For interop, we are only adding interfaces, so we do not have this problem.

I suppose this is a good callout of how to think about the need to add or remove interfaces

smart-contract-feature-development.md
```diff
@@ -0,0 +1,94 @@
...
+
+# Problem Statement + Context
+
+We want to move away from a world with long lived feature branches.
+To do so, we need to have some flexibility when it comes to composition of features.
+This is easy in off-chain code as the cost is minimal for toggling off certain features but still having them in the source code.
+This is not the case for Solidity. This document exists to outline how to think about composition of features for Solidity code.
+When the feature is ready to be shipped, the feature specific contract should be merged into the base contract.
+
+# Alternatives Considered
```
mds1, line 19: We probably should have a section for feature branches as an alternative also, given the "Feature branches result in a lot of effort spent keeping the feature branch up to date" aspect of the summary. I would be curious to hear if there are learnings you can share from past attempts with feature branches for solidity development

smart-contract-feature-development.md
tynes: Yes exactly this, we are doing this with interop, see this example diff for what it would look like in practice

Screenshot 2024-05-23 at 12 05 33 AM

smart-contract-feature-development.md
```diff
@@ -0,0 +1,94 @@
...
+of solidity inheritance and all of the logic lives in a single contract, making it easier to audit.
+
+# Proposed Solution
+
+Below includes general guidelines for how to develop solidity feature contracts that are meant to be merged into mainline at
+some point in the future.
+
+### Composition through Inheritance
+
+Solidity allows contacts to [inherit](https://docs.soliditylang.org/en/v0.8.17/contracts.html#inheritance) other contracts.
```
mslipper, line 35: Do we expect that this will make compile times worse/cause problems with the shift to `--via-ir`?

smart-contract-feature-development.md
tynes: from https://github.com/ethereum-optimism/optimism/pull/10586

smart-contract-feature-development.md
tynes: > Is composition through inheritance only meant to be a temporary solution during the development phase

Yes exactly this. We currently do not have a paradigm where we deploy different contracts to enable different features. This could be a design space we explore in the future, but I lean against it as we are already trying to handle versioning of a single set of contracts, see https://github.com/ethereum-optimism/design-docs/pull/21

tynes:
We should have a list of possible implementation strategies and give criteria for how to evaluate which strategy to use

tynes:
We want to be sure to run the full test suite with the feature enabled to ensure there are no regressions, for example the bridge tests will never run with the interop feature turned on in the current architecture

tynes:
These changes could go in the style guide after we come to consensus on them

mslipper (approved):
We agreed on this design doc last week - we said we'd update style guides and provide examples inside the style guide. @tynes please let me know once this is done.

PR #9, opened 2024-05-14
superchain-backend design doc

protolambda:
**Description**

Design for the superchain-backend, the next iteration of the idea after initial takes by @hamdiallam in Q1. 

This supports the cross-L2 message verification for superchain interop.


superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+the log-event contract-origin address and log-event topics and data can be hashed into one commitment.
+
+Since initiating messages and executing messages are both log events,
+and assuming intra-block messaging is allowed by the protocol, we can:
+
+- use an executing message as initiating message
+- execute the same message
+- have a loop of executing messages between chains
+
+We should version the hash, and type it, such that we can:
```
tynes, line 157: Do you suggest making this hash an `indexed` field, so that it can be used to identify a log?

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+- Wrapper around a shreds reader: if we maintain the tip of the chain in memory based on streamed data,
+  we can serve queries for that data, and otherwise fallback to another source.
+
+In a fault-proof, a trustless version of the `eth_getBlockReceipts` can support the query,
+as that verifies the log-index correctness relative to the start of the block, and inclusion-proof to the block etc.
+
+### Batching
+
+We may want to consider batch-queries as part of the interface;
+query `[]BatchElem` (`BatchElem = (Identifier, PayloadHash)`), return `[]SafetyLevel`.
```
tynes, line 257: You are right that its an optimization to support this directly in the interface rather than do a batch RPC call

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+- Hashing method to turn event-log (initiating or executing message) into versioned hash
+- Append-log entries as follows:
+  - Block metadata (prefixed to be distinguishable from event-log hashes)
+  - Event log hashes (concatenated)
+- Binary search to find a log entry by block-number
+- Reorgs by truncating the append-log to the last common block before the reorg divergence
+
+## Superchain Backend implementation: `op-resolver`
+
+> *Name suggestions welcome, `op-superchain` seems to generic however.*
```
tynes, line 364: typo: `to` -> `too`

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+
+This service resolves superchain dependencies, and implements the above proposed spec for a standalone implementation.
+
+A later iteration of this can support the ideated shreds streaming,
```
tynes, line 368: Perhaps in the future this could also support the p2p optimizations, ie participate in the p2p network and fetch data that way

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+We may introduce non-standard RPC methods,
+to avoid overhead like fully encoding the receipts and hydrating their metadata.
+The event-log-filter query is already as narrowly scoped as possible (single block / contract),
+but is still effectively backed by receipts-retrieval in op-geth.
+
+If this proves to be too slow, then there still is the standalone service that benefits from prefetching
+and aggregate data handling (no receipts, just efficient message-hash lookups).
+In the long-term, with the growth of interoperating chains, light-clients (for just the cross-L2 part)
+may support the under-resourced nodes that are unable to run this service
+in addition to the current rollup-node and execution-engine stack.
```
tynes, line 457: Eventually we will need to move to a more optimized protocol rather than RPC over HTTP 1

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+- OP Mainnet history (from Dune query) = 1.9B logs.
+- Size of just 32 bytes per log = ~ 60 GB.
+- Size of receipts in geth (from recent `db inspect`) = 80+ GB.
+
+Caching a day worth of event log hashes is ~130 MB for OP Mainnet.
+The worst-case, if a chain is attacked with unbalanced log-event usage, has significantly more logs than that.
+This would amount to `13k / 93 = ~ 140` times as many log-events,
+or about 18 GB of log-event hashes for a full-day full-capacity attack on one standard-throughput chain.
+
+A standalone-service that prefetches the logs should be able to support the above:
```
tynes, line 113: Do you think that the `An RPC-backed, utilizing L2 execution engines to surface the data just in time.` implementation could work for the first release of the devnet or do you prefer to go straight to the standalone service approach?

superchain-backend.md
protolambda: For now it's an offchain detail. We can include it onchain, but it would add overhead, and we can't expect it from initiating messages anyway. The event-hashing here applies to both executing and initiating messages, so there is less data to retain and process offchain.

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+- Group messages by their blocks / chains, to not duplicate metadata
+- Warning: executing events can also count as initiating events
+    - We need to specify if we allow an executing message to consume itself,
+      i.e. if initiating is always before executing, if log-index matches.
+    - We can also have cross-chain loops; executing message is used as initiating message and executed by other chain,
+      and executing message on other chain is what is consumed as initiating message by original executing message.
+
+## Embedded RPC-based implementation
+
+As alternative to running a full `op-resolver` service, we may support a fallback in the op-node.
```
tynes, line 405: What is `op-resolver`?

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+Message safety is checked by relying on the local execution-engine
+for serving full event-logs and block-safety labels of the local chain,
+and then attaching to another external backend 
+(possibly served by an op-node / execution-engine also) to verify the safety of external messages.
```
tynes, line 409: We can either enforce that additional infra is ran and have all messages be validated or we could "assume true" bitcoin softfork style when the infra isn't set up to do full verification. I am not sure if the assume true style would be able to work in practice due to the way that invalid blocks are mapped to noop blocks - does the noop block end up being batch submitted? I feel like it would need to be batch submitted and the fork choice would need to favor it so that the hash chain in batches can be preserved. If the noop block does get batch submitted, then the "assume true" style syncing should work. Am i thinking about this correctly?

superchain-backend.md
protolambda: See above in doc, it's the placeholder name for the standalone superchain backend service.

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+Loading of initiating messages can take quite a while, this should not be the default critical path.
+
+Ideally we prefetch the data, and persist it efficiently for later lookups and validation.
+
+We can entirely separate all the data loading from validation:
+
+- **Loading**: Load log-events for each new block, transform into interop-hashes,
+  and append the block of hashes to a append-log.
+- **Validating**: Traverse the append-log, and for each executing hash, check for matching initiating hashes.
+  If at any point invalid, then truncate the log to remove any data that depends on it.
```
ajsutton, line 179: I'm assuming we also truncate the log if there's a reorg caused by something other than logs not being valid (ie safe head update forces a reorg or there's an L1 reorg).

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+To avoid reorgs in the fault-proof, the block-processing and cross-L2 verification should be interleaved,
+to run one step at a time each.
+
+With an additional simplification of not allowing equivocation of batch-data,
+and falling back to a safe deposits-only L2 block, the proof is simplified to running one block of each
+L2 and then performing an aggregate cross-L2 verification step of this block-height.
+
+The "backend" here is backed by pre-image oracle queries that:
+- inspect the emitted tentative event-log hashes of the latest block-height, to extract executing messages
+- inspect the matching event-logs of the historical data, to retrieve initiating messages (if valid)
```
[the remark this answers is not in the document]
tynes: We could implement something in the interop predeploy contract that ensures that the log is relatively recent but we need to benchmark to figure out what that number should be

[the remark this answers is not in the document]
ajsutton: I think the key challenge is that as the number of chains we might receive messages from increases, the max age of those events has to decrease since we might have to walk back many different chains.  We may need something that lets us more efficiently walk back through the chain like the hash aggregation done in the beacon state. Could be quite a few ways to address it, just needs some thought I suspect.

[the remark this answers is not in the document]
tynes: It would be nice if there was a way that we could parallelize the fault proof vm execution for different chains. The traces that are cached to disk could be useful here, it would be like spin up multiple vms that all produce traces and then consume the various traces

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+When considering the global log of interop hashes per L2 chain,
+the safety level label just moves along that log, like it does on the chain of blocks.
+
+### Simple interface:
+
+The verifier side is really all just one interface:
+
+```
+type MessageVerifier interface {
+    Check(Identifier, PayloadHash) SafetyLevel
```
tynes, line 233: Couldn't this also be useful for the v0 block builder? It would reduce a lot of diff in op-geth

tynes: Spoke to proto, yes this will be used for block building

superchain-backend.md
mslipper, line 113: How will this service scale? Will it support some kind of single-writer multiple-reader pattern to allow us to use multiple k8s replicas, or will we need to load-balance multiple distinct instances of this service (and therefore have some "consensus-awareness" concept in the load balancer)? Alternatively, could it support some kind of shared database?

We don't have to build for scale now, but as we've seen with the recent blob archiver outages we will need some degree of redundancy before going to prod. This service is a mission-critical component of  each interoperable chain, and we need to ensure that it stays up.

hamdiallam: I want to add that prefetching initiating logs might make sense in a world where users are unpredictably pulling all sorts of events as executing messages.

Maybe this is the case but my hunch is that at least initially we probably will see an extremely small subset of origin addresses that are pulled in as executing messages (bridge & cdm contracts).

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+- inspect the emitted tentative event-log hashes of the latest block-height, to extract executing messages
+- inspect the matching event-logs of the historical data, to retrieve initiating messages (if valid)
+
+
+# Alternatives Considered
+
+<!-- *List out a short summary of each possible solution that was considered.
+Comparing the effort of each solution* --> 
+
+## Original superchain-backend draft, with recursive resolution, RPC backing
```
hamdiallam, line 285: Maybe i'm misinterpreting but the original draft didn't trigger messages resolution recursively. Resolution was only triggered as initiating data that a block was dependent on has become available

superchain-backend.md
```diff
@@ -0,0 +1,457 @@
...
+```
+type SafetyLevel # see design-doc
+
+type Verifier interface {
+    CheckMessage(Identifier, PayloadHash) SafetyLevel
+    CheckBlock(ChainID, BlockHash, BlockNumber) SafetyLevel
+}
+```
+
+Summary of simplified implementation to include in specs:
```
hamdiallam, line 352: How does this implementation avoid the temporary unresolved dependencies between blocks? If the L2 peer hasn't caught up yet for an initiating hash to be looked up. Does the resolution for the block requesting that initiating halt or rejects the resolution? If not does it somehow subscribe to needing that data at a later point?

superchain-backend.md
protolambda: Standalone first. Added to the notes.

protolambda: Added to the notes

superchain-backend.md
protolambda: It's been a while, I have lost some of the context on these drafts. But I believe it was 1-deep resolution, and then relied on the same system resolving itself the same way, thus effectively recursive.

protolambda:
Comments summary:
- Optimization of acquiring data via RPC, rather via P2P or some aggregate or optimized form.
- Protocol change to be able to verify initiating/executing message inclusion more easily (i.e. not as inefficient as full receipt proofs)
- Flexible trust model, not everyone may want to run this new service.
- Need to expand on redundancy and reliability. May need load-balancer to be aware of unsafe vs cross-unsafe.
- One remaining comment from Hamdi that we did not get to during the design review call.

Doc to be updated, and can then do a final review/merge.

superchain-backend.md
protolambda: The append-logs can provide the data to resolve any initiating-messages. I'll update the summary to describe this, good catch.

protolambda:
Updates on previous review comments and requests:

- Optimization: added optimization ideas to throughput sub-section in considered risks.
- Protocol change for message-inclusion proofs: listed as option under light-client idea, but out-of-scope for now.
- Flexible trust model: design-doc already offers a general interface and some optionality, I believe this forms sufficient forward-compatibility with other trust models.
- Redundancy and reliability: added a new section covering this, with proposal how to use multiple backends as rollup node.
- Comment from Hamdi about block-safety resolution: added missing details on implementation summary, nice catch.

Additionally:
- Naming: we are calling the standalone service the `op-supervisor`, since it monitors superchain activity for cross-L2 safety.
- Rebased on `main`
- Moved the design-doc to `protocol/` directory.

superchain-backend.md
protolambda: The current interop spec assumes that there is a complete available history of cross-L2 data when determining the canonical chain from batch-inputs. If we want to make this not the case, then interop requires deeper batch-submission changes, out of scope for this superchain backend doc. When syncing from a L2 chain directly, without derivation from L1, there will not be an alternative interpretation of inputs with noop-blocks, so then the "assume true" approach can work without superchain backend data (if the safety-level is accepted, but since it's already assuming the chain is canonical, I think this is fine).

mslipper (approved):

PR #13, opened 2024-05-19
interop: deposits

tynes:
**Description**

This design doc is meant to explain how we can safely support
deposit transactions that trigger executing messages. It adds
a new stage to the derivation pipeline as well as a new invariant.
It modifies the way that deposit events are mapped into deposit
transactions to guarantee integrity of deposited executing messages.

### DIFF
--- a/protocol/deposits-interop.md
+++ b/protocol/deposits-interop.md
@@ -78,7 +78,7 @@ can cause nodes to inconsistent chains.
 The stateless injective function for mapping deposit events into deposit transactions is updated to be a
 2-valued function, meaning that any input can be mapped to one of two inputs. In mathematical terms:
 
-$$f:A→{b1,b2}$$
+$$f:A→\{b1,b2\}$$
 
 A deposit transaction event can be mapped to either a valid deposit transaction or a noop deposit transaction.
 
--- a/protocol/deposits-interop.md
+++ b/protocol/deposits-interop.md
@@ -7,7 +7,9 @@ What this means in practice is that a deposit transaction can be used to create 
 
 # Summary
 
-Protocol modifications need to be made
+Some deep protocol modifications need to be made, particularly in the derivation pipeline. A new way of mapping deposit
+events into deposit transactions is introduced via a new invariant as well as way to check for executing message validity
+as part of the derivation pipeline.
 
 # Problem Statement + Context
 
@@ -119,6 +121,10 @@ for i, tx in enumerate(payload.transactions):
         txs.append(tx)
         break
 ```
+
+If static analysis was possible, then no execution would be required to derive the information required to check the
+validity of the executing messages. The same two checks `sequencing_window_invariant` and `is_valid_message` could
+be applied but solely with static analysis instead of needing to execute to get the executing message.
 
 ## Sequencing Window Invariant
 
--- a/protocol/deposits-interop.md
+++ b/protocol/deposits-interop.md
@@ -89,7 +89,7 @@ transactions that trigger executing messages.
 
 The derivation pipeline is updated to include EVM execution of the deposit transactions. A forking EVM provider backed
 by L2 tip state is used. This enables simulation of the deposit transactions before they are included in L2. If the deposit
-transaction fails the sequencing window invariant or creates an invalid deposit transaction, it is mapped into a noop deposit
+transaction fails the sequencing window invariant or creates an invalid executing message, it is mapped into a noop deposit
 transaction.
 
 A noop deposit transaction:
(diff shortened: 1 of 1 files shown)


protocol/deposits-interop.md
```diff
@@ -0,0 +1,189 @@
...
+race conditions with deposit transactions and the sequencer having a slow p2p network. Without this invariant,
+it would be possible to trigger reorgs by sending a deposit transaction that references an initiating message
+that is at the tip of a remote chain. If the sequencer doesn't immediately always have the remote chain, it
+could result in reorgs.
+
+The definition of the sequencing window invariant is as follows:
+
+```
+require(identifier.timestamp + sequencer_window <= block.timestamp)
+```
```
tynes, line 145: This requires either:
- standard sequencer window for all chains
- sequencer window of remote chains is legible to the derivation pipeline

protocol/deposits-interop.md
```diff
@@ -0,0 +1,189 @@
...
+and the `op-node` is responsible for building the `ExecutionPayload` correctly. It is possible to
+have the EL modify (malleate) the `ExecutionPayload`'s deposit transactions before considering
+them final. This would essentially be adding more rules to the state transition function.
+
+This would look like a diff to the state transition function that applies the sequencing
+window invariant and does the executing message validity checking during the state transition.
+This creates a leaky abstraction as now the state transition function needs to know about
+the dependency set, doing remote RPC calls during execution. This does not seem like
+an ideal solution for that reason.
+
```
tynes, line 184: This would also impact p2p networking as the execution payloads are sent via p2p, would need to make sure to define semantics around signing the payload before/after the malleation

tynes:
Some short term solutions for devnet-0:
- do nothing, permission deposits that have calldata
- do nothing, allow for deposits to cause reorgs
- make diff in state transition function as a quick hack

protocol/deposits-interop.md
```diff
@@ -0,0 +1,189 @@
...
+  logs = evm.call(tx.to_message())
+  for log in logs:
+    if is_executing_message(log):
+      if sequencing_window_invariant(log) == False or is_valid_message(log) == False:
+        tx.gas = 0
+        tx.data = ""
+        txs.append(tx)
+        break
+```
+
```
tynes, line 124: This logic can be added to the derivation pipeline pretty easily [here](https://github.com/ethereum-optimism/optimism/blob/93891de2176f6f997da80a4de8de84466164bb25/op-node/rollup/derive/attributes.go#L70)

tynes:
This design would result in the requirement of archival state for historical sync. We don't want to impose that, h/t @protolambda 

Thinking more on how to solve this

tynes:
Some follow up thoughts based on a conversation with @protolambda 

- It may be possible that we can encode information in the block body using [EIP-7685](https://eips.ethereum.org/EIPS/eip-7685) to remove the need for historical execution. EIP-7685 is being included in [Pectra](https://notes.ethereum.org/@ethpandaops/pectra-devnet-0)
- Another design space is around having "is deposit" introspection in the EVM
  - The simple approach would be to modify the STF to set state in a predeploy before execution so that when called, it can return a bool
  - Another approach involves adding a second system deposit transaction at the tail end of the deposits. This would store something in state that indicates that there are no more deposits, so its possible to differentiate between deposits and not
- The "is deposit" introspection could be tied to a "delayed inclusion system" where if its a deposit that is triggering a cross chain message, the codepath would be different and it would result in a follow up where the deposit is "finalized"
- Another design space would involve providing a proof if its a deposit, this proof would be done against some global accumulator that is updated slowly

protocol/deposits-interop.md
```diff
@@ -0,0 +1,189 @@
...
+$$f:A→\{b1,b2\}$$
+
+A deposit transaction event can be mapped to either a valid deposit transaction or a noop deposit transaction.
+
+A new invariant called the "sequencing window invariant" is introduced that enforces a delay for deposit
+transactions that trigger executing messages.
+
+## Derivation Pipeline Execution
+
+The derivation pipeline is updated to include EVM execution of the deposit transactions. A forking EVM provider backed
```
tynes, line 90: We don't want to add execution to the derivation pipeline. We instead want to use a 2 step process that registers the dependency into the L2 state and then force deposits to use state that has been registered ahead of time. This means that there needs to be an "is deposit" introspection from within the EVM. This can be accomplished by having 2 system txs, the first on toggles on `isDeposit() -> true` and the second `isDeposit() -> false`. Ideally we make `isDeposit` only callable by the `CrossL2Inbox` so that arbitrary applications cannot use it to censor

tynes:
Additional context on how we have been thinking about this problem can be found in https://github.com/ethereum-optimism/optimism/issues/10867#issuecomment-2183111697

tynes:
This is being worked on in https://github.com/ethereum-optimism/specs/pull/258

PR #14, opened 2024-05-19
configurable sequencing window

tynes:
**Description**

Adds a design doc to describe a more flexible sequencing window design
where it can be modified by the chain operator as a `SystemConfig`
config option, similar to other config options. The sequencing
window of remote chains in the derivation pipeline of the local chain
is required to enable the safety of deposits of executing messages in
the context of interop.

While it is possible that interop can ship with a hard assumption on the
value of the sequencing window, we should consider a better way to make
the value legible to the protocol and at the same time make L3s more
stable.

### DIFF
--- a/protocol/protocol-sequencing-window.md
+++ b/protocol/protocol-sequencing-window.md
@@ -1,0 +1,91 @@
+# Purpose
+
+The purpose of this document is to standardize the "sequencing window" from the point of view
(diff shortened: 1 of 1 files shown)


tynes:
This is important for finality in interop so that a single chain cannot slow down the finality of fast batching chains

tynes:
Closing this for now

PR #20, opened 2024-05-22
design doc: invalid block progression

tynes:
**Description**

A design doc describing the two invariants that can ensure
that the derivation pipeline can always progress without needing
to stall and do a lot of extra work when invalid batches are submitted.

Ideally this can be included in a hardfork before interop ships
to make the derivation pipeline much more hardened for interop.

### DIFF
--- a/protocol/invalid-block-progression.md
+++ b/protocol/invalid-block-progression.md
@@ -1,0 +1,60 @@
+# Purpose
+
+The purpose of this document is to progress a design originally proposed by @protolambda and contributed to
+by Hamdi [here](https://github.com/ethereum-optimism/specs/pull/88). The goal is to reduce resource usage
+by the derivation pipeline when invalid blocks are proposed. This is helpful for ensuring that the fault
+proof program cannot be denial of service attacked as well as reducing the latency of reorgs, which is
+especially important in the context of interop.
+
+# Summary
+
+New derivation pipeline invariants are introduced that are meant to reduce large resets of the derivation pipeline
+when handling invalid batches.
+
+# Problem Statement + Context
+
+The derivation pipeline is set up as a series of stages. It takes work to progress from one stage to the next.
+Sometimes this work involves many remote HTTP requests, other times it requires in memory manipulation of
+data structures. Its not ideal when there is an error in a later stage that causes a lot of additional work
+to be done in an earlier stage to recover. In practice, this looks like some sort of invalid batch was deserialized
+and errors during its processing.
+
+> The `DerivationPipeline` has a `Step()` function that is called over and over again to derive the chain,
+which ultimately looks like mapping raw batch data into `ExecutionPayload`s. `DerivationPipeline.Step()`
+wraps a call to `EngineQueue.Step()` which wraps a call to `AttributesQueue.NextAttributes()`.
+
+Examples of types of errors:
+- [Invalid transaction in block](https://github.com/ethereum-optimism/optimism/blob/8b5e3c264d2a3a114e7df5361acd8d4c1c700f81/op-node/rollup/derive/batches.go#L156)
+- [Invalid Execution Payload](https://github.com/ethereum-optimism/optimism/blob/8b5e3c264d2a3a114e7df5361acd8d4c1c700f81/op-node/rollup/derive/engine_update.go#L94)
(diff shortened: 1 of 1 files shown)


protocol/invalid-block-progression.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+# Proposed Solution
+
+The following 2 invariants are introduced to ensure timely progression of the derivation pipeline. In both cases, either a batch
+is correctly progressed or it is mapped into a deposits only batch.
+
+If the batch data [makes it](https://github.com/ethereum-optimism/optimism/blob/8b5e3c264d2a3a114e7df5361acd8d4c1c700f81/op-node/rollup/derive/batch_queue.go#L223)
+to the `BatchQueue`, then the batch is either properly deserialized and validated or is mapped into a "deposits only" batch.
+This specifically describes the `BatchDrop` cases returned from [CheckBatch](https://github.com/ethereum-optimism/optimism/blob/8b5e3c264d2a3a114e7df5361acd8d4c1c700f81/op-node/rollup/derive/batches.go#L34).
```
tynes, line 40: This needs defined semantics for span batches

protocol/invalid-block-progression.md
```diff
@@ -0,0 +1,60 @@
...
+## Smaller Solution
+
+It is possible to only enforce one of the two invariants, where we would want to drop the EL `INVALID` invariant in favor of
+the batch queue invariant.
+
+# Risks & Uncertainties
+
+- This adds risk around unsafe head reorgs. To date, there has not been an unsafe head reorg. Cases where batches would be
+dropped and then there is another chance to progress the safe head via submitting another batch would be turned into
+unsafe reorgs. The software and ops have been hardened after running it in production for some time.
```
protolambda, line 60: It would be worth clarifying that going from one failed batch to another different batch at the same L2 block height can be considered a form of "equivocation" (especially since it can be invalidated by a change of message validity on another chain), something that negatively impacts a lot of possible protocol designs, and is hence worth removing, despite the drawback in unsafe-chain recovery here.

sebastianst: I agree. Also, the actual current Go op-batcher implementation wouldn't resend a batch for the same block height if that batch failed to import during derivation. Only a batcher restart would cause this. I'm also not aware of a single case since the launch of Bedrock where an invalid batch was produced that then got replaces by a valid one for the same L2 block height.

protocol/invalid-block-progression.md
```diff
@@ -0,0 +1,60 @@
...
+The purpose of this document is to progress a design originally proposed by @protolambda and contributed to
+by Hamdi [here](https://github.com/ethereum-optimism/specs/pull/88). The goal is to reduce resource usage
+by the derivation pipeline when invalid blocks are proposed. This is helpful for ensuring that the fault
+proof program cannot be denial of service attacked as well as reducing the latency of reorgs, which is
+especially important in the context of interop.
+
+# Summary
+
+New derivation pipeline invariants are introduced that are meant to reduce large resets of the derivation pipeline
+when handling invalid batches.
```
sebastianst, line 12: It would be nice to more explicitly explain (probably below, not here) why the new rules avoid larger resets. This could include explaining what the _current_ behavior is under the preconditions of the new rules.

sebastianst (commented):
Thanks for the proposal! I'm supportive and the way I think about this is that the current rules allow to _resubmit_ a faulty batch which feels like a defensive mechanism to recover from a faulty sequencer (implementation). However, we've now established enough trust into our implementation that we can take this approach instead. After all, faulty batches shouldn't happen with a correct implementation.

* are there any other places in the DP where we also want to change the protocol to just derive a deposit-only block instead of causing a pipeline reset?

protocol/invalid-block-progression.md
sebastianst: I guess we could just say that we apply the rule for every singular batch reconstructed from the span batch? This should already give us the behavior that we need for interop, IIUC.

An alternative would be to drop the whole span batch, but I don't see an advantage from this over going over each singular batch case by case.

mslipper:
I believe that this is already an in-flight project, so I'll be merging this in.

mslipper (approved):

PR #21, opened 2024-05-22
Changes to Monorepo and Contract Versionings

mds1:
**Description**

This design doc proposes a new smart contracts versioning scheme to improve clarity and more closely align with typical semantic versioning practices.

### DIFF
--- a/protocol/smart-contract-versioning.md
+++ b/protocol/smart-contract-versioning.md
@@ -9,54 +9,86 @@ The monorepo releases will also follow semver, with the version bump determined 
 
 # Problem Statement + Context
 
-There are two facets to contract versioning: the versioning of monorepo releases, and the versioning of individual contracts.
+There are two aspects to contract versioning: the versioning of monorepo releases, and the versioning of individual contracts.
 
-- **Monorepo Releases**:  The current versioning scheme of monorepo contract releases does not follow a true semver. For example, the Fault Proofs release introduced breaking changes by changing the ABI, and we are releasing this as a minor version bump from `op-contracts/v1.3.0` to `op-contracts/v1.4.0`. This was not an intentional decision. We likely slid into this path because node releases cannot bump the major version (for go-specific reasons), and our current [contracts release process](https://www.notion.so/oplabs/Minimal-Contracts-Release-Process-ef8e91b1b13944e7a7f632a824895764?pvs=4) does not specify how to determine the version bump for an upcoming monorepo release.
+- **Monorepo Releases**:  The current versioning scheme of monorepo contract releases does not follow a true semver. For example, the Fault Proofs release introduced breaking changes (e.g. by changing the ABI of some contracts), and we are releasing this as a minor version bump from `op-contracts/v1.3.0` to `op-contracts/v1.4.0`. There was not a strong rationale for this—we likely default to a minor bump for consistency with how node releases bump their versions (they cannot bump the major version for go-specific reasons), and our current [contracts release process](https://www.notion.so/oplabs/Minimal-Contracts-Release-Process-ef8e91b1b13944e7a7f632a824895764?pvs=4) does not specify how to determine the version bump for an upcoming monorepo release.
 - **Individual Contracts**: The current versioning scheme of individual contracts is also not a true semver, but in a different way. We do bump major and minor versions appropriately, but we don't distinguish when a contract is in beta, release candidate, or feature status. This makes it difficult to determine the stability and compatibility of a given contract version.
 
-These have resulted in various issues, such as chains deploying in-development versions of contracts.
+These have resulted in various issues, such as users deploying in-development versions of contracts.
 
-Unlike typical software development, OP Stack chains cannot rely on "compatible semvers" to compose the chain of various contracts.
+Unlike typical software development, OP Stack chains should not rely on "compatible semvers" to compose the chain of various contract at different versions.
 Instead, a given monorepo contracts release must be defined by an exact set of contract versions, not a range of semver-compatible versions.
 
 # Alternatives Considered
 
-1. Keep the current versioning scheme: This alternative would maintain the status quo but would not address the issues of clarity, consistency, and alignment with semver practices.
+1. **Keep the current versioning schemes**. Although we are now familiar with the existing versioning schemes, they are not ideal for the reasons given above.
 
-2. Use semver for contracts but not for monorepo releases: This alternative would improve the versioning of individual contracts but would not provide a clear mapping between monorepo releases and contract versions.
+2. **Keep with the current contract versioning scheme but improve the monorepo release versioning scheme**: Because chains should only be deploying their contracts from a tag, we arguably do not need to be as granular with the versioning of individual contracts. However, the existing scheme is confusing because it is nonstandard, and since we are already using a form of semver on the contracts, it is preferable to improve that than to continue with the nonstandard approach. It is also simpler to have a similar versioning scheme for both contracts and monorepo releases instead of two different approaches.
 
 # Proposed Solution
 
 The proposed solution is to more closely adopt a true semver format for both individual contracts and monorepo releases.
+There are five parts to the solution:
 
-We preserve the existing rules for when to bump major, minor, and patch versions in individual contracts:
+- [Semver Rules](#semver-rules): Preserves the existing rules for when to bump major, minor, and patch versions in individual contracts.
+- [Individual Contract Versioning](#individual-contract-versioning): Introduces a new versioning scheme for individual contracts that includes beta, release candidate, and feature tags.
+- [Monorepo Contracts Release Versioning](#monorepo-contracts-release-versioning): Introduces a new versioning scheme for monorepo releases that follows semver.
+- [Release Process](#release-process): Describes the process for deploy contracts and proposing a release to governance.
+- [Changelog](#changelog): Describes how we will maintain a CHANGELOG for contract releases.
 
-- `patch` releases are to be used only for changes that do NOT modify contract bytecode (such as updating comments).
-- `minor` releases are to be used for changes that modify bytecode OR changes that expand the contract ABI provided that these changes do NOT break the existing interface.
-- `major` releases are to be used for changes that break the existing contract interface OR changes that modify the security model of a contract.
+## Semver Rules
 
-Versioning for individual contracts:
+We preserve the [existing rules](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/STYLE_GUIDE.md#versioning) for when to bump major, minor, and patch versions in individual contracts:
 
-- A contract is only `X.Y.Z` on `develop` if it's been governance approved. If it's `X.Y.Z` before that, it must be on a branch. More on this below.
+> - `patch` releases are to be used only for changes that do NOT modify contract bytecode (such as updating comments).
+> - `minor` releases are to be used for changes that modify bytecode OR changes that expand the contract ABI provided that these changes do NOT break the existing interface.
+> - `major` releases are to be used for changes that break the existing contract interface OR changes that modify the security model of a contract.
+
+Note that bumping the patch version does change the bytecode, so we carve out an exception for that and will update the style guide to reflect this.
+
+## Individual Contract Versioning
+
+Versioning for individual contracts works as follows:
+
+- A contract is only `X.Y.Z` on `develop` if it has been governance approved. If it's `X.Y.Z` before that, it must be on a branch. More on this below.
 - For contracts undergoing development, a `-beta.n` identifier must be appended to the version number.
 - For contracts in a release candidate state, an `-rc.n` identifier must be appended to the version number.
-- For contracts with feature-specific changes, a `+feature` identifier must be appended to the version number. See [ethereum-optimism/design-docs#8](https://github.com/ethereum-optimism/design-docs/pull/8) to learn more.
+- For contracts with feature-specific changes, a `+feature-name` identifier must be appended to the version number. See the [Smart Contract Feature Development](https://github.com/ethereum-optimism/design-docs/blob/main/smart-contract-feature-development.md) design doc to learn more.
 - When making changes to a contract, always bump to the lowest possible version based on the specific change you are making. We do not want to e.g. optimistically bump to a major version, because protocol sequencing may change unexpectedly. Use these examples to know how to bump the version:
-  - Example 1: A contract is currently on `1.2.3`. We don't yet know when the next release of this contract will be. However, you are simply fixing typos in comments so you bump the version to `1.2.4-beta.1`.
-    - The next PR that makes in that same contract but makes no functional changes, so it bumps the version to `1.2.4-beta.2`.
-    - The last PR introduces a breaking change, which bumps the version from `1.2.4-beta.2` to `2.0.0-beta.1`.
-  - Example 2: A contract is currently on `2.4.7`. We know the next release of this contract will be a breaking change. Regardless, as you start development by fixing typos in comments, bump the version to `2.4.8-beta.1`. This is because we may end up putting out a release before the breaking change is added.
+  - Example 1: A contract is currently on `1.2.3`.
+    - We don't yet know when the next release of this contract will be. However, you are simply fixing typos in comments so you bump the version to `1.2.4-beta.1`.
+    - The next PR made to that same contract clarifies some comments, so it bumps the version to `1.2.4-beta.2`.
+    - The last PR introduces a breaking change, which bumps the version from `1.2.4-beta.2` to `2.0.0-beta.1`. A `1.2.4-rc.1` and `1.2.4` version both never exist.
+  - Example 2: A contract is currently on `2.4.7`.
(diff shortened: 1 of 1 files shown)


protocol/smart-contract-versioning.md
```diff
@@ -0,0 +1,86 @@
...
+  - Example 2: The monorepo is at `op-contracts/v1.5.1`. Legacy `ALL_CAPS()` getter methods are removed from a contract, causing that contract to bump the major version. The next monorepo release will be `op-contracts/v2.0.0`.
+- A monorepo contracts release must map to an exact set of contract semvers, and this mapping must be defined in the contract release notes which are the source of truth. See [`op-contracts/v1.4.0-rc.4`](https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0-rc.4) for an example of what release notes should look like.
+
+The above gets us most of the way there, and there is one problem not yet addressed: When a release is proposed to governance, we propose a commit hash, and often those proposed contracts are already deployed to mainnet.
+For example, the [Fault Proofs governance proposal](https://gov.optimism.io/t/upgrade-proposal-fault-proofs/8161) provides specific addresses that will be used.
+This is ideal and we should continue to do that, but we need to fit this in to the above scheme.
+
+To accommodate this, once ready for governance approval the release flow is:
```
mds1, line 65: To me this is the weakest part of the proposal and I would to get some thoughts on feedback on how to handle this

protocol/smart-contract-versioning.md
```diff
@@ -0,0 +1,86 @@
...
+
+- `patch` releases are to be used only for changes that do NOT modify contract bytecode (such as updating comments).
+- `minor` releases are to be used for changes that modify bytecode OR changes that expand the contract ABI provided that these changes do NOT break the existing interface.
+- `major` releases are to be used for changes that break the existing contract interface OR changes that modify the security model of a contract.
+
+Versioning for individual contracts:
+
+- A contract is only `X.Y.Z` on `develop` if it's been governance approved. If it's `X.Y.Z` before that, it must be on a branch. More on this below.
+- For contracts undergoing development, a `-beta.n` identifier must be appended to the version number.
+- For contracts in a release candidate state, an `-rc.n` identifier must be appended to the version number.
```
maurelian, line 42: Does this prevent us from 'releasing the lock' on `develop` to other projects once a tag has been submitted as a governance proposal?
Or is the plan to remove the `-rc.n` suffix on a separate branch?

protocol/smart-contract-versioning.md
```diff
@@ -0,0 +1,86 @@
...
+- For contracts with feature-specific changes, a `+feature` identifier must be appended to the version number. See [ethereum-optimism/design-docs#8](https://github.com/ethereum-optimism/design-docs/pull/8) to learn more.
```
maurelian, line 43: Will there thus be `X.Y.Z-beta.n+feature-abc` tagged contracts?

protocol/smart-contract-versioning.md
```diff
@@ -0,0 +1,86 @@
...
+- A monorepo contracts release must map to an exact set of contract semvers, and this mapping must be defined in the contract release notes which are the source of truth. See [`op-contracts/v1.4.0-rc.4`](https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0-rc.4) for an example of what release notes should look like.
+
+The above gets us most of the way there, and there is one problem not yet addressed: When a release is proposed to governance, we propose a commit hash, and often those proposed contracts are already deployed to mainnet.
+For example, the [Fault Proofs governance proposal](https://gov.optimism.io/t/upgrade-proposal-fault-proofs/8161) provides specific addresses that will be used.
+This is ideal and we should continue to do that, but we need to fit this in to the above scheme.
+
+To accommodate this, once ready for governance approval the release flow is:
+
+- On the `develop` branch, bump the version of all contracts to their respective `X.Y.Z-rc.n`. The `X.Y.Z` here refers to the contract-specific versions so differs per-contract, whereas the `.rc.n` is the same across all contracts.
```
maurelian, line 67: Presumably the contracts have a `-beta.n+feature-abc` suffix that should be removed at this point?

protocol/smart-contract-versioning.md
mds1: The workflow would be like this:
- A PR into develop which replaces the `-beta.n` suffix with the `-rc.n` suffix for all contracts in your release. For this release, this is the last commit to develop.
  - At this point, the lock on develop is released.
- Create a branch from that commit called `proposal/op-contracts/vX.Y.Z`.
- Open a PR into that branch to remove all the `-rc.n` suffixes. Once merged, this commit is where contracts are deployed, and this commit is tagged as `op-contracts/vX.Y.Z-rc.1` for the governance proposal
- Subsequent changes only bump the `rc.n` identifier both at the contract and monorepo release level (possible exception: a change is at a higher version level than prior bumps, e.g. if this rc only bumped the minor version but now we introduce a breaking change for `rc.2` we need to bump the major version. But ideally once we're at an rc this should not be happening)
- Once the governance proposal passes, create the `op-contracts/vX.Y.Z` release from the latest `op-contracts/vX.Y.Z-rc.n` release, then merge the branch into develop

protocol/smart-contract-versioning.md
```diff
@@ -0,0 +1,86 @@
...
+- Branch off of `develop` and create a branch name `proposal/op-contracts/vX.Y.Z`. Here, `X.Y.Z` is the new version of the monorepo release.
+- On this branch, create a new commit that removes the `-rc.n` from all contracts. This is the commit hash that will be used to deploy the contracts and proposed to governance.
+- Once governance approves the proposal, merge the proposal branch into `develop` and bump the version of all contracts to their respective `X.Y.Z`. When there are conflicts, resolve them in favor of the `develop` branch. In other words:
+  - If a contract has not had any changes since the proposal, the version should be bumped to `X.Y.Z`.
+  - If a contract has had changes since the proposal, the version should be stay as whatever is the latest version on `develop`.
```
maurelian, line 72: I'm having trouble understanding this section. Specifically how to choose the `X.Y.Z`.

protocol/smart-contract-versioning.md
mds1: Sorry, `+feature` was meant to indicate `feature` as the variable name, maybe `+FEATURE-NAME` would be better? So in your example with a feature named `abc` it would be `X.Y.Z-beta.n+abc`

protocol/smart-contract-versioning.md
```diff
@@ -0,0 +1,86 @@
...
+
+Lastly, we will start maintaining a CHANGELOG for contract releases.
+The next expected release version will have a tracking issue that documents the new versions of each contract that will be included in the release, along with the links to the PRs that made the changes.
+Every contracts PR should have an accompanying changelog entry in the tracking issue once it is merged.
+See [ethereum-optimism/optimism#10592](https://github.com/ethereum-optimism/optimism/issues/10592) for an example of what this tracking issue should look like.
```
maurelian, line 77: Nit: that PR's title has a version number in it. It may actually be heard to predict the final version number once semver is driven by code changes rather than gov cycles. 
Perhaps we should enumerate changelog PRs with their expected governance cycle number, or just "upgrade N".

protocol/smart-contract-versioning.md
```diff
@@ -0,0 +1,86 @@
...
+    - Once you start working on the breaking change, bump the version to `3.0.0-beta.1`.
+    - Continue to bump the beta version as you make changes. When the contract is ready for release, bump the version to `3.0.0-rc.1`.
+- New contracts start at `0.y.z` and only become `1.0.0` when they are ready for production.
+
+Versioning for monorepo releases:
+
+- Monorepo releases continue to follow the `op-contracts/vX.Y.Z` naming convention.
+- The version used for the next release is determined by the highest version bump of any individual contract in the release.
+  - Example 1: The monorepo is at `op-contracts/v1.5.0`. Various tech debt is cleaned up in contracts, but no features are added, so all contracts only bump the patch version. The next monorepo release will be `op-contracts/v1.5.1`.
+  - Example 2: The monorepo is at `op-contracts/v1.5.1`. Legacy `ALL_CAPS()` getter methods are removed from a contract, causing that contract to bump the major version. The next monorepo release will be `op-contracts/v2.0.0`.
```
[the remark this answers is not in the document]
mds1: > Intuitively, it doesn't make sense to me that removing a legacy method should cause the same version change as adding fault proofs.

Technically this is how it would work by following a strict semver, from https://semver.org/:
image

So we just need to be thoughtful about breaking changes and bundle them together when appropriate (which we should be doing regardless of this proposal and how we handle semvers)

protocol/smart-contract-versioning.md
mds1: Good callout, given the flow in https://github.com/ethereum-optimism/design-docs/pull/21#discussion_r1610298474, I wonder if they should be released on the branch that's created, or on develop

protocol/smart-contract-versioning.md
mds1: I will try to clarify with examples, and can move them into the doc if helpful:

- ContractA is `1.2.3-rc1` on develop.
- We create the release branch, and the final ContractA version on that branch is `1.2.3`
- Governance approves the release so we want to merge the final versions of the contracts, including ContractA, back into develop.

There are two scenarios for this PR that merges the release branch into develop:

1. On develop, no changes have been made to ContractA. The PR therefore changes ContractA's version on develop from `1.2.3-rc.1` to `1.2.3`, and no other changes to ContractA occur.
2. On develop, breaking changes have been made to ContractA for a new feature, and it's currently versioned as `2.0.0-beta.3`. The PR should modify the source code of ContractA appropriately and bump the version to `2.0.0-beta.4`

protocol/smart-contract-versioning.md
mds1: Great point. The expected governance cycle can slip, and upgrade order can change, so something constant is ideal. Perhaps we just give it a descriptive name only? For example, "Contracts: Small improvements and Tech Debt Release Tracker"

protocol/smart-contract-versioning.md
```diff
@@ -0,0 +1,86 @@
...
+
+2. Use semver for contracts but not for monorepo releases: This alternative would improve the versioning of individual contracts but would not provide a clear mapping between monorepo releases and contract versions.
+
+# Proposed Solution
+
+The proposed solution is to more closely adopt a true semver format for both individual contracts and monorepo releases.
+
+We preserve the existing rules for when to bump major, minor, and patch versions in individual contracts:
+
+- `patch` releases are to be used only for changes that do NOT modify contract bytecode (such as updating comments).
```
tynes, line 34: This is so much better without the bindings, just fyi updating the version itself will modify the bytecode so this definition is impossible unless there is a clause where the version itself doesnt count as modifying the bytecode

tynes:
Is this solving specifically for L1 contracts? Any thoughts on L2 contracts?

protocol/smart-contract-versioning.md
```diff
@@ -0,0 +1,86 @@
...
+# Summary
+
+The proposed versioning scheme for contracts will follow a true semver format, with major, minor, and patch versions, as well as beta, release candidate, and feature tags.
+The monorepo releases will also follow semver, with the version bump determined by the highest version bump of any individual contract modified in the release.
+
+# Problem Statement + Context
+
+There are two facets to contract versioning: the versioning of monorepo releases, and the versioning of individual contracts.
+
+- **Monorepo Releases**:  The current versioning scheme of monorepo contract releases does not follow a true semver. For example, the Fault Proofs release introduced breaking changes by changing the ABI, and we are releasing this as a minor version bump from `op-contracts/v1.3.0` to `op-contracts/v1.4.0`. This was not an intentional decision. We likely slid into this path because node releases cannot bump the major version (for go-specific reasons), and our current [contracts release process](https://www.notion.so/oplabs/Minimal-Contracts-Release-Process-ef8e91b1b13944e7a7f632a824895764?pvs=4) does not specify how to determine the version bump for an upcoming monorepo release.
```
tynes, line 14: I notice that everything is `op-contracts`, what if we want to release L1 and L2 contracts differently? Or generally different subpackages of contracts. I would like to enshrine the idea of tying L2 predeploys to hardforks (related to [this](https://github.com/ethereum-optimism/specs/pull/194), proposal: the existing predeploys are all hardfork managed) so they would be on a different release schedule

protocol/smart-contract-versioning.md
mds1: Ah great point. Another idea is change the version from being immutable into a storage variable, that way initcode changes but runtime code does not, and we'd update this to say "that do NOT modify contract runtime bytecode". But I prefer your idea of adding an exception clause, because modifying initcode can still be a significant change

protocol/smart-contract-versioning.md
mds1: > Is this solving specifically for L1 contracts? Any thoughts on L2 contracts?

Moving https://github.com/ethereum-optimism/design-docs/pull/21#issuecomment-2125537843 here too since these are related.

---

Great callout, this is currently unspecified. Having separate `op-contracts-L1/vX.Y.Z` and `op-contracts-L2/vX.Y.Z` releases is a good direction and I'm onboard with that for this proposal.

For cases like custom gas token where both L1 and L2 contracts change, we can reference the corresponding L2 contracts release in the L1 release notes, and vice versa, and note that they must be used together.

I'm pretty sure all of our existing `op-contracts/` releases only released L1 contracts, so we'd also need a plan for the migration. We have a small set of tags, so we could:
- Create `op-contracts-L1/vX.Y.Z` tags at the same commit as all existing `op-contracts/vX.Y.Z`.
- Create releases for them, copy-pasting the existing release notes from the `op-contracts/vX.Y.Z` release into the `op-contracts-L1/vX.Y.Z`
- Update the `op-contracts/vX.Y.Z` to say "This release is deprecated in favor of the identical `op-contracts-L1/vX.Y.Z` release. See here to learn more about contract versioning"

protocol/smart-contract-versioning.md
mds1: This may actually be a strength. It provides a dedicated branch for each release, making it easy to deploy a patch or bug fix, regardless of other changes that may have occurred on `develop` since the release.

protocol/smart-contract-versioning.md
mds1: I have incorporated this info into the primary doc in 0f9f896

protocol/smart-contract-versioning.md
mds1: Changed to `+feature-name` in 0f9f896 to match https://github.com/ethereum-optimism/design-docs/blob/main/smart-contract-feature-development.md

protocol/smart-contract-versioning.md
mds1: Clarified in 0f9f896

protocol/smart-contract-versioning.md
mds1: Clarified this to use upgrade names in 0f9f896

protocol/smart-contract-versioning.md
mds1: Documented this exception in 0f9f896

protocol/smart-contract-versioning.md
```diff
@@ -0,0 +1,141 @@
...
+  - A subsequent release of the custom gas token feature that fixes bugs and introduces an additional breaking change would be `op-contracts/v2.0.0-beta.2`.
+  - This means `+feature-name` naming is not used for monorepo releases, only for individual contracts as described below.
+- A monorepo contracts release must map to an exact set of contract semvers, and this mapping must be defined in the contract release notes which are the source of truth. See [`op-contracts/v1.4.0-rc.4`](https://github.com/ethereum-optimism/optimism/releases/tag/op-contracts%2Fv1.4.0-rc.4) for an example of what release notes should look like.
+
+## Release Process
```
mds1, line 91: Proposed change from @ajsutton:
- Develop on `develop` with `-rc.1` version
- Create release branch which removes the `-rc` qualifier and create a tag for governance off that branch
- When we want to put out an `.rc.2`, first make the change with a PR into develop (assuming it's feasible, which it should be if following https://github.com/ethereum-optimism/design-docs/blob/main/smart-contract-feature-development.md
- Then merge the updated develop into the release branch.

This ensures develop stays up to date during the governance process, and mitigates the risk of forgetting to merge the release back into the develop branch

protocol/smart-contract-versioning.md
mds1: Agreed that we will:
- Not change release approach for this design doc. We'll continue doing `op-contracts` level and specify if it's L1 vs. L2. vs. some combination in the release notes
- A future effort will determine best way to handle L2 releases/upgrades in more detail

protocol/smart-contract-versioning.md
mds1: Incorporated in 3df49ea

protocol/smart-contract-versioning.md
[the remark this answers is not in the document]
maurelian: Good response from semver.org, makes sense.

tynes (approved):

PR #22, opened 2024-05-22
feat: added super usdc design doc

skeletor-spaceman:
**Description**

This PR adds the Super USDC design doc.
Super USDC proposes a new standard to deploy and manage Bridged USDC on OP-chains, it also focuses on facilitating a safe and swift migration to Native USDC.
This spec is also aimed to be *Interop* future-proof, achieving complete fungibility between bridged USDC tokens deployed on Interop-enabled op-chains.

**Tests**

N/A

**Additional context**

TODO: add reference to the relevant Github repo and public notion documents when they become public.

**Metadata**

N/A

### DIFF
--- a/ecopod/superUSDC.md
+++ b/ecopod/superUSDC.md
@@ -1,0 +1,129 @@
+# SuperUSDC design doc
+
+# Purpose
+
+> 📑 *This section is also sometimes called “Motivations” or “Goals”. It is fine to remove this section from the final document, but understanding the purpose of the doc when writing is very helpful.*
+
+This document proposes a solution for implementing a bridged USDC design that can be easily upgraded to native USDC. This solution leverages Superchain capabilities, including shared bridges, shared upgrades, and interoperability, providing an all-in-one solution.
+
+# Summary
+
+> 📑 *[Most (if not all) documents should have a summary. While the length will likely be proportional to the length of the full document, the summary should be as succinct as possible.]*
+
+
+**superUSDC**, is a modular and efficient refence implementation to expand USDC across the Superchain ecosystem. Leveraging the existing message-passing system and dedicated adapters, to ensure seamless access to native and bridged USDC liquidity, across all OP Chains.
+
+This implementation of superUSDC for the OP Stack allows for:
+
+- Utilizing the Bridged USDC standard without any modifications to it.
+- Enabling a chain operator to coordinate alongside OP an upgrade to native USDC.
+- Enabling a chain operator to upgrade to the superUSDC version of USDC (when available).
+- Moving between L1 and L2 via the OP Stack cross-domain messenger.
+- Moving between L2 and L2 through liquidity networks and eventually native interoperability (when available).
+- Deploy and setup SuperUSDC for new chains with a single transaction.
+- Fungibility of Bridged USDC between interop-chains.
+
+The goal is to provide a secure pathway for Circle's participation in any OP Chain ecosystem, facilitate [Cross-Chain Transfer Protocol (CCTP)](https://www.notion.so/CCTP-USDC-d464d146ce244502af760a9b9e4765f4?pvs=21) integration and enable future integration with native Superchain interoperability.
+
+# Problem Statement + Context
+
+
+> 📑 *Describe the specific problem that the document is seeking to address as well as information needed to understand the problem and design space. If more information is needed on the costs of the problem, this is a good place to that information.*
+
+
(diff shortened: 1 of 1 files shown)


zainbacchus:
LFG!

fede2442:
This is amazing, the Mode team is super excited. Thanks for all the work!

ecopod/superUSDC.md
```diff
@@ -0,0 +1,129 @@
...
+
+# Risks & Uncertainties
+
+> 📑 *An overview of what could go wrong. Also any open questions that need more work to resolve.*
+
+**1) OP Chain leaving the Superchain**
+
+In the edge case where an OP Chain chooses not to be part of the Superchain or unilaterally follows a different chain version than the one approved by governance and supported by other OP Chains, it could negatively impact asset compatibility and security, including superUSDC. This divergence can cause severe problems in the representation of the asset and necessitate a migration process, which remains unclear. Such an event could result in a painful experience for users.
+
+**2) Token path finding**
```
agileurbanite, line 122: This can be further exacerbated by line 118 in this doc

agileurbanite:
How can existing OP chains migrate to this new superUSDC standard? We should cover that in the doc. Any security implications of this migration etc?

skeletor-spaceman:
@agileurbanite 
> How can existing OP chains migrate to this new superUSDC standard? We should cover that in the doc. Any security implications of this migration etc?

This is a great question, we are currently researching different migration/standardization mechanisms for chains that already have a Bridge USDC token deployed.
We will add it as a link under a "Migration guidelines" when it's ready.

For now the 2 main paths are:
1) partial, total and/or bi-directional manual migration through a custom minter/burner for both Bridged USDC & SuperUSDC.
2) Bridged USDC ownership change, contract upgrade and deployment of SuperUSDC as a secondary entry-point contract.

would love to hear your thoughts and ideas too!

pegahcarter:
This standard should apply more than just to USDC as the requirement of native issuence conflicts with the vanilla OptimismMintableERC20 created upon bridging from mainnet.

USDC does, provide a standard requirement on how to fully issue their token natively, and is spec'd as modifying the bridge to give burning and pause rights for the token.

I believe the issue we have here is more based on the immutability and simplicity of OptimismMintableERC20 as all of these design challenges proposed would be circumvented if the token issuer on the native chain could define the message passed to the bridge before the first bridge, after which the custom token would be created on the destination chain with the message logic.

I know this doesn't really solve what you're proposing but ultimately there will need to be some external wrapper mint/burn contract to consolidate L2 token supply into 1 token address.  OptimismMintableERC20 does not currently provide this unless the token is bridged back to source, as the only permitted minter/burner is the bridge.

ecopod/superUSDC.md
```diff
@@ -0,0 +1,129 @@
...
+Getting the factory, upgrade-manager, bridge-adapter and the implementations of the L2 adapter and token on L1. This is necessary to securely deploy on existing and new chains:
+
+1. Anyone can deploy `SuperUSDCFactory`, `UpgradeManager`, the Proxy Adapter, and the initial implementation for the Adapter on L1. The `UpgradeManager` owns the Proxy Adapter on L1.
+2. Anyone set up the Trusted 3rd Party as the owner of the `UpgradeManager`, which can be done at deployment or via a change owner function.
+3. The Trusted 3rd Party whitelists `CrossDomainMessenger` addresses in the `UpgradeManager`. There are two possible flows:
+    1. Allow users to suggest addresses and let the Trusted 3rd Party approve them.
+    2. The Trusted 3rd Party sets addresses directly.
+4. Anyone can initialize a deployment to a chain with a whitelisted `CrossDomainMessenger` by calling the `SuperUSDCFactory` contract.
+    1. The `SuperUSDCFactory` calls the `CrossDomainMessenger` with the deployment calldata.
+    2. The calldata is used to deploy and set up the L2 contracts: Proxy Adapter, Implementation Adapter, Proxy `SuperUSDC`, and Implementation `SuperUSDC`.
```
pegahcarter, line 74: Is the `CrossDomainMessenger` able to deploy L2 contracts pre-deterministically?  Reason I'm asking is because if these tokes are the same address across op-stack, it would be simple to atomically bridge across supported Superchains.  If these tokens have different addresses, there would be an additional requirement to wrap/unwrap when jumping across Superchains.

skeletor-spaceman:
closing this PR in favor of: https://github.com/ethereum-optimism/design-docs/pull/43

PR #29, opened 2024-06-07
design-doc: advanced delegation

cairoeth:
**Description**

Adds a design doc for introducing advanced delegation in the [GovToken](https://github.com/voteagora/governor/blob/main/src/lib/OptimismToken.sol#L1869) via a hook-based implementation with [Alligator](https://github.com/voteagora/governor/blob/main/src/alligator/AlligatorOP_V5.sol).


tynes (approved):

governance/advanced-delegation.md
```diff
@@ -0,0 +1,52 @@
+# Purpose
+
+This design document aims to provide an overview and explanation of a hook-based integration between the [GovToken](https://github.com/voteagora/governor/blob/main/src/lib/OptimismToken.sol#L1863-L1901) and the [Alligator](https://github.com/voteagora/governor/blob/main/src/alligator/AlligatorOP_V5.sol) contract to enable advanced delegation functionality natively. The purpose is to understand how it will enable flexible and granular delegation of voting power among multiple delegatees, allowing for fractional delegation and subdelegations.
+
+# Summary
+
+The proposed feature will introduce a hook-based integration between the GovToken and the Alligator contract. By modifying the `_afterTokenTransfer` function in the GovToken to call the Alligator contract, the Alligator will be able to consume the hooks and update its delegation and checkpoint mappings accordingly. This approach simplifies the logic and state transitions related to delegation in the Alligator. Additionally, all delegation-related state will be shifted from the GovToken to the Alligator, allowing users and indexers to interact with a single contract for governance-related functionalities.
```
0xfuturistic, line 7: my big question is whether it makes more sense to call the Alligator not in `_afterTokenTransfer` but in `_beforeTokenTransfer`. this would allow any inheriting contract that overrides `_beforeTokenTransfer` to implement arbitrary logic with the Alligator's updates to delegation and checkpoint mappings already completed. seems like a better pattern. thoughts?

governance/advanced-delegation.md
```diff
@@ -0,0 +1,52 @@
...
+
+# Problem Statement + Context
+
+Advanced delegation, which includes partial delegation and subdelegations, is a superset of basic delegation. Partial delegation allows delegators to distribute their voting power among multiple delegatees in a fractional manner, while subdelegations enable delegatees to further delegate their received voting power to other delegatees based on predefined rules and constraints.
+
+Currently, delegators have to use the Alligator contract separately from the GovToken to access advanced delegation features. This separation of contracts complicates integration for indexers and front-end clients, as they need to track events and interact with both contracts. It also poses challenges for future interoperability plans. To streamline the delegation experience and reduce complexity, a hook-based integration between the GovToken and the Alligator is proposed, consolidating the delegation logic and state into a single contract.
```
0xfuturistic, line 13: it would be awesome to have more clarity on the last line of the paragraph
> consolidating the delegation logic and state into a single contract
which single contract?

governance/advanced-delegation.md
cairoeth: i think you mean for inheritng contracts to override `_afterTokenTransfer`, not `_beforeTokenTransfer` -- moving the voting power update to the before hook would go against the current standard implementation of ERC20 which has the voting power update on the after hook. could be interesting for a future update tho

0xfuturistic: ok I think the best argument is that it doesn't go against ERC20 (which offers both `_afterTokenTransfer` and `_beforeTokenTransfer`, but that it rather goes against `ERC20Votes`, which uses `_afterTokenTransfer`: https://github.com/voteagora/governor/blob/main/src/lib/OptimismToken.sol#L1880-L1886

PR #30, opened 2024-06-10
feat: sequencer-enforced invariants

sam-iamm:
**Description**

This is design doc to enable sequencers to enforce invariants on OP Stack chains related to inclusion and ordering.

### DIFF
--- a/protocol/sequencer-enforced-invariants.md
+++ b/protocol/sequencer-enforced-invariants.md
@@ -1,0 +1,119 @@
+# Purpose
+
+We ([Firewall](https://usefirewall.com/)) are building the Safe Sequencer for OP Stack chains. Safe sequencing uses exploit detection during block construction to exclude transactions from blocks that are identified as smart contract exploits. However, forced L1 → L2 messages can bypass the exploit detection of a safe sequencer and force include an exploit into a block.
+
+# Summary
+
+This is a design doc to enable (optional) sequencer-enforced invariants. We propose a design that uses L1 → L2 message validation hooks to delegate inclusion and ordering invariants to the sequencer without requiring changes to the derivation pipeline for handling deposits.
+
+# Problem Statement + Context
+
+The OP Stack does not allow computation in the sequencer to determine deposit inclusion since L1 → L2 messages are always force executed on L2. This forced execution can bypass additional inclusion and ordering invariants offloaded to the sequencer role.
+
+We believe delegating inclusion and ordering invariants to the sequencer can be a useful pattern, especially when heavy and complex computation is needed to enforce invariants, as seen in safe sequencing. This separation from the derivation pipeline and fault proof program of a rollup offers some nice benefits:
+
+- Sequencer-enforced invariants can be added or removed from rollups through the SystemConfig.
+- Sequencer-enforced invariants can be upgraded and managed at a pace appropriate for their own domain.
+- Sequencer-enforced invariants cannot create bugs in the rollup’s fault proof program that effect withdrawals.
+
+Sequencer-enforced invariants minimize complexity, yet still maintain rollup sovereignty. User sovereignty can also be maintained by allowing a fault or in(validity) proof of the delegated sequencer to [automatically trigger](https://x.com/sam___iamm/status/1798008895488827662) the SystemConfig to fall back to OP Stack defaults.
+
+### Related Problems
+
+MEV capture designs such as [Priority Is All You Need](https://www.paradigm.xyz/2024/06/priority-is-all-you-need) also use the sequencer role to enforce inclusion and ordering invariants and may require special handling of deposits. However, we note that sequencer-enforced invariants are more relevant for heavy and complex computation like exploit detection algorithms.
+
+# Proposed Solution
+
+## Hooks for configurable static analysis of L1 → L2 Messages
+
+Message validation hooks can be used within the bridging contracts to ensure that deposit messages will not violate sequencer-enforced invariants. These hooks are optional, turned off by default, and managed through the rollup’s SystemConfig.
+
+### Message Validation Hook Design
+
+Reference implementation: https://github.com/ethereum-optimism/optimism/pull/10791
+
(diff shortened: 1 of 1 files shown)


tynes:
> We ([Firewall](https://usefirewall.com/)) are building the Safe Sequencer for OP Stack chains. Safe sequencing uses exploit detection during block construction to exclude transactions from blocks that are identified as smart contract exploits. However, forced L1 → L2 messages can bypass the exploit detection of a safe sequencer and force include an exploit into a block.

This sort of logic should not live on chain. This sort of product could exist off chain as part of the block builder behind a PBS API as proposed in https://github.com/ethereum-optimism/specs/pull/116. Chain operators that want to opt into block builders that use exploit detection can opt into it

tynes:
If you can come up with a much more simple interface rather than adding logic into the bridge itself, it would be more palatable. It also adds uncertainty around what an exploit is, because its not objectively clear what that means. What is perfectly fine usage to one person is an exploit to another person.

protocol/sequencer-enforced-invariants.md
```diff
@@ -0,0 +1,119 @@
...
+    )
+        external
+        view
+        returns (bool);
+}
+```
+
+The `IL1MessageValidator` is static called within the `OptimismPortal` during the `_depositTransaction` function before the `TransactionDeposited` event is emitted. If the message validation fails, then the call reverts.
+
+The `IL2MessageValidator` is called within the `L2CrossDomainMessenger` during the `relayMessage` function before the external call of a given message. If message validation fails, then the message is added to the `failedMessages` mapping and the function returns.
```
tynes, line 73: You can get around this by withdrawing directly from the `L2ToL1MessagePasser`

tynes:
Can you build such a product on top of https://github.com/ethereum-optimism/specs/pull/27? This would be more aligned with the interfaces that we have been thinking about adding to the OP Stack

sam-iamm:
> > We ([Firewall](https://usefirewall.com/)) are building the Safe Sequencer for OP Stack chains. Safe sequencing uses exploit detection during block construction to exclude transactions from blocks that are identified as smart contract exploits. However, forced L1 → L2 messages can bypass the exploit detection of a safe sequencer and force include an exploit into a block.
> 
> This sort of logic should not live on chain. This sort of product could exist off chain as part of the block builder behind a PBS API as proposed in [ethereum-optimism/specs#116](https://github.com/ethereum-optimism/specs/pull/116). Chain operators that want to opt into block builders that use exploit detection can opt into it.
> 
> I see that you are trying to handle the deposits not being able to be censored as part of block building, but the complexity introduced by the implementation is very high. Not sure what the best solution to handle deposits is and if it makes sense as something we would want to include in the OP Stack long term as it basically introduces a way to censor arbitrary deposits.

Yes the product exists offchain and we are trying to handle the deposits problem because they can break any invariants enforced by the block builder. Appreciate your feedback a lot. Re complexity, I definitely see your concern. The reference implementation ended up being more complex than we had anticipated.

tynes:
This sort of feature needs to be very user research driven. Are chain operators asking for this? 

It also jeopardizes the ability to be a stage 1 rollup by allowing arbitrary censorship of deposits. The OP Stack is meant to commoditize stage 2 rollups when the multiproof is ready for production

sam-iamm:
> If you can come up with a much more simple interface rather than adding logic into the bridge itself, it would be more palatable. It also adds uncertainty around what an exploit is, because its not objectively clear what that means. What is perfectly fine usage to one person is an exploit to another person.

Will dive into the resource you linked below and look for an alternative, simpler interface.

Re exploit classification, ideally the interface doesn't embed any assumptions about what is an exploit and that is handled by an operator or algorithm chosen by the rollup.

protocol/sequencer-enforced-invariants.md
sam-iamm: It's not clear to me why that is the case, but it doesn't sound like we need dive deeper into the proposed solution anyway.

iamoptmstc:
> This sort of feature needs to be very user research driven. Are chain operators asking for this?
> 
> It also jeopardizes the ability to be a stage 1 rollup by allowing arbitrary censorship of deposits. The OP Stack is meant to commoditize stage 2 rollups when the multiproof is ready for production

Yes, there’s been two core reasons:

- (1) differentiation and enhancement of block-space - rollups are commoditized from a user POV, relative to each other they’re all cheap and fast

- (2) attraction of new liquidity - large capital allocators (whales, crypto funds, and eventually tradfi) can’t deploy capital today because of the smart contract risk. Chain operators are aware that if you mitigate that, your rollup becomes attractive to a new magnitude of assets — this is the core main driver for chain operator demand for this 

We’ve seen strong demand from rollups for exploit detection at the sequencer-level (e.g. [Concrete](https://www.concrete.xyz/), [Volmex](https://volmex.finance/), and [Mezzanine](https://www.mezzanine.xyz/). [Zircuit](https://www.zircuit.com/) is another proof-point, their unique differentiator is sequencer-level security and they have attracted $3.38B TVL within a few months. Zircuit uses a design similar to the “forced no-op deposits by the sequencer”, one of the alternative solutions summarized in the doc.

sam-iamm:
> This sort of feature needs to be very user research driven. Are chain operators asking for this?
> 
> It also jeopardizes the ability to be a stage 1 rollup by allowing arbitrary censorship of deposits. The OP Stack is meant to commoditize stage 2 rollups when the multiproof is ready for production

There are definitely some designs we’ll explore further that do not involve as much (or any) modification to the OptimismPortal and CrossDomainMessengers. Thanks for sharing this [spec](https://github.com/ethereum-optimism/specs/pull/27) as well.

Overall, this doc suggests a separation of concerns that allows a rollup to delegate inclusion and ordering invariants to the sequencer because it offers more flexibility. It lets a rollup easily add or remove sequencer-enforced invariants (in our case, ML detection algorithms) through the system config which enables it to enforce “good” censorship, invariants can be upgraded without upgrading the rollup itself, and there’s a clean separation of concerns for the rollup (e.g. invariant compute isn’t in the rollups FPP).

I think the core question for your feedback to keep in mind as we explore design trade-offs is: do we want the separation of concerns, or do we want to put all inclusion invariants (and their associated compute) within the derivation and execution client of a rollup’s full node and FPP?

tynes:
Closing this as we are not interested in censorship and this adds too much risk to being a stage 1+ rollup

PR #34, opened 2024-06-20
design-doc: supersim

fainashalts:
**Description**

Design document for supersim, a local multichain development environment for app developers building on the Superchain.

### DIFF
--- a/ecosystem/supersim.md
+++ b/ecosystem/supersim.md
@@ -120,26 +120,27 @@ flowchart TD
 
 ```mermaid
 flowchart TD
-   subgraph Op-Simulator Interface
-       direction LR
-       A[Instance Status] -->|Checks| A1[Status from Orchestrator]
-      
-       B[RPC Call Interception] -->|Intercepts| B1[eth_sendRawTransaction]
-       B -->|Intercepts| B2[eth_getLog]
-       B -->|Proxies to Anvil| B3[Other RPC Calls]
+  subgraph Op-Simulator Interface
+      direction LR
 
 
-       C[Event Subscription] -->|Subscribes to| C1[Events from Anvil Chains]
-       C1 -->|Determines Interception| B
-      
-       subgraph D[Invariant Checks]
-           D1[Check Interop Message Invariants] -->|If Pass| D2[Pass to Anvil Nodes]
-           D1 -->|If Fail| D3[Send Failure Message to Anvil Instance]
-       end
-      
-       B2 --> D1
-       D2 --> B3
-   end
+      C -->|any other RPC call| D2[call eth_sendRawTransaction \non Anvil]
+
+
+      C[ETH RPC API] -->|eth_sendRawTransaction| C1[Does TX have \n ExecutingMessage?]
+             C1 -->|no| D2
+
+
+      C1 -->|yes| D1
+
+
+    
+          D1[Check Interop Message Invariants] -->|If Pass|D2
+          D1 -->|If Fail| D3[Return TX failed]
+  
+    
+      D2 --> B3[Return result]
+  end
 ```
 
 ### Services
--- a/ecosystem/supersim.md
+++ b/ecosystem/supersim.md
@@ -67,9 +67,9 @@ The flow of this tool will differ depending on whether the user opts for **vanil
 2. **Forking mode**: 
     a. Orchestrator acquires L1 reference block to get the latest L2 height for forking using l2-fork-state-finder. 
     b. Orchestrator gets appropriate chain configs from the superchain registry using the chain-config-loader.
-    c. Orchestrator starts the requested chains at determined L2 block height
-    d. Orchestrator applies genesis to each chain using the genesis-applier
-    e. Orchestrator starts op-simulator and offers interface between anvil instances and op-simulator
+    c. Orchestrator starts the requested chains at determined L2 block height.
+    d. Orchestrator applies genesis to each chain using the genesis-applier.
+    e. Orchestrator starts op-simulator and offers interface between anvil instances and op-simulator.
     f. Orchestrator monitors status of running local anvil instances.
     
 
@@ -159,7 +159,8 @@ Given the application flow, the following services will be part of supersim:
 3. **Genesis-applier** : The `genesis-applier` takes genesis files output by the monorepo and applies them to anvil instances.
     a. The genesis files needed will differ depending on whether the user is in vanilla or forked mode, and forked mode is unlikely to need anything other than the interop contracts initially. 
     b. Can be multi-step: multiple genesis files can be applied idempotently, particularly in forking mode (when you fork a network that has bedrock contracts but not interop contracts, for instance).
-4. **Chain-config-loader**: Downloads the latest chain config from the superchain registry to allow the orchestrator to apply needed configs to anvil instances.
+    c. Genesis-applier should work as a standalone tool: if a developer is using Anvil outside the context of the orchestrator, they should be able to use genesis-applier to apply genesis state to their instance. In the context of supersim, the `orchestrator` will use the `genesis-applier` to apply the needed genesis state to the anvil instances it is running. 
+4. **Chain-config-loader**: Downloads the latest chain config from the superchain registry to allow the orchestrator to apply needed configs to anvil instances. Should be a standalone tool that works with the orchestrator but could be used directly on local testing nodes. 
 5. **L2-fork-state-finder**: From an L1 reference block, runs an algorithm to identify the latest L2 height derived from the L1 block for each network.
 6. **Op-simulator**: Proxy server in front of anvil instances that simulates op-stack services without the real derivation pipeline.
     a. Simulates deposits by listening to the `optimismPortal` and forwarding deposit txs via `eth_sendRawTransaction`.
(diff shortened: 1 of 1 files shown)


ecosystem/supersim.md
```diff
@@ -0,0 +1,319 @@
...
+
+      C[ETH RPC API] -->|eth_sendRawTransaction| C1[Does TX have \n ExecutingMessage?]
+             C1 -->|no| D2
+
+
+      C1 -->|yes| D1
+
+
+    
+          D1[Check Interop Message Invariants] -->|If Pass|D2
```
protolambda, line 149: Clear debug messages, so the user understands how/why their interop executing message is invalid, would be really useful.

ecosystem/supersim.md
```diff
@@ -0,0 +1,319 @@
...
+        * Initially a static interface.
+        * Medium-term, offer an API interface that the orchestrator exposes to make it easier to build new offchain services that hook into the runtime.
+
+2. **Anvil Instances**: Anvil instances are run by the orchestrator after initial prep work to allow them to mirror the Superchain interop environment.
+
+    * Should be run in `--optimism` mode so the extra fields in transactions are there.
+    * Orchestrator should also run an L1 instance, though L1 withdrawals won’t initially be supported.
+    * Eventually supersim can be extended to be used with HardHat and other local simulation tools as well.
+
+3. **Genesis-applier** : The `genesis-applier` takes genesis files output by the monorepo and applies them to anvil instances.
```
protolambda, line 175: The monorepo currently outputs `allocs-X.json` files, where `X` is a feature-set like a hardfork. E.g. `allocs-ecotone.json`. We might just want to output an additional `allocs-supersim.json` with any pre-configured things for the supersim, if more internal EVM things have to change.

Also note that some L2 contracts contain references to the L1, which would need to be updated, to make it work e.g. with different L1 deployments, chain-IDs, etc. Any improvements here, to make templating easier, like not using `immutable` but `storage` (so contract bytecode never ha to be touched) would be really helpful, but has not been prioritized yet (even though it's currently affecting the monorepo testing, the superchain-registry, and the supersim).

ecosystem/supersim.md
```diff
@@ -0,0 +1,319 @@
...
+
+    * The genesis files needed will differ depending on whether the user is in vanilla or forked mode, and forked mode is unlikely to need anything other than the interop contracts initially. 
+    * Can be multi-step: multiple genesis files can be applied idempotently, particularly in forking mode (when you fork a network that has bedrock contracts but not interop contracts, for instance).
+    * Genesis-applier should work as a standalone tool: if a developer is using Anvil outside the context of the orchestrator, they should be able to use genesis-applier to apply genesis state to their instance. In the context of supersim, the `orchestrator` will use the `genesis-applier` to apply the needed genesis state to the anvil instances it is running. 
+
+4. **Chain-config-loader**: Downloads the latest chain config from the superchain registry to allow the orchestrator to apply needed configs to anvil instances. Should be a standalone tool that works with the orchestrator but could be used directly on local testing nodes.
```
protolambda, line 181: If it's part of the other tooling, and written in Go or rust, then it should be relatively straight-forward to just import the superchain-registry as library. Like the op-node does.

ecosystem/supersim.md
```diff
@@ -0,0 +1,319 @@
...
+          D1 -->|If Fail| D3[Return TX failed]
+  
+    
+      D2 --> B3[Return result]
+  end
+```
+
+### Services
+
+Given the application flow, the following services will be part of supersim:
```
protolambda, line 159: Will this all still be packaged into the same executable? I don't see a user installing and running 6 different tools

ecosystem/supersim.md
fainashalts: yes! will add this to be an explicit requirement.

ecosystem/supersim.md
fainashalts: Yes 100%, the orchestrator spins these up but it is all "supersim."

ecosystem/supersim.md
fainashalts: Ooh the allocs approach is interesting! We have been talking about the best ways to identify what exactly should be used for genesis, and this could be a really helpful option. 

We'll look into the L1 references in the L2 contracts. Maybe we can help make the changes if it makes sense. I'd like to be able to use the contracts as is rather than have to edit them on our end, if at all possible.

ecosystem/supersim.md
```diff
@@ -0,0 +1,319 @@
...
+
+# High-Level Overview
+
+Supersim will allow developers to start multiple local Layer 2 chains with one command, and it will simulate the experience of interop via a command line interface. 
+
+# Proposed Solution
+
+Supersim will be a lightweight tool that simulates an interoperable Superchain environment locally. It will not require a complicated devnet setup and will be run using cli commands with configuration options that fall back to sensible defaults if they are not specified. Predefined state can be passed in at startup or the state can be forked based on specified block height.
+
+From a developer perspective, supersim will offer a command line interface through which to easily start up multiple local testnets that can be used interoperably via the [Optimism Interop Protocol](https://github.com/ethereum-optimism/specs/tree/main/specs/interop).
```
mslipper, line 20: How does it orchestrate each instance? Through Docker or something similar?

ecosystem/supersim.md
```diff
@@ -0,0 +1,319 @@
...
+    1. Orchestrator has preset chain configs and starts chains anew.
+    2. Orchestrator applies genesis state via the genesis-applier.
+    3. Orchestrator starts op-simulator and offers an interface between anvil instances and op-simulator.
+    4. Orchestrator monitors status of running local anvil instances.
+
+#### Forking mode 
+    
+    1. Orchestrator acquires L1 reference block to get the latest L2 height for forking using l2-fork-state-finder. 
+    2. Orchestrator gets appropriate chain configs from the superchain registry using the chain-config-loader.
+    3. Orchestrator starts the requested chains at determined L2 block height.
```
mslipper, line 73: Presumably this means that you'd need RPCs for all chains in your forked interop set, correct?

ecosystem/supersim.md
tynes: It would be useful to directly use the JSON allocs files to populate `anvil` instances in vanilla mode, would be surprised if they don't already support a CLI flag for this, where you can just pass a path to an allocs file and its applied. Then we can avoid the need to write another tool that converts an allocs file into JSON `anvil_setCode`/`anvil_setStorage` calls. Ideally we reuse as much of our existing tooling as possible here, so the deploy script with `--sig runWithStateDump()` so that the same genesis is created as in the production setup

ecosystem/supersim.md
```diff
@@ -0,0 +1,319 @@
...
+
+2. **Anvil Instances**: Anvil instances are run by the orchestrator after initial prep work to allow them to mirror the Superchain interop environment.
+
+    * Should be run in `--optimism` mode so the extra fields in transactions are there.
+    * Orchestrator should also run an L1 instance, though L1 withdrawals won’t initially be supported.
+    * Eventually supersim can be extended to be used with HardHat and other local simulation tools as well.
+
+3. **Genesis-applier** : The `genesis-applier` takes genesis files output by the monorepo and applies them to anvil instances.
+
+    * The genesis files needed will differ depending on whether the user is in vanilla or forked mode, and forked mode is unlikely to need anything other than the interop contracts initially.
```
tynes, line 177: Could create a small script to install the interop contracts in the short term

ecosystem/supersim.md
```diff
@@ -0,0 +1,319 @@
...
+    * Genesis-applier should work as a standalone tool: if a developer is using Anvil outside the context of the orchestrator, they should be able to use genesis-applier to apply genesis state to their instance. In the context of supersim, the `orchestrator` will use the `genesis-applier` to apply the needed genesis state to the anvil instances it is running. 
+
+4. **Chain-config-loader**: Downloads the latest chain config from the superchain registry to allow the orchestrator to apply needed configs to anvil instances. Should be a standalone tool that works with the orchestrator but could be used directly on local testing nodes. 
+
+5. **L2-fork-state-finder**: From an L1 reference block, runs an algorithm to identify the latest L2 height derived from the L1 block for each network.
+
+6. **Op-simulator**: Proxy server in front of anvil instances that simulates op-stack services without the real derivation pipeline.
+
+    * Simulates deposits by listening to the `optimismPortal` and forwarding deposit txs via `eth_sendRawTransaction`.
```
tynes, line 187: Screenshot 2024-07-17 at 4 13 39 PM

The foundry maintainers don't want to support deposit txs via RPC like this

ecosystem/supersim.md
jakim929: @tynes makes sense! we're using `anvil --init` to pass a pre-made `genesis.json`

in terms of generating the state dump, definitely alligned on trying to use the deploy script as is, but basically copied the logic in the python script in the monorepo https://github.com/ethereum-optimism/supersim/pull/53/files.

ecosystem/supersim.md
fainashalts: For forking, yes I don't see a way around that. We were planning to default to public RPCs where possible and allow users to override.

mslipper (approved):

PR #35, opened 2024-06-20
design-doc: liquidity migration

0xParticle:
**Description**
Added a design doc file discussing the possible approaches for the liquidity migration process for tokens locked in L1, deployed through the current and legacy `OptimismMintableERC20` Factories in L2 (also known as _legacy tokens_), to make them fungible and compatible with Superchain interoperability.

### DIFF
--- a/protocol/superc20/liquidity-migration.md
+++ b/protocol/superc20/liquidity-migration.md
@@ -8,53 +8,51 @@ This document proposes a solution for migrating the liquidity of tokens locked i
 
 ## Summary
 
-We must deploy new token representations because the current L2 tokens can't be upgraded to be compatible with Superchain interoperability. The main goal is to minimize any issues with liquidity migration.
+The current L2 tokens can't be upgraded to be compatible with Superchain interoperability. The main goal is to achieve compatibility whie minimizing liquidity migration issues.
 
-In this document, we will explore three approaches to create a representation that complies with `ISuperchainERC20`: Wrap, Convert, and Mirror. Among these, Mirror seems to be the best option as it doesn't force users to migrate to new token contracts.
+This desgin doc will explore multiple approaches to create a representation that complies with `ISuperchainERC20`. Among these, [Mirror](https://github.com/ethereum-optimism/design-docs/pull/36) seems to be the best option as it doesn't force users to migrate to new token contracts.
 
 ## Problem Statement + Context
 
-The current design of the `StandardBridge` allows for the transfer of standard ERC20 tokens between L1 and each L2 using a lock and mint mechanism. Token representations aren't fungible across different OP chains. We need a way to allow current liquidity to adopt the new `SuperchainERC20` standard to be Interop compatible.
-
-One possible approach is to create a new `SharedStandardBridge` from scratch. This would involve building a single contract connected to each `CrossDomainMessenger` of the Superchain, consolidating all liquidity. It could be implemented as a new contract for users to migrate their liquidity, or via a forced migration on L1 to the new contract. However, a manual migration could take months and a forced migration would introduce significant liability.
-
-Another approach is to make changes only on the L2 side to make the current `OptimismMintableERC20` tokens interoperable without many protocol-level changes. This could involve using the burn/mint permissions of the `L2StandardBridge` and introducing new token contracts to work alongside the existing legacy tokens. While this is not the only way to achieve interoperability, we believe it is the simplest, as the `L2StandardBridge` is upgradable and `OptimismMintableERC20` tokens are not.
+The current design of the `StandardBridge` allows for the transfer of standard ERC20 tokens between L1 and each L2 using a lock and mint mechanism. `OptimismMintableERC20` representations aren't fungible across different OP chains. These tokens need a way to adopt the new `SuperchainERC20` standard to be Interop compatible.
 
 ## Alternatives Considered
 
-We have considered multiple approaches:
+### L1 migration
 
-### Wrap
+One possible approach is to create a new `SharedStandardBridge` from scratch. This would involve building a single contract connected to each `CrossDomainMessenger` of the Superchain, consolidating all liquidity. It could be implemented as a new contract for users to migrate their liquidity, or via a forced migration on L1 to the new contract. However, a manual migration could take months and a forced migration would introduce significant liability.
+
+### L2 migration
+
+Another approach is to make changes only on the L2 side to make the current `OptimismMintableERC20` tokens interoperable without many protocol-level changes. This could involve using the burn/mint permissions of the `L2StandardBridge` and introducing new token contracts to work alongside the existing legacy tokens.
+
+**Wrap**
 
 Introduce a new contract that wraps one or multiple legacy tokens into an `ISuperchainERC20`.
 
-### Convert
+**Convert**
 
 Create a new `ISuperchainERC20` extension for the legacy tokens that can be exchanged with the legacy representation using the burn and mint rights of `L2StandardBridge` .
 
-### Mirror
+**Mirror**
 
 Create an `ISuperchainERC20` representation that behaves (with some exceptions) like the legacy "mirrored" token while allowing compatibility with Interop. Users will not realize that these are two distinct tokens, as the two contracts (legacy and Mirror) will act as secondary entry points to each other.
 
-## Tradeoffs and Comments
+A more detailed design doc can be found [here](https://github.com/ethereum-optimism/design-docs/pull/36).
 
-- The Convert and Wrap approaches require users to migrate to the new representation to support interoperability, while mirror representation requires no additional action from the user.
+## Tradeoffs and conclusion
+
+- The migration flow in L1 is very complex to achieve. An L2 migration seems like a simpler approach.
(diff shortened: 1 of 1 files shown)


tynes:
Can you include the pseudocode that you were working on for how the mirror token would work?

0xParticle:
> Can you include the pseudocode that you were working on for how the mirror token would work?
> 
> edit: ok i see that there is going to be a follow up doc for the mirror token itself

I just submitted a new PR with the pseudocode:
https://github.com/ethereum-optimism/design-docs/pull/36

We decided to split this into two design docs because we wanted to keep discussions separated, but we can merge them if you think that would make things more clear :)

protocol/superc20/liquidity-migration.md
```diff
@@ -0,0 +1,60 @@
...
+- When accepting multiple legacy tokens for the same interoperability representation, the Convert method is a bit cleaner than Wrap. This is because Wrap locks liquidity and differentiates each valid source, whereas Convert doesn't. Since legacy tokens are burned and minted, convert doesn't lock liquidity. Mirror can, in theory, support multiple representations, but we suggest sticking to a single one.
+
+### **Considerations**
+
+- Withdrawals to L1 will remain problematic when introducing interoperability without making changes to the `L1StandardBridge`. Users might attempt to withdraw an amount greater than what the `L1StandardBridge` can handle. In that case, several options can be considered:
+  - We can modify `L2StandardBridge` to conduct liquidity checks by comparing the total amount deposited with the total amount withdrawn and canceling the `withdrawal` if the amount is exceeded.
+  - Bridge UIs could monitor liquidity availability in both `L1StandardBridge` and choose the best path.
+
+## Proposed Solution: Mirror design
```
smartcontracts, line 52: I would like to see a more detailed outline of the mirror design and would want to open up a public discussion so that app teams can provide input to see if the mirror design will break anything critical for any major apps.

protocol/superc20/liquidity-migration.md
```diff
@@ -0,0 +1,60 @@
...
+
+This document proposes a solution for migrating the liquidity of tokens locked in L1, deployed through the current and legacy `OptimismMintableERC20` Factories in L2 (also known as _legacy tokens_), to make them fungible and compatible with Superchain interoperability.
+
+## Summary
+
+We must deploy new token representations because the current L2 tokens can't be upgraded to be compatible with Superchain interoperability. The main goal is to minimize any issues with liquidity migration.
+
+In this document, we will explore three approaches to create a representation that complies with `ISuperchainERC20`: Wrap, Convert, and Mirror. Among these, Mirror seems to be the best option as it doesn't force users to migrate to new token contracts.
+
+## Problem Statement + Context
```
smartcontracts, line 15: Does anyone have concrete data on the following things:

- TVL of immutable bridged tokens created with the `OptimismMintableERC20Factory`
- TVL of native L2 tokens
- TVL of bridged tokens that are upgradeable

This would be highly valuable IMO.

protocol/superc20/liquidity-migration.md
0xParticle: Absolutely! We just submitted a new PR with a more detailed outline:
https://github.com/ethereum-optimism/design-docs/pull/36

protocol/superc20/liquidity-migration.md
zainbacchus: TVL of bridged tokens that are upgradeable:
TVL of immutable bridged tokens created with the OptimismMintableERC20Factory: 
TVL of native L2 tokens: https://l2beat.com/scaling/projects/optimism/tvl-breakdown
TVL of L1StandardBridge: https://etherscan.io/address/0x99C9fc46f92E8a1c0deC1b1747d010903E884bE1#multichain-portfolio

MSilb7: One metapoint: Want to clarify when we mean TVL versus Assets Onchain

The L2Beat and Etherscan links to do show assets on chain, I can work on app TVL (assuming these are the assets that will actually move vs being stuck in CEXs / wallets).

---

Second, is the `bridged tokens that are upgradeable` / `immutable bridged tokens` distinction just if they're behind a proxy?

--

I also imagine we're talking about as many Superchain members as we can get, not just OP Mainnet right?

zainbacchus: @MSilb7 context of the question is: how much of assets bridged are upgradable / non-upgradable. If majority is non-upgradable then we need to prioritize mirroring, if its upgradable we might not need to pri mirroring for the initial release. 

Re members - starting with Base, OPM would be great as that's the majority of liquidity to inform the above. 

Proxy would work to determine upgradable!

MSilb7: Ok yeah, if you all have a good way to identify which tokens are proxies vs not, you can match that up against L2Beat.

I know USDC is a proxy as an example. We could ~try to approximate it like "did the token ever emit an upgrade event."

MSilb7: What contract(s) are the `OptimismMintableERC20Factory`?

I was looking at `0x4200000000000000000000000000000000000012` on OPM and Base, but I'm seeing some tokens on L2Beat that were minted on Base via `0xf23d369d7471bD9f6487E198723eEa023389f1d4`. Are the custom ERC20 factories, or are these custom ones not valid?

MSilb7: per [Base docs](https://docs.base.org/docs/base-contracts/#base-mainnet), seems like this is another token factory `0xF10122D428B4bc8A9d050D06a2037259b4c4B83B`

If chains are rolling their own token factories, we'll likely need a registry of this.

MSilb7: Looked at Assets Onchain for Base and OPM
- Proxy Tokens, Non-Proxy, Native (ETH) even about 25% each (~30% each excl. predeploys)
- Immutable token factory 8%
- OP (Predeploy) 14.7%

image


- Base rolled custom minting factories.
- "Unknown" = Token was deployed before regenesis, or does not have an onchain L2<>L1 mapping.
- Assets Onchain value sourced from L2Beat & Mapped to the L2 token address.
- Tokens created on OP Mainnet pre-regenesis were labeled as "Unknown," unless a proxy event was observed post-regenesis.

Data Issues:
- Some L1 tokens map to multiple L2 tokens by bridge events. In this case, we used amount of dex swaps as the indicator for which version of the token to treat as the adopted version.
- The immutable token factory events seem to be spoofable, which will cause token classification errors, if you go off of factory events (i.e. Native USDC appears to be [factory created](https://basescan.org/tx/0x04a19b13b48464557d394ab2cdcb7a578642a0e8d898ec4a632da92066b5acf4#eventlog), when it [is not](https://basescan.org/tx/0x8aa214f98bcf2984add809d10232135cccc4d6ab97d8477e66475d8bf68def34)). I manually overrode USDC, but there may be other bugs.
- ^ Request: **Can we make sure there's no ability for people to spoof token creations like this going forward?**

protocol/superc20/liquidity-migration.md
```diff
@@ -0,0 +1,60 @@
...
+
+One possible approach is to create a new `SharedStandardBridge` from scratch. This would involve building a single contract connected to each `CrossDomainMessenger` of the Superchain, consolidating all liquidity. It could be implemented as a new contract for users to migrate their liquidity, or via a forced migration on L1 to the new contract. However, a manual migration could take months and a forced migration would introduce significant liability.
+
+Another approach is to make changes only on the L2 side to make the current `OptimismMintableERC20` tokens interoperable without many protocol-level changes. This could involve using the burn/mint permissions of the `L2StandardBridge` and introducing new token contracts to work alongside the existing legacy tokens. While this is not the only way to achieve interoperability, we believe it is the simplest, as the `L2StandardBridge` is upgradable and `OptimismMintableERC20` tokens are not.
```
mds1, line 21: Should these paragraphs be moved to "Alternatives Considered"? They sound like designs, not part of the problem statement

0xParticle: Good point, yes. Will do

protocol/superc20/liquidity-migration.md
mds1, line 52: So this is basically a double entry point (DEP) token, where a user can interact with a token either from it's existing `OptimismMintableERC20` contract OR from the new `MirrorERC20` contract? If so my initial reaction is that this design makes me very nervous—DEP tokens have been the cause of at least two security vulnerabilities:
- Compound issue: [1](https://medium.com/chainsecurity/trueusd-compound-vulnerability-bc5b696d29e2), [2](https://blog.openzeppelin.com/compound-tusd-integration-issue-retrospective)
- Balancer issue: [1](https://forum.balancer.fi/t/medium-severity-bug-found/3161), [2](https://www.chainsecurity.com/blog/denial-of-service-attacks-in-defi-the-balancer-synthetix-case)

I suspect part of the reason there are "only" two instances of vulnerabilities from DEP tokens is because there aren't many DEP tokens that protocols integrate with. Now we are making every bridged token a DEP token—how many protocols would this introduce vulnerabilities for?

It's also possible I have misunderstood, and that there's something unique about this design that gives us confidence we are not introducing any vulnerabilities?

tynes (approved):

PR #36, opened 2024-06-20
design-doc: mirror standard

0xParticle:
**Description**

This document presents the `SuperchainERC20` Mirror Standard, which enables existing tokens deployed through `OptimismMintableERC20` Factory in L2 (also called legacy tokens) to be interoperable without any liquidity migration process.

### DIFF
--- a/protocol/superc20/mirror-standard.md
+++ b/protocol/superc20/mirror-standard.md
@@ -1,0 +1,114 @@
+# Mirror Standard Design Doc
+
+## Purpose
+
+_The work present below is part of the [Interoperability project](https://github.com/ethereum-optimism/optimism/issues/10899)._
+
+This document presents the `SuperchainERC20` Mirror Standard, which enables existing tokens deployed through `OptimismMintableERC20` Factory in L2 (also called legacy tokens) to be interoperable without any liquidity migration process.
+
+## Summary
+
+We discussed different approaches to making existing liquidity interoperable in the [Migrated Liquidity Design Doc](https://github.com/ethereum-optimism/design-docs/pull/35). We introduced the Mirror standard and highlighted it as the simplest solution, as it requires no liquidity migration at all.
+
+The Mirror approach deploys a new token contract that follows the `ISuperchainERC20` specifications and will work as a double-entry point for an appointed `OptimismMintableERC20`. We must modify the `L2StandardBridge` to enable the Mirror contract to burn and mint the underlying `OptimismMintableERC20` to implement the Mirror.
+
+## Proposed Design
+
+The proposed design includes a specification on implementing a double-entry point to mirror an existing `OptimismMintableERC20`, as well as approaches to the Factory and `L2StandardBridge`.
+
+### OptimismMintableERC20 proxiable functions
+
+The following functions will be implemented as simple proxies, that is, by having a mirror call to the appointed `OptimismMintableERC20` token:
+
+- `totalSupply()`
+- `balanceOf(account)`
+- `transfer(recipient, amount)`
+- `transferFrom(sender, recipient, amount)`
+
+### Local functions
+
+Approvals functions will be implemented in a conventional way, specific to the contract, and not as mirror calls:
+
+- `approve(spender, amount)`
+- `allowance(owner, spender)`
+
+### Code sample
+
+Below, we show what the implementation of the mirror contract would look like.
+
+```solidity
+contract Mirror is Superc20 {
+  IERC20 public immutable LEGACY_TOKEN;
+  IL2StandardBridge public constant L2_STANDARD_BRIDGE = IL2StandardBridge(0x4200000000000000000000000000000000000010);
+
+  constructor(IERC20 _legacyToken) Superc20(_name, _symbol, _decimals) {
+    LEGACY_TOKEN = _legacyToken;
+    _name = sting.concat("Mirror",_legacyToken.name());
+    _symbol = string.concat("m:",_legacyToken.symbol());
+    _decimals = _legacyToken.decimals();
+  }
+
+  function totalSupply() public view override returns (uint256) {
+    return LEGACY_TOKEN.totalSupply();
+  }
+
+  function balanceOf(address account) public view override returns (uint256) {
+    return LEGACY_TOKEN.balanceOf(account);
+  }
+
+  function transfer(address recipient, uint256 amount) public override returns (bool) {
+    L2_STANDARD_BRIDGE.mirrorBurn(msg.sender, amount);
+    L2_STANDARD_BRIDGE.mirrorMint(recipient, amount);
+
+    emit Transfer(msg.sender, recipient, amount);
+    return true;
+  }
+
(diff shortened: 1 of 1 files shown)


protocol/superc20/mirror-standard.md
```diff
@@ -0,0 +1,114 @@
...
+Deploying mirror versions will require introducing a new predeploy for`SuperchainERC20` factory. This factory would accomplish the following:
+
+- Designed to enable same-address deployments across OP Chains.
+- Address predictability for each token, we should know the corresponding Mirror.
+- Simple Factory code.
+- It may serve to create both `SuperchainERC20` standard tokens and mirror versions.
+
+### Upgrading `L2StandardBridge`
+
+We must update the `L2StandardBridge` to allow the mirror contract to mint and burn the mirrored legacy token. We can do this by adding a `mirrorMint()` and a `mirrorBurn()` functions that check the mirror is valid for the corresponding legacy token.
```
tynes, line 107: How do we know this is secure? It just checks the L1 token matches? What is the exact check this does?

0xng: The approach we were thinking to allow for proper validation here is to have the `L2StandardBridge` check if the address of the `msg.sender` (the mirror) corresponds to the correct mirror address deployed by the `Mirror` factory. In the `Mirror` factory design, each token is coupled with a single `Mirror`, so this can be done. I added a code example here: [Limited Mirror Factory](https://github.com/ethereum-optimism/design-docs/pull/37).

Here's an example code snippet:
```solidity
function mirrorMint(address _token, uint256 _amount) external {
	address _mirror = MIRROR_FACTORY.mirrors(_token);
	if (_mirror != msg.sender) revert();
	///...mint logic
}
```

protocol/superc20/mirror-standard.md
```diff
@@ -0,0 +1,114 @@
...
+
+- `approve(spender, amount)`
+- `allowance(owner, spender)`
+
+### Code sample
+
+Below, we show what the implementation of the mirror contract would look like.
+
+```solidity
+contract Mirror is Superc20 {
```
tynes, line 40: Since this references the design in https://github.com/ethereum-optimism/specs/pull/71, can we assume general consensus on the design in that PR?

protocol/superc20/mirror-standard.md
```diff
@@ -0,0 +1,114 @@
...
+
+  function totalSupply() public view override returns (uint256) {
+    return LEGACY_TOKEN.totalSupply();
+  }
+
+  function balanceOf(address account) public view override returns (uint256) {
+    return LEGACY_TOKEN.balanceOf(account);
+  }
+
+  function transfer(address recipient, uint256 amount) public override returns (bool) {
```
tynes, line 59: It would be good to have the flow that shows a cross chain transfer, perhaps as a mermaid diagram. My current understanding, assuming the mirror token is already deployed on both domains:
- holding `OptimismMintableERC20Token`
- call mirror `transfer` to self
- call mirror superchain erc20 `sendERC20` per https://github.com/ethereum-optimism/specs/pull/71#discussion_r1631562710, this burns the token
- on remote domain, send the executing transaction that mints the token

protocol/superc20/mirror-standard.md
```diff
@@ -0,0 +1,114 @@
...
+- `balanceOf(account)`
+- `transfer(recipient, amount)`
+- `transferFrom(sender, recipient, amount)`
+
+### Local functions
+
+Approvals functions will be implemented in a conventional way, specific to the contract, and not as mirror calls:
+
+- `approve(spender, amount)`
+- `allowance(owner, spender)`
```
mds1, line 33: Can you provide rationale for why the proxiable vs. local function split is the way it is? i.e. why is `approve` and `allowance` local but the rest proxiable? Also what about `name`, `symbol`, and `decimals`? From the code sample below it seems those are also proxied

protocol/superc20/mirror-standard.md
mds1: > an we assume general consensus on the design in that PR?

I was going to ask a similar question, and I think the answer is no we cannot assume general consensus, since that PR is a bit stale with a lot of open comments. It feels like that PR is a prerequisite for this one + interoperable ether (https://github.com/ethereum-optimism/design-docs/pull/25), so I would suggest reprioritizing the SuperchainERC20 standard so we can finalize that first

0xParticle: Yes. The mirror design will require a particular implementation for the cross-chain actions, but we can assume here no compatibility issues.

protocol/superc20/mirror-standard.md
mds1: +1 I would suggest removing code from this design doc, and replace it with diagrams showing the user flows and data flows. Per the [design doc template](https://github.com/ethereum-optimism/design-docs/blob/main/assets/design-doc-template.md):

> As a rule of thumb, including code snippets (except for defining an external API) is likely too low level.

protocol/superc20/mirror-standard.md
```diff
@@ -0,0 +1,114 @@
...
+
+    emit Transfer(_from, address(0), _amount);
+    emit Burn(_from, _amount);
+  }
+}
+```
+
+### Deployment
+
+Deploying mirror versions will require introducing a new predeploy for`SuperchainERC20` factory. This factory would accomplish the following:
```
mds1, line 98: What happens to the existing `OptimismMintableERC20Factory` on L1 and L2?

protocol/superc20/mirror-standard.md
0xParticle: yes, good catch. Will push an update today with more details on the rationale. You are right with `name`, `symbol` and `decimals`, will modify the design doc.

protocol/superc20/mirror-standard.md
```diff
@@ -0,0 +1,114 @@
...
+
+- Designed to enable same-address deployments across OP Chains.
+- Address predictability for each token, we should know the corresponding Mirror.
+- Simple Factory code.
+- It may serve to create both `SuperchainERC20` standard tokens and mirror versions.
```
mds1, line 103: It sounds like this is uncertain, we should finalize the design here before merging

0xParticle: The current [Factory design doc](https://github.com/ethereum-optimism/design-docs/pull/39) only covers Mirror, but we think it can be extended. We can discuss this.

0xParticle: So yes, agree we should cover and solve this before merging

protocol/superc20/mirror-standard.md
0xParticle: I just created some basic flows in Figma: https://www.figma.com/board/IXC6WnDaqrheplxbCZpw3l/%5BExternal%5D-Mirror-Standard?node-id=0-1&t=3tsA67ghaqocKaRv-1

will migrate these to Mermaid after discussion

tynes:
We will want to describe how to do withdrawals back to L1 using this standard

tynes:
Closing this in favor of the convert approach

PR #37, opened 2024-06-20
feat: adding limited space mirror factory

0xng:
**Description**

This document describes a possible design for a new Factory predeploy capable of deploying `Mirror` contracts for `OptimismMintableERC20Token`s. 

**Additional Context**
The space for this design is limited to only `OptimismMintableERC20Token`s.

### DIFF
--- a/protocol/superc20/mirror-factory-limited.md
+++ b/protocol/superc20/mirror-factory-limited.md
@@ -36,7 +36,7 @@ This gives us:
 - Simple factory code.
 
 > [!NOTE]
-> An important note is that L1 must also be able to deploy `Mirror` contracts. We don’t have control over the factory address on L1. Therefore, to ensure the same `Mirror` address for a given token is maintained across chains, the factory pre-deploy will not follow the `0x42.....N` standard, but it will have to match the factory’s deployment address on L1 in all chains.
+> 💡 An important note is that L1 must also be able to deploy `Mirror` contracts. We don’t have control over the factory address on L1. Therefore, to ensure the same `Mirror` address for a given token is maintained across chains, the factory pre-deploy will not follow the `0x42.....N` standard, but it will have to match the factory’s deployment address on L1 in all chains.
 
 The salt doesn't need to be more than the `OptimismMintableERC20Token` address. The `name` and `symbol` can be queried from the implementation code. The `Mirror` will take in the token it extends as an argument and the factory address as its admin to allow for upgrades.
 
--- a/protocol/superc20/mirror-factory-limited.md
+++ b/protocol/superc20/mirror-factory-limited.md
@@ -1,0 +1,103 @@
+# Summary
+
+*The work present below is part of the [Interoperability project](https://github.com/ethereum-optimism/optimism/issues/10899).*
+
+A factory to spin up `ISuperchainERC20` Mirrors of `OptimismMintableERC20Tokens` is required. This document aims to provide a design for such a factory.
+
+# Problem Statement + Context
+
+A vast amount of `OptimismMintableERC20Tokens` currently exists in the ecosystem. These tokens do not have interoperable properties. Therefore, they require a migration to become interoperable. The migration will be performed through Mirrors, as specified by [Mirror Standard](https://github.com/ethereum-optimism/design-docs/pull/36). These Mirrors must be easily deployed through a predefined factory to ensure all Mirrors share the same functionality and address across chains and also to verify their legitimacy.
+
+The factory must facilitate the creation of a single legitimate Mirror per `OptimistismMintableERC20Token` to avoid liquidity fragmentation and unnecessary complexity.
+
+# Alternatives Considered
+
+- Creation Methods:
+    - Using `CREATE`. Discarded due to nonce dependability.
+    - Using `CREATE3`. Discarded due to unnecessary complexity.
+- Upgradability
+    - Deploying the `Mirror` without a `Proxy` and simply updating the `mirrors`' mapping to point to a new address when the Mirror is upgraded was discarded. Maintaining the `Mirrors'` addresses across chains is crucial *at all times,* and this approach could cause synchronization issues where one `Mirror` was updated in a given chain but not in another, resulting in different mirror addresses for the same token.
+
+# Proposed Solution
+
+Limiting the scope to `OptimismMintableERC20Tokens` allows using `CREATE2` instead of `CREATE3`, which is more expensive and complex. This is because the variables CREATE2 uses to define the address are salt, the child's init code, and the sender's address, which provide us with all the needed properties.
+
+The `Mirror` needs to be upgradable. Therefore, a Proxy contract will take the `_token` as an argument and `delegatecall` to the `Mirror` implementation.
+
+> [!NOTE]
+>💡 From now on, when we say `Mirror`, we’re referring to the proxy, not the implementation.
+
+Using the `OptimismMintableERC20Token` address as the salt allows linking the `Mirror` to this specific token. The sender's address will be the factory. The factory will be a pre-deploy which must be maintained across chains. This means the same `Mirror` address for the given salt/token will be kept.
+
+This gives us:
+
+- Address predictability. For each token, we know the corresponding Mirror.
+- Consistent addresses across chains.
+- Simple factory code.
+
+> [!NOTE]
+> An important note is that L1 must also be able to deploy `Mirror` contracts. We don’t have control over the factory address on L1. Therefore, to ensure the same `Mirror` address for a given token is maintained across chains, the factory pre-deploy will not follow the `0x42.....N` standard, but it will have to match the factory’s deployment address on L1 in all chains.
+
+The salt doesn't need to be more than the `OptimismMintableERC20Token` address. The `name` and `symbol` can be queried from the implementation code. The `Mirror` will take in the token it extends as an argument and the factory address as its admin to allow for upgrades.
+
+The factory should be upgradable or have setters to allow the creation of `Mirrors` with new implementations. This means the factory should also allow old, already-deployed `Mirrors` to be updated to new ones. A function `upgradeMirror` is introduced to accomplish this.
+
+We can track versioning through numbers or the addresses of the implementations used.
+
+The mirror implementation contract should be deployed beforehand. The Proxy must properly match storage slots with the mirror implementation storage.
+
+This design allows the `StandardBridge` to call `mirrors(_token)` when a `Mirror` contract performs a `transfer` and validates the mirror belongs to the `_token`. As an illustrative example:
+
+```solidity
+function mirrorMint(_token, amount) external {
+	address _mirror = MIRROR_FACTORY.mirrors(_token);
+	if (_mirror != msg.sender) revert();
+	///...mint logic
+}
+
+```
+
+Example code:
+
+```solidity
+contract MirrorFactory {
+	string public constant MIRROR_VERSION = "1.0.0";
+	Mirror public constant MIRROR_IMPL = <PLACEHOLDER_ADDRESS>;
+
+	mapping(address token => address _mirror) public mirrors;
+	// this could be a mapping to IMPLs
+	mapping(address token => string version) public mirrorVersions;
+
+	function deployMirror(address _token) external returns (address) {
+		// Check if its `OptimismMintableERC20Token`?
+		if (mirrors[_token] != address(0)) revert();
+		Proxy _mirror = address(new Proxy{salt: keccak256(abi.encode(_token))}(address(this), _token));
+		_updateMirror(_mirror, _token);
+	}
+
+	function upgradeMirror(address _mirror) external {
+		if (mirror[token] == address(0)) revert();
+		if (mirrorVersions[mirror] == MIRROR_VERSION) revert();
+		_updateMirror(_mirror, _token);
+		//..........
+	}
+
+	function _updateMirror(address _mirror, address _token) internal {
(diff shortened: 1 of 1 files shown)


protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+- `Init Code`: The same proxy's init code must be used accross chains.
+
+The salt doesn't need to be more than the `OptimismMintableERC20Token` address. The `name` and `symbol`, and `decimals` can be fetched with the `token` address alone in the implementation. The factory will be the `Mirror`'s admin, and the proxy will have to either call `setStorage` immediately with the `token` at the time of deployment, or add a `TOKEN_SLOT` and update it on the constructor.
+
+The factory should be upgradable or have setters to allow the modification of the implementations for the current `Mirrors`. The deployment and update of already-deployed `Mirrors` to new implementations should be open to anyone.
+
+We can track versioning through numbers or the addresses of the implementations used.
+
+This design allows the `StandardBridge` or any contract that needs to check the legitimacy of a `Mirror` to call `MIRROR_FACTORY.getMirrorAddress()` and validate whether the calling user is a mirror that corresponds to a given `token`. As an illustrative example of a potential function of the `StandardBridge`:
```
tynes, line 31: We should be clear that this is only for the `L2StandardBridge` since the superchain erc20 standard is only for L2 to L2

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+```solidity
+contract MirrorFactory {
+	string public constant MIRROR_VERSION = "1.0.0";
+	Mirror public constant MIRROR_IMPL = <PLACEHOLDER_ADDRESS>;
+
+	function deployMirror(address _token) external returns (address) {
+		require(_isOptimismMintableERC20(_token), "not OptimismMintableERC20Token");
+
+		// Note: There's no _token slot in this proxy
+		Proxy _mirror = address(new L2ChugSplashProxy{salt: keccak256(abi.encode(_token))}(address(this), _token));
```
tynes, line 53: Perhaps we could just use `Proxy` rather than chugsplash proxy. There is always a balance between pseudocode and details, in practice we don't want to use chugsplash here. We should consider a [1967 beacon](https://eips.ethereum.org/EIPS/eip-1967) for this

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+function mirrorMint(address _token, uint256 _amount) external {
+	address _mirror = MIRROR_FACTORY.getMirrorAddress(_token);
+	if (_mirror != msg.sender) revert();
+	///...mint logic
+}
+
+```
+
+Example code:
+
```
tynes, line 43: small nit: using tabs here makes it a little harder to read

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+		require(_isOptimismMintableERC20(_token), "not OptimismMintableERC20Token");
+
+		// Note: There's no _token slot in this proxy
+		Proxy _mirror = address(new L2ChugSplashProxy{salt: keccak256(abi.encode(_token))}(address(this), _token));
+
+		// Note: This function does not exist on L2ChugSplashProxy.
+		_mirror.setImplementation(MIRROR_IMPL);
+	}
+
+	function upgradeMirror(address _mirror) external {
```
tynes, line 59: We should remove this function and think about a 1967 beacon based design for simplicity.

We don't yet have a predeploy system for 1967 proxies, but thinking of making a predeploy that looks like this:

```solidity
contract Mirror1967Beacon {
  function implementation() {
      return address(0x..).code;
  }
}
```

The above contract would be a predeploy and it would hardcode in an address that it returns code of. As part of a hardfork, we deploy smart contracts. We would deploy the implementation and we would be able to know the address ahead of time, see [here](https://github.com/ethereum-optimism/specs/blob/main/specs/fjord/derivation.md#gaspriceoracle-deployment) for previous work on how we can know the address ahead of time. We would be able to install the bytecode into the genesis for new chains

tynes: What this would allow us to do is to hardfork to update the `Mirror1967Beacon` predeploy and modify the hardcoded address that it returns

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+		// Note: This function does not exist on L2ChugSplashProxy.
+		_mirror.setImplementation(MIRROR_IMPL);
+	}
+
+	function _isOptimismMintableERC20(address _token) internal view returns (bool) {
+        return ERC165Checker.supportsInterface(_token, type(ILegacyMintableERC20).interfaceId)
+            || ERC165Checker.supportsInterface(_token, type(IOptimismMintableERC20).interfaceId);
+    }
+
+	function getMirrorAddress(address _token) external view returns (address _mirror) {
```
tynes, line 69: Can we use a mapping that is populated rather than computing the address on chain? Otherwise we will need to forever pin the specific version of solc

protocol/superc20/mirror-factory-limited.md
0xng: Perfect, this makes a lot of sense. Thanks for the suggestion, will update this.

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+- `Init Code`: The same proxy's init code must be used accross chains.
+
+The salt doesn't need to be more than the `OptimismMintableERC20Token` address. The `name` and `symbol`, and `decimals` can be fetched with the `token` address alone in the implementation. The factory will be the `Mirror`'s admin, and the proxy will have to either call `setStorage` immediately with the `token` at the time of deployment, or add a `TOKEN_SLOT` and update it on the constructor.
+
+The factory should be upgradable or have setters to allow the modification of the implementations for the current `Mirrors`. The deployment and update of already-deployed `Mirrors` to new implementations should be open to anyone.
+
+We can track versioning through numbers or the addresses of the implementations used.
+
+This design allows the `StandardBridge` or any contract that needs to check the legitimacy of a `Mirror` to call `MIRROR_FACTORY.getMirrorAddress()` and validate whether the calling user is a mirror that corresponds to a given `token`. As an illustrative example of a potential function of the `StandardBridge`:
+
```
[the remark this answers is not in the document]
0xng: I can add invariants, but you raise a great point. If modifying the init code used for create2 is a possibility, many issues appear:
- Theoretically--if there wasn't a check--one could deploy two mirrors for the same token, as the resulting address of the create2 would be different.
- The `getMirrorAddress` function would break for tokens with the old init code.
- If adding a chain later on that has different/upgraded init code, and we deploy a Mirror for a token deployed with the previous init code on another chain, the interoperable features would break as the address of these tokens would differ.

It seems to me that I need to rewrite this with the `CREATE3` approach to remove the init code dependency and rely solely on the salt.  I think this will be the most robust approach accompanying with the beacon and the populated mapping as you suggested.

protocol/superc20/mirror-factory-limited.md
tynes: Another thought, we could do something like this:

```solidity
contract MirrorFactory {
  address implementation;
  constructor() {
    implementation = new MirrorSuperchainERC20();
  }

  function deploy() {
    proxy = new Proxy({ beacon: address(this) });
  }
}
```

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
+# Summary
+
+*The work present below is part of the [Interoperability project](https://github.com/ethereum-optimism/optimism/issues/10899).*
+
+A factory to spin up `ISuperchainERC20` Mirrors of `OptimismMintableERC20Tokens` is required. This document aims to provide a design for such a factory. The Mirrors need to be proxied to address any future addition of features. The factory will deploy these proxies. Therefore, from now on, when we say `Mirror` we refer to the Proxy and not the implementation.
```
mds1, line 5: I would really appreciate an overall architecture/glossary section in one of these design docs, there are a lot of new terms and I want to make sure the overall design is unambiguous:
- OptimismMintableERC20Factory is a proxy, with a corresponding implementation
- OptimismMintableERC20Factory deploys OptimismMintableERC20 contracts, and the OptimismMintableERC20s are not proxied
- MirrorERC20Factory is a proxy, with a corresponding implementation
- It deploys MirrorERC20 tokens, which are proxied using the beacon pattern, so we can upgrade all implementations if needed

(Something like that, I don't know if that's exactly correct)

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+- Using `CREATE3`. Viable, yet not preferred due to increased complexity.
+
+# Proposed Solution
+
+Limiting the scope to `OptimismMintableERC20Tokens` allows using `CREATE2` instead of `CREATE3`, which is more expensive and complex. This is because the variables CREATE2 uses to define the address of the deployed contract provides all the properties we need to maintain the same address accross chains and easily verify whether a mirror was deployed by the official factory or not:
+
+- `Salt`: If we use the `token` address, we can link the `Mirror` to this `token`. This will allow other contracts to know whether a caller is a legitimate `Mirror` or not.
+- `Sender Address`: In this case, it will be the Factory address itself, or the address of the factory's proxy. This works well, as it the factory can be deployed at the same address accross different chains. For chains in the superchain, it will be a predeploy. This predeploy, however, won't have a pretty address. Instead, it has to be deployed at the address where the L1 Factory is deployed so the address is maintained for all chains.
+- `Init Code`: The same proxy's init code must be used accross chains.
```
mds1, line 23: If we change the solidity version, new chains deployed after that version change will have proxies with different initcode, losing the "same address" benefit?

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+
+The salt doesn't need to be more than the `OptimismMintableERC20Token` address. The `name` and `symbol`, and `decimals` can be fetched with the `token` address alone in the implementation. The factory will be the `Mirror`'s admin, and the proxy will have to either call `setStorage` immediately with the `token` at the time of deployment, or add a `TOKEN_SLOT` and update it on the constructor.
```
mds1, line 25: > and the proxy will have to either call `setStorage` immediately with the `token` at the time of deployment, or add a `TOKEN_SLOT` and update it on the constructor.

I'm not following what this sentence means, I don't see `setStorage` referenced anywhere else in this document

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+
+The factory should be upgradable or have setters to allow the modification of the implementations for the current `Mirrors`. The deployment and update of already-deployed `Mirrors` to new implementations should be open to anyone.
```
mds1, line 27: This sounds like implementations can be changed permissionlessly so I think I may be misunderstanding something here

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+
+We can track versioning through numbers or the addresses of the implementations used.
```
mds1, line 29: For contract versioning, wouldn't we just use the standard semver versioning scheme we use for other contracts? https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/VERSIONING.md

protocol/superc20/mirror-factory-limited.md
mds1: Agreed with @tynes, my read of this design doc makes it seem like we want to use a beacon pattern here

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+            bytes1(0xff),
+            address(this),
+            keccak256(abi.encode(_token)),
+            keccak256(abi.encodePacked(
+                type(L2ChugSplashProxy).creationCode,
+                abi.encode(_token)
+            ))
+        )))))
+	} 
+}
```
mds1, line 80: Similar to https://github.com/ethereum-optimism/design-docs/pull/36#discussion_r1653085937, we should focus less on specific code here and more on high level design and flows

protocol/superc20/mirror-factory-limited.md
```diff
@@ -0,0 +1,90 @@
...
+
+```
+
+# Risks & Uncertainties
+
+This only accounts for `OptimismMintableERC20Tokens`. The design space is limited for simplicity. Legacy tokens are not considered. New interoperable tokens using the `Superc20` standard without Mirrors, are not taken into account for this factory.
```
mds1, line 86: This is a user story I'd like to see fleshed out more, as the high level paths are unclear to me:
- Token A has existing OptimismMintableERC20 contract -> describe the path to create a SuperERC20 version
- Token B has no existing OptimismMintableERC20 contract -> describe the path to create a SuperERC20 version
  - Token B now has SuperERC20 version but no OptimismMintableERC20 version -> can someone create the OptimismMintableERC20 now? If yes, what happens? If no, what prevents it?

protocol/superc20/mirror-factory-limited.md
0xng: Perfect, agreed. Will add it.

tynes:
Closing this in favor of https://github.com/ethereum-optimism/design-docs/pull/50

PR #45, opened 2024-07-10
design-doc: convert isolated

0xParticle:
**Description**
This document presents a solution for migrating `OptimismMintableERC20`, which corresponds to locked liquidity in L1, to the `SuperchainERC20` standard to be interop compatible.

To do that, we suggest  using the `L2StandardBridge` mint/burn rights over the legacy representations to allow easy conversion between the `OptimismMintableERC20` and the corresponding `SuperchainERC20`.

### DIFF
--- a/protocol/superchainERC20/convert.md
+++ b/protocol/superchainERC20/convert.md
@@ -1,0 +1,176 @@
+# Purpose
+
+_This design doc is part of the [Interoperability project](https://github.com/ethereum-optimism/optimism/issues/10899)._
+
+This document presents a solution for migrating `OptimismMintableERC20`, which corresponds to locked liquidity in L1, to the `SuperchainERC20` standard to be interop compatible.
+
+# Summary
+
+The `OptimismMintableERC20` tokens (also known as legacy tokens) are not compatible with interop. This document identifies the migration process's potential challenges and suggests a solution called Convert.
+
+The Convert method will take advantage of the `L2StandardBridge` mint/burn rights over the legacy representations to allow easy conversion between the `OptimismMintableERC20` and the corresponding `SuperchainERC20`.
+
+# Problem Statement
+
+The current design of the `StandardBridge` enables the transfer of standard ERC20 tokens between L1 and each L2 using a lock and mint mechanism. Token representations aren't fungible across different OP chains. We would like to make current liquidity adopt the new `SuperchainERC20` standard to be Interop compatible.
+
+# Possible approaches
+
+### Liquidity Migration Methods
+
+**L1 migration**
+
+One possible approach is to create a new `SharedStandardBridge` from scratch, which would interact with each `CrossDomainMessenger` and share all the liquidity. This could be implemented as a new contract for users to migrate their liquidity or by forcing a migration through an unconventional process.
+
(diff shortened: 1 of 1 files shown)


protocol/superchainERC20/convert.md
```diff
@@ -0,0 +1,176 @@
...
+- **Check `SuperchainERC20` is valid**: There are two ways to implement access control in the Isolated design.
+  - **Registry**: If the Factory stores the `SuperchainERC20` address for each legacy address.
+    ```solidity
+    ISuperchainERC20Factory public superFactory = ISuperchainERC20Factory(factoryAddr);
+
+    function isValidPair(address _legacyAddr, address _superAddr) internal view returns (bool) {
+      return superFactory.deployment(_legacyAddr) == _superAddr;
+    }
+    ```
+  - **Deterministic address**: The `SuperchainERC20` factory will use the L2 representation address as a salt, so the address of the `SuperchainERC20` can be deduced from the address of the `OptimismMintableERC20` and the Factory (predeploy).
```
protolambda, line 108: Do we have any L2 chains with conflicting tokens at the same address, minted by the system factory? I remember Base running into something like this with a previous factory. If yes, then that breaks the deterministic-address assumption. Using the L1-address of the token may be more safe (if that can be authenticated to be correct).

protocol/superchainERC20/convert.md
```diff
@@ -0,0 +1,176 @@
...
+> - The `SuperchainERC20` representation for migrated liquidity will be as close as possible to newly deployed `SuperchainERC20`s.
+> - The `L2StandardBridge` needs to be involved because of burn and mint. Having the functionality in the token would have added additional external calls.
+
+### Access Control
+
+The contract will need to:
+
+- **Check `OptimismMintableERC20` address is valid**: there is no need to verify anything as any liquidity risk will be isolated.
+- **Check `SuperchainERC20` is valid**: There are two ways to implement access control in the Isolated design.
+  - **Registry**: If the Factory stores the `SuperchainERC20` address for each legacy address.
```
protolambda, line 100: How does the factory choose to store an L2 super-erc20 contract address value by a `_legacyAddr` L2-address key? Does it store it upon deployment of the erc20? What happens to previously deployed addresses?

protocol/superchainERC20/convert.md
```diff
@@ -0,0 +1,176 @@
...
+
+We prefer simplicity and will therefore focus on the Isolated approach, i.e., a single `SuperchainERC20` per legacy token. For a more detailed discussion, please refer to the Appendix.
+
+# Proposed Solution: Isolated Convert
+
+## Implementation
+
+The Convert Method will require modifying the `L2StandardBridge` to include the following:
+
+- `convertToSuper(address _*legacyAddr, address _superAddr, uint256 _amount)`: converts* `_amount` *legacy tokens with address `_*legacyAddr` to the same `_amount` of `SuperchainERC20` tokens in address `_superAddr`.
```
protolambda, line 58: Sanity check: is the `_legacyAddr` the same across all L2 chains, for the same L1 token?

protolambda (commented):
Generally looks good, but I have some questions about the token-address mapping between legacy and super.

tynes:
We decided to go with the approach in https://github.com/ethereum-optimism/design-docs/pull/46, so closing this

PR #46, opened 2024-07-10
design-doc: convert-unified

Joxess:
**Description**
This document presents a solution for migrating `OptimismMintableERC20`, which corresponds to locked liquidity in L1, to the `SuperchainERC20` standard to be interop compatible.

To do that, we suggest using the `L2StandardBridge` mint/burn rights over the legacy representations to allow easy conversion between the `OptimismMintableERC20` and the corresponding `SuperchainERC20`.

In contrast to the isolated approach presented here https://github.com/ethereum-optimism/design-docs/pull/45, this PR focuses on the unified approach, where multiple legacy tokens can be converted into one single `SuperchainERC20`.

### DIFF
--- a/protocol/superchainerc20/convert.md
+++ b/protocol/superchainerc20/convert.md
@@ -1,0 +1,196 @@
+# Purpose
+
+*This design doc is part of the [Interoperability project](https://github.com/ethereum-optimism/optimism/issues/10899).*
+
+This document presents a solution for migrating `OptimismMintableERC20`, which corresponds to locked liquidity in L1, to the `SuperchainERC20` standard to be interop compatible. 
+
+# Summary
+
+The `OptimismMintableERC20` tokens (also known as legacy tokens) are not compatible with interop. Several approaches are possible to make it possible. In this document, we identify the migration process's potential challenges and suggest a solution called Convert. 
+
+The Convert method will take advantage of the `L2StandardBridge` mint/burn rights over the legacy representations to allow easy conversion between the `OptimismMintableERC20` and the corresponding `SuperchainERC20`.
+
+# Problem Statement
+
+The current design of the `StandardBridge` enables the transfer of standard ERC20 tokens between L1 and each L2 using a lock and mint mechanism. Token representations aren't fungible across different OP chains. We would like to make current liquidity adopt the new `SuperchainERC20` standard to be Interop compatible.
+
+# Possible approaches
+
+### L1 migration
+
+One possible approach is to create a new `SharedStandardBridge` from scratch, which would interact with each `CrossDomainMessenger` and share all the liquidity. This could be implemented as a new contract for users to migrate their liquidity or by forcing a migration through an unconventional process.
+
+However, a manual migration could take months, and a forced migration would introduce significant liability. For these reasons, we advise against this path.
+
+### L2 migration
+
+Another approach is to make changes only on the L2 side to make the current `OptimismMintableERC20` tokens interoperable without many protocol-level changes. This approach is simpler and safer than L1 migration. 
+
+There are several ways to achieve this result:
+
+- **Wrap:** Introduce a new contract that wraps one or multiple legacy tokens into a `SuperchainERC20`.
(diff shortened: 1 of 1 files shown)


protocol/superchainerc20/convert.md
```diff
@@ -0,0 +1,196 @@
...
+    - **Use deterministic address**: This is only possible for some factories.
+        - The [new factory](https://github.com/ethereum-optimism/optimism/blob/v1.1.4/packages/contracts-bedrock/src/universal/OptimismMintableERC20Factory.sol) uses CREATE2 and will have a single representation. Deterministic address computation can verify the validity of the `OptimismMintableERC20`.
+        - The [old factory](https://github.com/ethereum-optimism/optimism-legacy/blob/develop/packages/contracts-bedrock/contracts/universal/OptimismMintableERC20Factory.sol) uses CREATE, meaning that the only possible approach is to iterate through the potential nonces. The maximum used nonce is not available in the contract, so the best approach would be to have some kind of registry (single-time update).
+- **Check `SuperchainERC20` address corresponds to the correct L1 address**: This is verified during the same validity check of the `SuperchainERC20`.
+
+A possible implementation would look like this:
+
+```solidity
+
```
tynes, line 112: I am leaning towards the idea of adding a storage key/value pair to the factory for deployed addresses. We could:
- Update the factory to set storage after doing a deployment
- Do manual state surgery to fill in the storage slots on OPM + Base using an atomic upgrade to `StorageSetter` that sets values based on the offchain token list.

This would be a one time action that moves the token list on chain, we would not need to manually update the tokenlist over time. This would work well for chains like Base where they didn't use a canonical factory for some time. Then new chains don't need to worry about this storage setting and we can not need to add extra contracts to the mix.

protocol/superchainerc20/convert.md
```diff
@@ -0,0 +1,196 @@
...
+ISuperchainERC20Factory public superFactory = ISuperchainERC20Factory(factoryAddr);
+
+function isValidPair(address _legacyAddr, address _superAddr) internal view returns (bool) {
+  IOptimismMintableERC20 legacyToken = IOptimismMintableERC20(_legacyAddr);
+  address l1address = legacyToken.l1Token();
+  if (!registry.legacyDeployments(l1address)) {
+      return false;
+  }
+  return superFactory.deployment(l1address) == _superAddr;
+}
```
protolambda, line 122: This solidity does not check the condition:

> Check `OptimismMintableERC20` is valid

Any L2 contract could implement a `l1Token` function that can claim whatever here, right? To be unified, and merge its own supply with that of the real L2 token?

The `_legacyAddr` needs to be checked to be registered by the contract, or to be generated from a valid `OptimismMintableERC20` salt, to verify the deployment. Maybe ideally with a code-hash sanity check.

protolambda (commented):
Similar to the other PR, I am concerned about the token address-mapping. The general super-erc20 part LGTM.

tynes (approved):

tynes:
We have consensus on this approach, its good to begin work on specs and break the project into 2 workstreams, one for the work itself and one for backporting the registry to the chain

protocol/superchainerc20/convert.md
tynes: We will want to figure out these details as part of the spec but if we go with the 1 time backport approach then we can get around needing to compute addresses on chain

protocol/superchainerc20/convert.md
ajit2903: Hi [Admin/Maintainer Name],

I would like to request write/maintain permissions on the ethereum-optimism/design-docs repository so I can help review and approve pull requests.

My GitHub username: ajit2903

Here are some of my recent contributions:
- [Link to your PR or commit, e.g., https://github.com/ethereum-optimism/design-docs/pull/46/commits/f8ecf603986491b840a75509cf2f45a95cffde91]

Thank you!

ajit2903:
Hi [Admin/Maintainer Name],

I would like to request write/maintain permissions on the ethereum-optimism/design-docs repository so I can help review and approve pull requests.

My GitHub username: ajit2903

Here are some of my recent contributions:
- [Link to your PR or commit, e.g., https://github.com/ethereum-optimism/design-docs/pull/46/commits/f8ecf603986491b840a75509cf2f45a95cffde91]

Thank you!

PR #50, opened 2024-07-19
design-doc: SuperchainERC20 Factory for L1 native tokens

0xParticle:
**Description**
This document presents the requirements for a factory that deploys `SuperchainERC20` tokens for tokens native to L1.

### DIFF
--- a/protocol/superchainerc20/L1-native-factory.md
+++ b/protocol/superchainerc20/L1-native-factory.md
@@ -130,7 +130,7 @@ contract OptimismSuperchainERC20Beacon is IBeacon {
 The implementation should include
 
 - An `initialize` function that takes `(address _remoteToken, string _name, string _symbol, uint8 decimals)` and stores these in the storage of the BeaconProxy.
-- A function to reinitialize after an upgrade.
+- A function to reinitialize after proxy deployments or upgrades.
 
 ## User Stories
 
--- a/protocol/superchainerc20/L1-native-factory.md
+++ b/protocol/superchainerc20/L1-native-factory.md
@@ -82,10 +82,10 @@ that maps each L1 address to the `DeploymentData` struct
 
 ```solidity
 struct DeploymentData {
-	address superchainERC20;
-	uint8 decimals;
-	string name;
-	string symbol;
+  address superchainERC20;
+  uint8 decimals;
+  string name;
+  string symbol;
 }
 ```
 
@@ -101,8 +101,8 @@ Will follow the [`Proxy.sol` implementation](https://github.com/ethereum-optimis
 contract SuperchainERC20Factory is Semver {
   address public constant NATIVE_INITIALIZER = 0x0...;
 
-	// address of superchainERC20 can be packed with decimals to save a slot
-	struct DeploymentData {
+  // address of superchainERC20 can be packed with decimals to save a slot
+  struct DeploymentData {
     address superchainERC20;
     uint8 decimals;
     string name;
@@ -115,9 +115,9 @@ contract SuperchainERC20Factory is Semver {
 
   constructor() Semver(1, 0, 0) {}
 
-	// This function can be called from L1 or L2. A real implementation would probably move the
-	// deployment logic to an internal function and split deploy() into deployFromL1() and deployFromL2()
-	// external functions to minimize branching.
+  // This function can be called from L1 or L2. A real implementation would probably move the
+  // deployment logic to an internal function and split deploy() into deployFromL1() and deployFromL2()
+  // external functions to minimize branching.
   function deploy(address _remoteToken, uint8 _decimals, string memory _name, string memory _symbol) external returns (address _superchainERC20) {
     if (msg.sender == Predeploys.L2CrossChainMessenger) {
         require(IL2CrossChainMessenger(Predeploys.L2CrossChainMessenger).xDomainMessageSender() == NATIVE_INITIALIZER, "Invalid initializer");
@@ -144,7 +144,7 @@ contract SuperchainERC20Factory is Semver {
     emit SuperchainERC20Deployed(_superchainERC20, _remoteToken, _decimals, _name, _symbol);
   }
 
-	// discuss having an array of chainIds and a for
+  // discuss having an array of chainIds and a for
   function deployWithMetadata(address _remoteToken, uint256 _chainId) external {
     DeploymentData storage _metadata = deploymentsData[_remoteToken];
     require(_metadata.superchainERC20 != address(0), "Token has not been deployed yet");
--- a/protocol/superchainerc20/L1-native-factory.md
+++ b/protocol/superchainerc20/L1-native-factory.md
@@ -4,7 +4,7 @@ This document presents the requirements for a factory that deploys `SuperchainER
 
 ## Overview
 
-The `SuperchainERC20Factory` will be a proxied predeploy. It will deploy proxied `SuperchainERC20` tokens using the Beacon pattern. This structure allows upgrading all `SuperchainERC20` implementations at once if needed.
+The `SuperchainERC20Factory` will be a proxied predeploy. It will deploy proxied `SuperchainERC20` tokens using the Beacon pattern. This structure allows upgrading all implementations at once if needed.
 
 Even though this document is specific for L1 native tokens, it should serve as a basis for other cases.
 
@@ -12,13 +12,13 @@ Even though this document is specific for L1 native tokens, it should serve as a
 
 The `SuperchainERC20` [standard](https://github.com/ethereum-optimism/specs/blob/aee0b2b2b45447daedef4b09bedc1fe7794d645d/specs/interop/token-bridging.md) will make ERC20s interoperable across the Superchain. However, the standard does not solve one of the main challenges in the future of Superchain: dealing with synchronized deployments and upgrades.
 
-Moreover, the `SuperchainERC20` corresponding to a legacy token will require a liquidity migration conversion method, as presented in the corresponding [spec document](https://github.com/ethereum-optimism/specs/pull/294). The system requires a registry to track the allowed representations for the conversion to be safe.
+Moreover, the `SuperchainERC20` corresponding to a legacy token (`OptimismSuperchainERC20`) will require a liquidity migration conversion method, as presented in the corresponding [spec document](https://github.com/ethereum-optimism/specs/pull/294). The system requires a registry to track the allowed representations for the conversion to be safe.
 
 Both these problems can be solved by introducing factories that comply with the following requirements:
 
 - **Same address deployment:** the `SuperchainERC20` standard requires the same address across chains.
   - Notice this is not strictly necessary: allowing different addresses of `SuperchainERC20`s to communicate using a registry is possible. However, maintaining a registry would be a demanding task.
-- **Implementation upgradability:** it should be easy to upgrade `SuperchainERC20` implementations across chains.
+- **Implementation upgradability:** it should be easy to upgrade `OptimismSuperchainERC20` implementations across chains.
 - **Factory upgradability:** The factory should be upgradable to follow OP Factories structure.
 - **Deployment history:** The factory should keep a history of deployed tokens, acting as a registry.
 
@@ -33,13 +33,13 @@ In what follows, we will list each requirement and detail how the factory should
 
 **Predeploys**
 
-Factories will live in each L2 as a predeploy on the same address, which is crucial to achieving the same address for `SuperchainERC20`s across chains.
+Factories will live in each L2 as a predeploy on the same address, which is crucial to achieving the same address for `OptimismSuperchainERC20`s across chains.
 
-If the same address for the `SuperchainERC20` on L1 is also needed, then the address at which the predeploy is initialized should match the factory's L1 address.
+If the same address for the `OptimismSuperchainERC20` on L1 is also needed, then the address at which the predeploy is initialized should match the factory's L1 address.
 
 **Creation method**
 
-The `SuperchainERC20` addresses should be independent of the implementations. `CREATE` was discarded due to nonce dependability, and `CREATE2` was discarded because of its dependence on the creation code. For these reasons, the factory should use `CREATE3`.
+The `OptimismSuperchainERC20` addresses should be independent of the implementations. `CREATE` was discarded due to nonce dependability, and `CREATE2` was discarded because of its dependence on the creation code. For these reasons, the factory should use `CREATE3`.
 
 The salt will be composed of the L1 address and the token metadata. This implies that the same L1 token can have multiple SuperchainERC20 representation, as long as the metadata also changes.
 
@@ -47,7 +47,7 @@ The salt will be composed of the L1 address and the token metadata. This implies
 
 **BeaconProxies**
 
-The factory will deploy `SuperchainERC20`s as BeaconProxies, as this is the easiest way to upgrade multiple proxies simultaneously. Each BeaconProxy will delegate-call to the implementation address provided by the Beacon Contract.
+The factory will deploy `OptimismSuperchainERC20`s as BeaconProxies, as this is the easiest way to upgrade multiple proxies simultaneously. Each BeaconProxy will delegate-call to the implementation address provided by the Beacon Contract.
 
 **Beacon Contract**
 
@@ -68,10 +68,10 @@ The factory will be a simple Proxy contract pointing to the current implementati
 The factory will store a `deployments` mapping that the `L2StandardBridge` will check for access control when converting between the SuperchainERC20 and their corresponding legacy representations (see the [liquidity migration specs](https://github.com/ethereum-optimism/specs/pull/294) for more details).
 
 ```solidity
-mapping(address => address) public deployments;
+mapping(address superc20 => address remoteToken) public deployments;
 ```
 
-that maps each deployed SuperchainERC20 address to the corresponding L1 address (remote address).
+that maps each deployed `OptimismSuperchainERC20` address to the corresponding L1 address (remote address).
 
 ## Implementation
 
@@ -82,17 +82,17 @@ Will follow the [`Proxy.sol` implementation](https://github.com/ethereum-optimis
 **Factory Implementation**
 
 ```solidity
-contract SuperchainERC20Factory is Semver {
-  mapping(address => address) public deployments;
+contract OptimismSuperchainERC20Factory is Semver {
+  mapping(address superc20 => address remoteToken) public deployments;
 
-  event SuperchainERC20Deployed(address indexed superchainERC20, address indexed remoteToken, string name, string symbol, uint8 decimals);
+  event OptimismSuperchainERC20Deployed(address indexed superchainERC20, address indexed remoteToken, string name, string symbol, uint8 decimals);
 
   constructor() Semver(1, 0, 0) {}
 
   function deploy(address _remoteToken, string memory _name, string memory _symbol, uint8 _decimals) external returns (address _superchainERC20) {
     bytes memory _creationCode = abi.encodePacked(
         type(BeaconProxy).creationCode,
-        abi.encode(Predeploy.SuperchainERC20Beacon, abi.encodeCall(SUPERCHAIN_IMPL.initialize.selector, (_remoteToken, _name, _symbol, _decimals)))
+        abi.encode(Predeploy.OptimismSuperchainERC20Beacon, abi.encodeCall(SUPERCHAIN_IMPL.initialize.selector, (_remoteToken, _name, _symbol, _decimals)))
     );
 
     bytes32 salt = keccak256(abi.encode(_remoteToken, _name, _symbol, _decimals));
@@ -100,14 +100,14 @@ contract SuperchainERC20Factory is Semver {
 
     deployments[_superchainERC20] = _remoteToken;
 
-    emit SuperchainERC20Deployed(_superchainERC20, _remoteToken _name, _symbol, _decimals);
+    emit OptimismSuperchainERC20Deployed(_superchainERC20, _remoteToken _name, _symbol, _decimals);
   }
 }
 ```
 
 The name `_remoteToken` was chosen over `_l1Token` to contemplate future L2 native cases.
 
-**BeaconProxy (SuperchainERC20)**
+**BeaconProxy (OptimismSuperchainERC20)**
 
 The Token should follow a simple BeaconProxy implementation, like Openzeppelin:
 
@@ -116,7 +116,7 @@ https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/pro
 **Beacon Contract**
 
 ```solidity
-contract SuperchainERC20Beacon is IBeacon {
+contract OptimismSuperchainERC20Beacon is IBeacon {
   /// TODO: Replace with real implementation address
   address internal constant IMPLEMENTATION_ADDRESS = 0x0000000000000000000000000000000000000000;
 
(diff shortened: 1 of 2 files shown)


tynes:
Are you sure that we need to deploy the L2 tokens via deposits from L1? That will increase the cost of creating these tokens. I believe we can get around this if the salt commits to the L1 token address + the erc20 properties fetched from the `OptimismMintableERC20` token

0xParticle:
> Are you sure that we need to deploy the L2 tokens via deposits from L1? That will increase the cost of creating these tokens. I believe we can get around this if the salt commits to the L1 token address + the erc20 properties fetched from the `OptimismMintableERC20` token

We think of L1 deployments as the simplest way to ensure the token metadata is correct while allowing permissionless deployments. An alternative could be to use the L1 block hash and build some sort of storage proof for the L1 token address (proving metadata against storage), but it's a more complex implementation (even UX wise).
Using the `REMOTE_TOKEN` from the `OptimismMintableERC20` as the L1 token address is not enough, as you still need a source of truth for the metadata.
We can also relax the metadata requisite and go with that approach.

0xParticle:
> Are you sure that we need to deploy the L2 tokens via deposits from L1? That will increase the cost of creating these tokens. I believe we can get around this if the salt commits to the L1 token address + the erc20 properties fetched from the `OptimismMintableERC20` token

After discussing this point, we will update the design doc to remove the need for L1 deployments. We will relax the metadata enforced condition and make it based on social consensous.

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,307 @@
...
+
+**Predeploys**
+
+Factories will live in each L2 as a predeploy on the same address, which is crucial to achieving the same address for `SuperchainERC20`s across chains.
+
+If the same address for the `SuperchainERC20` on L1 is also needed, then the address at which the predeploy is initialized should match the factory's L1 address.
+
+**Creation method**
+
+The `SuperchainERC20` addresses should be independent of the implementations. `CREATE` was discarded due to nonce dependability, and `CREATE2` was discarded because of its dependence on the creation code. For these reasons, the factory should use `CREATE3`.
```
smartcontracts, line 42: A link to the `CREATE3` spec would be very helpful here.

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,307 @@
...
+
+The factory will be a simple Proxy contract pointing to the current implementation. In case of an upgrade to the factory, the upgrading chains will need to deploy the new implementation and perform a hardfork to update the implementation address.
+
+### Conservation of Properties
+
+The first deployment should be initialized from a Deployer contract in L1 called `NATIVE_INITIALIZER`, with a message that includes name, symbol, and decimal information. The predeployed factory on L2 will check that the message comes from the whitelisted address of the Deployer (which will live as a state variable in the predeploy) and proceed with the rest of the `SuperchainERC20` deployment.
+
+After the first deployment, it is possible to deploy from L1 or from the factory that already deployed and stored the metadata using interop.
+
+An alternative design to deploy from L1 is to use the `L1 blockhash` along with storage proofs to prove the properties. This would allow to initialize deployments from L2 as well (the factory would need to implement two external functions in this case).
```
smartcontracts, line 72: Hmmm how would this work? Each ERC20 stores stuff in different locations and some ERC20s store their details in functions instead of storage variables.

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,307 @@
...
+
+In an implementation update, the upgraded chains must perform a hardfork to change the constant value in the Beacon Contract representing the implementation address.
+
+### Factory Upgradability
+
+The factory will be a simple Proxy contract pointing to the current implementation. In case of an upgrade to the factory, the upgrading chains will need to deploy the new implementation and perform a hardfork to update the implementation address.
+
+### Conservation of Properties
+
+The first deployment should be initialized from a Deployer contract in L1 called `NATIVE_INITIALIZER`, with a message that includes name, symbol, and decimal information. The predeployed factory on L2 will check that the message comes from the whitelisted address of the Deployer (which will live as a state variable in the predeploy) and proceed with the rest of the `SuperchainERC20` deployment.
```
smartcontracts, line 68: Not sure that the benefit here is worth the effort. It's a nice to have but it adds complexity and messes with the user experience. Clients will generally agree to pick the token that has the desired properties anyway.

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,307 @@
...
+  function deploy(address _remoteToken, uint8 _decimals, string memory _name, string memory _symbol) external returns (address _superchainERC20) {
+    if (msg.sender == Predeploys.L2CrossChainMessenger) {
+        require(IL2CrossChainMessenger(Predeploys.L2CrossChainMessenger).xDomainMessageSender() == NATIVE_INITIALIZER, "Invalid initializer");
+    } else if (msg.sender == Predeploys.L2ToL2CrossDomainMessenger) {
+        require(IL2toL2CrossDomainMessenger(Predeploys.L2ToL2CrossDomainMessenger).crossDomainMessageSender() == address(this), "Invalid cross-domain sender");
+    } else {
+        revert("Unauthorized caller");
+    }
+
+    require(deployments[_remoteToken] == address(0), "Token already deployed");
```
smartcontracts, line 132: Do we strictly want this property? Factory will only deploy 1 SuperchainERC20 token per L1 token? If so, why?

smartcontracts (commented):
I added some comments but then realized that the most fundamental question that must be answered is:

Do we really care about the 1:1 mapping between L1 tokens and SuperchainERC20 tokens? I'm skeptical that this is necessary and it seems to limit this design document quite a bit. In fact, this property does not seem to be documented as a design requirement in the problem statement but this property then dominates the design as a whole. Would be great to get a clear answer here which will help direct future review.

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -82,32 +82,32 @@ Will follow the [`Proxy.sol` implementation](https://github.com/ethereum-optimis
 **Factory Implementation**
 
 ```solidity
-contract SuperchainERC20Factory is Semver {
-  mapping(address => address) public deployments;
+contract OptimismSuperchainERC20Factory is Semver {
+  mapping(address superc20 => address remoteToken) public deployments;
 
-  event SuperchainERC20Deployed(address indexed superchainERC20, address indexed remoteToken, string name, string symbol, uint8 decimals);
+  event OptimismSuperchainERC20Deployed(address indexed superchainERC20, address indexed remoteToken, string name, string symbol, uint8 decimals);
```
Joxess, line 88: We’re looking for ways to establish common standards regarding deployments. For `SuperchainERC20` backed by an L2 native token, we’re considering using `chainId` as an additional input. Should we do the same for L1 bridged tokens (e.g., Ethereum `chainId` as 1)?

The same reasoning can be applied to parameters stored for each representation, based on having a `remoteToken`, `chainId`, and bridge.

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,170 @@
+## Summary
+
```
tynes, line 2: Can we call the implementation of the converted `SuperchainERC20` tokens `OptimismSuperchainERC20` tokens? So that the naming is consistent between `OptimismMintableERC20` and `OptimismSuperchainERC20`, its the optimism mintable erc20 that can be converted into optimism superchain erc20

Then we can keep `SuperchainERC20` as the name of the standard

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,170 @@
...
+In what follows, we will list each requirement and detail how the factory should look to accomplish them.
+
+### Same address: Predeploys + `CREATE3`
+
+**Predeploys**
+
+Factories will live in each L2 as a predeploy on the same address, which is crucial to achieving the same address for `OptimismSuperchainERC20`s across chains.
+
+If the same address for the `OptimismSuperchainERC20` on L1 is also needed, then the address at which the predeploy is initialized should match the factory's L1 address.
```
tynes, line 38: If its extra complexity around having the addresses be the same on L1, we can cut that scope

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Implementation Upgradability: Beacon Proxies
+
+**BeaconProxies**
+
+The factory will deploy `OptimismSuperchainERC20`s as BeaconProxies, as this is the easiest way to upgrade multiple proxies simultaneously. Each BeaconProxy will delegate-call to the implementation address provided by the Beacon Contract.
+
+**Beacon Contract**
+
+The Beacon Contract holds the implementation address for the BeaconProxies. It will be a predeploy (consistent with the current architecture of the OP stack).
```
tynes, line 54: It could be good to detail that the beacon contract has to implement a particular interface defined in erc 1976

https://eips.ethereum.org/EIPS/eip-1967

```solidity
function implementation() returns (address)
```

This means that the predeploy would implement this interface and then return an address of the actual implementation. The address of the actual implementation would have to be determined similarly to the [upgrade transactions](https://github.com/ethereum-optimism/specs/blob/1f8ace7f21e44ff028951965ab552c81b892f199/specs/fjord/derivation.md#gaspriceoracle-deployment) for hardforks.

It could be good to talk about upgradability, so we would likely want the address returned by `implementation()` to be an `immutable` and then we can have a hard fork that upgrades the predeploy implementation to include a different `immutable`

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Implementations**
+
+Each type of ERC20 (vanilla, vote token, rebasing, xERC20, etc) will require their unique implementation and, therefore, factory. Optimism should deploy and support the most common implementations. Anyone can deploy their custom implementations and factories but without predeploys.
```
tynes, line 58: We should scope this specifically to the `OptimismSuperchainERC20` factory and then we can revisit designs for more generic factories in the future

protocol/superchainerc20/L1-native-factory.md
tynes: I think we should focus specifically on L1 native tokens for this factory as they are a majority of the liquidity and then return to the problem of making a more generic factory in the future

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,170 @@
...
+## Implementation
+
+**Proxy Factory**
+
+Will follow the [`Proxy.sol` implementation](https://github.com/ethereum-optimism/optimism/blob/v1.1.4/packages/contracts-bedrock/src/universal/Proxy.sol) and `delegatecall()` to the factory implementation address.
+
+**Factory Implementation**
+
+```solidity
+contract OptimismSuperchainERC20Factory is Semver {
```
tynes, line 85: small nit: we migrated away from the `Semver` contract to `ISemver`, this style is a bit dated

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,170 @@
...
+    _superchainERC20 = CREATE3.deploy(salt, _creationCode);
+
+    deployments[_superchainERC20] = _remoteToken;
+
+    emit OptimismSuperchainERC20Deployed(_superchainERC20, _remoteToken _name, _symbol, _decimals);
+  }
+}
+```
+
+The name `_remoteToken` was chosen over `_l1Token` to contemplate future L2 native cases.
```
tynes, line 108: I think that this is good, we could add a new function to this factory or add another factory to solve this problem and have unified language

protocol/superchainerc20/L1-native-factory.md
tynes: I notice that you cover some of this later in the doc :)

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,170 @@
...
+
+### `SuperchainERC20` implementation upgrade
+
+1. Anyone can deploy a new implementation on each chain to be upgraded.
+2. Core developers and Foundation will organize a hardfork between the upgraded chains.
+3. The hardfork will update the implementation address on the Beacon Contract.
+
+## Open Questions
+
+- How should different ERC20 implementations be handled? One or multiple implementations per Beacon Contract predeploy?
```
tynes, line 169: We can add a new beacon contract predeploy in the future if necessary. For now we should focus on solving the convert problem really well. We also have this generic factory pattern for "1 click installs" of common contracts across the superchain which makes it permissionless to add contracts at deterministic addresses across all of the chains. We could ship this in the future as a basis for other types of erc20 tokens, which removes us as the bottleneck for adding new contract features to op stack. The addresses won't be predeploys, but it will be permissionless + deterministic, which is a major benefit

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,170 @@
...
+- How should native token deployment in L2 behave?
```
tynes, line 170: I think we should leave this out of scope for this particular project and spike out what the best devex/ux is in the future. My guess is that either minting happens on 1 chain or minting happens on any chain and admin comes from 1 chain

protocol/superchainerc20/L1-native-factory.md
0xParticle: I will update the naming in Summary and Overview as well

protocol/superchainerc20/L1-native-factory.md
0xParticle: will remove

protocol/superchainerc20/L1-native-factory.md
0xParticle: just to be clear, do you agree on using `_remoteToken` or would prefer `_l1token`?

protocol/superchainerc20/L1-native-factory.md
0xParticle: I will remove the open questions from this design doc. We might mention it on Future considerations for the Specs.

protocol/superchainerc20/L1-native-factory.md
0xParticle: Agree it's out of scope. Will remove from here.

protocol/superchainerc20/L1-native-factory.md
0xParticle: Added links to the EIP and references to the implementation address computation. 
Is there a reason to make the implementation address an `immutable` instead of a `constant`?

protocol/superchainerc20/L1-native-factory.md
tynes: I agree with `_remoteToken` here :)

tynes:
This looks ready for a design review meeting, the design at this point generally looks good to me

tynes:
@mds1 Would love to have your eyes on this if you have time

protocol/superchainerc20/L1-native-factory.md
tynes: Follow up context: https://www.notion.so/defi-wonderland/Extending-SuperchainERC20-to-L2-native-tokens-and-custom-e91ca08dd3eb4434834bf74f56237abf

tynes:
Implementation PR for the token itself: https://github.com/ethereum-optimism/optimism/pull/11256

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,166 @@
...
+The factory will deploy `OptimismSuperchainERC20`s as BeaconProxies, as this is the easiest way to upgrade multiple proxies simultaneously. Each BeaconProxy will delegate-call to the implementation address provided by the Beacon Contract.
+
+**Beacon Contract**
+
+The Beacon Contract holds the implementation address for the BeaconProxies. It will be a predeploy and implement the interface defined in [EIP-1967](https://eips.ethereum.org/EIPS/eip-1967).
+
+**Implementations**
+
+In an implementation update, the upgraded chains must perform a hardfork to change the constant value in the Beacon Contract representing the implementation address. 
+The address of the implementation will be determined similarly to [other upgrade transactions](https://github.com/ethereum-optimism/specs/blob/1f8ace7f21e44ff028951965ab552c81b892f199/specs/fjord/derivation.md#gaspriceoracle-deployment) .
```
mds1, line 57: What is the rationale for this? One consideration is incident response: if a buggy implementation is deployed, it is much harder to coordinate a hard fork than to send L1 deposit transactions that change a storage slot on each chain

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,166 @@
...
+
+
+
+### Factory Upgradability
+
+The factory will be a simple Proxy contract pointing to the current implementation. In case of an upgrade to the factory, the upgrading chains will need to deploy the new implementation and perform a hardfork to update the implementation address.
```
mds1, line 63: Same comment regarding incident response. I'm ok with hardfork being the default happy case path, but we should consider also supporting upgrades via deposit transaction to support quicker changes when required. The way this would work is similar to the current setup: The L2ProxyAdmin Owner is set to the aliased address of the L1ProxyAdmin owner. This way the L1ProxyAdmin owner can send a deposit transaction to upgrade predeploys.

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Deployment history
+
+The factory will store a `deployments` mapping that the `L2StandardBridge` will check for access control when converting between the SuperchainERC20 and their corresponding legacy representations (see the [liquidity migration specs](https://github.com/ethereum-optimism/specs/pull/294) for more details).
+
+```solidity
+mapping(address superc20 => address remoteToken) public deployments;
+```
+
+that maps each deployed `OptimismSuperchainERC20` address to the corresponding L1 address (remote address).
```
mds1, line 73: I need a refresher on the liquidity migration spec so I'm not totally following this section. Why does the L2StandardBridge need a mapping from an L2's SuperchainERC20 address to the RemoteToken address? Is an alternate design for the SuperchainERC20 contract itself expose a getter of it's L1 token address? 

Would be great if we can expand this section and clarify it. In general, diagrams showing the various flows and conversions would be useful (edit: I do see a nice deployment diagram below, it doesn't include the L2StandardBridge though)

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,166 @@
...
+}
+```
+
+The name `_remoteToken` was chosen over `_l1Token` to contemplate future L2 native cases.
+
+**BeaconProxy (OptimismSuperchainERC20)**
+
+The Token should follow a simple BeaconProxy implementation, like Openzeppelin:
+
+https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/beacon/BeaconProxy.sol
```
mds1, line 114: nit, let's use a permalink to to a specific OZ version so this link doesn't break in the future if things get moved around

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,166 @@
...
+}
+```
+
+**Implementation**
+The implementation should include
+
+- An `initialize` function that takes `(address _remoteToken, string _name, string _symbol, uint8 decimals)` and stores these in the storage of the BeaconProxy.
+- A function to reinitialize after proxy deployments or upgrades.
+
+## User Stories
```
mds1, line 135: It sounds like we also need a section here explaining the interactions with the L2StandardBridge

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,166 @@
...
+
+### `OptimismSuperchainERC20` deployment
+
+1. Anyone can deploy a `OptimismSuperchainERC20` by calling the `deploy` function in the `OptimismSuperchainERC20Factory` contract with `_remoteToken`, `_name`, `_symbol` and `decimals` as inputs.
+2. The factory stores the `_remoteToken` address for each deployed SuperchainERC20 address.
+
+```mermaid
+sequenceDiagram
```
mds1, line 143: Since we're deploying an OptimismSuperchainERC20 BeaconProxy, should this diagram also have an `OptimismSuperchainERC20BeaconProxy.initialize()` call? Or is there nothing that needs to be initialized?

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,166 @@
...
+
+**Predeploys**
+
+Factories will live in each L2 as a predeploy on the same address, which is crucial to achieving the same address for `OptimismSuperchainERC20`s across chains.
+
+**Creation method**
+
+The `OptimismSuperchainERC20` addresses should be independent of the implementations. `CREATE` was discarded due to nonce dependability, and `CREATE2` was discarded because of its dependence on the creation code. For these reasons, the factory should use `CREATE3`.
+
+The salt will be composed of the L1 address and the token metadata. This implies that the same L1 token can have multiple SuperchainERC20 representation, as long as the metadata also changes.
```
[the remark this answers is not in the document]
0xParticle: Doing so would imply having a single `OptimismSuperchainERC20` per L1 token. As the deployment is permissionless, anyone could "frontrun" and deploy the token with a poor metadata choice. There would be no way of changing this later on.
We initially thought about using just L1 address as the salt but initializing the first deployment from L1 to bring the correct metadata in, but we ended up following the structure the `OptimismMintableERC20Factory` currently use. Any metadata will be allowed, but social consensous will pick the winner.

protocol/superchainerc20/L1-native-factory.md
0xParticle: Yes, I think the correct word here would be "Protocol Upgrade", which could have a hardfork as a particular case. I will change the sentence 👍

protocol/superchainerc20/L1-native-factory.md
0xParticle: Agree, will make this more generic

protocol/superchainerc20/L1-native-factory.md
0xParticle: I will add a short refresher here. The TLDR is that the `L2StandardBridge` needs a way to check a certain superchain token address is valid, and the best way to do so is asking to the factory if that address was ever deployed. That check could also be done with a mapping from `address` to `bool`, but having the `remoteToken` here is very useful for the `L2StandardBridge` as well, as it needs to verify both tokens being converted correspond to the same `remoteToken`

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,166 @@
+## Summary
+
+This document presents the requirements for a factory that deploys `OptimismSuperchainERC20` tokens.
+
+## Overview
+
+The `SuperchainERC20Factory` will be a proxied predeploy. It will deploy proxied `OptimismSuperchainERC20` tokens using the Beacon pattern. This structure allows upgrading all implementations at once if needed.
+
+Even though this document is specific for L1 native tokens, it should serve as a basis for other cases.
```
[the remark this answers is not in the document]
0xParticle: My only concern is that this factory is specific for tokens that are native to L1. We can still name it like that and change notation for the others (l2 native, etc)

protocol/superchainerc20/L1-native-factory.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Same address: Predeploys + `CREATE3`
+
+**Predeploys**
+
+Factories will live in each L2 as a predeploy on the same address, which is crucial to achieving the same address for `OptimismSuperchainERC20`s across chains.
+
+**Creation method**
+
+The `OptimismSuperchainERC20` addresses should be independent of the implementations. `CREATE` was discarded due to nonce dependability, and `CREATE2` was discarded because of its dependence on the creation code. For these reasons, the factory should use `CREATE3`.
```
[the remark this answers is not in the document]
0xParticle: I like the idea of having a CreateX preinstall. We where discussing having the CREATE3 one. Let's discuss this in the review meeting

protocol/superchainerc20/L1-native-factory.md
0xParticle: correct

tynes (approved):

PR #51, opened 2024-07-23
Add transfer precompile design doc

karlb:
**Description**

Bring Celo's `transfer` precompile to OP-stack to allow Token Duality (native token via ERC20) without requiring an op-geth fork.

**Additional context**

* [Token Duality description in Celo L2 spec](https://specs.celo.org/token_duality.html)


protocol/transfer-precompile.md
```diff
@@ -0,0 +1,83 @@
...
+Since they can't be accessed by the same interface, they either have to be
+handled as two different cases or be explicitly wrapped by an ERC-20 token
+wrapping contract.
+Wrapping tokens is not an ideal solution either, since the wrapped token is
+of limited use (e.g. can't be used to pay gas) and wrapping/unwrapping is
+annoying and incurs additional gas charges.
+
+This can be solved by having a contract that provides an ERC-20 interface
+directly to the native token. To implement such a contract, the blockchain must
+provide a way to transfer native tokens on behalf of the tx sender.
```
palango, line 28:
```suggestion
provide a way to transfer native tokens on behalf of the transaction sender.
```

protocol/transfer-precompile.md
```diff
@@ -0,0 +1,83 @@
...
+# Proposed Solution
+
+Add one new blockchain parameter `TransferPrecompileAllowed` and a new precompile.
+The parameter defines the address from which the precompile must be called.
+
+Precompile name: `transfer`  
+Precompile address: 0x0000000000000000000000000000000000000101 (directly after the `P256VERIFY` precompile from Fjord)  
+Parameters (abi-encoded): `address from, address to, uint256 value`  
+Gas costs: 9000  
+Result: `value` native tokes will be transferred from `from` to `to`.
```
palango, line 39:
```suggestion
Result: `value` native tokens will be transferred from `from` to `to`.  
```

protocol/transfer-precompile.md
```diff
@@ -0,0 +1,83 @@
...
+of limited use (e.g. can't be used to pay gas) and wrapping/unwrapping is
+annoying and incurs additional gas charges.
+
+This can be solved by having a contract that provides an ERC-20 interface
+directly to the native token. To implement such a contract, the blockchain must
+provide a way to transfer native tokens on behalf of the tx sender.
+
+# Proposed Solution
+
+Add one new blockchain parameter `TransferPrecompileAllowed` and a new precompile.
```
palango, line 32: Do you think calling this `TransferPrecompileAllowedCaller` would make the option more clear?

protocol/transfer-precompile.md
```diff
@@ -0,0 +1,83 @@
...
+Precompile name: `transfer`  
+Precompile address: 0x0000000000000000000000000000000000000101 (directly after the `P256VERIFY` precompile from Fjord)  
+Parameters (abi-encoded): `address from, address to, uint256 value`  
+Gas costs: 9000  
+Result: `value` native tokens will be transferred from `from` to `to`.  
+No return value  
+
+The precompile can fail for the following reasons:
+* `TransferPrecompileAllowedCaller` is not configured for the blockchain
+* It is not called from the contract at the `TransferPrecompileAllowedCaller` address
```
tynes, line 44: Is the `TransferPrecompileAllowedCaller` address an arbitrary enshrined smart contract? In the context of the OP Stack, this could be a predeploy. Why can't anybody call this contract and instead of `msg.sender` being the first arg, the native EVM can look at the `CALLER` (which is `msg.sender`)?

protocol/transfer-precompile.md
```diff
@@ -0,0 +1,83 @@
...
+* The `from` address has less than `value` native tokens
+* The input is not exactly 96 bytes long
+* The input parameters can't be parsed correctly
+
+Example usage in solidity:
+
+```solidity
+address constant TRANSFER = address(0x0000000000000000000000000000000000000101);
+
+function _transfer(address to, uint256 value) internal {
```
tynes, line 54: What is the design space of possible `TransferPrecompileAllowedCaller` contracts? In practice would it mostly look like the pseudocode here?

protocol/transfer-precompile.md
```diff
@@ -0,0 +1,83 @@
...
+This can be solved by having a contract that provides an ERC-20 interface
+directly to the native token. To implement such a contract, the blockchain must
+provide a way to transfer native tokens on behalf of the transaction sender.
+
+# Proposed Solution
+
+Add one new blockchain parameter `TransferPrecompileAllowedCaller` and a new precompile.
+The parameter defines the address from which the precompile must be called.
+
+Precompile name: `transfer`
```
tynes, line 35: Are there any security concerns here? I feel like i've seen bugs based on having a dual interface into balance

protocol/transfer-precompile.md
tynes: Ok given the [geth code](https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go#L195) and your comment:

> Removing the `from` parameter from the precompile is not an option because 
`msg.sender` can't be accessed from inside the precompile.
There, the caller is set to the contract doing `TRANSFER.call`.

I understand this design decision more. It would be a larger diff to send through the evm context into the precompile

protocol/transfer-precompile.md
tynes: I feel like the allowed caller should just be a predeploy that is literally this code, perhaps optimized and its part of the design, the way that i read the design right now makes it sound like the allowed caller can be configured to different contracts

protocol/transfer-precompile.md
```diff
@@ -0,0 +1,83 @@
...
+
+Many users and contracts have to deal with both ERC20 tokens and native tokens.
+Since they can't be accessed by the same interface, they either have to be
+handled as two different cases or be explicitly wrapped by an ERC-20 token
+wrapping contract.
+Wrapping tokens is not an ideal solution either, since the wrapped token is
+of limited use (e.g. can't be used to pay gas) and wrapping/unwrapping is
+annoying and incurs additional gas charges.
+
+This can be solved by having a contract that provides an ERC-20 interface
```
tynes, line 26: So the main value prop here is that the unwrap/wrap steps are saved. In a world with smart contract wallets, users can do multicalls and save needing to send multiple transactions, but this is still cheaper as there is no wrap/unwrap step necessary

karlb: Yes, but there is plenty of benefit before everyone uses smart contract wallets and every app has implemented perfect wrap/unwrap handling (Should tokens stay wrapped after the action to save gas? Should existing wrapped tokens be used? Even if they are not enough so that you have to wrap anyway? Do you want to bulk wrap to save gas on the next actions? Should the wallet UI show wrapped tokens separately from unwrapped tokens?).

There is also the argument that it makes the app developers lives easier since they don't have to implement all these wrap/unwrap cases. But since most apps are deployed across many chains, most devs will have to do that anyway.

protocol/transfer-precompile.md
karlb: One more interface is certainly one more potential attack vector, especially when it provides access to everyone's balances. Fortunately, the interface is quite simple and low on interaction with other features. I don't know of any security issues we had with this feature since the Celo mainnet launch. The biggest danger I see is the privileged contract getting compromised.

protocol/transfer-precompile.md
karlb: > I understand this design decision more. It would be a larger diff to send through the evm context into the precompile

Yes, and it would also reduce the chances of the precompile being supported by third party tooling (e.g. foundry's anvil).

protocol/transfer-precompile.md
karlb: I would expect the contract to implement the all of ERC20 in a way that passes through the native balance. So in addition to the transfer, you would have other ERC20 functions like

```solidity
  function decimals() external view returns (uint8) {
    return 18;
  }

  function balanceOf(address _owner) public view returns (uint256) {
    return _owner.balance;
  }
```

and potentially other interfaces (it could be a proxy for upgradibility if you consider that safe enough, other token interfaces, etc). I assume it would be hard to find one contract that fits everyone here, and it would need to be configurable in every case to get the name and symbol right.

Having a predeploy with only the a transfer function that can be called from the full ERC20 contract would not solve anything AFAICS. That function would still need to receive the `from` since `msg.sender` can't be used for that due to the additional call. And it would still need to check that the calling contract is the privileged contract.

tynes:
I think this could make sense given the chain operator could turn on/off the smart contract that is authorized to interact with the predeploy. There is a non zero chance that an existing application has a bug due to this, so chains that want to be more conservative in adopting this would be able to not use it. Then chains that do want this feature could turn it on

PR #52, opened 2024-07-24
deployment-chains: Idea to resolve deployment composition and customization

protolambda:
**Description**

A deterministic incremental approach to chain deployment creates better composability and customization.
I.e. we organize it as a chain of smaller individually configurable deployments.
This enables multi-L2 Interop test setups, and aims to align with the deployment-approach changes of OPStackManager.

Draft, needs feedback from smart-contract and Optimism monorepo-testing stakeholders.

### DIFF
--- a/protocol/deployment-chains.md
+++ b/protocol/deployment-chains.md
@@ -1,0 +1,261 @@
+# Deployment chains
+
+A deterministic incremental approach to chain deployment creates better composability and customization.
+I.e. we organize it as a chain of smaller individually configurable deployments.
+This enables multi-L2 Interop test setups, and aims to align with the deployment-approach changes of OPStackManager.
+
+# Problem Statement + Context
+
+The current L2 chain deployment approach originates from a time with Hardhat,
+single L1 target, and a single monolithic set of features.
+
+Since then the system has migrated to Foundry and extended for more features,
+but remains centered around a single monolithic deploy-config for all its features.
+
+**With Interop we need to configure a new multi-L2 deployment**:
+the number of ways to compose L2s in tests grows past what a single legacy config template can support.
+
+Outside of interop, deployment also seems increasingly complex and opaque, while it does not have to be,
+due to the same configuration and composability troubles.
+
+## DeployConfig complexity
+
+The `DeployConfig` definition, as per the Go code, consists of 102 configurable variables.
+Each one of these may affect the chain deployment at any point during deployment
+due to lack of grouping and encapsulation.
+
+Within `op-chain-ops`, we have started introducing some grouping of features:
+[PR 11189](https://github.com/ethereum-optimism/optimism/pull/11189).
+
+But this is still limited to the Go side, and the complexity still remains on the solidity deploy scripts.
+
+## Stages of deployment
+
+A L2 chain deployment runs through the following steps:
+1. Configure `DeployConfig` (many choices).
+2. Create L1 chain, if developing.
+3. Deploy superchain-wide contracts to L1 (including a `ProxyAdmin` that should not be shared with step 4)
+4. Deploy L2-instance-specific contracts to L1.
+5. Collect deployments in one file, where there is 1 of everything (breaks interop).
+6. Run the L2-genesis script, depending on L1 deployments and the deploy config.
+7. Run tools to turn the L2 genesis state into a chain genesis and rollup configuration.
+
+The configuration of step (1) forces how all the following steps go,
+making it impractical to repeat step (2) per L2 chain.
+
+The L1 deployments artifact that bundles superchain and L2-instance-specific
+deployments also prevents multiple L2s from running.
+
+The main goal is to unbundle these steps, and make them customizable and composable for better test-coverage,
+including Interop parameters specifically.
+
+## OPStackManager
+
+Specs: https://specs.optimism.io/experimental/op-stack-manager.html
+
+Older implementatipn draft: https://github.com/ethereum-optimism/optimism/pull/9985
+
+The OP-Stack manager focuses on making the deployment of the L1 contracts of an L2 chain deterministic,
+by deploying proxies to deterministic addresses (salt = chainID),
+and coupling to superchain-provided implementation contracts.
+
+This implies a different L1-deployment flow than we have today,
+and limits it to just the L2-instance specific contracts.
+A deployment phase we also want to isolate the configuration and execution of.
+
+Note that the draft sources configuration inputs explicitly, rather than coupling to the (arguably bloated) 
+deploy config.
+
+## Deployer requirements
+
+### Op-e2e needs
+
+As an op-e2e test suite, I need to:
+- Customize the deploy-config deeply, including L1 parameters and upgrade parameters, to cover all functionality.
+- Create multiple L2s with a shared L1 (to go from isolated L2s to interoperating L2s through interop hardfork).
+- Create multiple L2s with a shared L1 and shared superchain (to test interop without emulating hardfork first).
+
+### Devnet needs
+
+As a devnet user, I need to:
+- Modify the chain template, for experimentation when building new features.
+- Support different "modes" without accidental configuration mixup,
+  e.g. I may to run different hardforks, or run with different DA types.
+
+### Production needs
+
+As a production superchain host, I need to:
+- Perform a new Superchain core deployment (global contracts and implementation-contract deployments).
+- Maintain the Superchain core deployment with upgrades.
+- Not deploy/upgrade an L2 at the same time.
+
+As a production L2 chain deployer, I may need to:
+- Build an L2 on top of a pre-existing superchain core deployment.
+- Upgrade the L2 by attaching to new superchain implementation contracts.
+- Customize just the L2 part of the deployment, while retaining compatibility with the general deployment flow.
+
+Production needs may change; there are many external users and new features. Flexibility is key.
+Making the deployment more modular fits this.
+
+# Alternatives Considered
+
+## Few enshrined deploy-configs
+
+Not all that different from status-quo.
+With copy-paste we can add 1 more chain, and suffix it with `-interop`, to produce a 2-chain test environment.
+For early testing this was drafted in [PR 10777](https://github.com/ethereum-optimism/optimism/pull/10777).
+
+This does not yet unify the common superchain part of the deployment, or any of the artifacts,
+nor supports test-specific changes to parameters.
+
+## Presets and configs
+
+In Eth2 there exist two classes of configuration:
+- "preset" (named, enshrined, compile-time)
+- "config" (customizable, run-time)
+
+This helped reduce complexity by encapsulating and limiting compile-time configuration,
+but also limited severely what could be changed.
+For a state-format it may be good to contain options,
+but for the amount of runtime configuration the OP-Stack has, it seems impractical.
+
+And each variable that remains in the runtime-config still cannot be customized in `op-e2e` tests,
+due to limits in the current `l1-allocs` and `l2-allocs` pre-generated files approach.
+
+## Templating
+
+By separating contract-code and contract-storage, the contract deployments can be reconfigured without recompiling.
+However, this pushes complexity to the user of the deployments data:
+it re-introduces smart-contract deployment concepts into the Go chain-ops.
+E.g. storage-layouts, storage packing, the occasional `immutable` variable, all make customization difficult.
+Not to mention cases where we need entirely different contracts, different hardforks, or more L2 deployments.
+
+The insight here is that we should strongly contain smart-contract configuration to the solidity-scripting side,
+and minimize post-processing in Go.
+
+## Mega test deployment
+
+Instead of generating many separate files per configuration,
+all deployments could be applied to a single `allocs` with all contracts, and many different parameter choices.
+It becomes bulky, but then allows the consumer to selectively choose what
+part of the deployment matters, and what is dead world state.
+
+This falls apart because the consumer has to track all the deployments,
+and it still leaves the issues of enshrined test-configurations.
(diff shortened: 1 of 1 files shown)


protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+With copy-paste we can add 1 more chain, and suffix it with `-interop`, to produce a 2-chain test environment.
+For early testing this was drafted in [PR 10777](https://github.com/ethereum-optimism/optimism/pull/10777).
+
+This does not yet unify the common superchain part of the deployment, or any of the artifacts,
+nor supports test-specific changes to parameters.
+
+## Presets and configs
+
+In Eth2 there exist two classes of configuration:
+- "preset" (named, enshrined, compile-time)
```
tynes, line 114: I would like to make more config optional like in https://github.com/ethereum-optimism/optimism/pull/10628 to reduce footguns and make it more simple to deploy a standard config chain

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+it re-introduces smart-contract deployment concepts into the Go chain-ops.
+E.g. storage-layouts, storage packing, the occasional `immutable` variable, all make customization difficult.
+Not to mention cases where we need entirely different contracts, different hardforks, or more L2 deployments.
+
+The insight here is that we should strongly contain smart-contract configuration to the solidity-scripting side,
+and minimize post-processing in Go.
+
+## Mega test deployment
+
+Instead of generating many separate files per configuration,
```
tynes, line 138: Ah i like this, so basically we would have a single L1 world state with many L2 chains deployed to it, each configured a bit differently (standard, plasma, plasma generic, custom gas token). The deploy script outputs a json file with key/value pairs of contract names -> addresses. That could be the view into a particular system. We could add a new entrypoint to the deploy script that uses `loadAllocs` to read in a previous state, we could also use `CONTRACT_ADDRESSES_PATHS` to populate the implementation addresses in the deploy script so that we don't have to redeploy the same implementations each time

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Result
+
+This approach enables us to get rid of "mega-feature" env-vars (Plasma, fault-proofs, Interop)
+as well as allocs hardfork-suffixes: everything is configuration,
+and we can just fork of existing configuration to tweak something.
+
+The main changes to the deploy/genesis-scripting are to:
+1. Get rid of L1 and L2 "allocs": every incremental output is a small "allocs", identified by the hash of inputs.
+2. Get rid of env-vars: we make all customization the same type of well-defined input.
```
tynes, line 231: Do you mean env-vars as in the env vars used in the docker compose devnet?

protolambda: toggles like plasma-mode and the FP<>OutputOracle difference in the devnet. But also env-vars that substitute in config changes / file-paths during the solidity script execution.

protocol/deployment-chains.md
protolambda: Agreed, I think we should do so through composition, to encapsulate things that are rarely modified. But not entirely freeze it either.

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+deployments also prevents multiple L2s from running.
+
+The main goal is to unbundle these steps, and make them customizable and composable for better test-coverage,
+including Interop parameters specifically.
+
+## OPStackManager
+
+Specs: https://specs.optimism.io/experimental/op-stack-manager.html
+
+Older implementatipn draft: https://github.com/ethereum-optimism/optimism/pull/9985
```
mds1, line 56:
```suggestion
Older implementation draft: https://github.com/ethereum-optimism/optimism/pull/9985
```

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+deploy config.
+
+## Deployer requirements
+
+### Op-e2e needs
+
+As an op-e2e test suite, I need to:
+- Customize the deploy-config deeply, including L1 parameters and upgrade parameters, to cover all functionality.
+- Create multiple L2s with a shared L1 (to go from isolated L2s to interoperating L2s through interop hardfork).
+- Create multiple L2s with a shared L1 and shared superchain (to test interop without emulating hardfork first).
```
mds1, line 76: What does "shared superchain" mean? My prior definition would have been "The L1 contracts for each L2 points to the same SuperchainConfig contract", but from these requirements I think you mean "The L2s are interoperable"

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Devnet needs
+
+As a devnet user, I need to:
+- Modify the chain template, for experimentation when building new features.
```
mds1, line 81: What is the chain template, is that what is currently the deploy config?

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+- Support different "modes" without accidental configuration mixup,
+  e.g. I may to run different hardforks, or run with different DA types.
+
+### Production needs
+
+As a production superchain host, I need to:
+- Perform a new Superchain core deployment (global contracts and implementation-contract deployments).
+- Maintain the Superchain core deployment with upgrades.
+- Not deploy/upgrade an L2 at the same time.
```
mds1, line 90: Can you expand on this one? Is this just saying we keep deployments and upgrades as distinct processes?

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+
+#### Addressing gaps
+
+##### How do we make this practical (fast) in tests?
+
+We can run it once, and cache the result, if it is deterministic, and if time-specifics can be adjusted later.
+
+##### How do we make this composable?
+
+We group the inputs, per deployment step, to not do everything at once.
```
mds1, line 167: This is great. One thing I really don't like about the current deploy config is that all types of inputs are treated equally—some inputs are set once at genesis and never change, others are config params that can be changed anytime by a privileged role, etc. A clean grouping of inputs will be very helpful

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+##### How does this work with production deployments?
+
+The caching, chaining, and commiting is not a part of production.
+As it is against a live chain, not a local state-dump.
+
+However, the isolated configuration and deployment-sub-routines do enable
+a deployer to plug into the superchain contracts, while customizing just what they need,
+without having to untangle a monolithic single deployment script.
+
+Ideally this is performed through the OPStackManager.
```
mds1, line 221: Agreed with OPSM usage here. In general we should architect things such that the default, least-effort way of deploying anything results in a standard chain at the end (this is what the OPSM spec currently does). Anything else should require additional work by the user to configure.

The rough idea I had in mind for the L1 contracts deployed by OPSM is to add hooks or flags that are enabled/disabled based on the config, which can be implemented in OPSM [feature contracts](https://github.com/ethereum-optimism/design-docs/blob/5d98f580358c87efc5e402d92027f0855b239af9/smart-contract-feature-development.md).

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Result
+
+This approach enables us to get rid of "mega-feature" env-vars (Plasma, fault-proofs, Interop)
+as well as allocs hardfork-suffixes: everything is configuration,
+and we can just fork of existing configuration to tweak something.
+
+The main changes to the deploy/genesis-scripting are to:
+1. Get rid of L1 and L2 "allocs": every incremental output is a small "allocs", identified by the hash of inputs.
```
mds1, line 230: Does this mean we're changing the genesis file format used by the node, or just changing the development format, and it gets postprocessed into the standard format for production deploys?

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+2. Get rid of env-vars: we make all customization the same type of well-defined input.
+3. Split the deploy-config: each deployment sub-routine should have its own well-contained config,
+  such that we can attribute the output to the change of input.
+
+The Go `op-e2e` side should then be able to produce the needed configuration input-groups, hash them,
+and then either hit the cache of outputs, or invoke the solidity-scripts to generate the outputs.
```
mds1, line 236: One thing to be careful about is how we handle the contract bytecode in genesis, and how to release it.

Currently, because of our monorepo structure, node releases (which encompass predeploys via genesis) and contracts releases (which to date have been only scoped to L1 contracts) are independently versioned and released. We want to make sure that as we change/add predeploys:
1. They are clearly released as either part of the node releases, the contract releases, or a new release set. We should decide on their release process and document it in the monorepo READMEs or [VERSIONING.md](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/VERSIONING.md)
2. When generating genesis files for development, we can always use the latest predeploy contract bytecode. But when generating genesis files for production, we must ensure only released versions are used.

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+1. Get rid of L1 and L2 "allocs": every incremental output is a small "allocs", identified by the hash of inputs.
+2. Get rid of env-vars: we make all customization the same type of well-defined input.
+3. Split the deploy-config: each deployment sub-routine should have its own well-contained config,
+  such that we can attribute the output to the change of input.
+
+The Go `op-e2e` side should then be able to produce the needed configuration input-groups, hash them,
+and then either hit the cache of outputs, or invoke the solidity-scripts to generate the outputs. 
+
+Backwards compatibility can be achieved by adding a function that loads a legacy deploy-config,
+and applies the sequence of all deployment steps, to generate what would have previously been generated.
```
mds1, line 239: Do we need backwards compatibility, or can just regenerate previous outputs by checking out an old monorepo commit?

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+
+The OPStackManager spec and draft already strongly point the L2-instance-specific
+deployment of L1 contracts to a subset of deploy-configuration.
+At this time that subset is under-specified (misses configuration, no fault-proof support, etc.),
+but it could utilize the L2-instance specific input-group and deployment routine, to achieve what it needs.
```
mds1, line 244: Yea this is correct, the spec is currently focused on deploying standard chains specifically. It removes all unnecessary inputs and only exposes allowable standard config params towards that goal.

An open question is how to expand it to support other configurations, with the rough idea mentioned above in https://github.com/ethereum-optimism/design-docs/pull/52/files?diff=unified&w=0#r1692407267. Architecting the hooks or flags based on the resulting input groups that come out of this is likely a good solution

protocol/deployment-chains.md
protolambda: There is a `devnetL1_template.json` file, that turns into a deploy-config, as part of the devnet. But it's a singular template, not possible to change per test.

protocol/deployment-chains.md
protolambda: Yes, distinct functions. Such that the following two are separate steps:
- the superchain global contracts (SuperchainConfig, implementations)
- the proxies for L2-specific chain, and proxyadmin

Step 1 is rare, only for conflicting superchain setups. Step 2 is what most chains only really need.
Also, I think we need to fix bad default overlaps like the current deployment a singular proxy-admin that is shared between these two different domains.

protocol/deployment-chains.md
protolambda: We need some form of compatibility, to preserve the current unit-test setup approach in solidity tests

protocol/deployment-chains.md
protolambda: This is all cached by content-hash, and testing is always on latest develop, the versioning part is out of scope.

protocol/deployment-chains.md
protolambda: Resolving this comment, as we discussed this outside of this PR. In summary it boils down to lots of book-keeping of contract addresses and their configurations, and remapping issues with predeploys on fixed addresses.

protocol/deployment-chains.md
protolambda: I'll clarify the wording. Yes it's meant to identify the L1 contracts that are common between different L2 instances. Like the `SuperchainConfig`, `ProtocolVersions`, and the implementations that proxies attach to.

protocol/deployment-chains.md
```diff
@@ -0,0 +1,261 @@
...
+deploy config.
+
+## Deployer requirements
+
+### Op-e2e needs
+
+As an op-e2e test suite, I need to:
+- Customize the deploy-config deeply, including L1 parameters and upgrade parameters, to cover all functionality.
+- Create multiple L2s with a shared L1 (to go from isolated L2s to interoperating L2s through interop hardfork).
+- Create multiple L2s with a shared L1 and shared superchain (to test interop without emulating hardfork first).
```
protolambda, line 76:
```suggestion
- Create multiple L2s with a shared L1 and shared superchain L1 contracts, like a shared SuperchainConfig and ProtocolVersions contract, and implementations list (to test interop without emulating hardfork first).
```

protocol/deployment-chains.md
protolambda: Changing an intermediate development format only. The solidity scripts would output a regular state-dump, the Go op-chain-ops tooling would turn that into cached state diffs, and the genesis-generation can then load those diffs into an actual full genesis.

protocol/deployment-chains.md
mds1: I see, thanls

tynes:
@protolambda should we close this in favor of #61?

PR #53, opened 2024-07-24
WIP: Builder Separation Design Doc

axelKingsley:
WIP

### DIFF
--- a/protocol/builder-separation.md
+++ b/protocol/builder-separation.md
@@ -1,0 +1,115 @@
+# Purpose
+The purpose of this document is to lay out and ratify all major technical aspects of Publisher-Builder-Separation (PBS)
+as it relates to the OP-Stack.
+
+# Summary
+# Problem Statement + Context
+Today, Chain Operators on OP use a Sequencer to publish new blocks to the tip of the chain. The creation of blocks happens through the Execution Engine which is paired to the Sequencer. When a new block is needed, the Sequencer issues commands to the Execution Engine to Start building a block, and will seal that block at the end of the 2s block window. Once done, this completed block is used to update the Execution Engine's chain, and then is published outward onto the network via P2P Gossip.
+
+Assigning the resposibility of block building to the Sequencer's Execution Engine makes sense. However, there are additional goals which can't be achieved in this arrangement. In particular, a customer who would like a *different* block building policy today must modify the behavior of the Execution Engine. This heavy lift is undesirable, and by Separating the Builder from the Publisher, we introduce a new dimension of modularity and extensibility which is desirable to the ecosystem.
+
+Publisher Builder Separation (PBS) is an architectual feature of a Sequencer which allows for block building to happen in a location *other than* the Sequencer's Execution Engine. By introducing this separation, we can create novel opporunities for innovation in block building on OP Chains.
+
+## What isn't PBS
+Because PBS is somtimes used ambiguously, it should be made clear: PBS, and this project, are not concerned with the particular behavior of a block builder, or builder(s), beyond the mechanisms the Sequencer uses to interface with them. Builder designs can be as simple or as intricate as operators would like. This design is meant to be supportive of all block building systems, but does not enshrine or plan for a particular style.
+
+PBS is *only* the abstraction of block-building as an activity, which *allows for* blocks to be used in the system when they originate from somewhere other than the Execution Engine. The "somewhere else" is considered opaque for the purposes of this document, except when the its properties inform the design of the interface.
+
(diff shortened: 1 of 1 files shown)


protocol/builder-separation.md
```diff
@@ -0,0 +1,115 @@
...
+    Block Builder->>Sequencer: Deliver Block Asynchronously
+        Note over Sequencer: Block Building Deadline Passes
+    Sequencer->>Execution Engine: Seal Block
+        Note over Sequencer: Pick which block should be used
+    Sequencer->>Execution Engine: FCU Block onto Execution Engine
+    Sequencer->>Replicas: Publish Block
+```
+To break down the components of this diagram:
+
+**Transactions are sent to the Execution Engine and Block Builder via Proxyd**.  Today, the Execution Engine is responsible for being informed of Transactions and maintaining a mempool. With PBS, the Block Builder would have the same responsibilities. The Sequencer *does not* forward Transactions. This is because the Sequencer is not involved in the block building activities directly, and does not benefit from acting as a pass-through
```
tynes, line 74: We don't necessarily need to be opinionated about the orderflow. This architecture makes sense for an OP Mainnet style network, but you could imagine many architectures around the order flow

protocol/builder-separation.md
```diff
@@ -0,0 +1,115 @@
...
+    participant Execution Engine
+    participant Sequencer
+    participant Replicas
+
+    Customer->>+ProxyD: Send Transaction
+    ProxyD->>+Block Builder: Forward Transaction
+    ProxyD->>+Execution Engine: Forward Transaction
+
+        Note over Sequencer: On a regular basis...
+    Sequencer->>Block Builder: Start Building <Attributes>
```
tynes, line 60: I am not sure what makes the most sense here with the particular flow to be honest, i naively thought that the block builder would ask the sequencer for the same info that the sequencer pushes to the EL when it tells it to build a block, then the block builder would push a block to the sequencer and the sequencer would then have a choice to seal either the local block or the block builders block

tynes:
It could be good to read + link the following doc: https://gov.optimism.io/t/optimism-pbs-proof-of-concept/6718

protocol/builder-separation.md
axelKingsley: I agree, referencing `proxyd` here is an over-constraint of the design which makes it OP Mainnet specific. The purpose of calling that mechanism out is:
* `proxyd` *is* the reference standard for delivering transactions to Execution Engines today, and the way in which transactions are delivered to Execution Engines should be matched when delivering to Block Builders.
* Sequencers `op-node` should not have to process or route transaction traffic to other components.

I will expand on this to make it clear, though it may still be useful to reference `proxyd` as the entry-point for customer transactions.

protocol/builder-separation.md
axelKingsley: Yeah, all sorts of sync/async push/pull permutations are possible for sure, and none of them are outright wrong. However, I think this particular arrangement is good and I justify it as:

*Sequencer should Push Block Attributes to BBs* **because** the sequencer has the most timely information regarding a new block needing to be built. If a Block Builder was expected to pull attributes from the Sequencer, it might accidentally do it too late, wasting time at the start of the block-period.

*BBs should Push Payloads to the Sequencer* **because** the sequencer should not add extra delay in the creation of blocks waiting for the BB. If the BB delivers a payload in time, great. If the BB is unperformant, it should not impact block production. This also gives the BB the opportunity to send out blocks iteratively, or to send the block early if it is done building. We *could* make this a pull from the sequencer instead, doing the call in a separate thread, and abandoning if the payload doesn't come back, but seeing as we have this nice Event-Driven Derivers architecture, I wanted the builder to be able to trigger an event workflow within the sequencer to consider the BB payload.

The "Notification of Deadline" message is a timing mechanism meant to help synchronize the BB and keep it from spending too long building, only to find that the sequencer has moved on.

We could totally consider other arrangements of these responsibilities, I would just want to consider what advantages they give us.

----

> the sequencer would then have a choice to seal either the local block or the block builders block

This might be a gap on my understanding of block-sealing. It was my impression that the BB could seal blocks it creates, and therefore both the EE and BB would be sealing their own blocks. But now that you point it out, I bet that there's a signing authority thing that has to happen with blocks, and in that case: *yes*, we should rearrange this diagram somewhat. The BB and EE should both provide *unsealed* payloads to the Sequencer, and then the sequencer chooses which one gets sealed by the EE, and all other steps continue as normal.

tynes:
Closing in favor of https://github.com/ethereum-optimism/design-docs/pull/86

PR #54, opened 2024-07-30
GovernanceDelegation Interop

corydickson:
**Description**

This design doc includes details about a potential update to the Governance Delegation contract allowing for 
interoperability of partial delegations.


tynes:
We will schedule this for design review on Thursday

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+a want to later batch updates of voting power to mainnet it would be costly to those operating governance.
+
+Another active consideration is the use of block timestamp of each L2 instead of block number while implementing this new solution. The primary reason for block numbers to be used in
+its stead, stems from a concern of manipulation by sequencers and the possible delays that might need to be enforced between events to ensure correctness.
+
+# Proposed Solution
+
+## Rationale
+
+Due to the fact that the `GovernanceDelegation` contract now has awareness of the block numbers and chain IDs of all of the voting power checkpoints each L2 can now be responsible for
```
tynes, line 34: Can you expand on this awareness a bit? I see some mention of it in the overview. It could be useful to have some simple data structures in here to make it easier to visualize.

Also i would like to avoid branching logic based on the chainid. It would be most ideal if the events could be imported into any chain but with no real incentive to import them anywhere besides op mainnet. This will simplify testing, remove edge cases and enable delegation power to be observed from any chain

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+their own checkpointing and thus make as many updates necessary without cost to relayers. From an implementation standpoint many of the changes are quite small and should minimize the
+risk of complication of an already intricate interface.
+
+The main goal is to ensure that the incentives are set up correctly so that relayers and indexers can manage and provide users the latest delegation changes when exercising their voting
+power.
+
+## Implementation
+
+Again, the implementation involves updates the Governor that will ensure that for voting power that has been migrated can be retrieved and updated based on the state on each of the L2 chains. It accomplishes this by keeping track of the token nonce and block numbers of each checkpoint in the mainnet contract. Instead of directly sending messages between the different instances, Governors on non-OP mainnet chains will now emit events regarding these blocks in the their checkpoint updates.
```
tynes, line 43: When you say `Governor`, do you mean `GovToken`?

smartcontracts: The `Governor` contract is an additional contract that already exists to tally up voting power, essentially. It's what handles voting outside of the `GovToken` contract. It makes the most sense to have the `Governor` be responsible for tallying up voting power for a particular vote.

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+However, pushing all of the logic onto the `GovToken` proves to be problematic given the potential flood of messages that may be out of sequence or have many smaller transfer events
+prior to a single delegation event. The following section aims to outline a solution to correct this as well as ensure that relayers still have the proper incentives to handle these
+messages.
+
+# Alternatives Considered
+
+When originally considering this problem, the idea was to create a hook during the `afterTokenTransfer` event. As alluded to above, this do not provide some of the desired functionality.It also makes a few assumptions about the behaviors of delegates that receive tokens and stifles their ability to create partial delegations in a low-cost manner. Moreover, if there was
+a want to later batch updates of voting power to mainnet it would be costly to those operating governance.
+
+Another active consideration is the use of block timestamp of each L2 instead of block number while implementing this new solution. The primary reason for block numbers to be used in
```
smartcontracts, line 27: We still need to determine if we're using L1 block number or timestamp.

Main reason to use L1 block number is that it makes the process of using L1 => L2 transactions for voting very easy to think about. This is important in the censorship case. If you want your vote to count, you know that you must include it in an L1 => L2 transaction by L1 block N.

Main reason to use timestamp is that it scales to chains where the L1 block number is not available in the same way.

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+to be processed on the mainnet contract. Each L2 will now have the total cumulative voting power for every partial delegation. Vote retrieval on mainnet will take in a parameter that specifies the height of the latest block it cares about of a given L2 chain.
+
+
+## Invariants
+
+The implementation should maintain the following invariants:
+
+1. **Awareness of migration status**: To ensure that existing voting power is preserved that do not have L1 block number.
+2. **Checkpoint mapping**: A new storage field that contains the block number and L2 chain IDs of the latest checkpoint.
+3. **Nonce on token transfers**: The governance token will now employ a nonce on each transfer that will be checked in the `afterTokenTransfer` hook.
```
smartcontracts, line 56: I don't think we need this anymore

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+
+# Risks & Uncertainties
+
+1. The biggest risk in implementing this change is ensuring that all existing voting power is preserved such that the checkpoints that already exist on the contract is accessible.
+2. Indexers must now accept a delay in updates if wishing to give an accurate voting status for a particular delegate.
+3. Minimize the griefing vector of incurring costs onto delegators.
+4. Estimation of costs of operating relayers to ensure they are still incentivized to relay messages.
```
smartcontracts, line 63: We don't need to handle this anymore

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+
+1. **Awareness of migration status**: To ensure that existing voting power is preserved that do not have L1 block number.
+2. **Checkpoint mapping**: A new storage field that contains the block number and L2 chain IDs of the latest checkpoint.
+3. **Nonce on token transfers**: The governance token will now employ a nonce on each transfer that will be checked in the `afterTokenTransfer` hook.
+
+# Risks & Uncertainties
+
+1. The biggest risk in implementing this change is ensuring that all existing voting power is preserved such that the checkpoints that already exist on the contract is accessible.
+2. Indexers must now accept a delay in updates if wishing to give an accurate voting status for a particular delegate.
+3. Minimize the griefing vector of incurring costs onto delegators.
```
smartcontracts, line 62: Also no need to handle this anymore with the new design

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+4. Estimation of costs of operating relayers to ensure they are still incentivized to relay messages.
+5. Batch ordering of checkpoints across all L2 chains.
```
smartcontracts, line 64: Hmmm not sure what this means, can you elaborate?

skeletor-spaceman:
@corydickson 
We've re-read the document a few times and it's still a bit confusing.
It'd be great to have a figma or detailed flows on how you are envisioning the technical implementation being.

governance/delegation-interop.md
fainashalts: Might be a noob question, but what do you mean the "L1 block number is not available in the same way"?

governance/delegation-interop.md
fainashalts: Perhaps this means to batch process the checkpoints on each L2 to reduce frequency & cost of updates (especially when multiple small transfers occur). At least I think that is what it means, and would like to see that as part of the design. :) I feel like this may be tricky though since batching could delay the update to `Governor` on mainnet maybe?

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+The implementation should maintain the following invariants:
+
+1. **Awareness of migration status**: To ensure that existing voting power is preserved that do not have L1 block number.
+2. **Checkpoint mapping**: A new storage field that contains the block number and L2 chain IDs of the latest checkpoint.
+3. **Nonce on token transfers**: The governance token will now employ a nonce on each transfer that will be checked in the `afterTokenTransfer` hook.
+
+# Risks & Uncertainties
+
+1. The biggest risk in implementing this change is ensuring that all existing voting power is preserved such that the checkpoints that already exist on the contract is accessible.
+2. Indexers must now accept a delay in updates if wishing to give an accurate voting status for a particular delegate.
```
fainashalts, line 61: Will there be a canonical indexer we run? Or is the idea that 3rd party services will index this data for us? If indexers track and store checkpoint into we should be sure this infra is available and reliable.

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+
+The main goal is to ensure that the incentives are set up correctly so that relayers and indexers can manage and provide users the latest delegation changes when exercising their voting
+power.
+
+## Implementation
+
+Again, the implementation involves updates the Governor that will ensure that for voting power that has been migrated can be retrieved and updated based on the state on each of the L2 chains. It accomplishes this by keeping track of the token nonce and block numbers of each checkpoint in the mainnet contract. Instead of directly sending messages between the different instances, Governors on non-OP mainnet chains will now emit events regarding these blocks in the their checkpoint updates.
+
+The new storage values will be used to keep track of both the voting power that has been accumulated on the native chain as well as others thus not breaking backwards compatibility for
+non-migrated tokens. Indexers will now need to be aware of the events being emitted on both OP mainnet and the other L2's that implement the `GovToken` to show checkpoints that have yet
```
fainashalts, line 46: So indexers need to watch every possible L2 that has governance token delegation? Just trying to understand the flow:
1. Event happens on L2 chain, changing delegation.
2. `GovToken` contract emits event in its checkpoint update.
3. `Governor` contract is aware of checkpoint updates(I'm not totally clear on how this works, is it using interop messaging to do this?)
4. `Governor` updates state and checkpoints and this info is indexed.
5. But indexer is _also_ listening for the L2 event emission because it should know about checkpoints from all L2s even if they are not on mainnet yet?

What happens if there is a delegation change on an L2 but it hasn't yet been submitted to mainnet? What is the source of truth if there is a vote in that scenario? It could be problematic if the onchain source of truth doesn't match the indexer/offchain source of truth, right? I think some sequence diagrams would be helpful here to visualize the various components and how they interact.

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+the latest checkpoint and the block number. After tokens are transferred using the `GovToken` contract a check is performed on the delegation contract that ensures that before this
+mapping is updated, the nonce of the token transfer received has been incremented as well as being sent from a block number greater than the current one.
+
+# Problem Statement + Context
+
+In the existing `GovernanceDelegation` and `GovToken` contract it does not support interoperability, thus making the voting power fragmented across all of the potential L2 chains. The negative consequence of this is that tracking delegations becomes a much harder task when users have partially delegated to many different users that may or may not be on the same chain.
+
+However, pushing all of the logic onto the `GovToken` proves to be problematic given the potential flood of messages that may be out of sequence or have many smaller transfer events
+prior to a single delegation event. The following section aims to outline a solution to correct this as well as ensure that relayers still have the proper incentives to handle these
```
mds1, line 19: Can you expand on this problem? I don't follow the "problematic given the potential flood of messages that may be out of sequence or have many smaller transfer events" aspect

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+messages.
+
+# Alternatives Considered
+
+When originally considering this problem, the idea was to create a hook during the `afterTokenTransfer` event. As alluded to above, this do not provide some of the desired functionality.It also makes a few assumptions about the behaviors of delegates that receive tokens and stifles their ability to create partial delegations in a low-cost manner. Moreover, if there was
+a want to later batch updates of voting power to mainnet it would be costly to those operating governance.
```
mds1, line 25: What would this `afterTokenTransfer` event have done? Would be great to expand on the alternatives in more detail for readers that may not have the full context (like me 😅)

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
...
+## Rationale
+
+Due to the fact that the `GovernanceDelegation` contract now has awareness of the block numbers and chain IDs of all of the voting power checkpoints each L2 can now be responsible for
+their own checkpointing and thus make as many updates necessary without cost to relayers. From an implementation standpoint many of the changes are quite small and should minimize the
+risk of complication of an already intricate interface.
+
+The main goal is to ensure that the incentives are set up correctly so that relayers and indexers can manage and provide users the latest delegation changes when exercising their voting
+power.
+
+## Implementation
```
mds1, line 41: > It'd be great to have a figma or detailed flows on how you are envisioning the technical implementation being.

+1 to this comment by @skeletor-spaceman, this would be a big help to understanding the proposal. What happens during some specific user stories (same chain token transfer, cross-chain token transfer, same-chain delegation, cross-chain delegation, voting, etc) would be really great

governance/delegation-interop.md
```diff
@@ -0,0 +1,65 @@
+# Purpose
+
+The primary purpose of this design document is to give an approach to support voting power delegation across multiple L2's that employ the `GovernanceDelegation` and `GovToken` as a
```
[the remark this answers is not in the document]
corydickson: Correct I'll be supplying links

governance/delegation-interop.md
corydickson: The goal here was to provide some context around the previous solution which had a hook `afterTokenTransfer` in which a message was sent between L2's to OP main net that then updated the checkpoints. This is no longer pertinent to the design but was trying to lay the groundwork of some prior thinking around supporting this feature. Here's the PR of the old spec that it looks like we are moving away from: https://github.com/ethereum-optimism/specs/pull/285

governance/delegation-interop.md
corydickson: I posted the link in the previous comment, I'll be sure to coalesce all the existing documentation here.

governance/delegation-interop.md
corydickson: Was unsure if pseudocode was bad style for a design doc, but specifically I was considering some type of mapping in solidity between `mapping ( uint16 => Checkpoint ) ` where the data structure is `ERC20Votes.Checkpoint[]` contains an additional field `blockNumber`

governance/delegation-interop.md
corydickson: I will include some diagrams in the document. But your understanding is correct, the source of truth is always the L2 but the block number inclusion allows votes to be counted on Mainnet for up to N specified blocks.

governance/delegation-interop.md
corydickson: Good question, I'm unsure if there is going to be a single indexer for all the chains but I do imagine at the very least each L2 runs their own indexer.

governance/delegation-interop.md
corydickson: Yes this would be exposing some batching function to process the messages in the cross message passing Inbox

governance/delegation-interop.md
```diff
@@ -0,0 +1,64 @@
+# Purpose
+
+The primary purpose of this design document is to give an approach to support voting power delegation across multiple L2's that employ the `GovernanceDelegation` and `GovernanceToken` as pre-deploys. It also gives context around the desired behavior of the interactions between the contracts that exist on both OP mainnet as well as each of these bespoke chains. Furthermore, this document serves as an exploration into a few choice mechanisms that allow users the ability to do partial delegation with tokens they posses on each of the respective chains.
```
cairoeth, line 3:
```suggestion
The primary purpose of this design document is to outline an approach to support cross-chain voting power delegations of the `GovernanceToken`. This document also aims to provide context around the desired behavior of the interactions between the contracts that exist on both OP mainnet as well as each of these bespoke chains. Furthermore, user flows and use cases are outlined to provide a more comprehensive understanding of the proposed changes.
```

governance/delegation-interop.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Summary
+
+There will be two main changes to the existing `GovernanceDelegation` [contract](https://github.com/ethereum-optimism/specs/blob/main/specs/experimental/gov-delegation.md) to support interoperability. First, the function in which the checkpoints can be queried from the contract will now check for the migration status prior to retrieving voting power and implement an announce function that will emit an event that contains the chain ID and the block number of the latest checkpoint. The contract on OP mainnet will then consume these events, and in its own storage provide a mapping between chainIds and an object containing both the latest checkpoint and the block number. After tokens are transferred using the `GovernanceToken` [contract](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/governance/GovernanceToken.sol) a check is performed on the delegation contract that ensures that before this mapping is updated, the nonce of the token transfer received has been incremented as well as being sent from a block number greater than the current one.
```
cairoeth, line 7:
```suggestion
There will be two main changes to the existing `GovernanceDelegation` [contract](https://github.com/ethereum-optimism/specs/blob/main/specs/experimental/gov-delegation.md) to support interoperability. First, the function in which the checkpoints can be queried from the contract will now check for the migration status prior to retrieving voting power and implement an announce function that will emit an event that contains the chain ID and the block number of the latest checkpoint. The contract on OP mainnet will then consume these events, and in its own storage provide a mapping between chainIds and an object containing both the latest checkpoint and the block number. After tokens are transferred using the `GovernanceToken` [contract](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/governance/GovernanceToken.sol) a check is performed on the delegation contract that ensures that before this mapping is updated, the nonce of the token transfer received has been incremented as well as being sent from a block number greater than the current one.
```

governance/delegation-interop.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Problem Statement + Context
+
+The existing `GovernanceDelegation` and `GovernanceToken` contracts do not support interoperability, thus making the voting power fragmented across all of the potential L2 chains. The negative consequence of this is that tracking delegations becomes a much harder task when users have partially delegated to many different users that may or may not be on the same chain.
+
+However, pushing all of the logic onto the `GovernanceToken` proves to be problematic given the potential flood of messages that may be out of sequence or have many smaller transfer events prior to a single delegation event. The following section aims to outline a solution to correct this as well as ensure that relayers still have the proper incentives to handle these
```
cairoeth, line 13:
```suggestion
As described in the [advanced delegation](advanced-delegation.md) design document, the `GovernanceToken` contract supports advanced delegation by integrating with the `GovernanceDelegation` contract. However, the `GovernanceDelegation` contract does not support interoperability, which means that when the `GovernanceToken` gets deployed across networks in the Superchain, the total voting supply in OP Mainnet may decrease as the token would be fragmented across other networks.

Therefore, this design document aims to modify the `GovernanceDelegation` contract to support cross-chain delegations of voting power, maximazing user experience and ensuring that relayers have the proper incentives to handle cross-chain messages of delegations.
```

cairoeth (commented):
some suggestions to clarify the first sections with more specific context and target problem

governance/delegation-interop.md
```diff
@@ -0,0 +1,64 @@
...
+
+The new storage values will be used to keep track of both the voting power that has been accumulated on the native chain as well as others thus not breaking backwards compatibility for non-migrated tokens. Indexers will now need to be aware of the events being emitted on both OP mainnet and the other L2's that implement the `GovernanceToken` to show checkpoints that have yet to be processed on the mainnet contract. Each L2 will now have the total cumulative voting power for every partial delegation. Vote retrieval on mainnet will take in a parameter that specifies the height of the latest block it cares about of a given L2 chain.
+
+```mermaid
+sequenceDiagram
+  participant delegate
+  participant GovernanceDelegationA as GovernanceDelegation (Chain A)
+  participant GovernanceDelegationB as GovernanceDelegation (Chain B)
+  participant Inbox as CrossL2Inbox
+  participant Messenger_A as L2ToL2CrossDomainMessenger (Chain A)
```
tynes, line 43: Its possible that we can design the system in a way where we don't need the `L2ToL2CrossDomainMessenger` since that contract gives domain binding + replay protection. Do we need both of those things? I think a UDP broadcast sort of design is better, where we can tally up the delegations on any chain

tynes:
We are going to want a design doc for Governor

governance/delegation-interop.md
smartcontracts: Yeah we don't need the `L2ToL2CrossDomainMessenger`

skeletor-spaceman:
I'd love it if we could include a section of the potential drawbacks and added complexity of this approach.
Since we are tracking the digested voting power on each chain separately, does this mean that as a token holder, I would need to re-execute the `delegate(...)` function on each chain to be sure those tokens are being delegated as well?
and if I choose to change them, i'd need to do so on all chains as well?

it'd be great to know which potential pitfalls or complexities we might encounter as we implement and integrate this.

tynes:
Will schedule this for design review on Monday

governance/delegation-interop.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Summary
+
+There will be two main changes to the existing `GovernanceDelegation` [contract](https://github.com/ethereum-optimism/specs/blob/main/specs/experimental/gov-delegation.md) to support interoperability. First, the function in which the checkpoints can be queried from the contract will now check for the migration status prior to retrieving voting power and implement an announce function that will emit an event that contains the chain ID and the block number of the latest checkpoint. The contract on OP mainnet will then consume these events, and in its own storage provide a mapping between chainIds and an object containing both the latest checkpoint and the block number. After tokens are transferred using the `GovernanceToken` [contract](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/governance/GovernanceToken.sol) a check is performed on the delegation contract that ensures that before this mapping is updated, the nonce of the token transfer received has been incremented as well as being sent from a block number greater than the current one.
```
tynes, line 7: > the function in which the checkpoints can be queried from the contract will now check for the migration status prior to retrieving voting power and implement an announce function that will emit an event that contains the chain ID and the block number of the latest checkpoint

I am confused about this, generally a query implies a `view` function (ie `STATICCALL`) meaning no state changes allowed, including events being emitted

tynes (approved):

PR #60, opened 2024-08-05
OP Stack Manager architecture to support interop

mds1:
This design doc details an architecture for OP Stack Manager that ideally fits the modular interop use case expressed in #52 that we move towards in https://github.com/ethereum-optimism/optimism/pull/11189.  It focuses on L1 contracts only. Our current deploy config and deploy script is a monolith, so here we break it up into 3 discrete steps, each with their own distinct inputs and outputs.

### DIFF
--- a/protocol/op-stack-manager-arch.md
+++ b/protocol/op-stack-manager-arch.md
@@ -42,11 +42,11 @@ This is not ideal for various reasons:
 
 # Proposed Solution
 
-There are two steps to deploying L1 contracts:
+There are three aspects to deploying L1 contracts:
 
-1. **Deploy Superchain Contracts**. This only occurs occasionally, as the Superchain contracts are shared between many OP chains.
+1. **Deploy Superchain Contracts**. Superchain contracts are shared between many OP chains, so this occurs only occasionally in production.
 2. **Deploy Shared Implementation Contracts**. This occurs once per [contracts release](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/VERSIONING.md).
-3. **Deploy OP Chain Contracts**. This occurs for once every OP chain deployment.
+3. **Deploy OP Chain Contracts**. This occurs for every OP chain deployment.
 
 For each step we define the step inputs (configuration), sub-steps that occur, and the step outputs (artifacts).
 For each we start by listing the sub-steps, then define the inputs and outputs for each sub-step, and use those to define the inputs and outputs for the step as a whole.
--- a/protocol/op-stack-manager-arch.md
+++ b/protocol/op-stack-manager-arch.md
@@ -1,0 +1,292 @@
+# Purpose
+
+OP Stack Manager is a contract that will deploy the L1 contracts for standard OP Stack chains in a single transaction.
+However, we also have use cases for deploying non-standard chains, such as deploying devnet and testnet contracts while developing interop.
+This document will outline an architecture that:
+
+- Enables the OP Stack Manager to deploy these non-standard chains as an initial milestone.
+- Gets us most of the way towards supporting standard chains deployments for the subsequent milestone.
+
+# Summary
+
+The original OP Stack Manager (OPSM) milestones were:
+
+- Milestone 1: Eliminating key handover complexity.
+- Milestone 2: Simplifying Superchain upgrades.
+- Milestone 3: Supporting interop upgrades.
+
+This design doc proposes the architecture for a Milestone 0.5 of "OPSM is used for all interop devnet & testnet L1 deployments".
+Adding this milestone will allow the interop team to build testing infrastructure that wraps OPSM, unifying our development, testnet, and production testing and deployment processes.
+
+# Problem Statement + Context
+
+*Some text copied from https://github.com/ethereum-optimism/design-docs/pull/52, which is related to this design doc.*
+
+The current L2 chain deployment approach originates from a time with Hardhat, single L1 target, and a single monolithic set of features.
+Since then the system has migrated to Foundry and extended for more features, but remains centered around a single monolithic deploy-config for all its features.
+
+The interop team needs a way to configure new multi-L2 deployments: The number of ways to compose L2s in tests grows past what a single legacy config template can support.
+
+Outside of interop, deployment also seems increasingly complex and opaque, while it does not have to be, due to the same configuration and composability troubles.
+
+Part of the solution to this is described in https://github.com/ethereum-optimism/design-docs/pull/52,and the other part involves restructuring our Solidity config processing and deploy scripts.
+This design doc focuses on the L1 contract configuration and deployment part of the solution.
+
+# Alternatives Considered
+
+The primary alternative is for the interop team to not use OPSM for devnet and testnet deployments.
+This is not ideal for various reasons:
+
+- Duplication of effort: The interop team would need to develop and maintain their own deployment scripts, even though new ones will already be written for OPSM.
+- Divergence: With duplication, it's likely the interop team's deployment scripts diverge from the production deployment scripts, leading to inconsistencies and bugs.
+
+# Proposed Solution
+
+There are two steps to deploying L1 contracts:
+
+1. **Deploy Superchain Contracts**. This only occurs occasionally, as the Superchain contracts are shared between many OP chains.
+2. **Deploy Shared Implementation Contracts**. This occurs once per [contracts release](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/VERSIONING.md).
+3. **Deploy OP Chain Contracts**. This occurs for once every OP chain deployment.
+
+For each step we define the step inputs (configuration), sub-steps that occur, and the step outputs (artifacts).
+For each we start by listing the sub-steps, then define the inputs and outputs for each sub-step, and use those to define the inputs and outputs for the step as a whole.
+
+Sample input and output TOML files are included for each step.
+The exact structure of these files is too low-level to be in-scope for this design doc, so just consider them as examples.
+However, feel free to comment on specifics if you think they're important.
+
+TOML is used because it is what the superchain-registry is standardizing on, it can be parsed by forge and go, it supports comments (unlike JSON), and it does not have the [quirks](https://noyaml.com/) of YAML.
+
+Each step is a separate forge script that takes the input TOML file as the only input, and outputs a TOML file.
+To make the artifact self-contained, the outputs also include the inputs.
+This allows a single artifact to be used to convey all information needed about a deployment.
+
+These forge scripts can be wrapped with go code to facilitate e2e testing by the interop team.
+The go code should leverage the same file-based interface to ensure equivalence between the forge script execution and the go execution.
+
+## Step 1: Deploy Superchain contracts
+
+The sub-steps are:
+
+1. Deploy a `ProxyAdmin`.
+2. Deploy a `Proxy` owned by `ProxyAdmin`, set the `SuperchainConfig` contract as its implementation, and initialize it.
+3. Deploy a `Proxy` owned by `ProxyAdmin`, set the `ProtocolVersions` contract as its implementation, and initialize it.
+
+The inputs for each sub-step are:
+
+1. A `proxyAdminOwner`.
+2. The Guardian address and the initial pause status. (The admin of its `Proxy` is always the `ProxyAdmin` from sub-step 1, so this is not an input).
+3. A `protocolVersionsOwner`, the `requiredVersion`, and the `recommendedVersion`. (The admin of its `Proxy` is always the `ProxyAdmin` from sub-step 1, so this is not an input).
+
+The outputs for each sub-step are:
+
+1. The `superchainProxyAdmin` contract address.
+2. The `superchainConfigProxy` and `superchainConfigImplementation` contract addresses.
+3. The `protocolVersionsProxy` and `protocolVersionsImplementation` contract addresses.
+
+A sample TOML deploy config input for this step might look like:
+
+```toml
+[roles]
+proxyAdminOwner = "0x1234..."
+protocolVersionsOwner = "0x1234..."
+
+[superchain_config]
+guardian = "0x1234..."
+paused = false
+
+[protocol_versions]
+requiredVersion = "0.1.0"
+recommendedVersion = "0.1.0"
+```
+
+A sample TOML output for this step might look like:
+
+```toml
+superchainProxyAdmin = "0x1234..."
+superchainConfigProxy = "0x1234..."
+superchainConfigImpl = "0x1234..."
+protocolVersionsProxy = "0x1234..."
+protocolVersionsImpl = "0x1234..."
+
+# Everything above this comment are the outputs, below are the inputs.
+[roles]
+proxyAdminOwner = "0x1234..."
+protocolVersionsOwner = "0x1234..."
+
+[superchain_config]
+guardian = "0x1234..."
+paused = false
+
+[protocol_versions]
+requiredVersion = "0.1.0"
+recommendedVersion = "0.1.0"
+```
+
+## Step 2: Deploy Shared Implementation Contracts
+
+We assume the latest Fault Proofs release `op-contracts/v1.5.0`.
+The exact set of sub-steps and inputs may change with future releases.
+
+The sub-steps are deploying the following contracts:
+
+1. DisputeGameFactory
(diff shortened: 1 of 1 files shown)


protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,292 @@
...
+*Some text copied from https://github.com/ethereum-optimism/design-docs/pull/52, which is related to this design doc.*
+
+The current L2 chain deployment approach originates from a time with Hardhat, single L1 target, and a single monolithic set of features.
+Since then the system has migrated to Foundry and extended for more features, but remains centered around a single monolithic deploy-config for all its features.
+
+The interop team needs a way to configure new multi-L2 deployments: The number of ways to compose L2s in tests grows past what a single legacy config template can support.
+
+Outside of interop, deployment also seems increasingly complex and opaque, while it does not have to be, due to the same configuration and composability troubles.
+
+Part of the solution to this is described in https://github.com/ethereum-optimism/design-docs/pull/52,and the other part involves restructuring our Solidity config processing and deploy scripts.
```
protolambda, line 32:
```suggestion
Part of the solution to this is described in [design doc 52](https://github.com/ethereum-optimism/design-docs/pull/52), and the other part involves restructuring our Solidity config processing and deploy scripts.
```

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,292 @@
...
+
+The primary alternative is for the interop team to not use OPSM for devnet and testnet deployments.
+This is not ideal for various reasons:
+
+- Duplication of effort: The interop team would need to develop and maintain their own deployment scripts, even though new ones will already be written for OPSM.
+- Divergence: With duplication, it's likely the interop team's deployment scripts diverge from the production deployment scripts, leading to inconsistencies and bugs.
+
+# Proposed Solution
+
+There are two steps to deploying L1 contracts:
```
protolambda, line 45: It lists three steps. The superchain-contracts and shared implementation contracts can (and should) be bundled (at high level at least, implementation can do it step-wise).

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,292 @@
...
+1. **Deploy Superchain Contracts**. This only occurs occasionally, as the Superchain contracts are shared between many OP chains.
+2. **Deploy Shared Implementation Contracts**. This occurs once per [contracts release](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/VERSIONING.md).
+3. **Deploy OP Chain Contracts**. This occurs for once every OP chain deployment.
+
+For each step we define the step inputs (configuration), sub-steps that occur, and the step outputs (artifacts).
+For each we start by listing the sub-steps, then define the inputs and outputs for each sub-step, and use those to define the inputs and outputs for the step as a whole.
+
+Sample input and output TOML files are included for each step.
+The exact structure of these files is too low-level to be in-scope for this design doc, so just consider them as examples.
+However, feel free to comment on specifics if you think they're important.
```
protolambda, line 56: I am fine with TOML, but I'd love some config abstraction, and not directly bundle TOML logic into deploy scripts. We can do small config contracts for usage in scripts, with a default implementation of the config interface that uses the TOML files. Having this interface can make the integration into Go tooling / testing a lot smoother, by not having to go through intermediate scripts on disk.

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,292 @@
...
+
+TOML is used because it is what the superchain-registry is standardizing on, it can be parsed by forge and go, it supports comments (unlike JSON), and it does not have the [quirks](https://noyaml.com/) of YAML.
+
+Each step is a separate forge script that takes the input TOML file as the only input, and outputs a TOML file.
+To make the artifact self-contained, the outputs also include the inputs.
+This allows a single artifact to be used to convey all information needed about a deployment.
```
protolambda, line 62: Like config inputs, using an interface for outputs would also be nice. I.e. an interface of `output(<metadata>)`, and then a contract implementation that does the JSON state dump and any post-processing.

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,292 @@
...
+
+These forge scripts can be wrapped with go code to facilitate e2e testing by the interop team.
```
protolambda, line 64: I am working on a new design doc for this, I believe this can be integrated pretty elegantly. And regardless of integration, the main initial value really is the modular configuration / deploy steps.

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,292 @@
...
+[superchain_config]
+guardian = "0x1234..."
+paused = false
+
+[protocol_versions]
+requiredVersion = "0.1.0"
+recommendedVersion = "0.1.0"
+```
+
+A sample TOML output for this step might look like:
```
protolambda, line 103: See comment about output interface, and separation of the TOML implementation.

For the Go tooling, if we go with the new upcoming design of script execution in Go, we can capture the outputs by handling cheatcode precompile calls and/or regular calls to a no-op contract. So it would be the Go tools interacting with TOML, and only in the genesis CLI tool case. In the op-e2e test case ideally everything stays in-process, and there's no need for intermediate config files on disk.

And for the Foundry tests, you would keep the state lightweight and in-process anyway, for the usage in the test, rather than interacting with disk.

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,292 @@
...
+disputeGameFinalityDelaySeconds = 1
+```
+
+## Step 3: Deploy OP Chain Contracts
+
+We assume the latest Fault Proofs release `op-contracts/v1.5.0`.
+The exact set of sub-steps and inputs may change with future releases.
+
+The sub-steps are:
+
```
protolambda, line 202: The Safe contract thing is missing I believe, there is some extra step in the `Deploy` contract for a named Safe, which we may need to integrate here. (Although I don't believe it's currently going into the op-e2e build of L1 chain state)

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,292 @@
...
+[fault_proofs]
+# left out for brevity
+```
+
+The initial version of these input files will be minimal and only include the required inputs to deploy standard chains.
+Future versions will include more inputs to support more complex deployments, such as custom gas tokens or alt-DA.
+The deploy scripts and OP Stack Manager code will be written in such a way to default to standard chains and only require additional inputs for non-standard chains.
+
+# Risks & Uncertainties
```
protolambda, line 289: To add:
- It needs to be somewhat extensible, to handle extra features like a challenge contract for Plasma
- We need a written summary of what current usage looks like, and figure out how we transition users from that. E.g. the need of a tool that converts a legacy `DeployConfig` into these smaller modular configs.

tynes:
A thought here is to have the ability to generate a `rollup.json` from a view function on the OPSM, not literally returning the json file but something that returns the required values that end up in the json file, it could be an update to the `op-node` command for generating config. This would create a cohesive story for OPSM tx -> operating L2

tynes:
The usage of toml makes sense to easily operate with the superchain registry

protocol/op-stack-manager-arch.md
mds1: I thought about whether or not those two steps should be bundled. I went with separating them out because:
- We'll often need to deploy new implementation contracts (for upgrades) independently from new superchain contracts
- You could have different superchains that use the same implementation contracts, though I don't think that's likely

I've fixed the text to say three steps in fb6bbb6

protocol/op-stack-manager-arch.md
mds1: Got it. I'm assuming this is expanded on more in https://github.com/ethereum-optimism/design-docs/pull/61, so I'll read that next

protocol/op-stack-manager-arch.md
mds1: Perhaps the forge scripts always postprocess to output a file like
```toml
[addresses]
# all output addresses here

[inputs]
# all inputs are nested under here
```

Since all outputs from these steps are just addresses

protocol/op-stack-manager-arch.md
mds1: This is an interesting point. The Safe architecture is technically peripheral to the protocol itself, and is a step you'd use to configure the ProxyAdmin owner role holder a certain way. I think we can leave it out of scope here for now, what do you think @maurelian?

protocol/op-stack-manager-arch.md
mds1: These are great, added in 7a22832

mds1: These are great, added in 7a22832

protocol/op-stack-manager-arch.md
mds1, line 1: > A thought here is to have the ability to generate a `rollup.json` from a view function on the OPSM, not literally returning the json file but something that returns the required values that end up in the json file, it could be an update to the `op-node` command for generating config. This would create a cohesive story for OPSM tx -> operating L2

@tynes I like this idea, though I'm not sure we can return 100% of the required data, for example the genesis key

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,294 @@
...
+Adding this milestone will allow the interop team to build testing infrastructure that wraps OPSM, unifying our development, testnet, and production testing and deployment processes.
+
+# Problem Statement + Context
+
+*Some text copied from https://github.com/ethereum-optimism/design-docs/pull/52, which is related to this design doc.*
+
+The current L2 chain deployment approach originates from a time with Hardhat, single L1 target, and a single monolithic set of features.
+Since then the system has migrated to Foundry and extended for more features, but remains centered around a single monolithic deploy-config for all its features.
+
+The interop team needs a way to configure new multi-L2 deployments: The number of ways to compose L2s in tests grows past what a single legacy config template can support.
```
maurelian, line 28: I'm curious what some of these ways are. Are we talking about different chains interoperating with different features activated?

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,294 @@
...
+
+Outside of interop, deployment also seems increasingly complex and opaque, while it does not have to be, due to the same configuration and composability troubles.
```
maurelian, line 30: What is meant by opaque here?  We should clarify that, so we can define requirements for addressing this opacity.

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,294 @@
...
+
+- Duplication of effort: The interop team would need to develop and maintain their own deployment scripts, even though new ones will already be written for OPSM.
+- Divergence: With duplication, it's likely the interop team's deployment scripts diverge from the production deployment scripts, leading to inconsistencies and bugs.
+
+# Proposed Solution
+
+There are three aspects to deploying L1 contracts:
+
+1. **Deploy Superchain Contracts**. Superchain contracts are shared between many OP chains, so this occurs only occasionally in production.
+2. **Deploy Shared Implementation Contracts**. This occurs once per [contracts release](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/VERSIONING.md).
```
maurelian, line 48: > This occurs once per contracts release

My read of the doc thus far suggests that for 0.5 this only occurs once, ie. this doc is limited to deploying new systems for testing. Is that correct?

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,294 @@
...
+A sample TOML deploy config input for this step might look like:
+
+```toml
+[roles]
+proxyAdminOwner = "0x1234..."
+protocolVersionsOwner = "0x1234..."
+
+[superchain_config]
+guardian = "0x1234..."
+paused = false
```
maurelian, line 96: could even go so far as to exclude the `paused` value. It should always be false to start.

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,294 @@
...
+6. OptimismPortal
+7. SystemConfig
+8. L1CrossDomainMessenger
+9. L1ERC721Bridge
+10. L1StandardBridge
+11. OptimismMintableERC20Factory
+
+The inputs for each sub-step are:
+
+1. None
```
maurelian, line 147: Will `DGF.setImplementation()` need to be called? If so that will require a dispute game impl, which is not yet MCP ready AFAIR.

protocol/op-stack-manager-arch.md
protolambda: The main issue is migrating into interop. Where some chains have their own different copy of some L1 contracts like the superchain config / implementations deployment. We want to be able to test how we get from those situations to the new system. And new tests / deployments following the op-stack-manager approach of deployment as much as possible.

And the secondary issue is that currently the L1 dev genesis creation itself is coupled to a single L2 deployment, so there's no composition of L2s in any form.

protocol/op-stack-manager-arch.md
protolambda: The deploy config has a 102+ config variables, and requires two build systems (Go + forge), with multiple steps in-between, no clear definition of input configuration and output artifacts, or deploy functions you can use and cannot use. We can improve this with docs + better deployment tooling. I have a design doc open for it.

protocol/op-stack-manager-arch.md
maurelian: Agree, it should be safer and simpler to exclude the Safes that control the system. Any inputs to the system that originate from the Safes (either the ProxyAdmin owner or the Guardian) can be simulated by an EOA.

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,294 @@
...
+[fees]
+basefeeScalar = 1
+blobBaseFeeScalar = 1
+
+[fault_proofs]
+# left out for brevity
+```
+
+The initial version of these input files will be minimal and only include the required inputs to deploy standard chains.
+Future versions will include more inputs to support more complex deployments, such as custom gas tokens or alt-DA.
```
maurelian, line 286: Does 'future versions' imply 'out of scope for 0.5'?

protocol/op-stack-manager-arch.md
```diff
@@ -0,0 +1,294 @@
...
+unsafeBlockSigner = "0x123..."
+proposer = "0x123..."
+challenger = "0x123..."
+
+[fees]
+basefeeScalar = 1
+blobBaseFeeScalar = 1
+
+[fault_proofs]
+# left out for brevity
```
maurelian, line 282: Does the FP system need to run at all for the sake of interop testing, or can the FDG just be mocked up?

protocol/op-stack-manager-arch.md
mds1: Good idea, I'm supportive of that

protocol/op-stack-manager-arch.md
mds1: decision: Enable passing config in-process (e.g. `Config` struct) and via input files

protocol/op-stack-manager-arch.md
mds1: same as https://github.com/ethereum-optimism/design-docs/pull/60#discussion_r1710274580 but for outputs

protocol/op-stack-manager-arch.md
mds1: Can consider this as a way to shrink scope for M0.5, but will still need it for M1

mds1:
Summary of design review call is we have general agreement on the design proposed. The following points and comments will be incorporated into the document:
- All forge scripts should have 2 ways of interacting with them: One way that takes a config filepath (this is what chains will use), another that takes the config object directly with file IO (this is what go e2e will use). 
- Similarly, we should enable the same flexibility for outputs, both output file based and return data based
- When talking about how frequently a step occurs, clarify that in dev/testing the steps occur every time vs. in production where some steps are rare
- The `paused` variable must be part of the interface for dev/testing, but we should have it default to `false` so it can be hidden from users
- It's TBD whether M0.5 elegantly handles nonstandard features like CGT and Alt-DA
- Consider mocking the FP system as a way to reduce scope for M0.5. It still needs to be implemented in M1, but this may help ship something usable for interop sooner

mds1:
All feedback incorporated in 117da12, @tynes this is now ready for final review + merge

tynes (approved):

PR #61, opened 2024-08-06
protocol: forge scripts deploy in Go

protolambda:
**Description**

This introduces the idea of running the Forge scripts, primarily the chain deployment / genesis setup scripts, in Go.
This improves the integration with Go tooling and tests, to unlock the customization we need for things like Interop, and to generally improve test coverage of our deployments.

**Metadata**

Fix https://github.com/ethereum-optimism/optimism/issues/11360

### DIFF
--- a/protocol/forge-scripts-deploy-in-go.md
+++ b/protocol/forge-scripts-deploy-in-go.md
@@ -1,0 +1,222 @@
+# Forge Scripts Deploy in Go
+
+# Purpose
+
+The purpose of this doc is to improve the integration between the `forge script` based chain deployment/genesis flow
+and the Go tooling and testing.
+
+# Summary
+
+By running the chain-deployment/genesis solidity scripts in an instrumented Go EVM we can reduce the roundtrip time,
+and greatly improve integration with inputs (configs, dependencies) and outputs (state, artifacts).
+
+This unlocks runtime-customization of deployments in Go tests,
+to increase coverage and create new multi-L2 chain deployments for Interop tests.
+
+# Problem Statement + Context
+
+From [Design doc 52](https://github.com/ethereum-optimism/design-docs/pull/52):
+
+> The current L2 chain deployment approach originates from a time with Hardhat,
+> single L1 target, and a single monolithic set of features.
+> 
+> Since then the system has migrated to Foundry and extended for more features,
+> but remains centered around a single monolithic deploy-config for all its features.
+> 
+> **With Interop we need to configure a new multi-L2 deployment**:
+> the number of ways to compose L2s in tests grows past what a single legacy config template can support.
+> 
+> Outside of interop, deployment also seems increasingly complex and opaque, while it does not have to be,
+> due to the same configuration and composability troubles.
+
+See the design-doc for further in-depth context on the problem.
+
+The integration between (1) Go testing/tooling and (2) Forge deployment/genesis needs to improve
+to reduce the complexity of deploying and testing.
+
+Specifically, Go tools need reliable and organized inputs for deployment/genesis,
+and interface with the forge scripts artifacts or scripts themselves
+in a way that is configurable but not error-prone.
+
+# Proposed Solution
+
+In [deployment-chains design doc 52](https://github.com/ethereum-optimism/design-docs/pull/52) and
+[OPStackManager design doc 60](https://github.com/ethereum-optimism/design-docs/pull/60) the importance
+of Modular configs and incremental deploy steps is outlined.
+
+To utilize those modular configs and deploy steps, we need a way to either 
+run and cache the individual Forge script functions (as outlined in 52),
+or integrate it closer such that no caching complexity is required.
+
+This document proposes the latter: run the Forge scripts within Go,
+such that we do not need multiple steps of Forge script sub-process to build a state for testing / genesis tooling.
+
+## Running Forge scripts in Go
+
+Forge scripts are really just solidity smart-contracts, with the addition of Forge cheat-codes.
+Running a contract in Go is relatively trivial: e.g. Cannon tests run the Cannon contracts in an instrumented Go EVM.
+
+To support the Forge cheatcodes, we need to emulate the cheatcode behavior, as the Geth EVM does not natively have support for it.
+Cheatcodes are essentially responses to `STATICCALL`s to a special system address, acting like a precompile,
+with functions that interact with the EVM environment.
+
+We do not have to support all Forge cheatcodes; just the subset used by the OP-Stack scripts would be sufficient.
+
+The most important cheat-codes are:
+- `vm.chainId`: change chain ID
+- `vm.load`, `vm.store`: storage getter/setter
+- `vm.etch`: write contract code
+- `vm.deal`: set balance
+- `vm.prank`: change sender
+- `vm.getNonce`/`vm.setNonce`: account nonce getter/setter
+- `vm.broadcast`,`vm.startBroadcast`, `vm.stopBroadcast`: capture calls as transaction candidates. (Can be no-op, if we do not want to go through Go for production deployments)
+- `vm.getCode`/`vm.getDeployedCode`: artifact inspection
+- `vm.env{Bool/Uint/Int/etc.}`: env var getters
+- `vm.keyExists{...}/parse{...}/writeJson/etc.`: encoding utils
+- `vm.addr`: priv key to addr
+- `vm.label`: name an address
+- `vm.dumpState`: export EVM state
+- `vm.loadAllocs`: import EVM state
+
+Note that with many of these, Go instrumentation can really improve integration:
+- `dumpState`/`loadAllocs`: no need to encode/decode state or read/write to disk -> faster Go tests
+- `getCode`/`getDeployedCode`: attach directly to artifacts, and *track which artifacts were used for a deployment*
+- `broadcast`: we can extend the deploy-tool functionality with transactions. Out of scope for now, but can be very useful.
+
+## Forge-artifacts as Go FS
+
+A Go FS is a simple filesystem abstraction: it provides read-only access to some source of files.
+A local directly can be wrapped into such FS, but also tarballs, or even data embedded in the Go binary, can be represented as Go FS.
+By using this FS abstraction, we can make the access to artifacts very simple, and "mount" the relevant FS into it.
+
+For Go tests, this would be the local FS.
+
+For Go genesis tooling, this might be a bundled FS, or one from a versioned release tarball.
+
+Using an FS helps simplify the way we interact with forge-artifacts.
+
+## Semver the scripts
+
+We have an existing `Semver` pattern for production contracts, but don't apply the same to deploy scripts, yet.
+If we introduce this, then the version of the scripts (part of the artifacts) can be inspected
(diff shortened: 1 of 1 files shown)


protocol/forge-scripts-deploy-in-go.md
```diff
@@ -0,0 +1,222 @@
...
+was completed successfully in the form of an interop side-project to remove tech-debt. This project scope looks quite similar.
+
+This functionality does block Interop op-e2e testing: without it,
+we are not able to customize deployments sufficiently and cleanly (avoiding many more `allocs` special cases),
+to get multi-L2 deployments into the op-e2e.
+
+## Devrel feedback
+
+Historically devrel has not been included sufficiently in the deployment-flow design.
+Known pain-points like inconsistency between the Go and forge genesis generation,
```
sbvegan, line 221: Good call out here, for example, a recent partner was having trouble using the solidity genesis allocs creation on the `op-contracts/v1.3.0` release so they opted for something ~200 commits into the future that played nicely with the tooling instead of using the legacy genesis creation method.

With that being called out, if this lands after a L1 contract release, it would be nice to have this type of functionality cut in a different minor release so its accessible for new chain deployments between say `op-contracts/v1.6.0` and `op-contracts/v1.7.0`.

protocol/forge-scripts-deploy-in-go.md
```diff
@@ -0,0 +1,222 @@
...
+unclear allocs, and monolithic deploy-config are being addressed, but more feedback may still improve the design.
```
sbvegan, line 222: The modular deploy configuration will be nice improvement, especially for our beta features (CGT and Alt-DA Mode). Similar to my comment above, it would be nice to have this worked into the CGT contract releases.

sbvegan (commented):
I'm very excited for any contract deployment quality of life improvements. Beyond my initial comments, I'm going to work on collecting additional customer contract deployment pain in this [notion document](https://www.notion.so/oplabs/Smart-Contract-Deployment-PAIN-29d8aa270d45413ea7ac879f1d2cb077). I'll reach out to developer support who may have additional insight.

protocol/forge-scripts-deploy-in-go.md
```diff
@@ -0,0 +1,222 @@
...
+Forge scripts are really just solidity smart-contracts, with the addition of Forge cheat-codes.
+Running a contract in Go is relatively trivial: e.g. Cannon tests run the Cannon contracts in an instrumented Go EVM.
+
+To support the Forge cheatcodes, we need to emulate the cheatcode behavior, as the Geth EVM does not natively have support for it.
+Cheatcodes are essentially responses to `STATICCALL`s to a special system address, acting like a precompile,
+with functions that interact with the EVM environment.
+
+We do not have to support all Forge cheatcodes; just the subset used by the OP-Stack scripts would be sufficient.
+
+The most important cheat-codes are:
```
tynes, line 65: Isn't the bottleneck the compilation time? We would still need to compile the sourcecode to bytecode. Basically you are saying that given the bytecode, we would be able to initialize the testing setup with various branching logic that occurs in the EVM based on different config

protocol/forge-scripts-deploy-in-go.md
```diff
@@ -0,0 +1,222 @@
...
+- `vm.chainId`: change chain ID
+- `vm.load`, `vm.store`: storage getter/setter
+- `vm.etch`: write contract code
+- `vm.deal`: set balance
+- `vm.prank`: change sender
+- `vm.getNonce`/`vm.setNonce`: account nonce getter/setter
+- `vm.broadcast`,`vm.startBroadcast`, `vm.stopBroadcast`: capture calls as transaction candidates. (Can be no-op, if we do not want to go through Go for production deployments)
```
tynes, line 72: I do not think its a good idea to go through go for production deployments. Adding a layer introduces a new risk for bugs and also makes it more difficult to onboard into the codebase

protocol/forge-scripts-deploy-in-go.md
```diff
@@ -0,0 +1,222 @@
...
+of Modular configs and incremental deploy steps is outlined.
+
+To utilize those modular configs and deploy steps, we need a way to either 
+run and cache the individual Forge script functions (as outlined in 52),
+or integrate it closer such that no caching complexity is required.
+
+This document proposes the latter: run the Forge scripts within Go,
+such that we do not need multiple steps of Forge script sub-process to build a state for testing / genesis tooling.
+
+## Running Forge scripts in Go
```
tynes, line 54: I have slight concern about possible bugs in the foundry Go precompile implementation resulting in e2e tests not running against "the same" contracts. I suppose the precompile isn't super complex, but adding layers of abstraction does make me feel worried

protocol/forge-scripts-deploy-in-go.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Forge-artifacts as Go FS
+
+A Go FS is a simple filesystem abstraction: it provides read-only access to some source of files.
+A local directly can be wrapped into such FS, but also tarballs, or even data embedded in the Go binary, can be represented as Go FS.
+By using this FS abstraction, we can make the access to artifacts very simple, and "mount" the relevant FS into it.
+
+For Go tests, this would be the local FS.
+
+For Go genesis tooling, this might be a bundled FS, or one from a versioned release tarball.
```
tynes, line 94: Are you proposing that this is used outside of testing, ie for L2 genesis generation?

tynes: I see in the follow up section this is mentioned

tynes:
I think that this could be a good approach although I would need more convincing to think its a good idea to use this for anything other than testing. I would much rather contribute features directly to foundry that help make our deployments better rather than us building tooling that subsumes foundry

protocol/forge-scripts-deploy-in-go.md
mds1, line 1: > A thought here is to have the ability to generate a `rollup.json` from a view function on the OPSM, not literally returning the json file but something that returns the required values that end up in the json file, it could be an update to the `op-node` command for generating config. This would create a cohesive story for OPSM tx -> operating L2

@tynes I like this idea, though I'm not sure we can return 100% of the required data, for example the genesis key

protocol/forge-scripts-deploy-in-go.md
mds1: Agreed. There are a lot of edge cases around deployments (e.g. resuming failed deployments, contract verification), so I do not think we should implement them in go. (The right way to make these a no-op is redirect them call `prank`, since a broadcast specifies the deployer's address)

protocol/forge-scripts-deploy-in-go.md
```diff
@@ -0,0 +1,222 @@
...
+## Running Forge scripts in Go
+
+Forge scripts are really just solidity smart-contracts, with the addition of Forge cheat-codes.
+Running a contract in Go is relatively trivial: e.g. Cannon tests run the Cannon contracts in an instrumented Go EVM.
+
+To support the Forge cheatcodes, we need to emulate the cheatcode behavior, as the Geth EVM does not natively have support for it.
+Cheatcodes are essentially responses to `STATICCALL`s to a special system address, acting like a precompile,
+with functions that interact with the EVM environment.
+
+We do not have to support all Forge cheatcodes; just the subset used by the OP-Stack scripts would be sufficient.
```
mds1, line 63: Aside from implementation risk, another risk is the subset used by OP Stack scripts will continue to grow, and I worry contract devs might want to use a new cheat, but then PRs get blocked by needing go support for that cheat.

Ultimately, is the resulting state from a script what the go tests need? If so, I think it will be a lot simpler to call the forge script from go, with a way conditionally call `vm.dumpState` at the end. Go can then read in the resulting file that was written and apply that state. This is effectively what [kontrol](https://docs.runtimeverification.com/kontrol) does to initialize proofs so it doesn't need to execute the EVM/cheats

If we don't want to use the filesystem as an interface, forge has a `--json` arg to output JSON to stdout. We can probably get a `forge script --json --dumpstate` feature added so that it automatically dumps state to the JSON as part of stdout

protocol/forge-scripts-deploy-in-go.md
```diff
@@ -0,0 +1,222 @@
...
+
+Note that with many of these, Go instrumentation can really improve integration:
+- `dumpState`/`loadAllocs`: no need to encode/decode state or read/write to disk -> faster Go tests
+- `getCode`/`getDeployedCode`: attach directly to artifacts, and *track which artifacts were used for a deployment*
+- `broadcast`: we can extend the deploy-tool functionality with transactions. Out of scope for now, but can be very useful.
+
+## Forge-artifacts as Go FS
+
+A Go FS is a simple filesystem abstraction: it provides read-only access to some source of files.
+A local directly can be wrapped into such FS, but also tarballs, or even data embedded in the Go binary, can be represented as Go FS.
```
mds1, line 89:
```suggestion
A local directory can be wrapped into such FS, but also tarballs, or even data embedded in the Go binary, can be represented as Go FS.
```

protocol/forge-scripts-deploy-in-go.md
```diff
@@ -0,0 +1,222 @@
...
+By using this FS abstraction, we can make the access to artifacts very simple, and "mount" the relevant FS into it.
+
+For Go tests, this would be the local FS.
+
+For Go genesis tooling, this might be a bundled FS, or one from a versioned release tarball.
+
+Using an FS helps simplify the way we interact with forge-artifacts.
+
+## Semver the scripts
```
mds1, line 98: Great idea

protolambda:
Summary of design-review call:

The primary concern is if whether or not we can use `forge script` as-is to serve and op-e2e.

Arguments for:
- Ideal to upstream changes to forge, if we can identify something that unlocks integration with Go
- Can be implemented (with caching to mitigate performance problems) as in https://github.com/ethereum-optimism/design-docs/pull/52
- Forge cheatcodes all work out of the box

Arguments against:
- Need for well-defined inputs/outputs
  - The more we side-channel data through files, the more the Go integration has to prepare and manage edge-cases of (caches, missing data, IO errors, etc.).
  - Difficult to bundle Go L2 genesis generation and Forge L2 genesis generation, do to multi-step process and intermediate files. Hurts devrel experience.
- Performance concerns
  - Intermediate disk access
  - Encoding the configs in JSON (or toml), split between 32 byte leafs, reconstructing them multiple times, JSON parsing again, etc.
  - Multiple state dumps / loads per test setup.
  - FFI calls to forge processes (and forge versioning etc. affecting tests)
- No cache-invalidation complexity

And L1 deployments can be unified by making a `broadcast` a no-op in the Go context, so we can run the same deploy functions, but generate state in-place.

There is some hesitation, but we can try the Go forge-script approach, and fall back to the other approach if the Go approach does not work.

protocol/forge-scripts-deploy-in-go.md
protolambda: Compilation is a separate step, runs once, and artifacts can then be reused by the Go execution for multiple different setups. The main bottleneck is runtime configuration and overhead/complexity of the intermediate steps, which is reduced by running the scripts in Go.

tynes (approved):

PR #64, opened 2024-08-12
Add cannon upgrade design doc

ajsutton:
**Description**

Draft of a doc to determine a roll out plan for cannon MT and cannon 64-bit.


protocol/cannon-upgrades.md
```diff
@@ -0,0 +1,134 @@
...
+### Specify Cannon Version in Prestate
+
+Add a flag to the cannon prestate that indicates which version of the VM is required. No new game type is used
+and `op-challenger` will simply provide the appropriate prestate to cannon which then operates in the appropriate mode.
+This makes the upgrade to cannon MT or 64bit work just like any other prestate upgrade.
+
+As the game type isn't changed, users would not need to reprove their withdrawals.
+
+This is really only feasible if we maintain support for legacy cannon in the monorepo until after the upgrade (and any
+existing games are completed). The legacy code could likely shortly after chains have upgraded to the new version. At
```
clabby, line 92: How so? We should be able to do this without keeping support in `develop`, if we cut the final tag of cannon

clabby (commented):
Thank you for putting this together! I agree on most points here - combining the upgrade, the want to not impact user withdrawals by treating this as a regular prestate upgrade.

I am a bit concerned that the want for backwards compatibility is too complex, and I think we can get around this in a much less painful and more straightforward way.

If we cut the final tag of cannon after granite releases, we can be done with that version of cannon barring major security vulnerabilities that need to be back ported on top of the tag. It should be trivial to build off of that branch and keep the old version of the VM in a deprecated state until we ultimately kill it.

For the upgrade, I like the idea of treating the combined 64bit MT cannon release as a regular prestate upgrade, but why do we have to change the snapshot format etc.? It may be much easier to just add an ugly if statement to the challenger for a few days during the roll over period, where certain prestate hashes must use old cannon, and anything else must use new cannon, for the `CANNON` game type. That way, there is only a single place where we need to worry about this intermediate period, and no debt that we add in the process which we may not need in the future.

My vote here would be to:

1. Cut a final tag of `cannon32` after the release of the Granite hardfork.
2. Merge the MIPS64 PRs, breaking backwards compatibility entirely in develop.
3. The final `cannon32` tag will serve as an ossified version of the old VM, where we can return to in order to backport fixes and/or build the old binary, during the time it is still being used in production.
4. Ongoing work will continue on the 64-bit FPVM, eventually taking the place of the 32-bit FPVM in production.
5. When we are ready for the MT cannon + 64 bit upgrade, we:
    5.i) Inform the challenger which cannon prestates belong to the old VM. During the roll over period, it will consult this switch statement to check which version it should use.
    5.ii) After the rollover period has completed, and all chains are upgraded, remove this switch from the challenger.

protocol/cannon-upgrades.md
ajsutton: You can't have cannon automatically swap which VM it uses based on the prestate if cannon doesn't have support for the legacy version.  We could in theory do that swap in op-challenger but then it needs even more detailed understanding of cannon's snapshot format and we're trying to go in the other direction and keep the snapshot format isolated in cannon.

protocol/cannon-upgrades.md
```diff
@@ -0,0 +1,134 @@
...
+
+The extra development time required to preserve backwards compatibility is likely to be offset by not being blocked on
+updating OptimismPortal to allow changing game types without invalidating withdrawals. We've also seen multiple times
+through fault proofs that developing in a backwards compatible way pays big dividends for testing and flexibility in
+deploying. With a major project underway to upgrade chains to fault proofs we want to make it as simple as possible to
+run `op-challenger` correctly and keep its deployment simple.
+
+# Risks & Uncertainties
+
+## Cost of Backwards Compatibility
```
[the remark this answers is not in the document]
ajsutton: I'm quite ok with trivial duplication - creating a tag is effectively the same duplication as we're still maintaining that fork, it just doesn't benefit from all our usual tooling and needs special cases for releases.

The big thing that worries me about just breaking backwards compatibility is that it means that our `develop` branch isn't really releasable - you have to do some extra stuff when deploying `op-challenger` to ensure it can play extra games which every one who runs a challenger has to then do. And we're very regularly frustrated by the fact that develop isn't always releasable because that complicates how you do a release. It affects more than just security releases of the legacy cannon, we wouldn't be testing how `op-challenger` and `op-program` work with the legacy cannon since it would exist in the monorepo anymore so we then can't release those which blocks any hard forks, or we have to build some testing framework for backwards compatibility.

Inphi:
Good doc. I'll add that we don't have to be fancy with this. 
Create a snapshot of cannon-32 today and keep it in a separate directory and Go module, `cannon32`, stored in the monorepo.
Then continue multithreaded and 64-bit development in `cannon`. We retain the benefits of running both VMs in our tests. And If need be, introduce a shim program that uses either VM.
Fixes to current cannon (`cannon32`) can be done upstream without having to backport anything.

Inphi:
And I'm onboard with implementing 64-bit and MT Cannon as a single feature.

ajsutton:
I spiked out supporting both versions in a single cannon version and it works out pretty well: https://github.com/ethereum-optimism/optimism/pull/11467 though the diff looks pretty intimidating.  We're essentially duplicating the `mipsevm` package to have `mipsevm32` and `mipsevm64` - the diff would be a lot smaller if we left the legacy version as just `mipsevm`.  I also removed the multithreaded impl from the 32 bit version and removed the single threaded version from the 64bit VM.

It's a little complicated to start from everything upgraded to 64bit and try and undo the single threaded upgrade - we could do it in smaller steps if we started from the current `develop`, introduce the abstractions and pull out code that can be shared, then split mipsevm into mipsevm32 and mipsevm64 and then apply the 64bit upgrade to mipsevm64. Will be much more reviewable that way.

ajsutton:
Going to close this - cannon now detects multithreaded or single threaded based on the prestate and we think rolling out 64bit with backwards compatibility will be doable.  I'm not sure this design doc actually reflects all the discussion that wound up happening elsewhere (GitHub is not a great venue for async discussion really) so won't merge it. The plan only affects the proofs team anyway so doesn't need wider buy-in.

PR #66, opened 2024-08-27
design-docs: add fee ratio

PinelliaC:
**Description**
For custom gas token chains, I added a design doc file discussing the possibility of considering the exchange rate between ether and the native token when calculating L2 and L1 fees.


tynes:
Left feedback here: https://github.com/ethereum-optimism/specs/discussions/140#discussioncomment-10491878

PR #70, opened 2024-08-29
protocol: block safety index

protolambda:
**Description**

Proposal to enshrine the idea of a "block safety index", such that other (existing and new) features can be built against this in a unified way.

In particular, as we enter Holocene steady-batch-derivation work, and Interop devnet 2, I believe there is a need for a shared block-safety-index feature, to reduce overall protocol complexity by unifying solutions to the same problem.

### DIFF
--- a/protocol/block-safety-index.md
+++ b/protocol/block-safety-index.md
@@ -1,0 +1,224 @@
+# Block safety index
+
+# Summary
+
+Unify the mapping between L2 blocks and the L1 blocks they were derived from,
+to address cross-protocol safety-validation challenges.
+
+This mapping conceptually exists in different forms in the OP-Stack,
+but should be described as a feature of the rollup-node, for other components to be built against.
+
+This does not directly affect the op-program (or Kona), but does affect how we source the data to enter a dispute game.
+
+# Problem
+
+The following things are all *the same*:
+- The op-node warmup UX problem
+- The span-batch recovery problem
+- The fault-proof pre-state problem
+- The interop reorg problem
+- The L2 finalization problem
+- The op-proposer consistency problem
+- The safety indexing problem
+
+At the core, all these issues stem from a single problem:
+determining which L2 block corresponds to which L1 block.
+
+Yet we currently have *different solutions* for each case.
+
+The reason why this is hard: the L2 chain, by design, does not pre-commit to how/where data gets confirmed on L1.
+This reduces reorg risk and DA cost by deferring optionality to the batcher, but introduces "offchain" state.
+
+Can we fix this, and simplify the OP-Stack?
+
+## Sub-problems
+
+### The op-node warmup UX problem
+
+The majority of op-node warmup is derivation-pipeline reset time.
+It traverses to an old L2 block guaranteed to still be canonical,
+then traverses to an old L1 block that is guaranteed to produce
+the inputs needed to continue derivation of next L2 block inputs.
+
+This traversal can be lengthy:
+- The L2 chain is traversed back, because we cannot with certainty say whether it was derived from a recent L1 block.
+  Often this means it "bottoms out" on the finalized L2 block, at least ~12 minutes old.
+- The L1 chain is traversed back a full channel-timeout; 300 blocks, i.e. an hour of L1 data!
+  (reduced in Granite to 50 blocks = 10 minutes).
+
+This delay slows down the process of making nodes fully operational after a restart,
+particularly with regard to derivation.
+
+But it can be so much faster, if we introduce state:
+- L2 traversal can go back to the last L2 block derived from a still-canonical L1 block.
+  Seconds, not minutes, of traversed time-span.
+- L1 traversal can go quickly find the relevant L1 data with a much shorter search space,
+  if we had stricter batch-ordering guarantees (being introduced with Holocene!) and the derived-from mapping.
+    - The last derived-from L1 block is an upper bound.
+    - The previous-to-last derived-from L1 block is a lower bound.
+
+### The span-batch recovery problem
+
+Currently, span-batch recovery (pre-Holocene) relies on 'pending-safe',
+where a span-batch may generate pending blocks that can be reorged out
+if later content of the span-batch is found invalid.
+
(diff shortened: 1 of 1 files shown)


protocol/block-safety-index.md
```diff
@@ -0,0 +1,224 @@
...
+    - The last derived-from L1 block is an upper bound.
+    - The previous-to-last derived-from L1 block is a lower bound.
+
+### The span-batch recovery problem
+
+Currently, span-batch recovery (pre-Holocene) relies on 'pending-safe',
+where a span-batch may generate pending blocks that can be reorged out
+if later content of the span-batch is found invalid.
+
+This is changing with Holocene (steady batch derivation, aka strict ordering): we do not want the complexity of having to revert data that was tentatively accepted.
```
sebastianst, line 66: So do you agree that the span batch recovery problem gets easier with Partial Span Batch Validity? If we only forward-invalidate, but not backward-invalidate in a span batch, the start of the span batch is less important as it won't be needed to reorg out on a next invalid batch.

protocol/block-safety-index.md
```diff
@@ -0,0 +1,224 @@
...
+For interop fault-proving, the L2 assumptions should not be larger than 1 block,
+since we want to consolidate cross-L2 after confirming each L2 optimistically.
+
+If we find a pre-existing L2 chain (simply on node restart, or on reorg),
+we currently cannot be sure it was derived from a certain L1; we have to recreate the derivation state to verify.
+And if the tip was part of a span-batch, we need to find the start of said span-batch.
+So while we can do away with the multi-block pending-safe reorg, we still have to "find" the start of a span-batch.
+If we had an index of L2-block to L1-derived-from, with span-batch bounds info,
+then finding that start would be much faster.
```
sebastianst, line 75: Don't we need to recreate the derivation state on a reorg or restart anyways? What's the difference to, say, a channel full of singular batches, where the reorg or node restart may have happened in the middle of deriving singular batches from the channel, so the L2 tip is in the middle of a channel? What I mean is that we always have to recreate the derivation state in a way that we get clarity on which L1 block a certain L2 block is derived from, and this will get easier with the set of changes we're introducing with Holocene.

This is not to say that the block safety index is still very useful.

protocol/block-safety-index.md
```diff
@@ -0,0 +1,224 @@
...
+
+The "safedb" in op-node can be replaced with a DB like the above, shared code with the op-supervisor.
+Once the above is done, we can remove the finality mechanism: the op-supervisor signals are used for finalization.
+
+The supervisor is Go, and modular enough that we can embed the single-DB solution into the op-node,
+to serve those less-standard chains with an interop-set of 1.
+This prevents these chains from entering a divergent hardfork path:
+very important to keep the code-base unified and tech-debt low.
+
+The main remaining question is how we bootstrap the data:
```
sebastianst, line 170: For how long do we actually need to maintain this db? Is there an expiry duration that satisfies all above cases, e.g. the FP challenge window? For most mentioned cases I think a soft rollout could work good enough.

protocol/block-safety-index.md
```diff
@@ -0,0 +1,224 @@
...
+- Chain-sync is on-chain information only, since it needs to verify retrieved data against authenticated commitments.
+  We cannot start from snap-sync if we still have to backfill the history data.
+- Nodes don't currently have this state (if they have not been running the safe-db),
+  we need to roll it out gradually.
+
+This feature could either be a "soft-rollout" type of thing,
+or a Holocene feature (if it turns out to be tightly integrated into the steady batch dervation work).
```
sebastianst, line 177: I don't think that we need to make it a Holocene feature. Holocene still works, like the pre-Holocene protocol, with a sync start protocol (that will actually get easier). But it's a helpful feature the moment it's there.

protocol/block-safety-index.md
```diff
@@ -0,0 +1,224 @@
...
+but gives an idea of the resource usage.
+
+In terms of compute it's very minimal to maintain:
+the op-node verifies L1 anyway, and writing the incremental updates is not expensive.
+
+In terms of disk-space it is a little more significant:
+at 32 bytes per L1 and L2 block hash, some L1 metadata, some L2 metadata, and both local-safe cross-safe,
+each entry may about 200 bytes. It does not compress well, as it largely contains blockhash data.
+
+Storing a week of this data, at 2 second blocks, would thus be `7 * 24 * 60 * 60 / 2 * 200 = 60,480,000`, or about 60 MB.
```
sebastianst, line 194: For what use case would we need more than ~a week of safe db history?

sebastianst (commented):
Great proposal!

One aspect I don't understand is the advantage of tracking additional info like span batch bounds, on top of just the L2/L1 derivation mapping.

Thankfully with strict batch ordering we won't have multiple buffered channels any more, but before Holocene, would we also want to track additional derivation pipeline state, like what L1s caused buffered frames/channels in the channel bank?

axelKingsley:
This is not really a blocker for the design itself, but worth noting a chore we should do: in various places in our docs, we talk about the statelessness of our components. That statelessness is much weaker already with things like the SafeDB, but this would definitely push us far enough that we'd want to amend documentation.

protocol/block-safety-index.md
```diff
@@ -0,0 +1,224 @@
...
+Storing a year of this data would be around ~3 GB.
```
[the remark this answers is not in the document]
protolambda: Explorers / indexers might want extended data, to show past L1<>L2 relations. But yes, 1 week should otherwise be enough. Perhaps we can add an archive-flag, where it writes the data that gets pruned to a separate DB, for archival purposes.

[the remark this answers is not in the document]
ajsutton: With clock extension games can be in progress for longer than a week (16 days I think is worst case but may have that wrong) and you typically want to monitor them for some period after they resolve so you have a chance to detect if they resolved incorrectly.  Currently dispute-mon and challenger have a default window of 28 days that they monitor. So I'd suggest we want to keep data for at least that long by default.

protolambda:
This is stale, a version of this called `fromda` package, to store L1 <> L2 increments in a stupid-simple append-only way (single type of entry), was implemented in the op-supervisor.

PR #72, opened 2024-09-04
Add Holocene design doc

sebastianst:
**Description**

Holocene design doc, to align on open questions.

**Additional context**

After getting alignment, the specs can be completed. They are currently in draft at https://github.com/ethereum-optimism/specs/pull/357

This has been reviewed in a public design review session on 2024-09-11, with a [public recording](https://discord.com/channels/1244729134312198194/1283534070252568636/1283692806094983199).

### DIFF
--- a/protocol/strict-derivation.md
+++ b/protocol/strict-derivation.md
@@ -47,25 +47,6 @@ and for simplicity it was decided to use the atomic validity model. However, abo
 focusses on implications of the atomic validity model for EL-sync. We're now concerned with behavior
 around CL client restarts and reorgs.
 
-### Mid-way restart
-
-One concern that was raised with partial span batch validity is how to deal with mid-way restarts,
-i.e. when the CL client restarts, loosing its internal derivation pipeline state while the safe head
-of the EL client is in the middle of a span batch. In particular, the correct behavior needs to
-cover the case where a L1 reorg happened in the meantime that reorged out the L1 block that
-contained the span batch.
-
-I fail to see currently how partial span batch validity makes this case worse or different. With
-atomic span batch validity, there's the pending-safe state of derived blocks that might be reorged
-out later by a backwards-invalidating singular batch derived from a span batch. So if we only
-forwards-invalidate, this becomes easier.
-
-And in any case, a restart or reorg needs to reconstruct the derivation pipeline state from the new
-(sync-start) L1 block, so I don't understand how reconstructing span batch derivation state is
-different from reconstructing the channel bank and batch queue in general. If a restart or reorg
-happens in the middle of deriving singular batches from a channel, we're in a similar situation that
-is solved by going back far enough for the sync-start.
-
 ## Proposed Design
 
 To resolve open questions regarding the implementation of Partial Span Batch Validity, I propose to
@@ -88,6 +69,25 @@ dropped, as it doesn't contain any new batches (this would also happen if applyi
 to each derived singular batch individually).
 
 Those checks are done once per span batch before the batch queue derives any singular batches.
+
+### Mid-way restart
+
+One concern that was raised with partial span batch validity is how to deal with mid-way restarts,
+i.e. when the CL client restarts, losing its internal derivation pipeline state while the safe head
+of the EL client is in the middle of a span batch. In particular, the correct behavior needs to
+cover the case where a L1 reorg happened in the meantime that reorged out the L1 block that
+contained the span batch.
+
+I fail to see currently how partial span batch validity makes this case worse or different. With
+atomic span batch validity, there's the pending-safe state of derived blocks that might be reorged
+out later by a backwards-invalidating singular batch derived from a span batch. So if we only
+forwards-invalidate, this becomes easier.
+
+And in any case, a restart or reorg needs to reconstruct the derivation pipeline state from the new
+(sync-start) L1 block, so I don't understand how reconstructing span batch derivation state is
+different from reconstructing the channel bank and batch queue in general. If a restart or reorg
+happens in the middle of deriving singular batches from a channel, we're in a similar situation that
+is solved by going back far enough for the sync-start.
 
 # Strict Batch Ordering
 
@@ -129,8 +129,12 @@ The batch queue also becomes simpler as batches have to arrive in order.
 - How to treat _older_ batches is easy: they are dropped.
 - Similarly to out of order frames, how to treat future batches (gaps) has two options:
   - Option 1: immediately fill the gap with deposit-only blocks, and then process the batch.
+  It is noteworthy that the future batch can then only be valid if it built on top of a gap of empty batches.
   - Option 2: drop future batches, and wait for up to a full sequencing window for the next
   batch to arrive.
+  - Note that the pre-Holocene behavior is to cache _future_ batches for later checking when the gap
+  has been closed. It is one goal of Holocene to remove as much buffering as possible, so this
+  behavior is not considered a valid option.
 
 ### Proposed solution
 
@@ -152,8 +156,10 @@ batches, as long as the sequencer is creating unsafe blocks that follow the same
 auto-derivation would for gaps. However, this would need to be investigated more deeply.
 
 Note that the new strict ordering rules of the batch queue will always lead to an empty batch queue
-when the origin of the derivation pipeline progresses to the next L1 block (what about
-`undecided` though?).
+when the origin of the derivation pipeline progresses to the next L1 block.
+In order to guarantee this invariant fully, we might need to add an extra rule to the derivation
+pipeline to only progress to the next L1 block for derivation if all `undecided` batches have been
+decided upon by retrieving the previously temporarily missing L1 and L2 blocks.
 
 ## Batcher hardening
 
@@ -190,7 +196,7 @@ Strict Batch Ordering leads to a simpler sync start algorithm. It may be suffici
 the L1 chain until finding a batcher transaction that starts a new channel, i.e., containing a first
 frame. The strict ordering rules, as detailed above, would in such a case drop any pre-existing
 staging channel, and the batch queue should always be empty right after the pipeline origin
-progresses (open question: is this true, in view of lingering `undecided` batches?).
+progresses.
 
 # Steady Batch Derivation
 
@@ -202,8 +208,23 @@ the derivation pipeline are treated.
   [`INVALID`](https://github.com/ethereum/go-ethereum/blob/7fd7c1f7dd9ba8d90399df2f080e4101ae37a255/beacon/engine/errors.go#L63)
   they are replaced by deposit-only/empty payload attributes and reinserted into the EL.
 
-At this point, there are no open questions about the design of this protocol change. I assume that
-the actual implementation will raise questions on certain details.
+## L1 origin selection
+
+There's an open design space about how to select the L1 origin when deriving one or more
+deposit-only blocks.
+
+- A simple default rule could be to generate blocks in a way to maintain a steady L1/L2 block ratio,
+e.g. bumping the L1 origin selection every 6 blocks in the case of mainnet.
+- Edge case: we were already very near the sequencer drift limit, and need to select a new origin
+faster.
+- Edge case: L1 missed a slot, and the L1 origin cannot advance as expected.
+- So maybe a better rule is to first eagerly advance the L1 origin as quickly as possible, and only
+if a newer L1 origin isn't available, keep it. This solves for missed slots and will implicitly and
+automatically maintain a good L1/L2 block ratio. We then just need a clear definition of "L1 origin
+is/isn't available".
+  - To mimic sequencer behavior, and avoid being hit by shallow L1 reorgs, we could add an
+  in-protocol L1 validation depth. So eagerly advancing the L1 origin while maintaining a timestamp
+  distance of this validation depth times the L1 bock time.
 
 # Activation rules & actions
 
@@ -228,14 +249,18 @@ The principle of eager derivation of later stages should guarantee at this point
 been applied directly on top of the current safe chain have already been exhausted from the batch
 queue. So what is left should be `undecided` and `future` batches. (Note: I'm not sure on this one.)
 I see two options on how to handle a non-empty batch queue at this point:
-- Option 1: Drop future batches, continue resolving undecided batches, if any are left, and apply
+- Option 1: When the L1 origin reaches the Holocene activation block, discard all batches in the
+batch queue. Could also be applied to the Channel Bank to give us a nice clean starting point.
+  - The batcher would have to be aware of that rule and consider any blocks it submitted in channels
+  that didn't close prior to the Holocene activation block as needing to be resubmitted.
+- Option 2: Drop future batches, continue resolving undecided batches, if any are left, and apply
 new Holocene rules.
-- Option 2: The Batch Queue will just start applying Holocene rules from this moment onwards. This will then
+- Option 3: The Batch Queue will just start applying Holocene rules from this moment onwards. This will then
 automatically derive gaps as empty batches, discard out of order batches, etc. A problem with this
 option is that the sync start for a while after Holocene would still need to apply the pre-Holocene
 algorithm to restore the derivation pipeline state.
 
-I'm undecided at this point which option is better. I'd prefer to pick the one that leads to a
+At this point I prefer Option 1. I'd like to pick the one that leads to a
 simpler implementation, minimizing complexity where possible. Since almost always a healthy batcher
 implementation during normal network conditions will already produce batches according to Holocene
 rules, we don't need to optimize for the prettiest activation rules, as long as they work
@@ -260,10 +285,36 @@ derivation changes in one go.
 
 ## Unsafe reorgs
 
-Before Steady Batch Derivation, invalid batches got chances to be replaced by valid future
-batches. Because they will now be immediately derived as deposit-only blocks, there is a theoretical
-heightened risk for unsafe head reorgs. However, we haven't experienced this on OP Mainnet or other mainnet
-OP Stack chains yet. The batcher wouldn't automatically detect this and resend 
+Before Steady Batch Derivation, invalid batches got chances to be replaced by valid future batches.
+Because they will now be immediately derived as deposit-only blocks, there is a theoretical
+heightened risk for unsafe head reorgs. To the best of our knowledge, we haven't
+experienced this on OP Mainnet or other mainnet OP Stack chains yet.
+
+The current rules of Steady Batch Derivation can indeed lead to a previously valid submitted batch
+to now be invalid, and then immediately be derived as deposit-only blocks, causing a long L2 unsafe
+reorg. The same batcher tx may still be included on L1. However, given that span batches are only
+forward-invalidated with Holocene, I think the L2 unsafe reorg would be limited to the L2 section
+that references the reorged-out L1 section. However, more batcher txs that might already have landed
+on L1 as well would cause more deposit-only blocks to be derived.
+
+I think this is the trade-off of Steady Batch Derivation.
+
+### Last L1 origin of channel
+
+One solution that I can think of that may alleviate this problem is to reference the last L1 origin
+in the channel metadata (in a new channel format), and then drop the channel directly in the channel
+bank before even decoding any batches from it, that would then at some point be derived as
+deposit-only blocks. This way, the batcher could get a "second chance" to submit a channel that
+includes the correct reorged-to L1 origin chain. This is very similar to how span batches contain
+the last L1 origin as `l1_origin_check` in their prefix, just moved one layer up to the channel
+container. Having the channel contain such an L1 origin check has the advantage that the derivation
+pipeline wouldn't too eagerly derive deposit-only blocks, and to recover from L1 reorgs. I think
+this solution would also still maintain the nice properties of Strict Batch Ordering, that there's
+only one staging channel, and that we don't buffer out of order frames or batches. However, this is
+also the disadvantage that it violates the above stated principle of fastest derivation.
+
+If Proofs and Interop experts can confirm that such a solution would still lead to the sought after
+improvements for Proofs and Interop, resp., we could consider including it as part of Holocene.
 
 ## Multiple batchers & strict ordering
 
--- a/protocol/strict-derivation.md
+++ b/protocol/strict-derivation.md
@@ -225,6 +225,10 @@ is/isn't available".
   - To mimic sequencer behavior, and avoid being hit by shallow L1 reorgs, we could add an
   in-protocol L1 validation depth. So eagerly advancing the L1 origin while maintaining a timestamp
   distance of this validation depth times the L1 bock time.
+  - I believe that, in reality, this validation depth would naturally be observed anyways, because
+  deposit-only blocks would only be derived for invalid or gapped batches, which must already have
+  landed in a channel on L1 by a sequencer that itself is observing a validation depth when creating
+  blocks. So this rule should check out nicely.
 
 # Activation rules & actions
 
--- a/protocol/strict-derivation.md
+++ b/protocol/strict-derivation.md
@@ -1,0 +1,269 @@
+# Purpose
+
+Holocene aims to introduce a set of changes to derivation that improve worst-case scenarios for
+Fault Proofs and Interop. The goal of this design doc is to align on the changes to derivation that
+will be part of the upcoming Holocene hardfork, and in particular to solve open questions to
+finalize the specs.
+
+The Holocene specs for these derivation changes are currently being written in [specs/#357](https://github.com/ethereum-optimism/specs/pull/357).
+
+# Summary
+
+The Holocene hardfork introduces several changes to batch derivation:
+
+- _Partial Span Batch Validity_ determines the validity of singular batches within a span batch
+individually, only invalidating the remaining span batch upon the first invalid batch.
+- _Steady Batch Derivation_ derives invalid singluar batches immediately as deposit-only
+blocks.
+- _Strict Batch Ordering_ requires batches within and across channels to be strictly ordered.
+
+The design space and some proposed solutions will be discussed, together with pratical implications
+for batcher implementations that have to satisfy the stricter rules.
+
+# Partial Span Batch Validity
+
+## Problem Statement + Context
+
+Before Holocene, Span Batch validity is handled atomically, i.e., either all singluar batches that
+can be derived from a span batch are valid, or if any of them is invalid (during batch queue
+checks), the whole span batch is dropped. This has negative implications for Fault Proofs: in a
+worst-case scenario, if the to-be-fault-proven block is the first batch in a span batch, to
+determine its validity at the batch queue stage, the validity of the full span batch has to be
+determined, as any later invalid singular batch would also invalidate all prior batches in the same
+span batch. So only invalidating the remaining span batch upon hitting an invalid derived singular
+batch vastly improves this scenario, since batch queue validity is then final for any singular
+batch, not just regular singular batches.
+
+In late 2023 while Delta was still in development, the problem of partial span batch validity was
+[already discussed](https://www.notion.so/oplabs/Handling-Invalid-Batches-48bb972368ce409c84bd9189eddd0577)
+and for simplicity it was decided to use the atomic validity model. However, above discussion
+focusses on implications of the atomic validity model for EL-sync. We're now concerned with behavior
+around CL client restarts and reorgs.
+
+### Mid-way restart
+
+One concern that was raised with partial span batch validity is how to deal with mid-way restarts,
+i.e. when the CL client restarts, loosing its internal derivation pipeline state while the safe head
+of the EL client is in the middle of a span batch. In particular, the correct behavior needs to
+cover the case where a L1 reorg happened in the meantime that reorged out the L1 block that
+contained the span batch.
+
+I fail to see currently how partial span batch validity makes this case worse or different. With
+atomic span batch validity, there's the pending-safe state of derived blocks that might be reorged
+out later by a backwards-invalidating singular batch derived from a span batch. So if we only
+forwards-invalidate, this becomes easier.
+
+And in any case, a restart or reorg needs to reconstruct the derivation pipeline state from the new
+(sync-start) L1 block, so I don't understand how reconstructing span batch derivation state is
+different from reconstructing the channel bank and batch queue in general. If a restart or reorg
+happens in the middle of deriving singular batches from a channel, we're in a similar situation that
+is solved by going back far enough for the sync-start.
+
+## Proposed Design
+
+To resolve open questions regarding the implementation of Partial Span Batch Validity, I propose to
+apply the principle to consistently change the validation rules to be only forward-invalidating, and
+thus treat a span batch like an additional (optional) stage in the derivation pipeline that sits
+before the batch queue, so that the batch queue pulls singular batches from this previous span batch
+stage. This way, we never backwards-invalidate the span batch, only forward.
+
+This means that sequencer drift validations are applied individually to each singular batch derived
+from the span batch. On a failed check, due to Steady Batch Derivation, the singular batch is
+immediately derived as a deposit-only block, and the remaining span batch is invalidated.
+
+There are still checks that naturally apply to the whole span batch due to its data structure:
+- A span batch only has one parent reference, since all internal batch references are implicit.
+Therefore, a failed parent check invalidates the whole span batch.
+- A failed sequencing window check will always apply to the oldest batches in a span batch, so a
+failed check of the first batch will invalidate the full span batch.
+- If `span_end.timestamp < next_timestamp` (as defined in the Span Batch specs), the whole batch is
+dropped, as it doesn't contain any new batches (this would also happen if applying timestamp checks
+to each derived singular batch individually).
+
+Those checks are done once per span batch before the batch queue derives any singular batches.
+
+# Strict Batch Ordering
+
+Strict batch ordering requires batches within and across channels to be strictly ordered.
+This removes the necessity to buffer more than one channel, and the channel bank effectively holds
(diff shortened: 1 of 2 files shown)


protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+out later by a backwards-invalidating singular batch derived from a span batch. So if we only
+forwards-invalidate, this becomes easier.
+
+And in any case, a restart or reorg needs to reconstruct the derivation pipeline state from the new
+(sync-start) L1 block, so I don't understand how reconstructing span batch derivation state is
+different from reconstructing the channel bank and batch queue in general. If a restart or reorg
+happens in the middle of deriving singular batches from a channel, we're in a similar situation that
+is solved by going back far enough for the sync-start.
+
+## Proposed Design
```
clabby, line 62: I'm a fan of this design - adding a new optional stage here to incrementally stream single batches into the batch queue seems to make sense.

Beyond the outer checks on the span batch prefix, is the validity check on the single batches within the span batch a superset of existing single batch validity rules, or is it the same verbatim?

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+Spoiler: Option 1 will always be the proposed option.
+
+## Out of order frames
+
+There's an open design space around how to handle some scenarios for missing or out of order frames:
+- How do we handle incoming _first_ frames while the current channel isn't closed yet?
+  - Option 1: discard the current channel and open a new one for the given frame.
+  - Option 2: discard the foreign frame and wait for the duration of the channel timeout for the
+    next correct frame and only then discard the current staging channel and open a new staging
+    channel for the next first frame.
```
clabby, line 104: Option 1 seems nice here - effectively acts as a way for the batcher to forcibly close the channel ahead of time in a situation where it made a mistake.

Is there a case you can think of where the batcher may accidentally land the next first frame prior to closing out its ongoing channel first, though? Unsure if we need the added protection against the early submission of a new channel in practice.

clabby: The q is addressed pretty well in the batcher hardening section, closing

protocol/strict-derivation.md
clabby: Beyond that, would the new flow generally look like:
1. Process span batch prefix validity checks
2. `checkSingleBatch` on each single batch streamed over to the batch queue
?

If so, that's awesome for proofs on top of fixing the invalidation issue - getting rid of the overlapped blocks check within the batch validity checking logic should be huge on cycle count / IO. That check was taking ~30% of cycles (per call, done twice re: https://github.com/ethereum-optimism/optimism/issues/11317) last time we profiled kona.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+checks), the whole span batch is dropped. This has negative implications for Fault Proofs: in a
+worst-case scenario, if the to-be-fault-proven block is the first batch in a span batch, to
+determine its validity at the batch queue stage, the validity of the full span batch has to be
+determined, as any later invalid singular batch would also invalidate all prior batches in the same
+span batch. So only invalidating the remaining span batch upon hitting an invalid derived singular
+batch vastly improves this scenario, since batch queue validity is then final for any singular
+batch, not just regular singular batches.
+
+In late 2023 while Delta was still in development, the problem of partial span batch validity was
+[already discussed](https://www.notion.so/oplabs/Handling-Invalid-Batches-48bb972368ce409c84bd9189eddd0577)
```
refcell, line 38: I'd suggest removing any notion links for posterity - folks external to op labs won't be able to access

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+and for simplicity it was decided to use the atomic validity model. However, above discussion
+focusses on implications of the atomic validity model for EL-sync. We're now concerned with behavior
+around CL client restarts and reorgs.
+
+### Mid-way restart
+
+One concern that was raised with partial span batch validity is how to deal with mid-way restarts,
+i.e. when the CL client restarts, loosing its internal derivation pipeline state while the safe head
```
refcell, line 46: small nit
```suggestion
i.e. when the CL client restarts, losing its internal derivation pipeline state while the safe head
```

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+of the EL client is in the middle of a span batch. In particular, the correct behavior needs to
+cover the case where a L1 reorg happened in the meantime that reorged out the L1 block that
+contained the span batch.
+
+I fail to see currently how partial span batch validity makes this case worse or different. With
+atomic span batch validity, there's the pending-safe state of derived blocks that might be reorged
+out later by a backwards-invalidating singular batch derived from a span batch. So if we only
+forwards-invalidate, this becomes easier.
```
refcell, line 54: Agreed.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+happens in the middle of deriving singular batches from a channel, we're in a similar situation that
+is solved by going back far enough for the sync-start.
+
+## Proposed Design
+
+To resolve open questions regarding the implementation of Partial Span Batch Validity, I propose to
+apply the principle to consistently change the validation rules to be only forward-invalidating, and
+thus treat a span batch like an additional (optional) stage in the derivation pipeline that sits
+before the batch queue, so that the batch queue pulls singular batches from this previous span batch
+stage. This way, we never backwards-invalidate the span batch, only forward.
```
refcell, line 68: Curious if the primary driver of this is to remain backwards compatible. It seems to me that this new stage would be very closely similar to the role of the batch queue itself as noted in [the inline docs](https://github.com/ethereum-optimism/optimism/blob/3f7a8ad1b8030a3d01d44bc57bf0553a249eae23/op-node/rollup/derive/batch_queue.go#L15-L28). I guess my question here would then be - is it worthwhile introducing a new stage as opposed to modifying the batch queue?

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+  3. This optional left-over channel
+     will be the start staging channel for the Holocene channel bank.
+
+The principle of eager derivation of later stages should guarantee at this point that all batches that could have
+been applied directly on top of the current safe chain have already been exhausted from the batch
+queue. So what is left should be `undecided` and `future` batches. (Note: I'm not sure on this one.)
+I see two options on how to handle a non-empty batch queue at this point:
+- Option 1: Drop future batches, continue resolving undecided batches, if any are left, and apply
+new Holocene rules.
+- Option 2: The Batch Queue will just start applying Holocene rules from this moment onwards. This will then
```
ajsutton, line 226: Option 3: When the L1 origin reaches the Holocene activation block, discard all batches in the batch queue.  Could also be applied to the Channel Bank to give us a nice clean starting point.

op-batcher would have to be aware of that rule and consider any blocks it submitted in channels that didn't close prior to the holocene activation block as needing to be resubmitted.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+I'm undecided at this point which option is better. I'd prefer to pick the one that leads to a
+simpler implementation, minimizing complexity where possible. Since almost always a healthy batcher
+implementation during normal network conditions will already produce batches according to Holocene
+rules, we don't need to optimize for the prettiest activation rules, as long as they work
+consistently.
+
+Another open question is how to handle span batches that come from pre-Holocene channels.
+I propose that, for simplicity of implementation, if they are found to be valid as a span batch, to
+just apply the new Partial Span Batch Validity rules even though those span batches were derived
+from pre-Holocene L1 blocks.
```
ajsutton, line 240: This would only apply to channels that span the Holocene activation right? Anything full pre-holocene would use the old rules and anything full post-Holocene would use the new rules.  If so I agree, though it wouldn't be an issue if we discarded any incomplete channels at Holocene activation.

protocol/strict-derivation.md
sebastianst: I also thought about this option, but deemed it too drastic. Hearing that you support it makes me consider it! I like its simplicity.

We just need to be careful that the batcher doesn't end up in a very unlikely situation where we're batching with calldata and have a single block that needs to be sent over two calldata frame txs that span across the Holocene activation.

sebastianst: Thanks for flagging, I also thought about this option, but deemed it too drastic. Hearing that you support this option, I certainly consider it! I like the simplicity of it. 
We just need to be careful with the batcher to arrive in a very unlikely situation where we're sending calldata txs and need to batch a single block that doesn't fit into a single calldata tx and then have that channel span across the Holocene activation. Very unlikely, but possible.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+I propose to let the batcher follow a fixed nonce to block-range assignment, once the first batcher
+transaction (which is almost always the only batcher transaction for a channel) starts being
+submitted. This should avoid out-of-order or gapped batches. It might require to implement some form
+of persistence in the transaction manager, since we cannot reliably recover all globally pending
+batcher transactions in the L1 network.
+
+Furthermore, the batcher will need to be made aware of the Steady Batch Derivation rules, namely that
+invalid batches will be derived as deposit-only blocks. So in case of a reorg, the batcher should
+e.g. wait on the sequencer it is connected to until it has derived all blocks from L1 in order to
+only start batching new blocks on top of the possibly derived deposit-only chain segment.
```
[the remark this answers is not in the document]
axelKingsley: Thinking through this a bit, I'm having trouble deciding if this is or isn't a problem.

If an L1 reorg occurs, it only affects the L2 if a batch is wiped out.

If a batch is wiped out, then the batcher's nonce is reverted. Per Seb's proposal "a fixed nonce to block-range assignment", the batcher wants to resubmit on the same range as before, so there isn't a risk of the batcher posting a batch-gap (which would create empty blocks).

🤔 🤔 🤔

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+discarded. This is intentional. This design allows for not having to submit empty batches to
+derive deposit-only blocks, but instead just pick the next non-deposit-only block.
+
+Writing this, I realize that this mechanism could even be used to encode large spans of empty
+batches, as long as the sequencer is creating unsafe blocks that follow the same L1 origins as the
+auto-derivation would for gaps. However, this would need to be investigated more deeply.
+
+Note that the new strict ordering rules of the batch queue will always lead to an empty batch queue
+when the origin of the derivation pipeline progresses to the next L1 block (what about
+`undecided` though?).
```
axelKingsley, line 149: I am not familiar with `undecided`, and I see another reference to it below. Can you explain how these behave briefly?

protocol/strict-derivation.md
axelKingsley: Even if we did find ourselves in that situation, discarding that in-progress submission would work right? It would need to be retried after the threshold.

Or, we could follow Adrian's suggestion and drop all batches, but do it some time *before* holocene activation, holding a ban on batches until holocene passes. This would allow for the unlikely big-block to resolve prior to passing the threshold.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+of changes over multiple hardforks would create more work that can be saved if applying all those
+derivation changes in one go.
+
+# Risks & Uncertainties
+
+## Unsafe reorgs
+
+Before Steady Batch Derivation, invalid batches got chances to be replaced by valid future
+batches. Because they will now be immediately derived as deposit-only blocks, there is a theoretical
+heightened risk for unsafe head reorgs. However, we haven't experienced this on OP Mainnet or other mainnet
```
axelKingsley, line 258: Would a search of op-mainnet logs help identify if we've had situations which would cause Unsafe Reorgs?

tynes:
One thing to consider here is that we have generally assumed that the batcher will not be malicious and submit batches that are expensive for derivation to process. This assumption can work well for a single chain that has a single sequencer, but in the world of interop the finality of a single chain becomes tied to the finality of another chain. This means that it could be possible for a batcher of a remote chain to influence the cost of proof generation for the local chain. This means that we need to start thinking about untrusted batchers in the context of interop.

There will be some reputation at play since the interop set will be managed by governance, but we should generally strive to minimize the amount of reputation required for security. This just means that we don't need to ship the absolute most denial of service proof thing in the world as the first iteration.

protocol/strict-derivation.md
tynes: We should definitely make it public

BlocksOnAChain: I don't see a reason not to share, after 1 more review of the doc, I made it public.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+
+The Holocene specs for these derivation changes are currently being written in [specs/#357](https://github.com/ethereum-optimism/specs/pull/357).
+
+# Summary
+
+The Holocene hardfork introduces several changes to batch derivation:
+
+- _Partial Span Batch Validity_ determines the validity of singular batches within a span batch
+individually, only invalidating the remaining span batch upon the first invalid batch.
+- _Steady Batch Derivation_ derives invalid singluar batches immediately as deposit-only
```
geoknee, line 16:
```suggestion
- _Steady Batch Derivation_ derives invalid singular batches immediately as deposit-only
```

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+- _Strict Batch Ordering_ requires batches within and across channels to be strictly ordered.
+
+The design space and some proposed solutions will be discussed, together with pratical implications
+for batcher implementations that have to satisfy the stricter rules.
+
+# Partial Span Batch Validity
+
+## Problem Statement + Context
+
+Before Holocene, Span Batch validity is handled atomically, i.e., either all singluar batches that
```
geoknee, line 27:
```suggestion
Before Holocene, Span Batch validity is handled atomically, i.e., either all singular batches that
```

protocol/strict-derivation.md
sebastianst: The singular batch checks are the same. The outer span batch checks are really just more of an early discard of the whole span batch because if these checks fail, there's no way that the singular batches pulled out of them could be valid.
The flow is correct. We could add _3. if the singular batch check fails, the remaining span batch is discarded_
This has to be because span batches don't encode the parent block hashes, so they are implicit. If any singular batch is invalid, it will be derived as a deposit-only block which will have a different block hash, so the implicit block hash must be assumed wrong.

protocol/strict-derivation.md
sebastianst: The primary driver of this concern was @protolambda so I'll leave it for him to mark this as resolved.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+I fail to see currently how partial span batch validity makes this case worse or different. With
+atomic span batch validity, there's the pending-safe state of derived blocks that might be reorged
+out later by a backwards-invalidating singular batch derived from a span batch. So if we only
+forwards-invalidate, this becomes easier.
+
+And in any case, a restart or reorg needs to reconstruct the derivation pipeline state from the new
+(sync-start) L1 block, so I don't understand how reconstructing span batch derivation state is
+different from reconstructing the channel bank and batch queue in general. If a restart or reorg
+happens in the middle of deriving singular batches from a channel, we're in a similar situation that
+is solved by going back far enough for the sync-start.
```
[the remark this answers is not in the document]
sebastianst: The primary driver of this concern was @protolambda so I'll leave it for him to mark this as resolved.

protocol/strict-derivation.md
sebastianst: I think an _implementation_ can choose to do either, have it as a separate stage, or part of the batch queue. I like the modularity of thinking of it as a separate stage _in the abstract_.

protocol/strict-derivation.md
sebastianst: Yes I think it is reasonable to implement some sort of safety behavior for the batcher close to the Holocene activation. We didn't do a good job of adding special hardfork activation logic to the batcher in the past but I think past batcher operating experience has shown that it may be worthwhile adding some.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+staged channel. Note that this will give the staging channel a chance to be continued or closed
+properly by the next in-order frame. If this doesn't happen, at the lastest the next first frame
+will drop the staging channel and start a new one.
+
+### Proposed solution
+
+Applied to above open question, this means choosing option 1. This has the
+additional advantage that the sync start algorithm simplifies significantly, because a fist frame
+_always_ marks the start of a new channel, and guarantees that there won't be any buffered channels
+in the channel bank.
```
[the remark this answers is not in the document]
sebastianst: 💯 this has also been part of the decision for option 1. It would otherwise be too risky to accidentally leave a channel open and cause a long chain halt.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+by gaps.
+
+It should be noted that if the blocks after the gap actually come from a chain where the gaps
+aren't deposit-only blocks, then these blocks would have mismatched parent hashes and be likely
+discarded. This is intentional. This design allows for not having to submit empty batches to
+derive deposit-only blocks, but instead just pick the next non-deposit-only block.
+
+Writing this, I realize that this mechanism could even be used to encode large spans of empty
+batches, as long as the sequencer is creating unsafe blocks that follow the same L1 origins as the
+auto-derivation would for gaps. However, this would need to be investigated more deeply.
```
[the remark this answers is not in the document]
sebastianst: Yes this is also a concern I have, and is probably the most underspecified bit. Random thoughts around that:
- A simple default rule could be to generate blocks in a way to maintain a steady L1/L2 block ratio, e.g. bumping the L1 origin selection every 6 blocks in the case of mainnet.
- Edge case: we were already very near the sequencer drift limit, and need to select a new origin faster.
- Edge case: L1 missed a slot, and the L1 origin cannot advance as expected.
- So maybe a better default rule is to first eagerly advance the L1 origin as quickly as possible, and only if a newer L1 origin isn't available, keep it. This solves for missed slots and will implicitly and automatically maintain a good L1/L2 block ratio. We then just need a clear definition of "L1 origin is/isn't available".

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+batch vastly improves this scenario, since batch queue validity is then final for any singular
+batch, not just regular singular batches.
+
+In late 2023 while Delta was still in development, the problem of partial span batch validity was
+[already discussed](https://www.notion.so/oplabs/Handling-Invalid-Batches-48bb972368ce409c84bd9189eddd0577)
+and for simplicity it was decided to use the atomic validity model. However, above discussion
+focusses on implications of the atomic validity model for EL-sync. We're now concerned with behavior
+around CL client restarts and reorgs.
+
+### Mid-way restart
```
geoknee, line 43: To me it would be more natural to put this section below the "Proposed Design" section, especially since it apparently raises a potential issue but then dismisses it.

protocol/strict-derivation.md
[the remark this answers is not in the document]
sebastianst: This is a valid concern, and has been flagged by @tynes in his original design doc's [risk section](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/invalid-block-progression.md#risks--uncertainties).

The current rules of Steady Batch Derivation can indeed lead to a previously valid submitted batch to now be invalid, and then immediately be derived as deposit-only blocks, causing a long L2 unsafe reorg. The same batcher tx may still be included on L1. However, given that span batches are only forward-invalidated with Holocene, I think the L2 unsafe reorg would be limited to the L2 section that references the reorged-out L1 section. However, more batcher txs that might already have landed on L1 as well would cause more deposit-only blocks to be derived.

I think this is the tradeoff of Steady Batch Derivation.

One solution that I can think of that may alleviate this problem is to reference the last L1 origin in the channel metadata (in a new channel format), and then drop the channel directly in the channel bank before even decoding any batches from it, that would then at some point be derived as deposit-only blocks. This way, the batcher could get a "second chance" to submit a channel that includes the correct reorged-to L1 origin chain. This is very similar to how span batches contain the last L1 origin as `l1_origin_check` in their prefix, just moved one layer up to the channel container. With such a new channel format, the L1 origin check could arguably be dropped from span batches. Having the channel, rather than any span or singular batch, contain such a L1 origin check has the advantage that the DP wouldn't too eagerly derive deposit-only blocks, and to recover from L1 reorgs. I think this solution would also still maintain the nice properties of Strict Batch Ordering, that there's only one staging channel, and that we don't buffer out of order frames or batches.

If Proofs and Interop experts can confirm that such a solution would still lead to the sought after improvements for Proofs and Interop, resp., we could consider including it as part of Holocene.

protocol/strict-derivation.md
sebastianst: It would technically also apply to full pre-Holocene span batches, if the channel that contained this span batch was already closed before Holocene activation, but included on L1 only after Holocene activation.

Since correct batcher implementations don't violate this property anyways, I just want to pick the rule with the lowest implementation lift.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+
+Before Holocene, Span Batch validity is handled atomically, i.e., either all singluar batches that
+can be derived from a span batch are valid, or if any of them is invalid (during batch queue
+checks), the whole span batch is dropped. This has negative implications for Fault Proofs: in a
+worst-case scenario, if the to-be-fault-proven block is the first batch in a span batch, to
+determine its validity at the batch queue stage, the validity of the full span batch has to be
+determined, as any later invalid singular batch would also invalidate all prior batches in the same
+span batch. So only invalidating the remaining span batch upon hitting an invalid derived singular
+batch vastly improves this scenario, since batch queue validity is then final for any singular
+batch, not just regular singular batches.
```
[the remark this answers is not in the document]
sebastianst: As mentioned in other discussions, singular batches within span batches don't explicitly state the parent hash, so they have to be assumed to implicitly reference the parent singular batch. So if that parent singular batch is replaced by an empty batch, for consistency reasons, all future singular batches in a span batch have to be considered invalid.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+
+The following will explore the open design space, presenting options, and propose a solution.
+The principle I would apply to answer open design questions is that the new rules should generally
+choose the design that leads to faster progess of derivation, even if it means deriving
+empty batches earlier.
+Spoiler: Option 1 will always be the proposed option.
+
+## Out of order frames
+
+There's an open design space around how to handle some scenarios for missing or out of order frames:
```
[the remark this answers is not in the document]
sebastianst: Yes, batcher restarts can in theory cause the batcher to not close an open channel and then submit a new one.
Then there's also the scenario where a batcher cannot get a tx on-chain and requeues the blocks and then attempts sending them again in a new channel. Depending on the specifics of the batcher implementation, this can also lead to out of order frames. All quite unlikely and almost never happening, but we need to take extra care to harden the batcher against such behavior in the future.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,269 @@
...
+  - Option 1: immediately fill the gap with deposit-only blocks, and then process the batch.
+  - Option 2: drop future batches, and wait for up to a full sequencing window for the next
+  batch to arrive.
+
+### Proposed solution
+
+Following the principle of fast derivation, option 1 is proposed to be chosen. Besides guaranteeing
+faster derivation, it also avoids waiting for full sequencing windows on gaps to be resolved, which
+is quite unlikely, unless the batcher implementation would be specially tuned to backfill gaps. If a
+gap occurs, it is actually much more likely that the next batches will be even further in the
```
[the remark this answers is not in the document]
sebastianst: Maybe some very unlikely of an L1 reorg, that doesn't cause an L2 reorg yet because the unsafe chain uses older L1 origins, then that L1 reorg removes the lastest batcher tx and the next channel is included by the txmgr automatically adjusting the nonce. Very constructed case, but not impossible.

protocol/strict-derivation.md
sebastianst: fixed in https://github.com/ethereum-optimism/design-docs/pull/73 right?

protocol/strict-derivation.md
sebastianst: fixed in https://github.com/ethereum-optimism/design-docs/pull/73 right?

protocol/strict-derivation.md
sebastianst: fixed in https://github.com/ethereum-optimism/design-docs/pull/73 right?

protocol/strict-derivation.md
sebastianst: I added it here because it first explains a problem, and then I just gave my reasoning as to why I believe it isn't an issue, but I don't know if it's the correct answer. I will consider moving it around.

protocol/strict-derivation.md
sebastianst: yes, but we retain logs for only 30 days unfortunately...

protocol/strict-derivation.md
sebastianst: The `undecided` status is still not 💯 clear to me, which is why I added those questions in brackets. What I understand from the implementation (files `batches.go` and `batch_queue.go`) is that if there's missing L1 or L2 data (e.g. due to temporary broken connections?), the batch is re-queued for later checking and no batch is processed at this point. I think we will just retain this behavior.

The difference to `future` is that a future batch is already determined to be out of order and lies in the future. The `future` case (a gap) will be treated differently in Holocene, and the gap will immediately be derived as deposit-only blocks. It is noteworthy that the future batch can then only be valid if it built on top of a gap of empty batches.

Added this to the design doc.

protocol/strict-derivation.md
sebastianst: Made this the new and preferred Option 1.

protocol/strict-derivation.md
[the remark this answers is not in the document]
sebastianst: Added this to the design doc.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,355 @@
...
+- _Steady Batch Derivation_ derives invalid singular batches immediately as deposit-only
+blocks.
+- _Strict Batch Ordering_ requires batches within and across channels to be strictly ordered.
+
+The design space and some proposed solutions will be discussed, together with practical implications
+for batcher implementations that have to satisfy the stricter rules.
+
+# Partial Span Batch Validity
+
+## Problem Statement + Context
```
protolambda, line 25: One aspect of the problem that is worth mentioning, although solved by the same partial-validity idea, is that interop fault-proofs require a block to be processed optimistically, and then the proof might later abort on cross-L2 interop-dependencies. After having aborted, the alternative chain continuation should be as straight-forward and minimal as possible, to avoid looping interop dependency checks. Falling back to a deposit-only block, when a batch is invalid, would serve this well.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,355 @@
...
+contained the span batch.
+
+I fail to see currently how partial span batch validity makes this case worse or different. With
+atomic span batch validity, there's the pending-safe state of derived blocks that might be reorged
+out later by a backwards-invalidating singular batch derived from a span batch. So if we only
+forwards-invalidate, this becomes easier.
+
+And in any case, a restart or reorg needs to reconstruct the derivation pipeline state from the new
+(sync-start) L1 block, so I don't understand how reconstructing span batch derivation state is
+different from reconstructing the channel bank and batch queue in general. If a restart or reorg
```
protolambda, line 88: I think you're correct here. The span-batch as a whole should be retrievable, assuming we always revisit the channel-window fully.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,355 @@
...
+_always_ marks the start of a new channel, and guarantees that there won't be any buffered channels
+in the channel bank.
+
+## Out of order batches
+
+The batch queue also becomes simpler as batches have to arrive in order. 
+
+- How to treat _older_ batches is easy: they are dropped.
+- Similarly to out of order frames, how to treat future batches (gaps) has two options:
+  - Option 1: immediately fill the gap with deposit-only blocks, and then process the batch.
```
protolambda, line 131: I agree with option 1, but an implementation detail we need to keep an eye on is that we don't run out of memory, and stream out the deposit-only blocks that fill the gap, rather than creating them all at once. Since the gap can be very large.

And the gap might have to be limited in size for fault-proofs still. Even with sufficient memory, and streaming rather than all-at-once allocation, it might still require a lot of processing.
However, without making the channel-window the effective new smaller sequencing window, there might still be a lot of deposit-only blocks to churn through.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,355 @@
...
+## Frame Queue, Channel Bank, Batch Queue
+
+However, we need to specify how existing buffered frames, channels and batches are treated at
+activation, since the stricter rules of Holocene take effect.
+There are multiple options to enter Holocene with a derivation pipeline state that is consistent
+with Holocene rules.
+
+- Option 1: When the L1 origin reaches the Holocene activation block, discard all frames in the
+  Frame Queue, channels in the Channel Bank, and batches in the Batch Queue.
+  This gives us a nice clean starting point.
```
protolambda, line 271: :+1: clean slate during upgrade is nice. We should add some warnings to the batch-submitter about this though, to prevent operational mistakes.

protolambda (approved):

protocol/strict-derivation.md
sebastianst: An implementation hint about this should be added to the specs

protocol/strict-derivation.md
```diff
@@ -0,0 +1,355 @@
...
+  - This is similar to Option 1, but rather maintains the current sequencer drift, instead of
+  possibly more eagerly advancing it to get closer to the latest L1 origin.
+  - Edge case: the safe head is already very near the sequencer drift limit, and a newer origin
+  needs to be selected faster. This complicates this solution.
+  - Edge case: L1 missed a slot, and the L1 origin cannot advance as expected. This also needs to be
+  taken care of in Option 1.
+- Option 3: keep last L1 origin as long as possible, so until hitting the sequencing window. This is
+the simplest option, but also means that a sequencer has to fast-advance the L1 origin to catch up
+again. And if the L1 origin already slipped out of the drift window, a sequencer would need to
+continue creating deposit-only blocks until inside the drift window again.
```
sebastianst, line 239: In the current implementation, it would actually already advance to keep within the drift window.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,355 @@
...
+deposit-only blocks. I see 3 sensible options:
+
+- Option 1: Eagerly advance the L1 origin. This solves for missing L1 slots and will implicitly and
+automatically maintain a good L1/L2 block ratio.
+  - This guaranteed fast deposit inclusion for users and is similar to how a live sequencer
+  operates.
+  - Option 1.a: is a variation of this, where an in-protocol validation depth is maintained, so that
+  there's a constant sequencer drift, e.g. of 4 L1 blocks. This is even more similar to how a live
+  sequencer operates. This has the benefit that there's minimal disruption to the way that L1
+  origins change when going in and out of auto-derivation.
```
sebastianst, line 223: Option 1 feels better than 1.a currently, we will see during implementation, but go with Option 1 for now.

protocol/strict-derivation.md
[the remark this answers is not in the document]
sebastianst: Instead of doing this channel L1 origin check, actually just throw away span batches with failing L1 origin check, but give a second chance to be replaced, and don't generate deposit-only blocks for the whole span batch range.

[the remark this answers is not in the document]
sebastianst: To fully benefit from this, only decode the prefix first, do the check.

protocol/strict-derivation.md
```diff
@@ -0,0 +1,355 @@
...
+    gaps as empty batches, discard out of order batches, etc. A problem with this option is that the
+    sync start for a while after Holocene would still need to apply the pre-Holocene algorithm to
+    restore the derivation pipeline state.
+
+At this point I prefer Option 1. I'd like to pick the one that leads to a simpler implementation,
+minimizing complexity where possible. Since almost always a healthy batcher implementation during
+normal network conditions will already produce batches according to Holocene rules, we don't need to
+optimize for the prettiest activation rules, as long as they work consistently.
+The batcher would have to be aware of those rules and consider any blocks it submitted in channels
+that didn't close prior to the Holocene activation block as needing to be resubmitted.
```
sebastianst, line 293: Also need to make sure that there aren't queued up batcher txs that are included after. These would cause gaps, that would auto-derive, then cause L2 reorg. Add special rules to the batcher.

sebastianst:
Further public discussion in [Discord](https://discord.com/channels/1244729134312198194/1283809784444751913) made us reconsider the idea of deriving invalid _batches_ as empty batches. The new proposal is to simply drop invalid, and future, batches, and instead only derive invalid _payloads_ at the engine stage as deposit-only payloads. An invalid payload wouldn't trigger the generation of future empty batches, but instead just forward-invalidate any remaining batches and the origin channel.

BlocksOnAChain (approved):
I think this design doc is now is "good-enough" state and that we can merge it, so we can move towards the actual development work for holocene.

BlocksOnAChain:
> > I think this design doc is now is "good-enough" state and that we can merge it, so we can move towards the actual development work for holocene.
> 
> @BlocksOnAChain The design changed significantly due to the async discussions after the design session. I want to first finish the spec, then adapt the design doc so it can serve as a historical reference, before we merge it.

Got it, made sense to me. I was just reviewing the current state, before the last edits and after we chatted on discord. fully agree.

PR #75, opened 2024-09-09
Holocene: protocol -  proposal for action-test based test-vectors

protolambda:
**Description**

Early draft of a proposal for using action-tests as base for test-vector generation.

@clabby and @refcell own the direction for multi-client test vectors (feel free to make design-doc edits).

### DIFF
--- a/protocol/exporting-action-tests.md
+++ b/protocol/exporting-action-tests.md
@@ -1,0 +1,151 @@
+# Purpose
+
+<!-- This section is also sometimes called “Motivations” or “Goals”. -->
+
+<!-- It is fine to remove this section from the final document,
+but understanding the purpose of the doc when writing is very helpful. -->
+
+Write state-transition tests against the Go stack, by reusing the action-test framework,
+and export the tests for verification of Fault-Proof programs, L2 CL nodes, and L2 EL nodes. 
+
+# Summary
+
+<!-- Most (if not all) documents should have a summary.
+While the length will likely be proportional to the length of the full document,
+the summary should be as succinct as possible. -->
+
+In the monorepo we have `op-e2e/actions`:
+a set of Go integration tests, that run against stripped-down in-process nodes.
+These tests run directly through the state-transition: no clocks or concurrent processes run.
+
+The involved L1 and L2 chains are valid, and could theoretically be exported,
+to reproduce the state-transition elsewhere.
+
+This can serve the multi-client onchain testing:
+the same derivation and EVM processing can be reproduced
+in an external client or fault-proof program.
+
+# Problem Statement + Context
+
+<!-- Describe the specific problem that the document is seeking to address as well
+as information needed to understand the problem and design space.
+If more information is needed on the costs of the problem,
+this is a good place to that information. -->
+
+Testing of the OP-Stack can be categorized in two categories:
+- `onchain`: focus on state-transition correctness. Includes derivation and EVM processing.
+- `offchain`: focus on RPC compatibility, syncing stability, scheduling behavior.
+
+Or in other words, `onchain` is primarily the Safety aspect of the classic security charter,
+whereas `offchain` is primarily the Liveness part of it.
+
+The `onchain` category is critical to the correctness of the fault-proof program.
+To reproduce all the aspects of it, we have need to run derivation:
+the process that turns L1 data into L2 inputs.
+
+This is more involved than an EVM process, as it relies on structured input on L1,
+including receipts, compression, batch-data channels, and blobs.
+
+Thus, to produce correct test-vectors,
+we need a process that is set up to create this type of data.
+
+# Proposed Solution
+
+<!-- A high level overview of the proposed solution.
+When there are multiple alternatives there should be an explanation
+of why one solution was picked over other solutions.
+As a rule of thumb, including code snippets (except for defining an external API)
+is likely too low level. -->
+
+Similar to how the Ethereum L1 Consensus-Layer test-vectors (beaconchain only) are generated,
+and more recently the Ethereum L1 Execution Layer test-vectors (EVM only),
+we can piggy-back on a reference implementation, and export the artifacts.
+
+The action tests in the Optimism monorepo can serve as the same kind of testing:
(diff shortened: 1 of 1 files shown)


protocol/exporting-action-tests.md
```diff
@@ -0,0 +1,151 @@
...
+- L2 blocks, receipts, pre/post-state.
+
+These modifications are likely special cases in the L1 and L2 engine actors,
+to export the produced blocks to e.g. JSON files, and any other data such as state.
+
+Potentially, the op-program DB can be reused in the action-tests,
+to capture witness data, such that individual blocks can be executed in a FP setting.
+
+A Fault-Proof program, when given a L1 chain and L2 chain,
+and potentially witness data, should be able to reproduce the state-transition,
```
sebastianst, line 100: What exactly comprises the "witness data"?

protocol/exporting-action-tests.md
```diff
@@ -0,0 +1,151 @@
...
+# Risks & Uncertainties
+
+<!-- An overview of what could go wrong.
+Also any open questions that need more work to resolve. -->
+
+## Time allocation
+
+The action-tests are a shared responsibility, but primarily maintained by Proto,
+who is occupied with Interop, and not directly involved in Holocene and Proofs work.
+There may be some onboarding required to be effective.
```
sebastianst, line 150: I think it makes a lot of sense for Holocene Derivation devs (@geoknee, me primarily at this point) and Proofs devs involved in multi-client testing (@clabby @refcell ?) to do the grunt work here, with only guidance provided by Proto. Holocene needs more features from action tests for testing derivation edge cases including out-of-order and invalid frames and batches, and multi-client testing needs the chain data exports.

protocol/exporting-action-tests.md
```diff
@@ -0,0 +1,151 @@
...
+
+<!-- What is the resource usage of the proposed solution?
+Does it consume a large amount of computational resources or time? -->
+
+The action-tests already run quite rapidly: there is no scheduling to wait for.
+
+However, exporting the test-data will likely add some overhead:
+DB features like witness-generation, and file-IO.
+
+By making this test-data export optional, the regular action-tests can still run quickly in CI.
```
sebastianst, line 132: I'm also not worried about added performance impact. The data export wouldn't run in the monorepo CI and would only be used sparingly to generate reference test vectors on-demand.

protocol/exporting-action-tests.md
```diff
@@ -0,0 +1,151 @@
...
+# Problem Statement + Context
+
+<!-- Describe the specific problem that the document is seeking to address as well
+as information needed to understand the problem and design space.
+If more information is needed on the costs of the problem,
+this is a good place to that information. -->
+
+Testing of the OP-Stack can be categorized in two categories:
+- `onchain`: focus on state-transition correctness. Includes derivation and EVM processing.
+- `offchain`: focus on RPC compatibility, syncing stability, scheduling behavior.
```
sebastianst, line 37: Helpful categorization!

sebastianst (approved):
This looks like a good path forward to provide multi-client test vectors. After understanding the action tests framework better now, I also believe that it is sufficient and the right tool for Holocene edge-case derivation rules testing, better than a custom test batcher that can be controlled to produce edge-cases.

The actions testing framework certainly deserves some love to improve its usability, but I regard it within the scope of Holocene to improve the framework. Plus, with help from multi-client devs @clabby and @refcell to provide the data export implementations as laid out in this design doc, this seems like the best path forward to me.

sebastianst:
One line of work that I'd like to invest in as part of Holocene is to extract bits out of the batcher and make them reusable in the action tests. There's currently too much custom batcher implementation in the action tests, that we should be able to deduplicate.

joohhnnn:
Hi @protolambda , could you clarify the scope of the data export? For instance, do we need to export all data, such as the pre and post states for all block mining and call executions? Or do we just need to record the pre and post states for transactions involving key components?

protocol/exporting-action-tests.md
```diff
@@ -0,0 +1,151 @@
...
+we need to also export the reorg signals and safety signals,
+so these offchain concepts can be acted on by the test target.
+
+For interop, ideally the export format includes test-actor names and chain-IDs,
+so we can support multi-chain tests.
+
+### action-tests DSL
+
+The current "actions" DSL has a few quirks, we may want to improve on:
+- Some actor functions produce an action, which then still has to be executed.
```
GrapeBaBa, line 113: Do you mean method like this https://github.com/ethereum-optimism/optimism/blob/develop/op-e2e/actions/l1_miner.go#L148, which becomes direct execution and does not return the function?

protocol/exporting-action-tests.md
```diff
@@ -0,0 +1,151 @@
...
+  The original idea here was to randomize/fuzz actions; this may still be on the table, but will require more time.
+- The test-params / genesis generation is quite limited, due to not being able to customize the allocs.
+    - Different genesis parameters can be applied by running the forge scripts as part of the setup.
+      See `interopgen` work for an example of producing a L1 state and multiple attached L2s, with custom configuration, within a second or two.
+    - Forge scripts could help produce custom pre-states. The test could run these scripts as part of the setup phase, using the same tooling as the above.
+
+These things would be worthwhile to improve as part of the testing effort, benefiting both the Go stack, and quality of exported test vectors.
```
GrapeBaBa, line 120: These things don't seem to be directly related to the export test vector, and it seems that they can be optimized now without any blocking

BlocksOnAChain:
@clabby @refcell  - Since I believe that we kinda made the decision on how we plan to proceed with testing for holocene and which approach we can use for action-tests, we can maybe close this PR now?
If you folks are ok with that?

BlocksOnAChain:
Closing this with a comments from our weekly:
- **Ownership for Action-testing framework  -**  For holocene: We believe this needs to be done together between.

@Ben Clabby -Proofs team already has everything they need. For us, it just reusing what we already have

@Sebastian Stammler On the other side, Holocene tests - we will create a framework that we need for Holocene, based on action tests.

PR #76, opened 2024-09-13
feat: AnchorStateRegistry poisoning protection

smartcontracts:
Proposes a change to the AnchorStateRegistry that would resolve the poisoning problem.

### DIFF
--- a/protocol/proofs/anchor-state-poison-protection.md
+++ b/protocol/proofs/anchor-state-poison-protection.md
@@ -1,0 +1,137 @@
+# AnchorStateRegistry Poisoning Mitigation
+
+## Purpose
+
+Provides a design for an updated version of the `AnchorStateRegistry` with a clearer role within
+the system as a whole that is not subject to the "poisoning" problem found in the current design.
+
+## Context
+
+The `FaultDisputeGame` contract is responsible for managing the process of disputing a fault in the
+computation of a state transition function. The `FaultDisputeGame` runs the state transition
+function from some starting state that is known to be valid. We refer to this starting state as the
+"anchor state". In the context of the OP Stack, the anchor state will originally either be the
+genesis state or the state of the system at some finalized L2 block height that is considered to be
+correct by social consensus.
+
+`FaultDisputeGame` contracts pull the current anchor state from the `AnchorStateRegistry` contract.
+As the system runs, the anchor state is continuously updated to reflect the latest finalized state
+as viewed by the proof system at any given time. `FaultDisputeGame` instances that resolve in favor
+of the defender will trigger the `AnchorStateRegistry` to update the anchor state to the newly
+finalized state.
+
+## Problem Statement
+
+The `AnchorStateRegistry` can become "poisoned" because the ultimate view of state validity
+currently actually lives inside of the `OptimismPortal` contract. An anchor state is considered
+"poisoned" if the `OptimismPortal` contract has blacklisted the dispute game that caused that
+anchor state to be recognized by the `AnchorStateRegistry`. Any new `FaultDisputeGame` contracts
+that load in this anchor state will operate over a starting state that the `OptimismPortal`
+considers to be invalid and will therefore also be considered invalid.
+
+Although a basic mitigation was added in the Granite upgrade that made it possible to reset the
+anchor state to some different (valid) state, the fundamental problem still remains that poisoned
+`FaultDisputeGame` contracts may continue to resolve and further poison the anchor state registry.
+In this cascading case, the only real solution at the moment is to switch to a new game type as
+each game type has its own anchor state mapping and therefore is not impacted by any poisoned
+anchor states in other game types.
+
+Our problem statement here is therefore that we wish to be able to prevent or mitigate anchor state
+poisoning more effectively so that we can reduce the surface for issues within the fault proof
+system as a whole.
+
+## Proposed Solution
+
+We propose solving this problem by updating the `AnchorStateRegistry` to more generally be the
+source of truth for the validity of states from the perspective of the L1 proof system. All
+functionality in the `OptimismPortal` related to state validity would be moved to the
+`AnchorStateRegistry` contract. This proposed solution would simplify the `OptimismPortal`
+contract, clarify the concept of state validity from the PoV of the proof system, and solve the
+state poisoning issue within the `AnchorStateRegistry`.
+
+### Changes to the AnchorStateRegistry
+
+The `AnchorStateRegistry` would be modified to take over the blacklisting, respected game type,
+and air-gap roles that are currently handled by the `OptimismPortal` contract. The
+`AnchorStateRegistry` would expose several new getter functions that can be used by the
+`OptimismPortal` contract to determine if a given dispute game can be permitted to prove or
+finalize a withdrawal.
+
+The `AnchorStateRegistry` would continue to keep track of the results of `FaultDisputeGame`
+contracts via the current "push" model where the `FaultDisputeGame` attempts to trigger an update
+in the `AnchorStateRegistry` on game finalization. However, instead of immediately updating the
+anchor state when a game is finalized, the `AnchorStateRegistry` would instead store the result in
+a log of game results. Game results would only be promoted to anchor states if they've passed the
+air-gap window without being blacklisted.
+
+We can avoid making significant changes to the current challenger setup by using the call to the
+`tryUpdateAnchorState` function that the `FaultDisputeGame` already makes to both store the game
+result in the game array. The `AnchorStateRegistry` could then iterate over some number of elements
+in this array every time `tryUpdateAnchorState` is triggered to try to find a new anchor state. As
+games tend to be published at a relatively regular interval, this means that the
+`AnchorStateRegistry` will tend to find new anchor states regularly without too much trouble. We
+can also expose a new permissionless method on the `AnchorStateRegistry` that allows anyone to show
+that a newer anchor state is valid to manually push up the anchor state if necessary.
+
+The `AnchorStateRegistry` would no longer keep track of different anchor states for different game
+types. Each game type would share a single anchor state. The `AnchorStateRegistry` would only use
+the results from the currently respected game type to update the anchor state. Note that this
+proposal does NOT support the ability for multiple proof systems to independently update the anchor
+state (i.e., possibility for multiple respected game types at the same time) as that feature was
+not discussed during the design review meetings.
+
+### Changes to the OptimismPortal
+
+State validity rules like blacklisting and air-gap window inside of the `OptimismPortal` would be
+removed and outsourced to the `AnchorStateRegistry`. New public getter functions in the
+`AnchorStateRegistry` would provide the information that the `OptimismPortal` needs to determine
+whether a given dispute game contract can be used to finalize a withdrawal.
+
+### Changes to the FaultDisputeGame
+
+No changes to the `FaultDisputeGame` contract would necessarily be required. `anchors` function can
+be made backwards compatible to always return the single anchor state now that anchor states will
+not be tracked on a per-game basis. We can optionally update the `FaultDisputeGame` to use whatever
+new function exposes the single `anchor` if desired.
+
+### Changes to the DeputyGuardianModule
+
+The `DeputyGuardianModule` will have a few light changes that will mean that the DGM will need to
+be redeployed. Any redeploy of the DGM breaks PSPs and means we would need to regenerate new PSPs.
+
+## Upgrade and Migration Path
+
+We must guarantee that the security properties of the previous system are maintained into the
+updated system. In particular, any games that have been blacklisted in the old system must also be
+blacklisted in the updated system. Additionally, it may be necessary to forward certain getter
+functions currently available on the `OptimismPortal` contract like `disputeGameBlacklist` over to
+the `AnchorStateRegistry`.
+
+## Alternatives Considered
+
+### Use OptimismPortal in AnchorStateRegistry
+
+We considered updating the `AnchorStateRegistry` to ask the `OptimismPortal` if a particular game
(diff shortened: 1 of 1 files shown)


Inphi (commented):
The proposal looks good. I'd like to see a section that discusses the upgrade and migration path of the new design.

Since this proposal doesn't update the `respectedGameType`, we should be careful to not alter the validity of withdrawals between the portal+asr upgrade.
For example, games that were blacklisted prior to the proposed upgrade will have to be blacklisted again directly on the `AnchorStateRegistry`. Otherwise, unfinalized withdrawals using the blacklisted games will become valid.
We could reuse the `respectedGameTypeUpdatedAt` to fix this; by resetting the game type with the current implementation. This ensures that games prior to the upgrade are unsuitable for withdrawals. We will be technically changing the security property of the portal so this seems appropriate even though it's kinda hacky.

smartcontracts:
@Inphi I've added a section for this, let me know if that looks good to you!

ajsutton (commented):
Is it worth also considering making the `AnchorStateRegistry` not game specific? The anchor state is a block number and output root pair which is the same data for every game type and could easily be shared.  We'd just need to add the rule that the game we choose as the new anchor point needs to be the current respected game type.  That concept is probably harder to move out of the portal, though it wouldn't be bad to have all the code to decide if an output root is "trusted" in the AnchorStateRegistry and not spread across multiple contracts.

That would mean that we have a single source of the latest trusted anchor state and avoid starting from a really old anchor state again when we switch game types. I think @tynes has suggested making the idea of "is this a trusted root" more of a plugin approach in the portal which I think this is at least moving towards.

protocol/proofs/anchor-state-poison-protection.md
```diff
@@ -0,0 +1,121 @@
...
+air-gap window without being blacklisted.
+
+We can avoid making significant changes to the current challenger setup by using the call to the
+`tryUpdateAnchorState` function that the `FaultDisputeGame` already makes to both store the game
+result in the game array and to use a small binary search to update the anchor state. A binary
+search is likely to be acceptable here because we can bound the starting end of the search range to
+the index of the current anchor state and only search over states that finalized after the current
+anchor state was finalized. Although this can potentially miss a few eligible anchor states under
+certain circumstances, this is an acceptable trade-off as new anchor states *will* eventually be
+found after the current anchor state.
```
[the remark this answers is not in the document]
smartcontracts: Yeah I think I like this better because it's simpler. Worst case you have a bunch of games to process, but I think we can potentially mitigate this with a few tricks like:

- First check X number of games ahead based on average publish rate, then only iterate if the game X number of games ahead is invalid.
- Add a function that allows someone to explicitly poke to update the pointer.

smartcontracts:
> Is it worth also considering making the AnchorStateRegistry not game specific? 

I think this is actually very interesting. It fully brings the concept of "state validity" into the `AnchorStateRegistry` - let me meditate on this for a sec to see if there's anything weird with it.

protocol/proofs/anchor-state-poison-protection.md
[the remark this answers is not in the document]
ajsutton: I don't think I'd want to encode an expected rate of games into the contract as it could change at any time based on who happens to propose or more likely what the proposer config is.  But the typical case is that games resolve at a reasonably constant rate, so the normal case is actually that the game you promote to the new anchor state is the one immediately after the pointer. In rare cases there's some extra games published and so you want to skip a few games ahead but if we limit to checking 5 or 10 games we still cap the gas limit and it will catch up eventually.

A permissionless method that allows attempting to advance the pointer would also be useful so if it falls a long way behind we could manually poke it more frequently to get it to catch up.  Would be safe for that method to allow specifying a start for the search as well - worst case if you specify something too recent it just won't find anything to promote and will do nothing.

clabby (commented):
I'm a fan of this change. It's also a nice step towards decoupling the proof system from the `OptimismPortal`, which is pretty desirable.

re: 
> Is it worth also considering making the AnchorStateRegistry not game specific?

I think that with this design, we should. The reason why the ASR is game-specific right now is predominantly because of the potential that it could be poisoned. The way I look at it is that we already have the assumption that we will never let a bad withdrawal go through. If we make the anchor state promotion requirements equivalent to the requirements of a dispute game being used for a withdrawal, we effectively move anchor state updates into the same category. This effectively is us saying "we will always manually intervene before a bad withdrawal or bad anchor state update goes through," which is pretty reasonable. In that model, games relying on the same anchor-state model would be able to re-use the same block num / output root pairs np.

The main thing there would be around the respected game type, and figuring out a nice way to have the ASR only consider anchor state updates for game types that are currently respected. This would move even more out of the `OptimismPortal` (the current respected game type), but I think that's a good thing to work towards on the path towards fully abstracting over the proof system in the portal.

protocol/proofs/anchor-state-poison-protection.md
```diff
@@ -0,0 +1,121 @@
...
+As the system runs, the anchor state is continuously updated to reflect the latest finalized state
+as viewed by the proof system at any given time. `FaultDisputeGame` instances that resolve in favor
+of the defender will trigger the `AnchorStateRegistry` to update the anchor state to the newly
+finalized state.
+
+## Problem Statement
+
+The `AnchorStateRegistry` can become "poisoned" because the ultimate view of state validity
+currently actually lives inside of the `OptimismPortal` contract. An anchor state is considered
+"poisoned" if the `OptimismPortal` contract has blacklisted the dispute game that caused that
```
Ethnical, line 27: Newbie Questions about the current design: 

> The Ultimate view of the state validity lives into the OptimismPortal contract.

- [ ] does  the anchor can be poisoned if the game is blacklisted **before** resolved? Does the game will still `resolve()` and write into the `AnchorStateRegistry` if blacklisted? If yes, does there is a specific reason that allowing blacklisted game to resolve? More generally, It that possible to have a fonction callable by the guardian that is `BlacklistGameAndStopGame` exactly like the previous one but would stop a malicious game for being resolved. 

Edit: there is only 3 status for the current `FaultDisputeGame`

```solidity
enum GameStatus {
    // The game is currently in progress, and has not been resolved.
    IN_PROGRESS,
    // The game has concluded, and the `rootClaim` was challenged successfully.
    CHALLENGER_WINS,
    // The game has concluded, and the `rootClaim` could not be contested.
    DEFENDER_WINS
}
```

protocol/proofs/anchor-state-poison-protection.md
```diff
@@ -0,0 +1,121 @@
...
+We can avoid making significant changes to the current challenger setup by using the call to the
+`tryUpdateAnchorState` function that the `FaultDisputeGame` already makes to both store the game
+result in the game array and to use a small binary search to update the anchor state. A binary
+search is likely to be acceptable here because we can bound the starting end of the search range to
+the index of the current anchor state and only search over states that finalized after the current
+anchor state was finalized. Although this can potentially miss a few eligible anchor states under
+certain circumstances, this is an acceptable trade-off as new anchor states *will* eventually be
+found after the current anchor state.
+
+### Changes to the OptimismPortal
```
Ethnical, line 75: Do you think that If we are moving the logic into the contract `AnchorStateRegistry`, the functions `setRespectedGameType()`, and `blacklistDisputeGame()` of the `DeputyGuardianModule` will require to be upgraded?
Should be fine with the 4bytes as we can already call another contract but the documentation would likely be incorrect like here: 
```solidity
    /// @notice Calls the Security Council Safe's `execTransactionFromModuleReturnData()`, with the arguments
    ///      necessary to call `blacklistDisputeGame()` on the `OptimismPortal2` contract.
    ///      Only the deputy guardian can call this function.
    /// @param _portal The `OptimismPortal2` contract instance.
    /// @param _game The `IDisputeGame` contract instance.
    function blacklistDisputeGame(IOptimismPortal2 _portal, IDisputeGame _game) external {
        _onlyDeputyGuardian();

        bytes memory data = abi.encodeCall(IOptimismPortal2.blacklistDisputeGame, (_game));
        (bool success, bytes memory returnData) =
            SAFE.execTransactionFromModuleReturnData(address(_portal), 0, data, Enum.Operation.Call);
        if (!success) {
            revert ExecutionFailed(string(returnData));
        }
        emit DisputeGameBlacklisted(_game);
    }
```
If some real changes are necessaries, as the `DGM` is not a proxy we will need to redeploy the contract and regenerate PSPs from this new one.

protocol/proofs/anchor-state-poison-protection.md
```diff
@@ -0,0 +1,121 @@
...
+used for the anchor state as long as they've been blacklisted before the air-gap window starts.
+Since we do expect some games to be blacklisted *after* the air-gap window starts (but before it
+expires) this does not fully solve the problem.
+
+We can also make the changes to the `AnchorStateRegistry` described above *without* moving the
+blacklisting and air-gap logic out of the `OptimismPortal` contract. This would still fix the
+poisoning issue but would introduce additional complexity into the `AnchorStateRegistry` without
+reducing any complexity in the `OptimismPortal` contract. Would additionally spread responsibility
+for anchor state validity across multiple contracts, which isn't ideal. However, this would have
+the benefit of not needing to touch the `OptimismPortal` at all which reduces bug surface and risk.
```
Ethnical, line 109: > Since we do expect some games to be blacklisted *after* the air-gap window starts (but before it
expires) this does not fully solve the problem.

Just to make sure my understanding is correct. 
This will also apply in the second case more hybrid  case with the logic into the OptimismPortal2 but also some changes into the `AnchorStateRegistry`. 
So, here we won't be able to blacklist game after the air-gap window starts, without having the anchor possibly poisoned.

protocol/proofs/anchor-state-poison-protection.md
smartcontracts: So we generally want to create a distinction between the underlying games and a subjective view of validity on top of those games. This is why we allow dispute games to resolve, blacklisting is a subjective thing that the `OptimismPortal` does on top of games. Poisoning happens because the `OptimismPortal` blacklists a game but the `AnchorStateRegistry` is not aware of this and allows new games to be created on top of blacklisted games, which then leads to a situation where the `OptimismPortal` should logically also consider these new games to be blacklisted.

smartcontracts: What we want to do in this proposal is make it no longer possible for new games to be created from blacklisted states so that this implicit invalidity doesn't happen anymore.

protocol/proofs/anchor-state-poison-protection.md
smartcontracts: Yes, that's correct. We would need to upgrade the `DGM` and regenerate PSPs.

protocol/proofs/anchor-state-poison-protection.md
smartcontracts: Yeah, correct. I don't really think this hybrid thing is very useful. It adds more complexity for no good reason.

Ethnical: I see thanks!
Btw, completely agreed on the complexity opinion, less complex implicitly make it more safer for us.

protocol/proofs/anchor-state-poison-protection.md
```diff
@@ -0,0 +1,121 @@
...
+
+State validity rules like blacklisting and air-gap window inside of the `OptimismPortal` would be
+removed and outsourced to the `AnchorStateRegistry`. New public getter functions in the
+`AnchorStateRegistry` would provide the information that the `OptimismPortal` needs to determine
+whether a given dispute game contract can be used to finalize a withdrawal.
+
+### Changes to the FaultDisputeGame
+
+No changes to the FaultDisputeGame contract would be required.
+
```
Ethnical, line 85: As we will need to upgrade the `DGM` Should we add a part into the current doc to also informed about this breaking changes for future readers?

```suggestion
### Changes to the DeputyGuardianModule
The DeputyGuardianModule will have few light changes but this would require a total redeployment on L1. Thus, breaking the current PSPs and would need to regenerate them. 
```

smartcontracts: Good point, thank you.

protocol/proofs/anchor-state-poison-protection.md
```diff
@@ -0,0 +1,121 @@
...
+are currently handled by the `OptimismPortal` contract. The `AnchorStateRegistry` would expose
+several new getter functions that can be used by the `OptimismPortal` contract to determine if a
+given dispute game can be permitted to finalize a withdrawal.
+
+The `AnchorStateRegistry` would continue to keep track of the results of `FaultDisputeGame`
+contracts via the current "push" model where the `FaultDisputeGame` attempts to trigger an update
+in the `AnchorStateRegistry` on game finalization. However, instead of immediately updating the
+anchor state when a game is finalized, the `AnchorStateRegistry` would instead store the result in
+a log of game results. Game results would only be promoted to anchor states if they've passed the
+air-gap window without being blacklisted.
```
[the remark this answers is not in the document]
smartcontracts: Yes, no changes to air-gap period.

Ethnical (dismissed):
Approved, it would make more sense to me to go with the alternate approach of @ajsutton with permissionless method. If we go with this one I suggest we update the doc? 
Big thanks to @Inphi for the sync help on some points definitely need to that more often ;)

wildmolasses:
re: making the AnchorStateRegistry not game specific, will that be articulated in this design doc, or is it a separate consideration?

smartcontracts:
> re: making the AnchorStateRegistry not game specific, will that be articulated in this design doc, or is it a separate consideration?

Design doc has been updated to include this

ajsutton (approved):

wildmolasses (approved):
lgtm

PR #77, opened 2024-09-13
feat: OptimismPortal wd invalidation mitigation

smartcontracts:
Introduces a proposal that would prevent user withdrawals in the OptimismPortal from being invalidated any time that the fallback is utilized.

### DIFF
--- a/protocol/proofs/withdrawal-invalidation-mitigation.md
+++ b/protocol/proofs/withdrawal-invalidation-mitigation.md
@@ -25,7 +25,7 @@ against `DisputeGame` contracts that were created with a timestamp greater than 
 submitted prior to the activation of the fallback mechanism.
 
 Withdrawal invalidation is a powerful tool in the incident response playbook as it can mitigate the
-impact of a large number of invalid dispute games at the same time. However, the it also forces
+impact of a large number of invalid dispute games at the same time. However, it also forces
 users to resubmit their withdrawal proofs and wait an additional 7 days to execute a withdrawal.
 Users have reported that this invalidation and additional delay period is a pain point. This user
 impact means that the fallback is highly disruptive as a standard security mechanism.
--- a/protocol/proofs/withdrawal-invalidation-mitigation.md
+++ b/protocol/proofs/withdrawal-invalidation-mitigation.md
@@ -1,0 +1,83 @@
+# Withdrawal Proof Invalidation Mitigation
+
+## Context
+
+The `OptimismPortal` contract allows users to prove withdrawals by demonstrating that those
+withdrawals exist inside of the state root of a `FaultDisputeGame` instance. Withdrawals can be
+finalized if the `FaultDisputeGame` instance resolves in favor of the defender, the instance is not
+blacklisted, and the game has sat around for the air-gap period.
+
+The `DisputeGameFactory` can create different implementations of the `DisputeGame` which typically
+have different code and are identified by an integer "game type". The `OptimismPortal` currently
+defines a variable `respectedGameType` which determines the particular `DisputeGame` code that
+users can utilize when proving and finalizing withdrawals. The "fallback" safety net action is the
+ability for the Guardian or Deputy Guardian address to change the `respectedGameType` variable to
+some alternative game type (e.g., the `PermissionedDisputeGame`) in the case that the
+`FaultDisputeGame` contract is found to be buggy.
+
+## Problem Statement
+
+The fallback action of changing the `respectedGameType` variable updates another variable called
+`respectedGameTypeUpdatedAt` that keeps track of exactly when the respected game type was changed.
+Currently, the `OptimismPortal` contract enforces that withdrawals can only be proven or finalized
+against `DisputeGame` contracts that were created with a timestamp greater than the
+`respectedGameTypeUpdatedAt` variable. This effectively invalidates any withdrawal proofs that were
+submitted prior to the activation of the fallback mechanism.
+
+Withdrawal invalidation is a powerful tool in the incident response playbook as it can mitigate the
+impact of a large number of invalid dispute games at the same time. However, the it also forces
+users to resubmit their withdrawal proofs and wait an additional 7 days to execute a withdrawal.
+Users have reported that this invalidation and additional delay period is a pain point. This user
+impact means that the fallback is highly disruptive as a standard security mechanism.
+
+We are in the process of further assessing the impact and severity of this pain point but are
+creating a design document in the meanwhile to determine the lift to resolve this issue.
+
+## Proposed Solution
+
+We propose solving this problem by making withdrawal invalidation an optional part of the fallback
+mechanism. We can achieve this by slightly modifying the rules for withdrawal verification.
+
+We currently enforce the following set of rules (not including other rules like blacklist checks):
+
+- Withdrawals must be proven against a dispute game of the `respectedGameType`.
+- Withdrawals must be finalized against a dispute game of the `respectedGameType`.
+- Withdrawals must be finalized against a dispute game created after `respectedGameTypeUpdatedAt`.
+
+We propose removing the requirement that withdrawals must be finalized against a dispute game of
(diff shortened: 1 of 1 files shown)


ajsutton (commented):
I'm in favour of this change. The validity checks are still really simple and it gives the guardian significantly more flexibility.  The concern about user impact affects the guardian decision making process even if the actual impact winds up being small so providing better tools to simplify the decision making process seems well worth it to me.

protocol/proofs/withdrawal-invalidation-mitigation.md
```diff
@@ -0,0 +1,83 @@
...
+
+We currently enforce the following set of rules (not including other rules like blacklist checks):
+
+- Withdrawals must be proven against a dispute game of the `respectedGameType`.
+- Withdrawals must be finalized against a dispute game of the `respectedGameType`.
+- Withdrawals must be finalized against a dispute game created after `respectedGameTypeUpdatedAt`.
+
+We propose removing the requirement that withdrawals must be finalized against a dispute game of
+the `respectedGameType`. Withdrawals that have been previously proven against a dispute game when
+that game *was* respected can be finalized even if that dispute game is no longer of the respected
```
wildmolasses, line 49: Looking at the impl, although `checkWithdrawal` no longer asserts against `respectedGameType`,  we continue to assert against it in `proveWithdrawalTransaction`, implying the withdrawal `GameType` was once the respected one. If a `GameType` needs to be invalidated, the `respectedGameType` is changed **and** `respectedGameTypeUpdatedAt` is updated. This means another `DisputeGame` must be created, of the new `respectedGameType` and not the invalid `GameType`. 👍

protocol/proofs/withdrawal-invalidation-mitigation.md
```diff
@@ -0,0 +1,83 @@
...
+could never actually be used to finalize a withdrawal. Although this is not strictly necessary, it
+blocks off a footgun and we typically like to block off footguns where possible.
+
+### Naming
+
+We may want to consider renaming `respectedGameTypeUpdatedAt` to something more representative of
+the new system given that the variable would not be updating every time that the
+`respectedGameType` variable is changed. Alternatively, we could keep `respectedGameTypeUpdatedAt`
+untouched and instead introduce a new variable that more accurately reflects the system. This would
+also avoid a breaking change to the contract ABI and the semantics of the original variable.
```
wildmolasses, line 72: this variable could be decoupled from `respectedGameType`. Instead it could be called something like `gameCreationValidFrom`, and then the check would be like

```solidity
require(
    createdAt >= gameCreationValidFrom,
    "OptimismPortal: dispute game created before latest creation validity timestamp"
);
```

what do you think?

the occurrences of `respectedGameTypeUpdatedAt` in the monorepo are low. sorry for the noob question but are there many concerns outside of the monorepo? perhaps integrators would be happy for the breaking change here as their withdrawal validity logic should be updated.

smartcontracts: Realistically I think you're correct here -- it's possible we could even keep the semantic meaning of `respectedGameTypeUpdatedAt` (as in, we do continue to update it when `setRespectedGameType` is called) but add this new variable `gameCreationValidFrom` with these new semantics.

clabby (commented):
This looks good to me. Simple enough to change up. Also will include some changes to the deputy guardian.

smartcontracts:
Reminder to self: do not allow invalidation timestamp to be greater than current timestamp

protocol/proofs/withdrawal-invalidation-mitigation.md
```diff
@@ -0,0 +1,83 @@
...
+We propose removing the requirement that withdrawals must be finalized against a dispute game of
+the `respectedGameType`. Withdrawals that have been previously proven against a dispute game when
+that game *was* respected can be finalized even if that dispute game is no longer of the respected
+game type.
+
+We additionally propose that the `respectedGameTypeUpdatedAt` timestamp not be updated by default.
+Instead, the timestamp can be selectively updated if withdrawal invalidation is actually required.
+Combined with the first proposed change, this proposal means that older withdrawal proofs can
+usually be finalized when the respected game type is changed but withdrawals can still be blanket
+invalidated if necessary.
```
Ethnical, line 56: An issue can possibly arise if the second case for the withdrawal invalidation without checking the `respectGameType` like in the past.

For example, this might be a possible attack vector:

1. The guardian make a tx to set the `respectedGameType` to 1 (permissioned game).
2. The attacker monitors the mempool and frontruns **in the same block** the call to `setRespectGame(1)`.
3. In the same block, the attacker create a game using `create()` on the Fault Dispute Factory contract. Because the `createAt` is a timestamp from (`block.timestamp`).
4. The attacker call `proveWithdrawalTransaction()` and then wait for the `PROOF_MATURITY_DELAY_SECONDS` submit the `finalizeWithdrawalTransaction()`.
5. As the check (below in solidity) is using ` createdAt >= respectedGameTypeUpdatedAt` the attacker might be able to finalize a transaction with the incorrect `gameType`.
6. The impact seems to be really low, but this would make an inconsistent state (as the `optimismPortal` and the operator will think that only certain gameType are valid during the withdrawal process).

```solidity

if (disputeGameProxy.gameType().raw() != respectedGameType.raw()) //@audit: The current design review propose to not use this anymore?
    revert InvalidGameType();

require(
    createdAt >= respectedGameTypeUpdatedAt, //@audit: Usage of the >= instead of > is a potential issue here?
    "OptimismPortal: dispute game created before respected game type was updated"
);

```

This scenario might be invalid, don't hesitate to let me know about it.

Ethnical (approved):
Agreed with this let some comment but the rest looks cool to me.

protocol/proofs/withdrawal-invalidation-mitigation.md
ajsutton: Yeah would be worth making that a strict `>`.  While that will mean that any withdrawals proven against the new game type after the change but in the same block are invalidated when they probably shouldn't be, that's a very unlikely corner case and much lower impact than having a withdrawal be unexpectedly valid.

smartcontracts: Let’s make sure that we account for this in the spec.

smartcontracts: > While that will mean that any withdrawals proven against the new game type after the change but in the same block are invalidated when they probably shouldn't be, that's a very unlikely corner case and much lower impact than having a withdrawal be unexpectedly valid.

We can get around this by reverting if someone tries to prove a withdrawal within the same block. Better to have an edge-case revert where you get immediate feedback than an edge-case invalid withdrawal where you only get feedback later.

ajsutton (approved):

wildmolasses (approved):
lgtm; looks like this doc defers on variable naming, i'll assume it comes during spec/impl phase :+1:

anikaraghu:
@smartcontracts a bit late to the game here, but wondering if there will be a way to have presigned fallbacks to the permissioned game that **do** invalidate all existing games (for incident response scenario). was chatting with @elihaims about this today and this could be really helpful to prevent a scenario in which we need to pause the whole bridge to remedy a widespread issue

smartcontracts:
> @smartcontracts a bit late to the game here, but wondering if there will be a way to have presigned fallbacks to the permissioned game that **do** invalidate all existing games (for incident response scenario). was chatting with @elihaims about this today and this could be really helpful to prevent a scenario in which we need to pause the whole bridge to remedy a widespread issue

Fallback can be triggered by the `Guardian` so it should be possible to have pre-signed fallbacks that do invalidate existing games. Can account for this in the spec with a function that triggers the fallback + sets the invalidation timestamp to the current timestamp (this way you don't need to know the timestamp that you'll be using ahead of time).

Worth noting that you would need to have pre-signed fallbacks that have the same nonces as pre-signed pauses, but that's fine imo.

PR #79, opened 2024-09-16
proofs: Avoid Cyclic Dependencies When Deploying New Chains

ajsutton:
**Description**

Explains the problem of cyclic dependencies encountered when deploying new chains using fault proofs and explores potential options.


protocol/proofs/new-chains.md
```diff
@@ -0,0 +1,70 @@
...
+* Chains that are not supporting permissionless games yet could deploy only the permissioned game, preventing creating
+  permissionless games entirely.
+
+### Cons
+
+* The `ProxyAdminOwner` is required to perform the `setImplementation` calls because ownership of the L1 contracts has
+  already been transferred. In the current system, the L1 contract deployment can be performed by anyone and the system
+  is fully configured (albeit incorrectly for proofs) before ownership is transferred.
+* It remains unclear at what point a chain can be added to the superchain-registry or how a version of `op-program` that
+  supports the new chain would be created.
```
ajsutton, line 65: @geoknee This is why you got tagged in as a reviewer. Things are simpler for `op-program` if the config is always coming from the superchain-registry but I suspect there's a bunch of checks in the superchain-registry that could be affected by not yet having the dispute game implementations deployed and in general it seems awkward to have adding the chain to the superchain registry on the critical path for deploying a new chain.

We have similar weirdness with op-challenger, op-node, op-geth etc which pull config from the superchain-registry as new chains have to configure everything the hard way by specifying the rollup config/genesis manually and then later switch to using `--network` once the chain is in the superchain-registry which introduces a lot of potential for errors and winding up with mismatched configs. But at least for those it's doable.

protocol/proofs/new-chains.md
```diff
@@ -0,0 +1,70 @@
...
+
+Finally, the `ProxyAdminOwner` would deploy the `FaultDisputeGame` and dependent contracts, providing the correct
+`faultGameGenesisOutputRoot` and `faultGameAbsolutePrestate` values and call `setImplementation` on the
+`DisputeGameFactory`. Withdrawals would then be possible.
+
+### Pros
+
+* Neatly avoids the cyclic dependency.
+* Chains that are not supporting permissionless games yet could deploy only the permissioned game, preventing creating
+  permissionless games entirely.
```
mds1, line 57: Can't we already do this? e.g. do you need the `faultGameGenesisOutputRoot` and `faultGameAbsolutePrestate` for permissioned games, given that games won't need to be played?

protocol/proofs/new-chains.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Cons
+
+* The `ProxyAdminOwner` is required to perform the `setImplementation` calls because ownership of the L1 contracts has
+  already been transferred. In the current system, the L1 contract deployment can be performed by anyone and the system
+  is fully configured (albeit incorrectly for proofs) before ownership is transferred.
```
mds1, line 63: This is a big downside, because we want the ability to deploy chains with the ProxyAdminOwner configured as the 2/2, meaning the Security Council needs to send a transaction for every new chain

Perhaps you could spin this as a positive, as it means new chains cannot permissionlessly give their keys to the SC, and this lets the `setImplementation` call as a formal way to signify acceptance of governing the chain. (Though they still can permissionlessly give their keys to SC by deploying themselves to v1.3.0 (MCP), then upgrading to FPs))

protocol/proofs/new-chains.md
```diff
@@ -0,0 +1,70 @@
...
+   to be used as part of permissioned and permissionless dispute games. `op-program` requires the L2 chain
+   configuration (rollup config and L2 genesis state) to be available in the version of the superchain-registry that it
+   includes to operate correctly. Similar to above, the L2 chain configuration is unknowable prior to deploying the L1
+   contracts. Additionally, chains cannot be added to the superchain-registry until _after_ they have launched.
+
+# Proposed Solution
+
+TBD
+
+# Alternatives Considered
```
mds1, line 32: Trying to brainstorm other ideas. I don't know enough about the implementation details to know if these are feasible, but:
1. Would an MCP L2 solve the problem? In an MCP L2 world, all code at all address in genesis becomes identical, but storage will still vary because of chain-specific L1 addresses that live in predeploy storage. Because of the storage differences at genesis I suspect this solution will not work.
    1. What if the genesis was 100% identical? This would be possible if storage values were pushed from L1 to L2 in the first block via an L1 attributes transaction, though I know @tynes wants to minimize the amount of data we pass that way, and this increases it.
3. What about something like: "If `faultGameGenesisOutputRoot` and `faultGameAbsolutePrestate` are both zero, ask the node for the ouput root and compute the prestate". In other words, get these values on-demand the first time a game is played

protocol/proofs/new-chains.md
ajsutton: Yes you need to specify those values for the permissioned game as well. In theory the permissioned game could be played down to the point where they are used, though in practice that would mean the CHALLENGER role is disagreeing with themselves which is weird.

I don't believe our deployment scripts have an option to not deploy the permissionless game though.

protocol/proofs/new-chains.md
ajsutton: I don't think it changes anything about being able to permissionlessly assign. You could deploy the whole system assigning ownership to yourself, run it for a few years and then change the owner to the SC without permission as I understand it.

A possible work around for this is to not reassign ownership in the first transaction that deploys. So you deploy the L1 contracts except the dispute game implementations, retaining ownership, generate the L2 genesis, deploy the game implementations and then assign the final owner. Not having atomic assignment of ownership does open up some possible security implications (if the deployer's key is compromised or if the deployer messes with things before assigning) but the deployer could just mess with the deployment script now.  With OPSM it would be different but OPSM could keep ownership for itself and be able to finish off the deployment in a trusted way I think (though not sure how it would verify the L2 genesis root....)

protocol/proofs/new-chains.md
ajsutton: 1. If the L2 genesis was always identical I think that would help. The superchain-registry part of this I think would still be an issue but that would be more administrative without having a fundamental cyclic dependency. Not sure if we can make the L2 genesis identical for all new chains though - seems like quite a challenge.
2. The game is created on the L1 chain so it can't reach out to a node for data. `faultGameGenesisOutputRoot` and `faultGameAbsolutePrestate` are the key sources of truth the fault system is rooted in and they aren't available on-chain otherwise.
3. I suspect this gets a lot worse post-interop. You'd need the L2 genesis of all chains in the dependency set. And adding a new chain into the dependency set would be an interesting challenge too - as would migrating an existing chain to be part of the shared DisputeGameFactory (and migrating it out again if needed). There's quite a lot of open questions there to be honest.

protocol/proofs/new-chains.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Defer Deployment of Dispute Game Types
+
+Remove the `FaultDisputeGame` and its dependent contracts (including `AnchorStateRegistry`) from the initial L1
+deployment. The `DisputeGameFactory` would still be deployment, but with no supported game types. It would be impossible
```
geoknee, line 37:
```suggestion
deployment. The `DisputeGameFactory` would still be deployed, but with no supported game types. It would be impossible
```

protocol/proofs/new-chains.md
geoknee: The registry does support adding a chain as a "standard candidate" chain, which will later become a fully standard chain (in fact this is the only way to add a chain at present). This means that when a chain is added, not all of the standard checks run on it (yet). So adding the chain to the registry is likely not going to be blocked on having the chain fully deployed and set up. Only "promoting the chain to standard" would be blocked, this is just a small PR bumping the superchain_level and potentially adding a couple of contract addresses.

I agree it is awkward to have the dependency the other way around (blocking chain deployment on adding the chain to the registry). OTOH maybe it is reasonable, for chains opting into a shared governance system with coordinated hardforks and so on? If we leaned into this, it might help i) adoption of the `--network` flags (these are not widely used) and ii) reduce the chance for mismatched configs in the end.

protocol/proofs/new-chains.md
mds1: > If the L2 genesis was always identical I think that would help. The superchain-registry part of this I think would still be an issue but that would be more administrative without having a fundamental cyclic dependency. Not sure if we can make the L2 genesis identical for all new chains though - seems like quite a challenge.

We are planning to have identical L2 genesis as part of Isthmus in Cycle 33, dates [here](https://discord.com/channels/1244729134312198194/1290809960732692534/1291317994177433673) (https://github.com/ethereum-optimism/specs/discussions/373#discussioncomment-10814320). This would remove the fundamental cyclic dependency. 

Do we have a need for deploying chains straight to permissionless before then? If not, we can probably close this design doc, and work out the superchain registry sequencing separately

ajsutton: Yeah I think that solves the fundamental issue.  I'm happy to close this in the meantime then. Honestly I just miss having a place I could put a doc for discussion rather than having to put it through an approval process. This PR will do for that. :)

ajsutton:
Going to close this for now. There's a separate proposal and work being done on having a standard L2 genesis which fixes the fundamental dependency cycle. There's still a question of how to fit the superchain-registry update into the process of launching a new chain but we can revisit that once we've got a standard genesis.

PR #80, opened 2024-09-16
feat: improve DelayedWETH response capabilities

smartcontracts:
Introduces a proposal to improve some of the incident response capabilities present in the DelayedWETH contract.

### DIFF
--- a/protocol/proofs/delayed-weth-expansion.md
+++ b/protocol/proofs/delayed-weth-expansion.md
@@ -1,0 +1,78 @@
+# Expanded DelayedWETH Capabilities
+
+## Context
+
+The `DelayedWETH` contract acts as a holding contract for the bonded ETH submitted by any
+participant in a `DisputeGame` contract. As of the Granite upgrade, each implementation of the
+`DisputeGame` uses its own `DelayedWETH` contract. Each `DelayedWETH` contract is `Ownable` and is
+subject to a number of safety net actions by the owner address (holding balances from specific
+addresses or removing all ETH held in the `DelayedWETH` contract). `DelayedWETH` is additionally
+subject to the Superchain-wide pause mechanism within the `SuperchainConfig`. Bonds cannot be
+retrieved by game participants if the Superchain pause is active.
+
+## Problem Statement
+
+Existing security mechanisms within the `DelayedWETH` contract are inconvenient and slow down the
+incident response process. Specifically:
+
+- The `DelayedWETH.hold(address target)` function creates an approval from the `target` address to
+  the `owner` address. Since the `target` can simply remove this approval, `hold` MUST be triggered
+  alongside a `transferFrom` in the same transaction. This adds unnecessary complexity to something
+  that could simply be a single transaction.
+- The various security mechanisms inside of `DelayedWETH` can only be triggered by the System Owner
+  which has a slow SLA. The Deputy Guardian can act to trigger the Superchain-wide pause, but this
+  is highly disruptive to the entire Superchain ecosystem.
+
+Additionally, it should be noted that all ETH inside a `DelayedWETH` contract is pooled together
+regardless of which games are using the contract. We currently mitigate this by using one
+`DelayedWETH` contract per game type, but this expands the number of contracts that the System
+Owner and Guardian roles must manage by at least 2 per additional chain. However, the modifications
+required to appropriately address this issue are sufficiently extensive that they are considered
+out of scope for this proposal.
+
+## Proposed Solution
+
+Here we propose some modifications to `DelayedWETH` that help to resolve the two noted limitations:
+
+1. We propose that `DelayedWETH.hold` executes a `transferFrom` automatically to reduce the number
+  of required transactions from 2 to 1 when holding ETH from a single dispute game. `hold` will
+  specifically transfer to the `owner` address.
+2. We propose a new `pause` function on the `DelayedWETH` contract itself that pauses *just* the
+  `DelayedWETH` contract and supplements the Superchain-wide pause functionality. This `pause`
+  function would be executable by the Guardian role and could therefore be triggered quickly in
+  case the System Owner is unable to execute another security action in time (e.g., holding bonds
+  from a particular dispute game).
+
+### Updated Pause Behavior
+
+We propose slightly modifying the behavior of the `pause` functionality for the `DelayedWETH`
+contract as follows:
+
+- As previously stated, the `Guardian` role will also be able to trigger the `pause` functionality
+  for a particular `DelayedWETH` contract. The `DelayedWETH` contract will be in a paused state if
+  the `Guardian` triggers the pause OR if the Superchain-wide pause is triggered.
+- When the `pause` is active, users will not be able to trigger the `withdraw` function (same as
+  before) EXCEPT the `owner` will still be able to trigger `withdraw` (may be useful in certain
+  situations).
+- When the `pause` is active, users other than the `owner` will not be able to transfer any tokens
+  within the contract.
+
+## Alternatives Considered
+
+### Wider Rewrite
+
+We considered the alternative to rewrite the `DelayedWETH` contract more broadly to be able to
+easily manage multiple game types across multiple chains within a single `DelayedWETH` instance.
+However, this would require significant modifications to the `DelayedWETH` contract and would act
+as an iterative improvement to the status quo. We still believe that a larger rewrite should be
+considered for the near future.
+
+## Risks & Uncertainties
+
+### Security Model
+
+We see little risk with the modification to the `hold` function. Introducing a new `pause` function
+slightly changes the existing security model (though not significantly). The Guardian role can
+already pause the `DelayedWETH` contract by triggering the Superchain-wide pause. Allowing this
+role to pause the `DelayedWETH` contract without pausing the rest of the system should not open up
+capabilities that the Guardian does not already have. Comments on this subject are welcome.


ajsutton (dismissed):
This sounds good to me. If we operating DelayedWETH is easier I think we'd be fine to go back to having a single DelayedWETH per chain. Sharing across multiple chains is another level again that we'd probably want the ability to pause per chain still and I'm not too sure how to do that.

protocol/proofs/delayed-weth-expansion.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Risks & Uncertainties
+
+### Security Model
+
+We see little risk with the modification to the `hold` function. Introducing a new `pause` function
+slightly changes the existing security model (though not significantly). The Guardian role can
+already pause the `DelayedWETH` contract by triggering the Superchain-wide pause. Allowing this
+role to pause the `DelayedWETH` contract without pausing the rest of the system should not open up
+capabilities that the Guardian does not already have. Comments on this subject are welcome.
```
mds1, line 62: Regarding the `DelayedWETH.pause` function, this would also require a change to the DeputyGuardianModule (DGM), this way both the Guardian and Deputy Guardian can call `DelayedWETH.pause`. Changing the DGM invalidates all presigned pauses. The Granite changes also changed the DGM and therefore invalidated the pauses, so they were resigned.

Since it seems the DGM changes more frequently than we thought, we should consider proxying as part of this scope, to avoid having to resign presigned pauses with each DGM change. Assuming proxying it is required additional work, does that change the decision on this vs. the alternative rewrite? I'm unsure how large the scope of "Wider Rewrite" is

smartcontracts: Ah, interesting. Good point. DGM starts to get increasingly bloated, especially if there are a bunch of different `DelayedWETH` contracts all over the place. I still think proxying the DGM would be much easier than rewriting `DelayedWETH` for now. I do think rewriting `DelayedWETH` is a good idea but the goal is to be able to ship these improvements within a few months and a rewrite would take a lot longer.

clabby: +1 on proxying the DGM. Having to update presigned pauses each time we change it will be a bit painful.

clabby (commented):
This looks good in general, not against these changes.

But at the same time, this still doesn't address the issue with "poisoned state" I don't think. There's still a decent risk of state corruption upon recovery here. The `recover` function is still in a really weird spot, as it just transfers ETH straight out of the contract without actually subtracting balance or allowance.

We should figure out a way to ensure that when funds are recovered, the contract's state isn't entirely corrupted imo (it sounds like just removing `recover` and using the `hold` + `transferFrom` route is the right way to go.)

Another pretty annoying thing about `DelayedWETH` is the manual reimbursements by the owner if funds are rescued. This used to be a lot easier to reason about prior to `DelayedWETH`, because the ETH was stored within the FDG, and we could make decisions about reimbursements on-chain based off of the state within the FDG. Not sure if we want to solve that here, but the "recover -> manually send ETH out to accounts" path feels like it would really suck in a high-stress scenario.

smartcontracts:
Yeah I fully agree that this response mechanism has the potential to be a huge nuisance. I think these changes are strictly improvements over the status quo but I would like to explore a more automated system that minimizes the amount of effort here.

protocol/proofs/delayed-weth-expansion.md
```diff
@@ -0,0 +1,62 @@
+# Expanded DelayedWETH Capabilities
+
+## Context
+
+The `DelayedWETH` contract acts as a holding contract for the bonded ETH submitted by any
+participant in a `DisputeGame` contract. As of the Granite upgrade, each implementation of the
+`DisputeGame` uses its own `DelayedWETH` contract. Each `DelayedWETH` contract is `Ownable` and is
+subject to a number of safety net actions by the owner address (holding balances from specific
```
Ethnical, line 8: > As of the Granite upgrade, each implementation of the
`DisputeGame` uses its own `DelayedWETH` contract.

I checked briefly onchain but seems to have only one `DelayedWETH` used currently -> https://etherscan.io/address/0x82511d494B5C942BE57498a70Fdd7184Ee33B975 
Does this means that since Granite each time we `create()` a `DisputeGame` contract,  we also create a `DelayedWETH` contract for the `DisputeGame` instance?

protocol/proofs/delayed-weth-expansion.md
Ethnical: Clearly Agreed about with the fact we need to proxify the DGM!  
I would also suggest to inform the reader that this would require to deploy a new DGM.  
```Markdown
### Changes to the DeputyGuardianModule
The `DeputyGuardianModule` will have few light changes but this would require a total redeployment on L1. Thus, breaking the current PSPs and would need to regenerate new PSPs from the new `DeputyGuardianModule`.
```

protocol/proofs/delayed-weth-expansion.md
```diff
@@ -0,0 +1,62 @@
...
+easily manage multiple game types across multiple chains within a single `DelayedWETH` instance.
+However, this would require significant modifications to the `DelayedWETH` contract and would act
+as an iterative improvement to the status quo. We still believe that a larger rewrite should be
+considered for the near future.
+
+## Risks & Uncertainties
+
+### Security Model
+
+We see little risk with the modification to the `hold` function. Introducing a new `pause` function
```
Ethnical, line 58: Small nits: I don't know if the `pause` function name is not about to be confusing especially inside the DGM, as we already have the wider `pause()`/`unpause()`? 
Just small thoughts about the naming convention.

Ethnical (dismissed):
Overall, really nice improvement to avoid to widely pause the Superchain. In case a of a targeted action into `DelayedWETH` contract.
I have let comments/suggestions but looks promising to me.
As some people already suggested the `recover()` can lead to a brick of the contract `DelayedWETH` as the `balanceOf[]` is not updated when the ETH are transfered, but seems to out of scope for this current design.  

Also the function `recover()` is sending native ETH to the `msg.sender` that would be the `owner()` the [proxyAdminOwner](https://etherscan.io/address/0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A), I don't know if in the future we would like to control the receiver as sometimes safe like [proxyAdminOwner](https://etherscan.io/address/0x5a0Aae59D09fccBdDb6C6CcEB07B7279367C3d2A) are tidious to move the funds to send them to a Multisig dedicated for this effect.

protocol/proofs/delayed-weth-expansion.md
smartcontracts: We have two `DelayedWETH` contracts ever since Granite. Each dispute game type has its own `DelayedWETH` contract. So every game of the same game type uses the same `DelayedWETH` contract.

protocol/proofs/delayed-weth-expansion.md
smartcontracts: Yeah, could consider that. I will leave that for the spec.

protocol/proofs/delayed-weth-expansion.md
Ethnical: Got it make more sense to me here!

protocol/proofs/delayed-weth-expansion.md
```diff
@@ -0,0 +1,62 @@
...
+  is highly disruptive to the entire Superchain ecosystem.
+
+Additionally, it should be noted that all ETH inside a `DelayedWETH` contract is pooled together
+regardless of which games are using the contract. We currently mitigate this by using one
+`DelayedWETH` contract per game type, but this expands the number of contracts that the System
+Owner and Guardian roles must manage by at least 2 per additional chain. However, the modifications
+required to appropriately address this issue are sufficiently extensive that they are considered
+out of scope for this proposal.
+
+## Propsed Solution
```
Ethnical, line 33:
```suggestion
## Proposed Solution
```

protocol/proofs/delayed-weth-expansion.md
```diff
@@ -0,0 +1,62 @@
...
+
+Here we propose some modifications to `DelayedWETH` that help to resolve the two noted limitations:
+
+1. We propose that `DelayedWETH.hold` executes a `transferFrom` automatically to reduce the number
```
wildmolasses, line 37: what should the destination of the `transferFrom` be?

wildmolasses: answer from @smartcontracts: destination is the owner

protocol/proofs/delayed-weth-expansion.md
```diff
@@ -0,0 +1,62 @@
...
+  of required transactions from 2 to 1 when holding ETH from a single dispute game.
+2. We propose a new `pause` function on the `DelayedWETH` contract itself that pauses *just* the
+  `DelayedWETH` contract and supplements the Superchain-wide pause functionality. This `pause`
+  function would be executable by the Guardian role and could therefore be triggered quickly in
+  case the System Owner is unable to execute another security action in time.
```
mds1, line 42: 1. When DelayedWETH is paused, what specifically is paused? I assume all state changing functions except `unpause`, but it might be a smaller set of functions?
2. What "other security action" would the DelayedWETH owner be expected to execute first? The intent of this question is just to make sure the incident response runbook is fleshed out

protocol/proofs/delayed-weth-expansion.md
```diff
@@ -0,0 +1,62 @@
...
+as an iterative improvement to the status quo. We still believe that a larger rewrite should be
+considered for the near future.
+
+## Risks & Uncertainties
+
+### Security Model
+
+We see little risk with the modification to the `hold` function. Introducing a new `pause` function
+slightly changes the existing security model (though not significantly). The Guardian role can
+already pause the `DelayedWETH` contract by triggering the Superchain-wide pause. Allowing this
```
mds1, line 60: > The Guardian role can already pause the `DelayedWETH` contract by triggering the Superchain-wide pause

What do you mean by this? The Superchain-wide pause only pauses all withdrawals (ETH, ERC20, ERC721, and arbitrary calls) from the bridge contracts, but it does not affect the DelayedWETH contract. So I'm assuming you mean that it effectively indirectly pauses DelayedWETH because a withdrawal cannot be executed until unpaused? (I am not too familiar with the current DelayedWETH mechanics and usage)

dajuguan: The Superchain-wide pause affects DelayedWETH contract too, see here https://github.com/ethereum-optimism/optimism/blob/6f41bac618826b7669ba584876586fb17ac5e9b0/packages/contracts-bedrock/src/dispute/DelayedWETH.sol#L92

protocol/proofs/delayed-weth-expansion.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Propsed Solution
+
+Here we propose some modifications to `DelayedWETH` that help to resolve the two noted limitations:
+
+1. We propose that `DelayedWETH.hold` executes a `transferFrom` automatically to reduce the number
+  of required transactions from 2 to 1 when holding ETH from a single dispute game.
+2. We propose a new `pause` function on the `DelayedWETH` contract itself that pauses *just* the
+  `DelayedWETH` contract and supplements the Superchain-wide pause functionality. This `pause`
+  function would be executable by the Guardian role and could therefore be triggered quickly in
```
Ethnical, line 41: > function would be executable by the Guardian role and could therefore be triggered quickly. 
We can also add a new method into the DGM that can call this (I don't know if this can impact the stage of l2beat).

protocol/proofs/delayed-weth-expansion.md
smartcontracts: Updated this to answer your questions.

ajsutton (approved):

wildmolasses (approved):
lgtm

protocol/proofs/delayed-weth-expansion.md
```diff
@@ -0,0 +1,78 @@
...
+Here we propose some modifications to `DelayedWETH` that help to resolve the two noted limitations:
+
+1. We propose that `DelayedWETH.hold` executes a `transferFrom` automatically to reduce the number
+  of required transactions from 2 to 1 when holding ETH from a single dispute game. `hold` will
+  specifically transfer to the `owner` address.
+2. We propose a new `pause` function on the `DelayedWETH` contract itself that pauses *just* the
+  `DelayedWETH` contract and supplements the Superchain-wide pause functionality. This `pause`
+  function would be executable by the Guardian role and could therefore be triggered quickly in
+  case the System Owner is unable to execute another security action in time (e.g., holding bonds
+  from a particular dispute game).
```
mds1, line 44: What is the use case for pausing DelayedWETH, e.g. what situation do we expect to do it? I ask to confirm if placing the `pause` at the DelayedWETH level what we want—the implication here is that if we need to pause DelayedWETH across the full superchain, it's `n` calls and we must ensure we don't miss a chain. Is the expected scenario that we only need to pause a single DelayedWETH for a single chain?

smartcontracts: The main reason to pause `DelayedWETH` for a single chain is because our only other option is to pause the entire superchain and doing so is highly disruptive. We want the option to do both. If it's just a single game type being exploited then we want a way to quickly pause withdrawals out of `DelayedWETH` (so that we can potentially recover bonds if needed) without pausing the entire superchain.

ajsutton: I'd also say that if we're in the situation where we need to pause DelayedWETH across every chain, there's a pretty good chance things are really messed up and we would just pause everything.  But much more likely is that there's a bunch of games on one chain that have resolved incorrectly (e.g. because op-challenger for that chain was mis configured) and we just need more time to sort that out.  We could invalidate all games on that chain with `setRespectedGameType` (with the option to invalidate since that will be optional in the future) and pause `DelayedWETH` to give us more time to work through how the bonds should be paid out.

mds1: Thank you both! One follow up:

> We could invalidate all games on that chain with setRespectedGameType 

On this point, do we have a plan / is it desirable to make the respected game type part of the superchain config (or some other analogous solution that makes all chains follow the same game type)? It’s of scope for this design doc, but it seems eventually we’d want to enforce the same game being used for all interop chains (and maybe all standard chains too) to ensure consistent security properties

smartcontracts: I do think this probably makes sense but I suspect it would probably come with a follow-up set of changes that allows multiple game types to be accepted. And possibly for the same contracts to be used everywhere with an anchor state keyed by chain ID.

mds1: Makes sense. I do want to make sure this is tracked somewhere, and will defer you on where to track

mds1 (approved):

PR #83, opened 2024-09-18
interop: added superc20Redesign design doc

0xParticle:
**Description**

We introduce a design doc to discuss a design change in the `SuperchainERC20` standard, as described in the [interop specs](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/token-bridging.md). The goal of this PR is to open the discussion and ensure we are on the best possible path.

### DIFF
--- a/protocol/superchainerc20/superc20Redesign.md
+++ b/protocol/superchainerc20/superc20Redesign.md
@@ -86,6 +86,7 @@ It’s important to notice that, even though this functions get implemented in t
     - If bridging functions are implemented on the ERC20, updates to the bridge might require updates to every ERC20. This might be easy to do for implementations that use a BeaconProxy structure, but not in other cases.
 - **Easier for integrators:** integrators must deal with a single entrypoint instead of many. This would improve their devX considerably.
 - **Improved indexing and monitoring:** only monitoring events and traces coming from a single source would be necessary instead of many. Improved monitoring correlates with improved security.
+- **Improved backwards compatibility:** the only modifications required to implement interop for a token are including `remoteToken` and giving mint and burn permissions to the `L2StandardBridge`, which can be implemented before the hardfork. This would not be possible with the current design. 
 - **Homogeneous implementation:** tokens on other chains do not implement bridging functionalities, which leads to a fragmented developer experience.
 
 ### Cons of using the alternative design
--- a/protocol/superchainerc20/superc20Redesign.md
+++ b/protocol/superchainerc20/superc20Redesign.md
@@ -1,6 +1,6 @@
 ## Summary
 
-This design doc evaluates the pros and cons of moving the interop specific functions from the [`SuperchainERC20` standard](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/token-bridging.md) to the `L2StandardBridge`. The goal of this document is to open up the discussion and ensure that the best design path is taken.
+This design doc evaluates the pros and cons of moving the interop specific functions from the [`SuperchainERC20` standard](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/token-bridging.md) to a new `SuperchainERC20Bridge` predeploy. The goal of this document is to open up the discussion and ensure that the best design path is taken.
 
 ## The current design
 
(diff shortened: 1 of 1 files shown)


protocol/superchainerc20/superc20Redesign.md
```diff
@@ -0,0 +1,101 @@
...
+
+```
+
+It’s important to notice that, even though this functions get implemented in the bridge, tokens will still have agency to reimplement the bridging functions on their contract if they want so. The alternative design gives the optionality.
+
+## Comparison
+
+### Pros of using the alternative design
+
+- **Compatible with the current design:** adding the `sendERC20` and `relayERC20` functions to the bridge does not imply that these should no longer exist in the token. It only gives the option to token developers not to include them.
```
tynes, line 84: Would this be adding these functions to the `L2StandardBridge`? I think yes, but just double checking

tynes (commented):
This is quite reasonable and also very flexible. I am supportive of this, the one thing that I want to question is around whether or not there should be a separate contract that isn't the `L2StandardBridge` that has this functionality. I think it makes sense to put it in the `L2StandardBridge`, just want to be sure that it doesn't make more sense to have a `L2ToL2StandardBridge`. Perhaps I am over engineering, just want to think thru that scenario before committing.

protocol/superchainerc20/superc20Redesign.md
0xParticle: yes, should have been more clear

protocol/superchainerc20/superc20Redesign.md
```diff
@@ -0,0 +1,101 @@
...
+  L2SBA->>Messenger_A: sendMessage(chainId, message)
+  L2SBA-->L2SBA: emit SentERC20(tokenAddr, from, to, amount, destination)
+  Inbox->>Messenger_B: relayMessage()
+  Messenger_B->>L2SBB: relayERC20(tokenAddr, from, to, amount)
+  L2SBB->>SuperERC20_B: mint(to, amount)
+  L2SBB-->L2SBB: emit RelayedERC20(from, to, amount, source)
+
+```
+
+It’s important to notice that, even though this functions get implemented in the bridge, tokens will still have agency to reimplement the bridging functions on their contract if they want so. The alternative design gives the optionality.
```
mds1, line 78: Is this saying the same thing as the "Compatible with the current design" pro below, or did this mean something different?

protocol/superchainerc20/superc20Redesign.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Comparison
+
+### Pros of using the alternative design
+
+- **Compatible with the current design:** adding the `sendERC20` and `relayERC20` functions to the bridge does not imply that these should no longer exist in the token. It only gives the option to token developers not to include them.
+- **Separation of functionalities:** bridging lives in the bridge, token functions live in the token. This fosters clarity but also comes with improved upgrade flows.
+    - If bridging functions are implemented on the ERC20, updates to the bridge might require updates to every ERC20. This might be easy to do for implementations that use a BeaconProxy structure, but not in other cases.
```
mds1, line 86: > If bridging functions are implemented on the ERC20, updates to the bridge might require updates to every ERC20. This might be easy to do for implementations that use a BeaconProxy structure, but not in other cases.

What do you mean here? I thought we do use the beacon pattern so I'm unsure what this bullet means exactly

protocol/superchainerc20/superc20Redesign.md
```diff
@@ -0,0 +1,101 @@
...
+- **Easier for integrators:** integrators must deal with a single entrypoint instead of many. This would improve their devX considerably.
```
mds1, line 87: Can you expand on this point? Wouldn't an integrator still need all the token addresses, so calling `L2StandardBridge.sendERC20(tokenAddr)` seems similarly easy to calling `tokenAddr.sendERC20`

protocol/superchainerc20/superc20Redesign.md
```diff
@@ -0,0 +1,101 @@
...
+- **Improved indexing and monitoring:** only monitoring events and traces coming from a single source would be necessary instead of many. Improved monitoring correlates with improved security.
+- **Improved backwards compatibility:** the only modifications required to implement interop for a token are including `remoteToken` and giving mint and burn permissions to the `L2StandardBridge`, which can be implemented before the hardfork. This would not be possible with the current design.
```
mds1, line 89: These two points are really nice 👌

protocol/superchainerc20/superc20Redesign.md
```diff
@@ -0,0 +1,101 @@
...
+    - If bridging functions are implemented on the ERC20, updates to the bridge might require updates to every ERC20. This might be easy to do for implementations that use a BeaconProxy structure, but not in other cases.
+- **Easier for integrators:** integrators must deal with a single entrypoint instead of many. This would improve their devX considerably.
+- **Improved indexing and monitoring:** only monitoring events and traces coming from a single source would be necessary instead of many. Improved monitoring correlates with improved security.
+- **Improved backwards compatibility:** the only modifications required to implement interop for a token are including `remoteToken` and giving mint and burn permissions to the `L2StandardBridge`, which can be implemented before the hardfork. This would not be possible with the current design. 
+- **Homogeneous implementation:** most tokens on other chains do not implement bridging functionalities (except for OFTs and some others), which leads to a fragmented developer experience.
+
+### Cons of using the alternative design
+
+- **Timing:**
```
mds1, line 94: Any cons of the design itself, or do we feel this is strictly superior? I could be wrong but I thought we did consider this approach originally (since xERC20 also has its methods on the bridge contract), so I'm curious why we didn't like it originally and what changed

mds1 (commented):
Overall I like this design and have no objections to it. Just left a few small comments around clarifying a few things and better understanding the history of the decision + motivation for this change

protocol/superchainerc20/superc20Redesign.md
tynes: The idea is hinting towards the philosophy of not enshrining a particular superchain ERC20 factory, since there is no one fits all token contract. Rather any token simply needs to give minting rights to the new bridge predeploy to be superchain erc20 compatible. We could also enshrine the same address constraint as well, not sure if we should or not, i think we need to do a little implementation work to find out the tradeoffs

protocol/superchainerc20/superc20Redesign.md
tynes: The integrator would need the token addresses to use as inputs to the function on the bridge. They now no longer care about the interface on those tokens, you could imagine the xERC20 standard using native interop vs OFT vs superchain erc20 having different interfaces but using the interop message passing under the hood

protocol/superchainerc20/superc20Redesign.md
tynes: We did originally think of this approach but it was discarded in favor of putting the interface on the token, this decision was made towards the early stages of the project

protocol/superchainerc20/superc20Redesign.md
0xParticle: Yes, its the same

protocol/superchainerc20/superc20Redesign.md
0xParticle: We will use the Beacon patter for `OptimismSuperchainERC20`, aka tokens that convert back and forth with legacy. Those tokens are gonna be generated by the `OptimismSuperchainERC20Factory` predeploy.

But then for generic `SuperchainERC20` we will not enforce any pattern nor create an implementation specific factory. We think the best path is to suggest using the `Create2Deployer` preinstall with their `init_code`, but they can do whatever they want. Devs might implement a Beacon pattern or not. Or even use a Beacon but point to a different implementation.

In that scenario, a potential upgrade to the `L2toL2CrossDomainMessenger` for example would imply that every single implementation would need to be upgraded in order to be compatible; or would give us very little flexibility in potential changes to the `L2toL2CrossDomainMessenger`.
The new design doesn't have that problem.

To me, this is the most important pro of the list.

protocol/superchainerc20/superc20Redesign.md
0xParticle: We thought about this and could not identify any strong con tbh. I do recognize there is some beauty on having the function directly accesible in the token, like a native behaviour of the token itself. But couldn't formalize that into a proper argument. 
Happy to open up the discussion though.

protocol/superchainerc20/superc20Redesign.md
mds1: I see now, thank you both. So now my thinking is that we should not encourage devs to implement bridging functions on their own ERC20, because:
- It might break depending on changes we make to `L2toL2CrossDomainMessenger`, or it may at least restrict what changes we feel comfortable making
- It fragments UX, which might confuse users as to why some tokens have a different entry point
- Added risk that the ERC20 methods are not implemented correctly

Is there value to having devs implement bridging functionality on the token?

protocol/superchainerc20/superc20Redesign.md
mds1: The main downside I can think of is that it may be a less familiar or less intuitive design—normally to interact with a token, you call the token contract, but now you call a different contract. xERC20 is similar though, so that precedent helps bring the UX back towards something familiar. So overall I am not worried about that downside.

I am curious why we discarded this idea originally though. I think it might have just been for that UX reason?

tynes (approved):

protocol/superchainerc20/superc20Redesign.md
tynes: > I am curious why we discarded this idea originally though. I think it might have just been for that UX reason?

It was originally discarded because we had strong intuitions on defining it as an interface originally and after learning more we dropped those desires

protocol/superchainerc20/superc20Redesign.md
tynes: The only real value that I can think of at this point is the user would interact directly with the token rather than interacting with the bridge

tynes:
I believe this PR is good to be merged now

mds1 (approved):

protocol/superchainerc20/superc20Redesign.md
parseb: I hope "we should not encourage" does not mean "we give no choice".

PR #84, opened 2024-09-18
interop: add Dependency Set Manager & Shared Lockbox design doc

Joxess:
Co-authored-by: ng <ng@defi.sucks>, agusduha <agusduha@defi.sucks>

**Description**

The following design doc cover a previous discussed issue on [ETH withdrawals](https://github.com/ethereum-optimism/specs/issues/362) introduced by interop. The Shared lockbox is presented as the most convenient solution to solve this problem. However, it is known that the design also is the interest for various aspects of the Superchain design, such as chain management, core interop, and others, which need to be carefully discussed.

**Additional context**

This doesn't address the case for L1 bridged tokens.

### DIFF
--- a/protocol/eth-shared-lockbox.md
+++ b/protocol/eth-shared-lockbox.md
@@ -47,7 +47,7 @@ The core proposed changes are as follows:
 - **Introduce the `SharedLockbox` contract**: This contract acts as an escrow for ETH, receiving deposits and allowing withdrawals from approved `OptimismPortal` contracts. The `SuperchainConfig` contract serves as the source of truth of the lockbox.
 - **Modify the `OptimismPortal`**: To forward ETH into the `SharedLockbox` when `depositTransaction` is called, with the inverse process applying when `finalizeWithdrawal` is called.
 
-### Managing op-governed dependency set through `SuperchainConfig` 
+### Managing op-governed dependency set through `SuperchainConfig`
 
 The `SuperchainConfig` contract will serve as the single point for managing the dependency set of a cluster and will be managed by the admin as occurs in other L1 OP contracts. This contract assumes the role of `dependencyManager` for every `SystemConfig` contract involved. Since the dependency graph follows a simple dependency, it only stores a mapping (or array) of chains added, e.g. given a `chainId`.
 
@@ -74,32 +74,32 @@ A code example would look like this:
 
 ```solidity
 function addChain(uint256 _chainId, address _systemConfig) external {
-    require(msg.sender == updater(), "Unauthorized");
+  require(msg.sender == updater(), "Unauthorized");
+  // Add to the dependency set and check it is not already added (`add()` returns false if it already exists)
+  require(_dependencySet.add(_chainId), "Chain already added");
 
-    // Add to the dependency set and check it is not already added (`add()` returns false if it already exists)
-    require(_dependencySet.add(_chainId), "Chain already added");
+  // Store the system config
+  systemConfigs[_chainId] = _systemConfig;
 
-		// Store the system config
-    systemConfigs[_chainId] = _systemConfig;
+  // Loop through the dependency set and update the dependency for each chain
+  for (uint256 i; i < _dependencySet.length() - 1; i++) {
 
-    // Loop through the dependency set and update the dependency for each chain
-    for (uint256 i; i < _dependencySet.length() - 1; i++) {
-		    uint256 currentId = _dependencySet.at(i);
-        
-        // Skip recently added chain
-        if (_chainId == currentId) continue;
-        
-        systemConfigs[currentId].addDependency(_chainId);
-        
-        systemConfigs[_chainId].addDependency(currentId);
-    }
+    uint256 currentId = _dependencySet.at(i);
 
-		address portal = _systemConfig.optimismPortal();
+    // Skip recently added chain
+    if (_chainId == currentId) continue;
 
-		// Authorize the portal on the shared lockbox
-		SHARED_LOCKBOX.authorizePortal(portal);
+    systemConfigs[currentId].addDependency(_chainId);
 
-    emit ChainAdded(_chainId, _systemConfig, portal);
+    systemConfigs[_chainId].addDependency(currentId);
+  }
+
+  address portal = _systemConfig.optimismPortal();
+
+  // Authorize the portal on the shared lockbox
+  SHARED_LOCKBOX.authorizePortal(portal);
+
+  emit ChainAdded(_chainId, _systemConfig, portal);
 }
 ```
 
@@ -121,6 +121,45 @@ A one-time L1 liquidity migration is required for each approved chain. By using 
 Importantly, the entire upgrade process—including migrating the ETH to the `SharedLockbox` and updating the `OptimismPortal` to the latest version—can be executed in a single batched transaction. This approach ensures migration without the necessity of maintaining a persistent migration function in the final contract.
 
 Note that migration processes may not be uniform and may vary according to the status of the chain seeking to join the OP-governed interoperable set. As stated in the previous section, chains are expected to ensure they reach the approved, OP Stack version. In practice, different migration paths could be implemented, for example, to allow chains to use the `SharedLockbox` from inception or migrate from a previous one, for example.
+
+```mermaid
+sequenceDiagram
+    participant L1PAO as L1 ProxyAdmin Owner
+    participant ProxyAdmin as ProxyAdmin
+    participant SystemConfigProxy as SystemConfig
+    participant StorageSetter
+    participant SuperchainConfig
+    participant OptimismPortalProxy as OptimismPortal
+    participant LiquidityMigrator
+    participant SharedLockbox
+
+    Note over L1PAO: Start batch
+
+    %% Step 1: Upgrade SystemConfig to StorageSetter to zero out initialized slot
+    L1PAO->>ProxyAdmin: upgradeAndCall()
+    ProxyAdmin->>SystemConfigProxy: Upgrade to StorageSetter
+    SystemConfigProxy->>StorageSetter: Call to set initialized slot to zero
+
+    %% Step 2: Upgrade SystemConfig and initialize with SuperchainConfig
+    L1PAO->>ProxyAdmin: upgradeAndCall()
+    ProxyAdmin->>SystemConfigProxy: Upgrade to new SystemConfig implementation
+    ProxyAdmin->>SystemConfigProxy: Call initialize(...SuperchainConfig address)
+
+    %% Step 3: Add chain to SuperchainConfig
+    L1PAO->>SuperchainConfig: addChain(chainId, SystemConfig address)
+
+    %% Step 4: Upgrade OptimismPortal to intermediate implementation that transfers ETH
+    L1PAO->>ProxyAdmin: upgradeAndCall()
+    ProxyAdmin->>OptimismPortalProxy: Upgrade to LiquidityMigrator
+    OptimismPortalProxy->>LiquidityMigrator: Call migrateETH()
+    OptimismPortalProxy->>SharedLockbox: Transfer entire ETH balance
+
+    %% Step 5: Upgrade OptimismPortal to final implementation
+    L1PAO->>ProxyAdmin: upgrade()
+    ProxyAdmin->>OptimismPortalProxy: Upgrade to new OptimismPortal implementation
+
+    Note over L1PAO: End batch
+```
 
 ## Impact
 
--- a/protocol/eth-shared-lockbox.md
+++ b/protocol/eth-shared-lockbox.md
@@ -1,6 +1,6 @@
 # Purpose
 
-*The document presented below is part of the [Interoperability project](https://github.com/orgs/ethereum-optimism/projects/71/views/1?sliceBy%5Bvalue%5D=skeletor-spaceman).*
+_The document presented below is part of the [Interoperability project](https://github.com/orgs/ethereum-optimism/projects/71/views/1?sliceBy%5Bvalue%5D=skeletor-spaceman)._
 
 This document discusses possible solutions to address the constraints on ETH withdrawals resulting from the introduction of interop and `SuperchainWETH` that share ETH liquidity across a set of interoperable OP Chains.
 
@@ -30,7 +30,7 @@ OP Chains will be governed within a common Chain Cluster governance entity (the 
 
 - **Ensuring chains are consistent at the implementation side**, achieved either through trusted deployment methods (e.g., OP Contracts Manager) or by approval after security checks are performed.
 - **Approving protocol upgrades** for all chains involved.
-- **Adding new chains** to the interoperable set, and managing the `dependencyManager` role. The final decision is up to governance and chains cannot be removed afterwards.
+- **Adding new chains** to the interoperable set. The final decision is up to governance and chains cannot be removed afterwards.
 - **Replacing chain servicers** for existing chains if they fail to satisfy technical requirements.
 
 ### Shared Bridging and SuperchainWETH usage
@@ -42,14 +42,16 @@ In any OP Chain that joins the cluster, it is assumed that `SuperchainWETH` ha
 The existing problem and considerations motivate us to propose an L1 shared liquidity design through the introduction of a new `SharedLockbox` on L1, which serves as a singleton contract for ETH, given a defined set of interoperable chains. To ensure consistency, the `SuperchainConfig` is extended to act as the *dependency set manager*, controlling the op-governed dependency set that the lockbox uses as a source of truth. New ETH deposits will be directed to the lockbox, with the same process applied to ETH withdrawals.
 
 ### Spec changes
+
 The core proposed changes are as follows:
-- **Modify the `SuperchainConfig` contract**: Extend this contract to manage the dependency set of each chain, taking ownership of the `dependencyManager` role for each `SystemConfig`.
+
+- **Modify the `SuperchainConfig` contract**: Extend this contract to manage the dependency set of each chain, interacting with each `SystemConfig`.
 - **Introduce the `SharedLockbox` contract**: This contract acts as an escrow for ETH, receiving deposits and allowing withdrawals from approved `OptimismPortal` contracts. The `SuperchainConfig` contract serves as the source of truth of the lockbox.
 - **Modify the `OptimismPortal`**: To forward ETH into the `SharedLockbox` when `depositTransaction` is called, with the inverse process applying when `finalizeWithdrawal` is called.
 
 ### Managing op-governed dependency set through `SuperchainConfig`
 
-The `SuperchainConfig` contract will serve as the single point for managing the dependency set of a cluster and will be managed by the admin as occurs in other L1 OP contracts. This contract assumes the role of `dependencyManager` for every `SystemConfig` contract involved. Since the dependency graph follows a simple dependency, it only stores a mapping (or array) of chains added, e.g. given a `chainId`.
+The `SuperchainConfig` contract will serve as the single point for managing the dependency set of a cluster and will be managed by the admin as occurs in other L1 OP contracts. This contract interacts with every `SystemConfig` contract involved. Since the dependency graph follows a simple dependency, it only stores a mapping (or array) of chains added, e.g. given a `chainId`.
 
 ```solidity
 // Mapping from chainId to SystemConfig address
@@ -126,8 +128,6 @@ Note that migration processes may not be uniform and may vary according to the s
 sequenceDiagram
     participant L1PAO as L1 ProxyAdmin Owner
     participant ProxyAdmin as ProxyAdmin
-    participant SystemConfigProxy as SystemConfig
-    participant StorageSetter
     participant SuperchainConfig
     participant OptimismPortalProxy as OptimismPortal
     participant LiquidityMigrator
@@ -135,26 +135,16 @@ sequenceDiagram
 
     Note over L1PAO: Start batch
 
-    %% Step 1: Upgrade SystemConfig to StorageSetter to zero out initialized slot
-    L1PAO->>ProxyAdmin: upgradeAndCall()
-    ProxyAdmin->>SystemConfigProxy: Upgrade to StorageSetter
-    SystemConfigProxy->>StorageSetter: Call to set initialized slot to zero
-
-    %% Step 2: Upgrade SystemConfig and initialize with SuperchainConfig
-    L1PAO->>ProxyAdmin: upgradeAndCall()
-    ProxyAdmin->>SystemConfigProxy: Upgrade to new SystemConfig implementation
-    ProxyAdmin->>SystemConfigProxy: Call initialize(...SuperchainConfig address)
-
-    %% Step 3: Add chain to SuperchainConfig
+    %% Step 1: Add chain to SuperchainConfig
     L1PAO->>SuperchainConfig: addChain(chainId, SystemConfig address)
 
-    %% Step 4: Upgrade OptimismPortal to intermediate implementation that transfers ETH
+    %% Step 2: Upgrade OptimismPortal to intermediate implementation that transfers ETH
     L1PAO->>ProxyAdmin: upgradeAndCall()
     ProxyAdmin->>OptimismPortalProxy: Upgrade to LiquidityMigrator
     OptimismPortalProxy->>LiquidityMigrator: Call migrateETH()
     OptimismPortalProxy->>SharedLockbox: Transfer entire ETH balance
 
-    %% Step 5: Upgrade OptimismPortal to final implementation
+    %% Step 3: Upgrade OptimismPortal to final implementation
     L1PAO->>ProxyAdmin: upgrade()
     ProxyAdmin->>OptimismPortalProxy: Upgrade to new OptimismPortal implementation
 
@@ -177,12 +167,11 @@ The problem with L2 reverts is that it breaks the ETH withdrawal guarantee invar
 
 ### **Permission to withdraw ETH from a different Portal**
 
-[Another solution](https://github.com/ethereum-optimism/specs/issues/362#issuecomment-2332481041) involves allowing ETH withdrawals to be finalized by taking ETH from one or more `OptimismPortal` contracts. In a cluster, this is done by authorizing withdrawals across the set of chains. For example, if  we have a dependency set composed by Chain A, B and C, a withdrawal initiated from A could be finalized by using funds from B and C if needed.
+[Another solution](https://github.com/ethereum-optimism/specs/issues/362#issuecomment-2332481041) involves allowing ETH withdrawals to be finalized by taking ETH from one or more `OptimismPortal` contracts. In a cluster, this is done by authorizing withdrawals across the set of chains. For example, if we have a dependency set composed by Chain A, B and C, a withdrawal initiated from A could be finalized by using funds from B and C if needed.
 
 The implementation would require iterating through the dependency set, determined on L1 to find the next `OptimismPortal` with available funds. This also means `OptimismPortal` would need to have authorization and validation logic to allow to extraction ETH given an arbitrary request dictated by the loop.
 **Implementation and security considerations**
 Since both approaches rely on multiple `OptimismPortal` contracts and access controls, the Shared Lockbox design stays as the minimal way to implement a shared liquidity approach. Instead, allowing a looping withdrawal into each `OptimismPortal` increases the code complexity and surface of bugs during the iterative checking.
-
 
 # Risks & Uncertainties
 
--- a/protocol/eth-shared-lockbox.md
+++ b/protocol/eth-shared-lockbox.md
@@ -1,0 +1,120 @@
+# Purpose
+
+*The document presented below is part of the [Interoperability project](https://github.com/ethereum-optimism/optimism/issues/10899).*
+
+This document discusses possible solutions to address the constraints on ETH withdrawals resulting from the introduction of interop and `SuperchainWETH` that share ETH liquidity across a set of OP Chains.
+
+# Summary
+
+With interoperable ETH, withdrawals may fail if the referenced `OptimismPortal`lacks sufficient ETH—especially for large amounts or on OP Chains with low liquidity—since ETH liquidity isn't shared at the L1 level. Several solutions have been explored to prevent users from getting stuck. Currently, the Superchain design favors the `SharedLockbox` as the most effective solution. Key considerations are detailed at the end of the document.
+
+# Problem Statement + Context
+
+With the introduction of interop, `SuperchainWETH` is created as a medium to move ETH across a set of interoperable chains, instead of using native ETH. To enable users to convert `SuperchainWETH` into native ETH, `ETHLiquidity` is also introduced, containing a sufficiently large virtual pool of native ETH, which can only be used for this purpose. As a result, ETH can be shared through all interconnected chains.
+
+However, one remaining problem to solve relates to [ETH withdrawals](https://github.com/ethereum-optimism/specs/issues/362). Currently, with Interop, the amount of ETH on an L2 can differ from what is deposited in the respective `OptimismPortal`. This mismatch can interfere with the finalization of withdrawals, especially if a request requires more ETH than is actually deposited at a given time.
+
+As a side note, this problem is also closely related to the process of how an OP Chain enters and leaves an interoperable set of chains. Not defining this scenario could create a problem in the future.
+
+# Prior considerations
+
+Before explaining the proposed solution, it is important to note that there are essential assumptions that underpin the whole design.
+
+### Chain Management
+
+This refers to how OP Chains will be governed within an interoperable set of chains (also called a "cluster"). Once governance principles are well-established from the perspective of the technological stack, state transitions can be secured, enabling ETH to be uniformly integrated and secured across the entire cluster. Currently, this is achieved by addressing the following:
+
+- The interoperable set of chains will be governed by an entity responsible for:
+    - Approving chains to join the cluster. This is done either through trusted deployment methods (e.g., OP Stack Manager) or by approval after security checks are performed.
+    - Managing (adding and removing) chains within the interoperable graph.
+    - Approving protocol upgrades for all involved.
+    - Taking action if an OP Chain decides not to follow the established standard (e.g. Expresses a desire to leave the cluster.).
+- Each chain is capable of defending against any maliciously claimed state transition. Ideally, this is accomplished through a set of security features at the proof level, such as permissionless proofs, capability of proving executions that are related to interop executions, and the presence of the Guardian role for the whole system. This is indifferent to what proof system is used (ZK, fault proofs, etc.).
+- Joining a cluster will automatically enable the use of `SuperchainWETH` for the specific set of chains defined by the `dependencyManager`.
+
+### SuperchainWETH usage
+
+As mentioned above, in any OP Chain within the cluster where interop is enabled, the use of `SuperchainWETH` will be activated. As a result, equivalence between ETH deposits and the actual ETH supply will vary from the outset. In a world with freedom of movement, all ETH liquidity is theoretically shared across the entire cluster eventually, regardless of how deposits are handled. This is an unavoidable scenario that deserves careful consideration and is crucial for understanding the following sections.
+
+# Solution
+
+We propose an L1 shared liquidity design, achieved by introducing a new `SharedLockbox` in L1 that serves as a singleton ETH contract for the entire interoperable set of chains. New ETH deposits will be forwarded to the lockbox, and the same applies to ETH withdrawals. This could also serve as the singleton contract for gas tokens in the case of CGT chains.
+
+### Spec changes
+
+The core changes proposed are the following:
+
+- Introduce `SharedLockbox` which accepts deposits and withdrawals of ETH and Gas Tokens as well.
+- Modify `OptimismPortal` to forward ETH deposits into `SharedLockbox` and allow extraction of ETH when a withdrawal is finalized.
+- Maintain an on-chain list (through `SystemConfigInterop` or another contract if already defined) that registers the chains that are part of the cluster. This allows `SharedLockbox` to validate a withdrawal from a chain that is part of the cluster.
+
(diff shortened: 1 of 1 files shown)


K-Ho:
From a product perspective - the ETH shared lockbox design solves a lot of user pain, by maintaining the invariant that "any amount of ETH can always be withdrawn from L2 to L1 safely", which many protocols have built upon. I think the only concern at this point is the time and resources required to actually ship this securely. Is there some rough scoping of the amount of time this adds to the engineering timeline for interoperable ETH? Who would we need review/support from to feel confident in the security of this upgrade? It's certainly a major change and would want to make sure we get a diverse set of reviews/eyes on this + put together a very thorough Failure Mode Analysis

K-Ho:
Also a stream of work will be making sure that external tooling that relies on the OptimismPortal to calculate ETH stored in a chain is updated as well (e.g. update L2Beat, Defillama, etc.)

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,120 @@
...
+
+The core changes proposed are the following:
+
+- Introduce `SharedLockbox` which accepts deposits and withdrawals of ETH and Gas Tokens as well.
+- Modify `OptimismPortal` to forward ETH deposits into `SharedLockbox` and allow extraction of ETH when a withdrawal is finalized.
+- Maintain an on-chain list (through `SystemConfigInterop` or another contract if already defined) that registers the chains that are part of the cluster. This allows `SharedLockbox` to validate a withdrawal from a chain that is part of the cluster.
+
+This would require a one-time L1 liquidity migration process for each chain. Newly deployed chains via `OPSM` should use the `SharedLockbox` from the beginning.
+
+### Joining the Cluster
```
K-Ho, line 53: I'm not 100% sure here, but I dont think it makes sense to define a separate "cluster" for the lockbox, when we already have a separate SystemConfig that lists out dependency sets. Would love to get all teams together during design review to align on a single on chain set of chains that are both: in bi-directional dependency set, AND in the shared lockbox.

Joxess:
> Also a stream of work will be making sure that external tooling that relies on the OptimismPortal to calculate ETH stored in a chain is updated as well (e.g. update L2Beat, Defillama, etc.)

This is a very interesting consideration. Actually, when making ETH interoperable, a shift in the interpretation of liquidity distribution is needed. To show the actual TVL, `OptimismPortal` isn't the only data to reference anymore; it is also necessary to consider the new contributions whenever ETH enters or exits in the form of `SuperchainWETH`. Probably, similar reasoning as in the case of a migration (deposit mapping + SuperchainWETH net flows) will need to be followed.

protocol/eth-shared-lockbox.md
Joxess: It makes sense. It definitely requires `SharedLockbox` to serve the entire Superchain cluster, no matter how the dependency graph ends up, so the contract only needs to know if the chain where the withdrawal is requested is part of it.

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,120 @@
...
+
+# Solution
+
+We propose an L1 shared liquidity design, achieved by introducing a new `SharedLockbox` in L1 that serves as a singleton ETH contract for the entire interoperable set of chains. New ETH deposits will be forwarded to the lockbox, and the same applies to ETH withdrawals. This could also serve as the singleton contract for gas tokens in the case of CGT chains.
+
+### Spec changes
+
+The core changes proposed are the following:
+
+- Introduce `SharedLockbox` which accepts deposits and withdrawals of ETH and Gas Tokens as well.
```
tynes, line 47: Must this also custody gas tokens? Seems like it would create some accounting complexity. I think we could skip this feature for now since we aren't planning on adding any custom gas token chains during the first release of interop, we could add this feature in the future when there is a custom gas token chain that becomes popular

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,120 @@
...
+- Modify `OptimismPortal` to forward ETH deposits into `SharedLockbox` and allow extraction of ETH when a withdrawal is finalized.
+- Maintain an on-chain list (through `SystemConfigInterop` or another contract if already defined) that registers the chains that are part of the cluster. This allows `SharedLockbox` to validate a withdrawal from a chain that is part of the cluster.
```
tynes, line 49: This control mechanism is really important. I would love to enumerate some possible designs. Right now there is a "dependency set manager" role in the `SystemConfig`. It was originally thought that this role would be governance/security council, but now maybe the default config should define it as this contract and then this contract is owned by gov/security council and it holds the aggregate dependency set and calls each individual chain's `SystemConfig` to manage the dependency set. Or we should rethink how the dependencies are managed from first principles, another thing that I realized recently is that the dependencies are not maintained within the `SystemConfig` on L1, they only exist at the moment on L2 in the `L1Block` contract.

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,120 @@
...
+
+This would require a one-time L1 liquidity migration process for each chain. Newly deployed chains via `OPSM` should use the `SharedLockbox` from the beginning.
+
+### Joining the Cluster
+
+From the point of view of a chain deployed via `OPSM`, we suggest adopting `SharedLockbox` even when there is no interoperable set established for it. Since the security is guarded by the security council and governance, it should not represent a risk.
+
+For existing chains, the `OptimismPortal` should incorporate a new gated function that hardcodes the sending of ETH to the `SharedLockbox` contract and that can only be called by a trusted entity, e.g., the Security Council. This function should also emit an event which the off-chain components can use to track how much ETH was initially moved from a chain to the `SharedLockbox`, which may be required in case the chain wishes to opt out of the cluster in the future.
```
tynes, line 57: I would recommend not adding a function and instead do an intermediate upgrade to a contract that simply sends all balance in its `initialize` and then upgrade to the implementation of the new `OptimismPortal`. This removes risk of having that function there forever

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,120 @@
...
+function joinInteropCluster() external onlySecurityCouncil {
+	uint256 balance = address(this).balance;
+	address(SHARED_LOCKBOX).call{value: balance}("");
+	emit JoinedInteropCluster(chainId, balance);
+}
+
+```
+
+### Opt-out the cluster
+
```
tynes, line 69: Being able to opt out is nice but to be honest I think that this adds too much complexity. We could just say that it's one way, once you join there is no leaving. I could be convinced otherwise because freedom is important, but there isn't a good solution for L2 native assets or upgraded deposited tokens

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,120 @@
...
+# Alternatives Considered
+
+### Reverts on L2
+
+One alternative is not to share liquidity at the L1 level and instead prevent withdrawals from being stuck by reverting them on L2. This approach requires tracking the exact ETH balance in L2 via deposits and withdrawals. An `ETHBalance` counter would increment with new mints from deposits and decrease with successful `initiateWithdrawal` calls.
+
+This method would require minimal changes to `L1Block` and adjustments to how `TransactionDeposited` is processed to validate new ETH counts. Additionally, it necessitates porting the initial balance as part of an upgrade transaction.
+
+The problem with L2 reverts is that it breaks the ETH withdrawal guarantee invariant and still exposes the system to solvency issues. Based on previous feedback, this would affect existing applications and pre-deployed contracts such as the current `FeeVault` withdrawal contract logic.
```
tynes, line 107: This logic would be implemented in the `L2ToL1MessagePasser` in practice, it's the lowest level abstraction. I am confident in being able to get the "start balance" into L2, but I do agree that it would break applications that assume they can withdraw, but at least their ether wouldn't be stuck

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Spec changes
+
+The core changes proposed are the following:
+
+- Introduce `SharedLockbox` which accepts deposits and withdrawals of ETH and Gas Tokens as well.
+- Modify `OptimismPortal` to forward ETH deposits into `SharedLockbox` and allow extraction of ETH when a withdrawal is finalized.
+- Maintain an on-chain list (through `SystemConfigInterop` or another contract if already defined) that registers the chains that are part of the cluster. This allows `SharedLockbox` to validate a withdrawal from a chain that is part of the cluster.
+
+This would require a one-time L1 liquidity migration process for each chain. Newly deployed chains via `OPSM` should use the `SharedLockbox` from the beginning.
```
Ethnical, line 51: To be compliant with the name of the OPSM -> OPCM now?
```suggestion
This would require a one-time L1 liquidity migration process for each chain. Newly deployed chains via `OPCM` should use the `SharedLockbox` from the beginning.
```

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Joining the Cluster
+
+From the point of view of a chain deployed via `OPSM`, we suggest adopting `SharedLockbox` even when there is no interoperable set established for it. Since the security is guarded by the security council and governance, it should not represent a risk.
```
Ethnical, line 55: Same here to avoid confusion?
```suggestion
From the point of view of a chain deployed via `OPCM`, we suggest adopting `SharedLockbox` even when there is no interoperable set established for it. Since the security is guarded by the security council and governance, it should not represent a risk.
```

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,120 @@
...
+- Joining a cluster will automatically enable the use of `SuperchainWETH` for the specific set of chains defined by the `dependencyManager`.
+
+### SuperchainWETH usage
+
+As mentioned above, in any OP Chain within the cluster where interop is enabled, the use of `SuperchainWETH` will be activated. As a result, equivalence between ETH deposits and the actual ETH supply will vary from the outset. In a world with freedom of movement, all ETH liquidity is theoretically shared across the entire cluster eventually, regardless of how deposits are handled. This is an unavoidable scenario that deserves careful consideration and is crucial for understanding the following sections.
+
+# Solution
+
+We propose an L1 shared liquidity design, achieved by introducing a new `SharedLockbox` in L1 that serves as a singleton ETH contract for the entire interoperable set of chains. New ETH deposits will be forwarded to the lockbox, and the same applies to ETH withdrawals. This could also serve as the singleton contract for gas tokens in the case of CGT chains.
```
Ethnical, line 41: > in L1 that serves as a singleton ETH contract
So every interop chains are about to forward ETH into the`SharedLockbox` so we should be capable to track the amount of total ETH on the Superchain by checking the balance of this contract?
Or there is still chains that will keep ETH in the previous `OptimismPortal`?

protocol/eth-shared-lockbox.md
Joxess: Changes incorporated.

protocol/eth-shared-lockbox.md
Joxess: I have taken this feedback and incorporated a new contract called `DependencySetManager` to manage the `dependencyManager` role.

protocol/eth-shared-lockbox.md
Joxess: Updated!

protocol/eth-shared-lockbox.md
Joxess: Yes, I have updated the doc and removed the section.

protocol/eth-shared-lockbox.md
Joxess: Good question. In this design, chains that are part of the interoperable graph must adhere to the `SharedLockbox` by default, so the total amount of ETH will be there. Although, in principle, one could design this solution as opt-in, it could revert to the main problem and still make them susceptible to liquidity constraints.

Joxess:
I have updated the doc with substantial changes:

- Updates to the Considerations section
- Introduction of the `DependencySetManager` contract as a replacement for the old "Joining the Cluster" section
- Proposed changes to `SystemConfigInterop`
- Modifications to the upgrade process
- Removal of the "Opt-out of the Cluster" section and liquidity trackability features

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+
+This means a solution is needed to achieve a truly shared `SuperchainWETH` via shared ETH liquidity given an interoperable set of chains.
+
+# Prior considerations
+
+Before explaining the proposed solution, it is important to note that there are essential assumptions that underpin the whole design: the Shared Security Model, Unified Chain Governance and Superchain WETH activation.
+
+### Shared Security Model
+
+The security model for the set of interoperable chains (commonly called a "cluster") is shared across all involved chains, ensuring that all state transitions are secured. For example, the shared security model allows for defending against any maliciously claimed state transition for any chain within the cluster. This is accomplished through a set of security features at the proof level, such as permissionless proposing, interop-provable proofs (also called shared proofs), and the presence of a single Guardian role across the entire system. This model is independent of the proof system used (e.g., ZK, fault proofs, etc.).
```
mds1, line 25: It sounds like this means we will have to move the respected game type out of the portal, where it can be different per-chain, and into the SuperchainConfig, where it's the same for all chains?

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Unified Chain Governance
+
+OP Chains will be governed within a common Chain Cluster governance entity (the Collective). Currently, this entity is responsible for:
```
mds1, line 29: Is this just the standard existing governance process, or is this something new?

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+
+- **Ensuring chains are consistent at the implementation side**, achieved either through trusted deployment methods (e.g., OP Contracts Manager) or by approval after security checks are performed.
+- **Approving protocol upgrades** for all chains involved.
+- **Adding new chains** to join into the interoperable set, managing the `dependencyManager` role. The final decision is up to governance.
+- **Remove existing chains** from the interoperable set. However, such a removal should not be executed without a contingency plan due to the significant implications this could entail.
```
mds1, line 34: What are the implications? Can chains be removed from the interoperable set?

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Shared Bridging and SuperchainWETH usage
+
+In any OP Chain that joins the cluster, the use of `SuperchainWETH` is activated. As a result, the equivalence between ETH deposits and withdrawal history and the actual ETH supply will vary from the outset. In a world with freedom of movement, all real ETH liquidity is theoretically shared across the entire cluster eventually, regardless of how deposits are handled.
```
mds1, line 38: What do you mean that it's "activated", how does this occur at the same time as updating the dependency set?

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+
+# Solution
+
+The existing problem and considerations motivate us to propose an L1 shared liquidity design through the introduction of a new `SharedLockbox` on L1, which serves as a singleton contract for ETH, given a defined set of interoperable chains. To ensure consistency, the `DependencySetManager` is also introduced to manage the dependency set that the lockbox uses as a source of truth. New ETH deposits will be directed to the lockbox, with the same process applying to ETH withdrawals.
```
mds1, line 42: Why is this lockbox only for ETH, and not tokens? Don't tokens have the same withdrawal liquidity problem when moved between chains?

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+- Introduce the `SharedLockbox` contract: It acts as an escrow for ETH, receiving deposits and allows withdrawals from approved `OptimismPortal2` contracts. The `DependencySetManager` contract serves as the source of truth of the lockbox.
+- Modify the `SystemConfigInterop`: Add `addDependencies` and `removeDependencies` functions to accept an array of `_chainId` values.
+- Modify the `L1BlockInterop`: To maintain consistency with `SystemConfigInterop`.
+- Modify the `OptimismPortal2`/`OptimismPortalInterop`: To forward ETH into the `SharedLockbox` when `depositTransaction` is called, with the same process applying when `finalizeWithdrawal` is called.
+
+### Managing `DependencySetManager`
+
+This contract serves as the single point for managing the dependency set of a cluster and is expected to be managed by an admin in the same manner as other L1 OP contracts (proxiable). This contract assumes the role of `dependencyManager` for every `SystemConfigInterop` contract involved. In the case of a simple dependency, the `DependencySetManager` only stores a mapping (or array) of chains added, e.g. given a `chainId`.
```
mds1, line 56: > In the case of a simple dependency

What does this mean, and what is the other case?

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+
+Adding a new chain can be done as follows:
+
+1. `registerChain` is called, which adds the chain to the registry (`chainId` value, `SystemConfig` address, `OptimismPortal2` address) but does not add it to the dependency set. The [Superchain Registry](https://github.com/ethereum-optimism/superchain-registry) or the [OP Contracts Manager](https://specs.optimism.io/experimental/op-contracts-manager.html?highlight=chain#chain-id-source-of-truth) can be used as the source of truth to add chains that have been verified as compatible with this integration.
+2. An OP-governed address, who has the role to call `addChain` with the new `chainId`.
```
mds1, line 61: Why do we need to store the dependency set twice, in the shared `DependencySetManager` / OPCM source of truth (unsure which it is from the text), and again in the system configs?

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+// Mapping from chainId to SystemConfigInterop address
+mapping(uint256 => address) public systemConfigInterops;
+// Mapping to check if a chainId is in the dependency set
+mapping(uint256 => bool) public alreadyAdded;
+// Current dependency set list
+uint256[] public dependencySet;
+
+// Function to add a chain to the dependency set
+function addChain(uint256 chainId) external onlyOwner {
+		require(systemConfigInterops[chainId] != address(0), "DependencySetManager: Chain not registered");
```
mds1, line 80: nit, indentation looks right in the code block text, but off when rendered
image

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+### `SharedLockbox` implementation
+
+A minimal set of functions should include:
+
+- `lockETH`: Accepts ETH from the `depositTransaction` originating from a valid `OptimismPortal2`.
+- `unlockETH`: Releases ETH from the `finalizeWithdrawalTransaction` originating from a valid `OptimismPortal2`.
+
+Access control for `lockETH` and `unlockETH` is validated against the list reated by `registerChain`.
```
mds1, line 130: typo, "reated"

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+
+### `OptimismPortal2` upgrade process
+
+A one-time L1 liquidity migration is required for each approved chain. By using an intermediate contract during the upgrade, all ETH held by the `OptimismPortal` can be transferred to the `SharedLockbox` within the `initialize` function. After this migration, the `OptimismPortal` is upgraded to the new version with the desired functionality.
```
mds1, line 134: What alternative migration flows have we considered and what are the tradeoffs? For example, we could just add a `forwardETH()` method to the portals that forwards all ETH balance to the lockbox whenever it's called, and call it from the `initialize` function

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+# Solution
+
+The existing problem and considerations motivate us to propose an L1 shared liquidity design through the introduction of a new `SharedLockbox` on L1, which serves as a singleton contract for ETH, given a defined set of interoperable chains. To ensure consistency, the `DependencySetManager` is also introduced to manage the dependency set that the lockbox uses as a source of truth. New ETH deposits will be directed to the lockbox, with the same process applying to ETH withdrawals.
+
+## Spec changes
+
+The core changes proposed are as follows:
+
+- Introduce the `DependencySetManager` contract: This contract manages and tracks the dependency graph.
+- Introduce the `SharedLockbox` contract: It acts as an escrow for ETH, receiving deposits and allows withdrawals from approved `OptimismPortal2` contracts. The `DependencySetManager` contract serves as the source of truth of the lockbox.
```
mds1, line 49: What alternatives are there to adding these two contracts? For example, can the functionality of the `DependencySetManager` instead live in `SuperchainConfig` or OPCM?

My main concern is that we have a lot of L1 contracts already so our architecture + auth can be confusing. I'd like to work towards reducing the number of contracts (when it makes sense to, i.e. good separation of concerns is still important). But both `SuperchainConfig` and OPCM feel like suitable candidates for the role of `DependencySetManager`

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+
+Importantly, the entire upgrade process—including migrating the ETH to the `SharedLockbox` and updating the `OptimismPortal` to the latest version—can be executed in a single transaction. This approach ensures a migration without the necessity of maintaining a persistent migration function in the final contract.
+
+Only chains registered in the `DependencySetManager` should be approved to join the `SharedLockbox`. Once approved, the `SharedLockbox` address is added in the `constructor` and `initialize` functions for each `OptimismPortal2`.
+
+## Impact
+
+The following components require an audit of the new and modified contracts: `OptimismPortal2`, `SharedLockbox`, `SystemConfigInterop`, `L1BlockInterop`, and `DependencySetManager`. This also includes the scripts necessary to perform the migration.
+
+# Alternatives Considered
```
mds1, line 144: Related to some of my questions about (why do we need dependency set info in the system config and the DependencySetManager, and do we really need the DependencySetManager), I'd like to see more alternatives and justification for this specific lockbox design, compared to other lockbox designs. 

An example: Have we considered deprecating the portal, and having the lockbox emit deposit events with a chain ID? The obvious downside is now this requires changes to the derivation path, but it does mean we can get rid of an entire contract which is nice

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+
+The problem with L2 reverts is that it breaks the ETH withdrawal guarantee invariant and still exposes the system to solvency issues. Based on previous feedback, this would affect existing applications and pre-deployed contracts such as the current `FeeVault` withdrawal contract logic.
+
+### **Permission to withdraw ETH from a different Portal**
+
+[Another solution](https://github.com/ethereum-optimism/specs/issues/362#issuecomment-2332481041) involves allowing ETH withdrawals to be finalized by taking ETH from one or more `OptimismPortal` contracts. In a cluster, this is done by authorizing withdrawals across the set of chains. For example, if  we have a dependency set composed by Chain A, B and C, a withdrawal initiated from A could be finalized by using funds from B and C if needed.
+
+The implementation would require iterating through the dependency set, determined on L1 to find the next `OptimismPortal` with available funds. This approach incurs more modifications to the `OptimismPortal`, increasing complexity.
+
+# Risks & Uncertainties
```
mds1, line 160: I'd like to see security considerations fleshed out a bit more. Arguably the risks aren't that much different than storing all this ETH in separate portal proxies that point to the same implementation, but there are risks with both the ETH migration and having a giant honeypot of funds. What is the simplest and most secure migration approach and architecture we can come up with? How do we gain confidence in the implementations?

protocol/eth-shared-lockbox.md
Joxess: This is based in a recent draft on op-governed interop set

protocol/eth-shared-lockbox.md
Joxess: Same as above

protocol/eth-shared-lockbox.md
Joxess: This means that we assume when a chain receives the interop upgrade (and then joins the dependency set), `SuperchainWETH` will also be deployed

protocol/eth-shared-lockbox.md
Joxess: Oh, not the best phrasing; we’re assuming a simple dependency only

protocol/eth-shared-lockbox.md
tynes: Yes, I believe that this is required given that there is an idea of shared security between all of the chains

protocol/eth-shared-lockbox.md
tynes: Why limit `SuperchainWETH` to being part of the cluster? It seems exclusive, the contract should always just exist on any chain post interop hardfork being activated

protocol/eth-shared-lockbox.md
tynes: The idea is to keep the scope small for now

protocol/eth-shared-lockbox.md
tynes: A way to think about this is that there is a difference from the ways that the protocol may work, and the way that we want to the protocol to work. Right now we want to enforce a bidirectional relationship for all dependencies. Whatever contract is the `dependencyManager` role should enforce this

protocol/eth-shared-lockbox.md
tynes, line 80: We will rename `SystemConfigInterop` to `SystemConfig`, its a temporary naming scheme. Same with renaming `OptimismPortal2` to `OptimismPortal`, so you can refer to them with those names

protocol/eth-shared-lockbox.md
Joxess: That's ok; it should not be a problem

protocol/eth-shared-lockbox.md
smartcontracts: RespectedGameType is already being moved into AnchorStateRegistry which may begin to resolve this problem

tynes: At first thought, we are going to need to have the `RespectedGameType` be the same for all interoperable chains, otherwise we risk heterogeneous (weakest link) security, so it may need to move again, or we will need to use social consensus to make sure all the values are the same and have playbooks for modifying them all atomically. Doable at a small number of chains, but definitely not ideal, so I do think it makes sense to put it in the `SuperchainRegistry` given the `AnchorStateRegistry` will have its own proxy for each chain

protocol/eth-shared-lockbox.md
tynes: I am not opinionated on what contract does this enforcement, whether its the `SuperchainConfig` or not, we should inspect where the roles live and come to a shared optimization target

protocol/eth-shared-lockbox.md
zainbacchus: We expect this to be the standard existing governance process, as outlined here: https://docs.google.com/document/d/1zeDO241CHl1H8Xl3XqdkmMedP4KQwT4AMRJQ32Hs_HE/edit

protocol/eth-shared-lockbox.md
zainbacchus: Chains should not be able to removed. https://docs.google.com/document/d/1zeDO241CHl1H8Xl3XqdkmMedP4KQwT4AMRJQ32Hs_HE/edit

zainbacchus:
- We expect the Superchain interop set to have a mesh topology with complete graphs. This is being discussed here: https://docs.google.com/document/d/1zeDO241CHl1H8Xl3XqdkmMedP4KQwT4AMRJQ32Hs_HE/edit?tab=t.0
- Even w/in the mesh topology of the Superchain interop set, there is a risk here that too much ETH ends up on a different L2 than it was originally deposited into and can break existing smart contracts from bridge and CEXes.  Problem outlined here: https://docs.google.com/presentation/d/13AHhuoxqOzte_Ev4s67NKix80H23r6Jv-lN_I1Egrok/edit#slide=id.g2ff36f95630_1_3

tynes:
One consideration now that we have a contract for managing the dependency set - should we make the full dependency set legible from within each L2? This would prevent the need to send to chains that do not have the local chain in their dependency set

We could assume this by observing the dependency set locally when sending a message through the `L2ToL2CrossDomainMessenger`, but this would have implications about the topology about all possible chain clusters. I do not love the idea of enforcing that all possible clusters MUST be a mesh

protocol/eth-shared-lockbox.md
zainbacchus: ETH is the priority for the initial release. 

We do expect to come back to deposited tokens but not for the inital release to limit scope.

protocol/eth-shared-lockbox.md
smartcontracts: Want to get back to this, I'm worried that not being able to remove chains from the interop set is possibly very bad strategy long-term

tynes:
We need to align on whether or not we want to be supporting non-complete graphs in the dependency set. We can make assumptions that it must be a full mesh. Right now we do not make this assumption

protocol/eth-shared-lockbox.md
agusduha: We are considering consolidating all new functionality into the `SuperchainConfig`, though having a specific contract doesn't sound like a bad option either. 

This contract would manage the dependency set and the ability to add a chain to it. 

Also we should consider that this contract will have to be set as the dependency manager of the `SystemConfig`.

protocol/eth-shared-lockbox.md
agusduha: Deprecating the portal in favor of developing a new SharedPortal sounds like a good long-term goal, but it would be far beyond the scope of the ETH Lockbox feature. Creating this new SharedPortal looks like a massive and highly sensitive project that would need debate and coordination across multiple teams to bring it to life. We think this is a valid but complex question that could significantly alter the direction of this task. We're interested to hear what others think about it.

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+				ISystemConfigInterop(newChainSystemConfig).addDependencies(dependencySet);
+	}		
+	// ...
+```
+
+### `SharedLockbox` implementation
+
+A minimal set of functions should include:
+
+- `lockETH`: Accepts ETH from the `depositTransaction` originating from a valid `OptimismPortal2`.
```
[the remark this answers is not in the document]
agusduha: Once we define the migration approach, we will try to apply the code changes needed and check if it reach the size limit. If we are able to avoid hitting the limit, we will leave it like that and apply the changes to the portal. If we hit the size limit, we will need to revisit and tackle the possible portal decoupling.

protocol/eth-shared-lockbox.md
gotzenx: We are exploring various migration options, with the goal of minimizing the amount of permanently added code. We’ll share some insights on this soon.

tynes:
> One consideration now that we have a contract for managing the dependency set - should we make the full dependency set legible from within each L2? This would prevent the need to send to chains that do not have the local chain in their dependency set
> 
> We could assume this by observing the dependency set locally when sending a message through the `L2ToL2CrossDomainMessenger`, but this would have implications about the topology about all possible chain clusters. I do not love the idea of enforcing that all possible clusters MUST be a mesh

Lets not worry about this as it adds scope and we can revisit in the future

protocol/eth-shared-lockbox.md
tynes: I would prefer to not add more derivation changes. We have considered a "beacon" style deposit contract, this could be on the roadmap in the future but this feature doesn't make or break the design. I am in favor of the minimal diff that maintains the portal contracts and simply holds its ether in a shared lockbox

tynes: We do need a dependency set manager that orchestrates the dependency set of the entire cluster, unless we do have a shared system config

protocol/eth-shared-lockbox.md
tynes: I would be in favor of either the `SuperchainConfig` or adding another contract, with slight preference for the `SuperchainConfig` to reduce the number of contracts that we need to manage/deploy

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,164 @@
...
+    emit ChainAdded(chainId);
+}
+    
+```
+
+For `removeChain`, the process would follows the same logic.
+
+Note that, under the specified flow, the dependency set consistently maintains the form of a [complete graph](https://en.wikipedia.org/wiki/Complete_graph) at all times.
+
+### `addDependencies` and `removeDependencies` in `SystemConfigInterop` and `L1BlockInterop`
```
tynes, line 107: This still mentions removal

tynes (approved):

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Spec changes
+The core proposed changes are as follows:
+- **Modify the `SuperchainConfig` contract**: Extend this contract to manage the dependency set of each chain, taking ownership of the `dependencyManager` role for each `SystemConfig`.
+- **Introduce the `SharedLockbox` contract**: This contract acts as an escrow for ETH, receiving deposits and allowing withdrawals from approved `OptimismPortal` contracts. The `SuperchainConfig` contract serves as the source of truth of the lockbox.
+- **Modify the `OptimismPortal`**: To forward ETH into the `SharedLockbox` when `depositTransaction` is called, with the inverse process applying when `finalizeWithdrawal` is called.
+
+### Managing op-governed dependency set through `SuperchainConfig`
+
+The `SuperchainConfig` contract will serve as the single point for managing the dependency set of a cluster and will be managed by the admin as occurs in other L1 OP contracts. This contract assumes the role of `dependencyManager` for every `SystemConfig` contract involved. Since the dependency graph follows a simple dependency, it only stores a mapping (or array) of chains added, e.g. given a `chainId`.
```
tynes, line 52: We could simplify and remove the concept of `dependencyManager` and instead just hardcode it to be  the `SuperchainConfig`. Unclear what the tradeoffs are here

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,192 @@
...
+
+```solidity
+// Mapping from chainId to SystemConfig address
+mapping(uint256 _chainId => ISystemConfig) public systemConfigs;
```
tynes, line 56: Keeping a mapping of chain id to system configs is related to standardizing on how L2s put their chain configs on chain, as mentioned at eth interop forum. My guess is that we will need to use the interface we are defining now to populate a standardized interface in the future

protocol/eth-shared-lockbox.md
```diff
@@ -0,0 +1,192 @@
...
+
+// Current dependency set list
+EnumerableSet.UintSet private _dependencySet;
+```
+
+Adding a new chain can be done by introducing a new function called `addChain`. The process would look as follows:
+
+1. The `updater` calls the function with `chainId` and `SystemConfig`.
+2. An external call is made to each `SystemConfig` from already added chains with the new `chainId`.
+3. The new `SystemConfig` is called to add all previously existing `chainId` from `SuperchainConfig` to itself.
```
tynes, line 66: Would we need to enforce via social that the dependency set is empty before `addChain` is called? Perhaps we keep dependency set size in the `SystemConfig` for this reason, so you can `require(config.depSetSize() == 0)`

tynes (approved):

tynes:
In favor of this as its designed. Nice work!

There is definitely going to be some integration work with OP Stack Contracts Manager and the upgrade process, but nothing sticks out as being blocking based on the latest OPCM + upgrades designs

tynes:
Regarding https://github.com/ethereum-optimism/design-docs/pull/84#discussion_r1828063782, there is some unscoped work around guaranteeing that there is the same respected game type set for all chains, but that can be ignored for now and scoped as part of the proof design.

protocol/eth-shared-lockbox.md
tynes: I think the 2 step process (similar to `StorageSetter` but instead transfer all ether out of the contract) is the most secure process to do this upgrade and it is mentioned in the doc. Decouple schema migration and implementation code always

protocol/eth-shared-lockbox.md
agusduha: We would need to track the dependency set size for each chain in `SystemConfig`, updating it by adding or subtracting in the `addDependency` and `removeDependency` functions. The size check will apply only to the new chain being added at `addChain` in the `SuperchainConfig`. Does this approach make sense?

protocol/eth-shared-lockbox.md
agusduha: The change involves hardcoding the current L1 `SuperchainConfig` address in the `SystemConfigInterop` contract and removing `dependencyManager` as an `initialize` parameter.

Pros:
- Reduces code complexity.
- Simplifies the migration process by minimizing steps, avoiding the need to reset the `initialized` slot.

Cons:
- Reduces flexibility.
- New chains deployed through the OPCM will share the same `SuperchainConfig`, which might not be desirable for clusters outside the Superchain.

About cons, we always tend to think about future clusters but we are not sure if we should do it and we will like to know you vision on this

protocol/eth-shared-lockbox.md
tynes: Yes this makes sense, its double storage on L1 since the set is stored in both sys config + superchain config under this model. Not the end of the world but we generally need a way to make sure that the dependency set size is 0 before the chain joins

protocol/eth-shared-lockbox.md
tynes: Lets enshrine the `SuperchainConfig` rather than having a `dependencyManager`, this will enforce the assumptions in https://github.com/ethereum-optimism/specs/issues/460 and make the core product always be a highly connected graph. It makes no sense having another topology until transitive deps don't need to be ran to fully validate the network. This also lets us remove a role from the `SystemConfig`, we may need to move it to the `SuperchainConfig`

tynes: Depending on the design of OPCM, it may mean that there is 1 OPCM per cluster, if the `SuperchainConfig` is `immutable`

tynes:
This is now good to move on to the spec phase

PR #86, opened 2024-09-19
External Block Production Sidecar

dmarzzz:
**Description**

This document proposes a sidecar to `op-node` for requesting block production from an external party.

**Tests**

No tests have been added, but the Software Maintenance section details a path to incorporating in standard OP Stack e2e testing after an initial trial.

**Additional context**

By decoupling the block construction process from the Sequencer's Execution Engine, operators can tailor transaction sequencing rules without diverging from the standard Optimism Protocol Client. This flexibility allows individual chains to experiment on seqeuncing features, providing a means for differentiation. This minimum viable design also includes a local block production fallback as a training wheel to ensure liveness and network performance in the event of local Block Builder failure.

**Metadata**

- This work started with this [PR in the spec](https://github.com/ethereum-optimism/specs/pull/116) repo but has been moved here as the work is experimental

### DIFF
--- a/protocol/external-block-production.md
+++ b/protocol/external-block-production.md
@@ -1,0 +1,116 @@
+# External Block Production
+
+# Purpose
+
+The purpose of this design-doc is to propose and get buy-in in on a first step towards building the ideal software and API for connecting proposers in the OP Stack (`op-node`) to an external block builder.
+
+# Summary
+
+This document proposes a sidecar to `op-node` for requesting block production from an external party. This sidecar has two roles: 1) obfuscate the presence of builder software from the `op-node` and `op-geth` and 2) manage communication with a block builder and handle block delivery to `op-node`. The first role is achieved via the sidecar forwarding all engine API calls to it's local `op-geth` and delivering 1 block exactly for each block request from `op-node`. The second role is achieved by the sidecar implementing the communication protocol with the builder, including authentication, and payload selection rules.
+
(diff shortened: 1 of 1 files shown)


tynes (commented):
Generally in favor of this approach. There is a thought that eventually we will want to embed the EL and CL in a single binary when we are trying to push the gas limits and block times. Using the proxy approach works great for now, enables teams to work in parallel without blocking each other on code review/merge

Would be super curious to see if you have a link to the proxy that you are working on. This would be really nice to include the proxy and config to spin up the system in the kurtosis OP Stack package so that people are aware that it exists and can play around with their own builders on a devnet

ferranbt:
We have an experimental version in Go [here](https://github.com/ferranbt/optimism/tree/ferranbt/experiment-sidecar/op-sidecar) and a more production-ready version in Rust [here](https://github.com/flashbots/rollup-boost).

dmarzzz:
Additionally our goal with the rust version is to migrate it to a reth module/extension so that should totally work in the single binary world in the event this isn't merged upstream by then

tynes (approved):

odyslam:
Hey @dmarzzz @tynes we are building our solutions based on the experimental Local Builder API PR, assuming that it would get merged eventually.

Does that mean that the sidecar approach is the preferred method for now instead of the other spec? If yes, that means we should change our approach to interface with the proxy here and not expecting the API support, is that correct?

PR #87, opened 2024-09-19
Governor Interop Design Document 

corydickson:
Description

This document outlines changes to the Governor contract to support interop voting. With the additional changes and considerations outlined from @skeletor-spaceman and other members that include possible alternative solutions.

### DIFF
--- a/governance/interop-governor.md
+++ b/governance/interop-governor.md
@@ -1,0 +1,138 @@
+# Purpose
+
+The main goal of this document is to address the changes to the `Governor` [contract](https://github.com/ethereum-optimism/specs/blob/main/specs/governance/governor.md) to support the
+capability of handling cross-chain partial delegation utilizing the interop protocol. As outlined in previous [specifications](https://github.com/ethereum-optimism/specs/pull/285)
+and design [documents](https://github.com/ethereum-optimism/design-docs/blob/main/governance/advanced-delegation.md), there is an ask to support tracking of voting power across several
+L2's once the `GovernanceToken` and `GovernanceDelegation` contract (fka. Alligator) predeploys become available. So far, the considerations around handling this use-case require
+changes to the `Governor` contract itself to properly support the checkpointing system on OP mainnet so that token holders can delegate their votes cross-chain.
+
+# Summary
+
+Below you will find an outline of changes to the functionality of the governor as well as the stated purpose of such a mechanism to support vote aggregation. It forwards a solution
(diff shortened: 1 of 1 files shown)


gotzenx:
After talking with different stakeholders we have proposed a new and simpler path as the first version of GovToken that should respect the following points:
- We will only make **one and only** upgrade to the OP Token to make it interoperable.
- This MVP version should only move the **current delegation logic** to an external contract separated from the token contract.
- Delegation contract needs to be compatible to be upgraded into it's interop evolution.
- Interop governance will **not** be part of this MPV.
- No change is required in snapshot key now.
- Delegation and Voting will **only** happen on OP-mainnet for now.
- Nothing from future iterations must force a token upgrade.

Any extra features like partial-delegation or interop-delegation are not a priority right now and will happen on future iterations after the token upgrade.

This current design doc aligns with the interop-governance version that will come after the MVP.
So we suggest to have a new doc (or modify this one) that reflects the MVP version, and keep this one as a future version.

kent:
Just got off a call between Agora <> Wonderland and we've got some news and a great path forward.

**Decision 👩‍⚖️:**  Focus on Interop Gov Token for liquidity unlock and single chain governance with a separate delegation contract to keep upgrades easier. This will allow Governance to function well and de-risk any token upgrade work.

## Plan
**Step 1**: Ship Interop Gov Token and make sure it's compatible with single chain, OP Mainnet Governance as the V1 for Devnet 2 in 2024. 
**Step 2:** Kick-off more in depth design sessions around how to bring Interop Governance to OP in Q4 2024
**Step 3:** Ship Interop with Single Chain Governance to OP Mainnet to get the liduidty unlock that @k-ho / Zain did such a good job pitching last week
**Step 4**: Build and ship Interop Multi-chain Governance in 2025

### Pros
Simpler rollout for 2024
More time to consider complex multichain governance architectures without holding up token upgrade
One token upgrade

### Cons:
Takes us a bit longer to get to the multi-chain governance promise land but gives us more time to measure twice and cut once and bring in more governance stakeholders.

## Division of labour:
Interop Gov Token & Token Upgrade Logistics -> Wonderland
External Delegation Contract, Updated Governor, Support for Partial Delegation -> Agora

We will be tightly collaborating and reviewing each others specs / implementation so that we can land the plane together. 

## Next actions on this PR

@gotzenx + @corydickson + @cairoeth ->  let's keep this one open and then we can come back to it. 

We will also create docs for:
* GovernanceDelegation Contract
* OP Governor w/ Interop Token and External GovernanceDelegation

See link to discord for full convo too: https://discord.com/channels/1244729134312198194/1253112573377904682/1286462075484377139

PR #89, opened 2024-09-23
 Change `DeputyGuardianModule` to use a Proxy to be `DeputyGuardianModuleProxy`

Ethnical:
Introduce a proposal that will prevent to regenerate the PSPs and make some unnecessary changes to the guardian contract each time we update the `DeputyGuardianModule` code. 
As discussed [here on discord.](https://discord.com/channels/1244729134312198194/1278362089597046846/1286684655407464498)

### DIFF
--- a/security/DeputyGuardianModule-Proxy.md
+++ b/security/DeputyGuardianModule-Proxy.md
@@ -2,10 +2,11 @@
 
 ## Context
 
-The `DeputyGuardianModule` is an important component for the Superchain security model.
-The `DeputyGuardianModule`, also known as _DGM_, is a smart contract deployed on L1 (currently [here](https://etherscan.io/address/0xc6901f65369fc59fc1b4d6d6be7a2318ff38db5b)) that has, for example, the ability to widely pause the Superchain.
-The _DGM_ is owned by the _Foundation Operation Safe_ so only operation run from the _Foundation Operation Safe_ can be executed into the _DGM_.
-To make sure we can widely pause the Superchain rapidly enough in case of an emergency, we need to presign the pause transaction from the signers of the _FoS_. These transactions are also known as `PSP` for PreSigned Pause transactions.
+The `DeputyGuardianModule` is an important component for the Superchain security model. \
+The `DeputyGuardianModule`, also known as _DGM_, is a smart contract deployed on L1 (currently [here](https://etherscan.io/address/0xc6901f65369fc59fc1b4d6d6be7a2318ff38db5b)) that has, for example, the ability to widely pause the Superchain. \
+The `DGM` is owned by the _Foundation Operation Safe_ so only operation run from the _Foundation Operation Safe_ can be executed into the `DGM`. \
+To make sure we can widely pause the Superchain rapidly enough in case of an emergency, we need to presign the pause transaction from the signers of the _FoS_. \
+These transactions are also known as `PSP` for PreSigned Pause transactions. \
 This process requires a ceremony and this is slow and tedious to put in place.
 
 The list of available actions that the `DeputyGuardianModule` can execute are:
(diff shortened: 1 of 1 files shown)


Inphi (approved):
LGTM. Pretty slick.

It appears that we'll never have a need to destroy or replace a PSP. If not, when could that happens?

ajsutton (approved):
This makes a lot of sense to me - we should be able to modify the deputy guardian module without invalidating PSPs.  Ensuring only the security council can perform that upgrade preserves the security properties of the system so the deputy guardian is still restricted in the actions they can perform and the security council is in control.

Ethnical:
> It appears that we'll never have a need to destroy or replace a PSP. If not, when could that happens?
@Inphi Good point, I will highlight the way to have invalidated PSPs into the new system.
1. The first cause of invalidated the PSPs is  _Nonces of the FoS increase too much_  so made monitoring for this [op-defender](https://github.com/ethereum-optimism/monitorism/blob/d14768998fcc64512ba7cd87603b1143f7ca4dbe/op-defender/psp_executor/simulator.go#L71) and oncall should be aware about it.
2. If there is breaking changes on the smart contracts on L1. For example, the SuperchainConfig is updated. Or the Guardian overwrite the list of the modules that can execute action. Really unlikely, but worth noting.

security/DeputyGuardianModule-Proxy.md
```diff
@@ -0,0 +1,63 @@
...
+We are clearly seeing that this is not a sustainable solution for the long term. 
+
+Additionally, we have to simulate the new PSPs and share them with other member of the Superchain.
+
+## Proposed Solution
+
+![image](https://github.com/user-attachments/assets/099258f4-19fb-44b8-84e9-670b78a2f868)
+
+Here we propose some modification of the architecture with a proxy contract on top of the `DeputyGuardianModule`, this will allow us to upgrade the `DeputyGuardianModule` and keeping the same address.
+Thus, this will not invalidate the PSPs and we will not need to regenerate new PSPs each time we upgrade the `DeputyGuardianModule`.
```
maurelian, line 43: It will invalidate the PSPs when we replace the DGM with the proxy though right? Perhaps we should wait until we need to upgrade the DGM for another reason, and then move to a proxy at the same time.

security/DeputyGuardianModule-Proxy.md
```diff
@@ -0,0 +1,63 @@
...
+
+The owner of the proxy contract `DeputyGuardianModuleProxy` would be the `Security Council` as the `DeputyGuardianModule` is doing action into the `Guardian` contract that is owned by the `Security Council`.
```
maurelian, line 45: We should think a bit more about whether the proxy owner should be: 
1. The L1 ProxyAdmin: This has the benefit of an upgrade path that is most like the other contracts.
2. The 2 of 2: This has the same upgrade auth as other contracts, but a different path as it skips the L1 ProxyAdmin. I don't like it. 
3. The Security Council: This would makes sense as the Security Council should have full control over the DGM, but it can already disable/replace the DGM via the Safe's `ModuleManager` functionality.

security/DeputyGuardianModule-Proxy.md
```diff
@@ -0,0 +1,63 @@
...
+## Risks & Uncertainties
+
+### Security Model
+
+We don't see any particular risk with this current design for the long term as this is a common pattern to use a proxy contract for upgrading a contract.
+
+A potential issue can occur with the PSPs coverage.
+During the upgrade the PSPs are going to be invalid so we need to make sure to presign new PSPs with the new `DeputyGuardianModuleProxy` address and simulate the PSPs accordingly.
+
+Now, we are using a proxy executing the same actions than before will cost a more gas. We should ensure there no previous call that can run out of gas (really unlikely).
```
maurelian, line 63: What do you mean by "previous call that can run out of gas"?

security/DeputyGuardianModule-Proxy.md
Ethnical: > It will invalidate the PSPs when we replace the DGM with the proxy though right? 

This is correct. 
> Perhaps we should wait until we need to upgrade the DGM for another reason

Yes this is the idea! With [_Incident Response Improvements_](https://github.com/ethereum-optimism/design-docs/issues/120) We will do the upgrade during the improvements of our incident response SC on L1. 
I should write this down somewhere.

security/DeputyGuardianModule-Proxy.md
Ethnical: > previous call that can run out of gas?

First, this should **not** happen (this is extremely unlikely). 
But, as I don't know the full components used with the DGM, it is possible that some hardcoded call are made with a certain amount of gas to the DGM (with automated code for switching automatically to certain game type or automatically call `pause()` or `unpause()` like Automated Pause or automated game switcher).
Since the `delegatecall` will now be replacing the previous normal call into the DGM as this is a proxy this will be more expensive and can lead to DoS if there is **hardcoded gas limit** in some code that called the DGM methods automatically.

security/DeputyGuardianModule-Proxy.md
Ethnical: >  it can already disable/replace the DGM via the Safe's ModuleManager functionality.

I think we should discuss about it during the review, I am not sure to follow this point there.

maurelian: Decision: we will make it the 2of2 (L1 ProxyAdmin owner), this prevents us from needing a separate upgrade transaction for the DGM, without weakening the Security Council's ability to remove/replace the module.

mds1: Sorry, I missed this design review so not sure if this was discussed—Do we need to run this decision by @sammcingvale or L2Beat to verify there's no reason it might cause a change in stage 1 status? Intuitively it feels like the SC is the right owner, since it's the SC's module, and the rationale is we're choosing the 2 of 2 to simplify operations by reducing the number of playbooks?

Ethnical: Good point @mds1! 
I didn't thought about the L2Beat. We should definitely make sure we are not going in the correct decision will reach @sammcingvale in the meanwhile.

Ethnical:
I have merged the change for the gap PSPs: 
https://github.com/ethereum-optimism/superchain-ops/pull/347
Also, Eli told us this is not suppose the cause issue with l2beat so we are good to merge this final doc review.

PR #93, opened 2024-09-25
Add opcm test integration

maurelian:
Outlines an approach to integrating the OPCM into the `Deploy.s.sol` script used in the forge test suite.

### DIFF
--- a/protocol/opcm-test-integration.md
+++ b/protocol/opcm-test-integration.md
@@ -42,10 +42,12 @@ A plausible PR stack for achieving this might be:
     Chain that it deploys. That needs to be fixed to prepare for the OPCM.
 3. Deploy the Superchain contracts using the OPCM's `DeploySuperchain.s.sol`. With the previous PRs,
     this will hopefully be straight-forward.
-4. Deploy the OP Chain Contracts using the `OPContractsManager` contract. This will be much simpler
-    if we can delete the `OptimismPortal`, `L2OutputOracle` and `useFaultProofs` flag. Otherwise we
-    will accomplish this using two entirely distinct flows depending on the value of
-    `useFaultProofs`:
+4.  Replace `deployImplementations()` and `initializeImplementations()` with the equivalent OPCM's
+    `DeployImplementations`.
+5. Deploy the OP Chain Contracts using the `OPContractsManager` contract (replacing
+    `deployProxies()` with the equivalent OPCM's `DeployOpChain`). This will be much simpler if we
+    can delete the `OptimismPortal`, `L2OutputOracle` and `useFaultProofs` flag. Otherwise we will
+    accomplish this using two entirely distinct flows depending on the value of `useFaultProofs`:
       1. `true` -> deploy using OPCM
       2. `false` -> deploy using the existing `deploy{ContractName}()` functions. It maybe possible
          to reuse some of the contracts deployed by the OPCM which do not vary pre and post fault
--- a/protocol/opcm-test-integration.md
+++ b/protocol/opcm-test-integration.md
@@ -1,0 +1,130 @@
+# Direction for OPCM test integration
+
+## Purpose
+
+In order to integrate the OPCM into the forge test setup (ie. `Deploy.s.sol`), we must first
+simplify the deploy scripts while keeping the desired properties and removing what we don't need.
+
+## Goals
+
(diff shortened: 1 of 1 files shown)


protocol/opcm-test-integration.md
```diff
@@ -0,0 +1,130 @@
...
+
+Nice to haves:
+- logs emitted during deployment:
+  - This is not essential, but comparing before and after logs is a quick and dirty way to ensure
+    that the config is roughly the same.
+- Deploying contracts to the same address as before.
+  - This might be achievable based on the use of
+    [`create2`](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/scripts/deploy/Deploy.s.sol#L1703)
+    in the existing scripts. If so it would enable us to use statedumps to diff the before and after
+    deployments.
```
mds1, line 28: I think this one is unlikely, because get deployed through OPCM and not from the test contract. However it is possible we can at least achieve this from the superchain and implementation deploys, so create2 with any salt can be used

protocol/opcm-test-integration.md
```diff
@@ -0,0 +1,130 @@
...
+    the `ProxyAdmin` While maintaining the `viaSafe()` helper methods (they are also needed in
+  `DeployOwnership` for now), we should remove all usages in `Deploy.s.sol`, perform all actions
+    through the deployer key, then finally transfer the `ProxyAdmin` to the `finalSystemOwner`
+    (which the current scripts don't even do at the moment).
+2. Deploy a separate `ProxyAdmin` contract for the Superchain. A key diff between the OPCM approach
+    and `Deploy.s.sol` is that the latter shares the same `ProxyAdmin` for the Superchain and the OP
+    Chain that it deploys. That needs to be fixed to prepare for the OPCM.
+3. Deploy the Superchain contracts using the OPCM's `DeploySuperchain.s.sol`. With the previous PRs,
+    this will hopefully be straight-forward.
+4. Deploy the OP Chain Contracts using the `OPContractsManager` contract. This will be much simpler
```
mds1, line 45: Is there a missing `DeployImplementations` step in between here?

protocol/opcm-test-integration.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Appendix
+
+The following content was created while investigating the current test setup flow, and
+is provided here for reference.
+
+##### Summary of the current Test Setup
+
+The Test contracts (`UltimateTestHarness` below) inherit from `CommonTest`, and implement a
+`setUp()` method which calls to `CommonTest.setUp()`. This results in an approximate call flow of:
```
mds1, line 64: What does "approximate" mean here?

maurelian: Just that I've distilled it down to the most important details and added my own annotations which would not appear in a normal trace.

protocol/opcm-test-integration.md
maurelian: Yes, good call out. I'm admittedly skipping a bit of detail by step 4 here. Will add that in.

mds1 (approved):
Overall this plan LGTM, will be nice to clean up the legacy deploy script and replace it with OPCM

protocol/opcm-test-integration.md
maurelian: We could potentially make a small change to set the existing deployer into the same address as the OPCM will eventually go. But I don't think that is worthwhile unless we get really concerned about the ability to compare the before and after deployment.

mds1: Ah true, that's a good idea. Alternatively we can etch OPCM into the existing deployer address. Foundry now has a `copyStorage` cheat that can ensure all storage gets migrated also: https://github.com/foundry-rs/forge-std/blob/5a802d7c10abb4bbfb3e7214c75052ef9e6a06f8/src/Vm.sol#L1552-L1553

protocol/opcm-test-integration.md
maurelian: 8d9901144613

protocol/opcm-test-integration.md
maurelian: Nice. Will see about that as I get into it.

tynes:
Merging this as its been approved

PR #95, opened 2024-10-01
Add FMA SuperchainERC20

Joxess:
**Description**

This PR introduces the FMA for `SuperchainERC20`. It contains the analysis for the `SuperchainERC20` standard, the changes introduced, including the work done for migrated liquidity (`OptimismSuperchainERC20`). Additionally, it expands on potential scenarios where `SuperchainERC20` could be used for custom functionalities, highlighting the potential restrictions.

### DIFF
--- a/security/fma-superchainerc20.md
+++ b/security/fma-superchainerc20.md
@@ -1,0 +1,147 @@
+## **SuperchainERC20 FMA (Failure Modes and Recovery Path Analysis)**
+
+| Author | Particle, Skeletor, Joxes |
+| --- | --- |
+| Created at | 2024-09-30 |
+| Needs Approval From | Matt Solomon, Mark Tyneway |
+| Other Reviewers |  |
+| Status | Review |
+
+## Introduction
+
+---
+
+This document is intended to be shared publicly for reviews and visibility purposes. It covers the following:
+
+1. Introduction of the `SuperchainERC20` standard:
+    - Introducing the `SuperchainERC20` baseline implementation.
+    - Introducing the `SuperchainERC20Bridge`.
+2. Introduction of the liquidity migration implementation for L1 bridged tokens as a `SuperchainERC20` extension:
+    - Updating the `L2StandardBridge` to allow token conversion.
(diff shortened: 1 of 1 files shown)


security/fma-superchainerc20.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Message unable to finalize in Cross-Chain Transfers
+
+- **Description:** During a cross-chain transaction, tokens are burned on the source chain via `sendERC20` and minted on the destination chain via `relayERC20` through `SuperchainERC20Bridge` and utilizing the `L2toL2CrossDomainMessenger`. However, if the destination chain doesn’t listen to messages from the origin chain, the message will never be included in `CrossL2Inbox` and therefore the tokens will be burned on the origin but not minted on the destination. This results in users’ funds being stuck or perceived as lost.
+- **Risk Assessment**: High
+    - Potential impact: High. Users may lose access to their funds, indefinitely.
+    - Likelihood: Medium. Currently there is no way to stop a cross-chain transaction whose destination doesn’t include the origin chain in their dependency set.
+- Mitigation: To be defined at the architecture level. In the meantime, user interfaces should prevent users from initiating a message that cannot be relayed.
+- **Detection**: Monitor cross-chain message queues and maintain an updated list of chain dependencies to identify potential incompatibilities proactively.
```
mds1, line 53: What does "To be defined at the architecture level" mean? Similarly for the detection, what tooling are we using and how exactly would it detect incompatibilities?

These are just sample questions to get at some general guidance for mitigations and detections—we typically want to be as detailed as possible, so if this failure mode occurs and we perform a retro, we can look back at the FMA and clearly identify why our mitigations or detections were insufficient.

We should also link to specific mitigations, for example if we say "we have tests to mitigate this", or "we have a runbook for this recovery path", let's link to those artifacts. If those tests or runbooks are not yet written, you would add an action item to the FMA to write them and link them afterwards

Joxess:
After discussion with teams, we're gonna reduce the scope for this FMA and make a PR with the new design doc.

PR #97, opened 2024-10-01
feat: standard l2 genesis

tynes:
**Description**

Add design doc for standard L2 genesis

Tracking ticket: https://github.com/ethereum-optimism/optimism/issues/12302


protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,128 @@
...
+The specs can be found [here](https://github.com/ethereum-optimism/specs/tree/17ef36cdc3bb9893b206a93464122d56730d30fb/specs/protocol/holocene).
+
+Similar to the L1 MCP project, we move all of the network specific configuration out of the individual contracts
+themselves and instead place all of it in a single place. The contracts will make a `CALL` rather than using
+`sload` to read the values. These values will be sourced from L1 via deposit transactions that come from
+the `SystemConfig.initialize` call. We need to make sure that the max deposit gas limit is at least able to
+fullfill these deposit transactions.
+
+The general flow is as follows:
+
```
tynes, line 61: One downside to this approach is that it expands the ABI of the `L1Block` contract, causing every system transaction to consume a bit more gas due to 4 byte selector comparisons in the jumptable. The alternative would be to introduce a new contract that stores these values. Custom gas token is in the `L1Block` contract already, we cannot remove that ABI due to modifications of the `WETH` predeploy being used in production, the `WETH` predeploy is not proxied so it cannot be updated easily

maurelian: In order to avoid that, we could just add:

```
function set(string _valueName, bytes32 value);
function get(string _valuename);
```

This will allow us to keep the ABI constant and still easily add new values.

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,128 @@
...
+consensus on using the `DEPOSITOR_ACCOUNT` as the owner.
+
+When we do a contract release, we commit to that bytecode as part of consensus. That is the bytecode used
+with deposit transactions doing upgrades to the network. In the L2 genesis creation script, we could have
+a library for each release of the predeploys. The genesis script would take the bytecode from the library if
+configured for a specific hardfork at genesis, otherwise it would use the compiled source code. This gives
+us a lot of flexibility and simplicity when it comes to being able to recreate an L2 genesis deterministically.
+
+The block history integrity check becomes as simple as observing a 32 byte state root matches in the genesis 
+block matches the expected value.
```
maurelian, line 91: To be clear, this means that every L2 will have the same initial output root, prior to the first deposit transactions.

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Resource Usage
+
+<!-- What is the resource usage of the proposed solution?
+Does it consume a large amount of computational resources or time? -->
+
+The additional deposit gas is the only additional resource usage and its covered in the risks section
+at the bottom of this document.
```
maurelian, line 99: I'm not terribly concerned about it, but what about the additional cost of retrieving these values?

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,128 @@
...
+is likely too low level. -->
+
+A WIP implementation can be found [here](https://github.com/ethereum-optimism/optimism/pull/12057).
+The specs can be found [here](https://github.com/ethereum-optimism/specs/tree/17ef36cdc3bb9893b206a93464122d56730d30fb/specs/protocol/holocene).
+
+Similar to the L1 MCP project, we move all of the network specific configuration out of the individual contracts
+themselves and instead place all of it in a single place. The contracts will make a `CALL` rather than using
+`sload` to read the values. These values will be sourced from L1 via deposit transactions that come from
+the `SystemConfig.initialize` call. We need to make sure that the max deposit gas limit is at least able to
+fullfill these deposit transactions.
```
mds1, line 58: Do we also need to increase `l1InfoGasOverhead`?

https://github.com/ethereum-optimism/op-geth/blob/d5a96613c22bc46238a21d6c0f805399c26c9d4c/core/txpool/validation.go#L37

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+
+The `data` and `gasLimit` are allowed to be specified since we don't fully know what sorts of calls we may have to do.
+We may only need to do simple `upgradeTo` calls, but we may also need to do `upgradeToAndCall`. To support the
+[liquidity migration](https://github.com/ethereum-optimism/design-docs/blob/4b62eb12eceb8e4867ac101134730102c0f5a989/protocol/superchainerc20/liquidity-migration.md), we need to backport storage slots into the `OptimismMintableERC20Factory`
+contract. We may need to introduce multicall support into the `L2ProxyAdmin` as part of this.
+
+#### L2ProxyAdmin
+
+A new contract exists called the `L2ProxyAdmin`, it simply inherits from the `ProxyAdmin` and overrides the
+`owner()(address)` function to return `DEPOSITOR_ACCOUNT`.
```
maurelian, line 118: to be clear, it's a new implementation, but the L2ProxyAdmin Proxy will still be at the same predeploy address `0x4200000000000000000000000000000000000018` right?

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+The block history integrity check becomes as simple as observing a 32 byte state root matches in the genesis 
+block matches the expected value.
+
+### Rationale Behind Certain Changes
+
+#### SuperchainConfig "Upgrader" Role
+
+This new role is to allow for operational flexibility going into the future. If its not desired to add this role,
+we could make the `guardian` able to faciliate the upgrade transactions. In practice, we may end up setting both
+the `guardian` and the `upgrader` to the same account for the Superchain.
```
maurelian, line 99: I think its unlikely we'll want to set these to the same value, as the Guardian should not be able to make upgrades.

But I really like having a new `upgrader` role, and a positive future with that added would be for the L1 ProxyAdmin to call `superchainConfig.upgrade()` rather than being `Ownable`.

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+
+This is taken from the [specs](https://github.com/ethereum-optimism/specs/blob/17ef36cdc3bb9893b206a93464122d56730d30fb/specs/protocol/holocene/predeploys.md) and misses the `L2ProxyAdmin`. The `L2ProxyAdmin` must also be deterministic
+and is explored in the following issue: https://github.com/ethereum-optimism/specs/issues/388. There is general
+consensus on using the `DEPOSITOR_ACCOUNT` as the owner.
+
+When we do a contract release, we commit to that bytecode as part of consensus. That is the bytecode used
+with deposit transactions doing upgrades to the network. In the L2 genesis creation script, we could have
+a library for each release of the predeploys. The genesis script would take the bytecode from the library if
+configured for a specific hardfork at genesis, otherwise it would use the compiled source code. This gives
+us a lot of flexibility and simplicity when it comes to being able to recreate an L2 genesis deterministically.
```
mds1, line 88: Before committing to this, we should sync with slipper when he's back. In practice most chains will use op-deployer to generate their genesis, and I believe that would pull in the L2 contract bytecode of a release from a published docker image for that release

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+  L2StandardBridge -- "l1StandardBridge()(address)" --> L1Block
+  L2ERC721Bridge -- "l1ERC721Bridge()(address)" --> L1Block
+  OptimismMintableERC721Factory -- "remoteChainId()(uint256)" --> L1Block
+  end
+  OptimismPortal -- "setConfig(uint8,bytes)" --> L1Block
+```
+
+This is taken from the [specs](https://github.com/ethereum-optimism/specs/blob/17ef36cdc3bb9893b206a93464122d56730d30fb/specs/protocol/holocene/predeploys.md) and misses the `L2ProxyAdmin`. The `L2ProxyAdmin` must also be deterministic
+and is explored in the following issue: https://github.com/ethereum-optimism/specs/issues/388. There is general
+consensus on using the `DEPOSITOR_ACCOUNT` as the owner.
```
mds1, line 82: Why is the L2ProxyAdmin not included in this design doc? It seems like a key part of Standard L2 Genesis, because without that we won't have identical genesis due to the L2 L2ProxyAdmin storage that is set at genesis

protocol/standard-l2-genesis.md
tynes: This would only reduce the dependency of having the L1 deployments before and change nothing else about the op-deployer flow. What would you like to hear from Slipper specifically?

protocol/standard-l2-genesis.md
tynes: Yes exactly

protocol/standard-l2-genesis.md
tynes: In the specs it is still called `ProxyAdmin`, i am open to renaming it to be `L2ProxyAdmin`

protocol/standard-l2-genesis.md
tynes: Yes exactly

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+to source this config, it would likely need to be hardcoded in the binary and that isn't super scalable.
+
+# Risks & Uncertainties
+
+<!-- An overview of what could go wrong.
+Also any open questions that need more work to resolve. -->
+
+There is a concern that the sequencer can include transactions before these values are set on L2.
+If we define the `SystemConfig.startBlock` as the [first block to start derivation in](https://github.com/ethereum-optimism/optimism/blob/d05fb505809717282d5cee7264a09d26002a4ddd/op-node/cmd/genesis/cmd.go#L174C30-L174C40),
+which is set on `SystemConfig.initialize` and also the deposit transactions that set these values are
```
maurelian, line 177: What options do we have for enforcing this property? ie. Is there a good code path that would reject a sequenced tx if these values are not set? Or perhaps the op-node can just read the fork block height from `SystemConfig.startBlock`?

protocol/standard-l2-genesis.md
mds1: Ah sorry I don't care much about the name, I just wasn't sure if the ProxyAdmin on L2 was in scope here. It looks like you just pushed an update though with a new `Rationale Behind Certain Changes` section I need to read, so maybe that ProxyAdmin is now in scope

protocol/standard-l2-genesis.md
tynes: Ah this is a great idea

protocol/standard-l2-genesis.md
mds1: > This would only reduce the dependency of having the L1 deployments before and change nothing else about the op-deployer flow.

Sorry, not following this sentence 😅

But basically I want to make sure we have a single canonical way of generating genesis. So assuming the genesis is done via op-deployer, you could either:
- Fetch bytecode from docker and use it (which I think is what the plan is), or
- Fetch bytecode from libraries in the repo that do not change

protocol/standard-l2-genesis.md
mds1: Like @maurelian said this should not be the guardian—general (but I think imperfect) principal is that the guardian is responsible for liveness, not safety (which includes contract upgrades)

This role will likely have to be set to the same role as the L1 ProxyAdmin Owner (i.e. 2/2 with Security Council) to meet stage 1 requirements.

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+
+A new contract exists called the `L2ProxyAdmin`, it simply inherits from the `ProxyAdmin` and overrides the
+`owner()(address)` function to return `DEPOSITOR_ACCOUNT`.
+
+Ideally we can remove the need for legacy proxy types since they don't exist on L2 eventually, but
+that is considered a bonus when we do get around to it.
+
+#### SystemConfig
+
+The `SystemConfig`
```
mds1, line 125: Incomplete section

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+Ideally we can remove the need for legacy proxy types since they don't exist on L2 eventually, but
+that is considered a bonus when we do get around to it.
+
+#### SystemConfig
+
+The `SystemConfig` 
+
+#### Initializable Predeploys Removed
+
+All predeploys are no longer initializable. This allows for upgrades issued by deposit transactions to be very smooth.
```
mds1, line 129: > All predeploys are no longer initializable

Good candidate for a semgrep rule

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+- `ERC721Bridge`
+
+#### CrossDomainMessenger
+
+Since the `CrossDomainMessenger` is no longer `initializable` we need to slightly modify the semantics around
+the `xDomainMsgSender`. There is actually no need to set the value in storage during `initialize`, we could modify
+the semantics such that if its `address(0)` in storage, then return the default value, otherwise return the
+actual sender value. This should be safe since there is no way to be a sender from `address(0)`.
+
+Given this insight and the fact that there is reentrency check on `relayMessage`, it should be safe to use transient
```
mds1, line 143:
```suggestion
Given this insight and the fact that there is reentrancy check on `relayMessage`, it should be safe to use transient
```

protocol/standard-l2-genesis.md
tynes: Ah I see the difference we are talking about now. The solution that I propose around hardcoding the released bytecode into a library just makes generating historical genesis easy on any given commit. This prevents implicit deps, like foundry changing something and breaking the old script. Since we put the bytecode into Go code as part of consensus, it makes sense to put it in solidity too. The artifacts will contain the bytecode as well, the way the `L2Genesis` script is called should have minimal change if any. But I see your point with regards to looping in Slipper now, I don't think we need to commit to this when coming to consensus on the overall design.

tynes:
I am adding more information, thank you for the review so far

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+
+Given all of the work for making releases and upgrades nice for the L1 contracts and the client software,
+it is at a waste if we cannot also have good releases of the L2 predeploys.
+
+Right now, OP Mainnet is running contracts at various versions of the software. It is actually very difficult
+to reproduce the exact OP Mainnet set of contracts being used. It would require cherry picking bytecode from many
+different commits. We run no tests against this particular combination of contracts. We believe it is safe
+given our development practices, but every time that we do want to do a release it results in a lot of time
+being spent trying to make sure that we are upgrading to a compatible set of contracts.
+
```
tynes, line 42: How do we do releases? The tests need to cover client versions, L1 contract versions and L2 contract versions

maurelian: Can you say more about your thoughts behind this comment?

While the change outlined here will get us closer to having the onchain L1/L2 contract versions in sync with a single monorepo commit, IMO this doc should only be concerned with how L2 contracts are upgraded.

protocol/standard-l2-genesis.md
maurelian: It looks like the graph below and the PR reflect this. Resolving.

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+If more information is needed on the costs of the problem,
+this is a good place to that information. -->
+
+There is currently no good way to do releases of L2 predeploys. Historically, chains
+have launched with an arbitrary commit for their L2 genesis, making the block history integrity
+checks that are part of the superchain registry very difficult. We tell chains that are
+trying to launch to use governance approved L1 contracts, the same should apply for L2.
+
+Given all of the work for making releases and upgrades nice for the L1 contracts and the client software,
+it is at a waste if we cannot also have good releases of the L2 predeploys.
```
maurelian, line 35:
```suggestion
it is all a waste if we cannot also have good releases of the L2 predeploys.
```

protocol/standard-l2-genesis.md
maurelian: @tynes why do we need to benchmark?

I think @mds1 is correct here: 

> ... we do not since these are dedicated `setConfig` calls

ie. we are not calling `setL1BlockValues()` or `setL1BlockValuesEcotone()`, and are simply emitting a `TransactionDeposited()` event, so the upgrade call is only limited by the [`maxResourceLimit`](https://github.com/ethereum-optimism/optimism/blob/feat/holocene-contracts/packages/contracts-bedrock/src/L1/ResourceMetering.sol#L36).

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+the `guardian` and the `upgrader` to the same account for the Superchain.
+
+Open to removing this from the spec and just using the `guardian` if preferred, but the possible operational
+flexibility seems useful.
+
+The `upgrader` role can call the `OptimismPortal.upgrade(bytes memory data, uint32 gasLimit)` function
+and it emits a deposit tx from the `DEPOSITOR_ACCOUNT` that calls the `L2ProxyAdmin`. Sourcing the auth
+from the `SuperchainConfig` allows for simple management of this very important role, given that it impacts
+stage 1 status. This is meant to simplify operations, ie remove the concept of the aliased L1 `ProxyAdmin` owner
+being set as the L2 `ProxyAdmin`.
```
maurelian, line 108: Just to be clear, this means that for each predeploy being upgraded, we'll need to make N calls to each OptimismPortal in the Superchain?

protocol/standard-l2-genesis.md
maurelian: > hardcoding the released bytecode into a library

This will need to be documented in the release process

tynes:
Here is a link to the audit plan for this feature: https://www.notion.so/oplabs/Isthmus-Smart-Contract-Audit-113f153ee16280448ebbd1b628c26ff2?pvs=4

protocol/standard-l2-genesis.md
tynes: Its way too big of a lift to add a new event source and we want to reduce number of event sources rather than increase, ideally there is a single event source and it can be observed statelessly, see https://github.com/ethereum-optimism/specs/issues/330. The more event sources required, the more complex the proof is

protocol/standard-l2-genesis.md
tynes: We can leave other things out of scope for this doc but generally we need to solve the problem of knowing which releases work together and be able to test different releases together as one unit

protocol/standard-l2-genesis.md
tynes: Its not safe to not meter any deposit tx until we have deposit queue since it could be possible to create a block that uses more than the block gas limit. Technically you could not meter up to a certain point but what that point is is related to the chain config which is dynamic and potentially arbitrary up to certain points

protocol/standard-l2-genesis.md
mds1: Our release process is definitely getting complex. Related: https://www.notion.so/oplabs/Our-Release-Process-Is-Like-This-Title-A-Wee-Bit-Too-Long-and-Complex-a63f8a71f8454401b69b4877a9ad42df?pvs=4

We should consider a follow up design doc to automate as much as possible into a minimal set of `just commands`

protocol/standard-l2-genesis.md
maurelian: got it.

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+
+The `data` and `gasLimit` are allowed to be specified since we don't fully know what sorts of calls we may have to do.
+We may only need to do simple `upgradeTo` calls, but we may also need to do `upgradeToAndCall`. To support the
+[liquidity migration](https://github.com/ethereum-optimism/design-docs/blob/4b62eb12eceb8e4867ac101134730102c0f5a989/protocol/superchainerc20/liquidity-migration.md), we need to backport storage slots into the `OptimismMintableERC20Factory`
+contract. We may need to introduce multicall support into the `L2ProxyAdmin` as part of this.
```
maurelian, line 113: > We may need to introduce multicall support into the `L2ProxyAdmin` as part of this.

Would that be in scope here?

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+There is a concern that the sequencer can include transactions before these values are set on L2.
+If we define the `SystemConfig.startBlock` as the [first block to start derivation in](https://github.com/ethereum-optimism/optimism/blob/d05fb505809717282d5cee7264a09d26002a4ddd/op-node/cmd/genesis/cmd.go#L174C30-L174C40),
+which is set on `SystemConfig.initialize` and also the deposit transactions that set these values are
+sent in the same block, then we should have the guarantee that no user transactions are included before
+the deposit transactions.
+
+There is a concern around the max deposit gas limit being too small so that the `SystemConfig` cannot
+deposit all of these values, we should have logic that reverts if the deposit gas limit is too small
+in the `ResourceConfig` sanity check function. Since the `ResourceConfig` can be modified during
+`initialize`, its not that big of a deal and chain operators will see that they need to increase that
```
maurelian, line 184: > Since the `ResourceConfig` can be modified during `initialize`, its not that big of a deal

Is the idea that it can be increased in the event that the cost of the setConfig() calls becomes to large?

> and chain operators will see that they need to increase that

how do chain operators do that, and why do they need to?

tynes: Yes that is the idea, it will revert during `initialize` if the deposit gas limit is too small, so we should make sure that the resource config is updated before any of the deposit txs are created, so that the chain operator can increase the deposit gas limit as part of the same initialize call. They may need to increase the deposit gas limit since deposits are issued as part of the call to `initialize`

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,185 @@
...
+Also any open questions that need more work to resolve. -->
+
+There is a concern that the sequencer can include transactions before these values are set on L2.
+If we define the `SystemConfig.startBlock` as the [first block to start derivation in](https://github.com/ethereum-optimism/optimism/blob/d05fb505809717282d5cee7264a09d26002a4ddd/op-node/cmd/genesis/cmd.go#L174C30-L174C40),
+which is set on `SystemConfig.initialize` and also the deposit transactions that set these values are
+sent in the same block, then we should have the guarantee that no user transactions are included before
+the deposit transactions.
+
+There is a concern around the max deposit gas limit being too small so that the `SystemConfig` cannot
+deposit all of these values, we should have logic that reverts if the deposit gas limit is too small
```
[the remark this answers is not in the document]
tynes: This is why we need the [deposit queue](https://github.com/ethereum-optimism/specs/pull/82), it would solve this problem

protocol/standard-l2-genesis.md
tynes: That does not need to be in scope here but may be a nice future follow up given we have user stories around how to issue an emergency upgrade via deposit txs

protocol/standard-l2-genesis.md
tynes: There is general consensus on the approach, see [here](https://discord.com/channels/1244729134312198194/1295923022389313646/1296537510012260374) for context

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,253 @@
...
+The `data` and `gasLimit` are allowed to be specified since we don't fully know what sorts of calls we may have to do.
+We may only need to do simple `upgradeTo` calls, but we may also need to do `upgradeToAndCall`. To support the
+[liquidity migration](https://github.com/ethereum-optimism/design-docs/blob/4b62eb12eceb8e4867ac101134730102c0f5a989/protocol/superchainerc20/liquidity-migration.md), we need to backport storage slots into the `OptimismMintableERC20Factory`
+contract. We may need to introduce multicall support into the `L2ProxyAdmin` as part of this.
+
+#### FeeAdmin role
+
+The entity which authorized to modify the various `FeeVault` configs must be able to vary from chain
+to chain. Therefore a new `feeAdmin` role will be added to the `SystemConfig` contract. This role
+can call a new `SystemConfig.setFeeConfig()` function which forwards config updates to
```
tynes, line 118: Right now per the open PR, there are functions for each fee vault, like `setL1FeeVaultConfig` and `setBaseFeeVaultConfig`. We could simplify to `setFeeVaultConfig` where the first arg is an enum for which fee vault it is

protocol/standard-l2-genesis.md
```diff
@@ -0,0 +1,253 @@
...
+`OptimismPortal.setConfig()` with the appropriate `ConfigType`.
+
+This role will be set in `SystemConfig.initialize()`, meaning that it can only be updated by an upgrade.
+
+In summary:
```
tynes, line 123: We need to guarantee 100% backwards compatibility in the roles, so for example the same 2/2 multisig that owns the L2 `ProxyAdmin` on base should be the `FeeAdmin` in base's `SystemConfig`

protocol/standard-l2-genesis.md
tynes: We should really remove the ability to configure this value and just fetch the value in `op-node` from the `SystemConfig`

protocol/standard-l2-genesis.md
```diff
@@ -95,24 +95,36 @@ block matches the expected value.
...
 and it emits a deposit tx from the `DEPOSITOR_ACCOUNT` that calls the `L2ProxyAdmin`. Sourcing the auth
 from the `SuperchainConfig` allows for simple management of this very important role, given that it impacts
-stage 1 status. This is meant to simplify operations, ie remove the concept of the aliased L1 `ProxyAdmin` owner
+stage 1 status. This is meant to simplify operations by removing the aliased L1 `ProxyAdmin` owner
 being set as the L2 `ProxyAdmin`.
 
+Since the the L1 and L2 `ProxyAdmin` contracts are intended to have the same owner,
+an additional improvement (which may be excluded to limit scope creep), would be to remove the
+`Ownable` dependency on the L1 `ProxyAdmin` contract, and instead have it read the
+`SuperchainConfig.upgrader()` role to authorize upgrades.
```
mds1, line 107: I do like this, but one issue is that it prevents a chain from deploying with their own upgrader to start, and only later transferring that role

protocol/standard-l2-genesis.md
```diff
@@ -173,14 +230,24 @@ to source this config, it would likely need to be hardcoded in the binary and th
...
-in the `ResourceConfig` sanity check function. Since the `ResourceConfig` can be modified during
+in the `setResourceConfig()` function's [sanity checks](https://github.com/ethereum-optimism/optimism/blob/feat/holocene-contracts/packages/contracts-bedrock/src/L1/SystemConfig.sol#L538-L556). Since the `ResourceConfig` can be modified during
 `initialize`, its not that big of a deal and chain operators will see that they need to increase that
 value.
+
+A related concern is that the `useGas()`
+[function](https://github.com/ethereum-optimism/optimism/blob/f99424ded3917ddc0c4ef14355d61e50a38d4d0d/packages/contracts-bedrock/src/L1/ResourceMetering.sol#L156)
+which is called in the `upgrade()` and `setConfig()` functions does not check that the max resource limit is not exceeded by
+the additional `prevBoughtGas`. This is in contrast with
+[`_metered()`](https://github.com/ethereum-optimism/optimism/blob/f99424ded3917ddc0c4ef14355d61e50a38d4d0d/packages/contracts-bedrock/src/L1/ResourceMetering.sol#L128C1-L132C10) which does check `prevBoughtGas`. This requires investigation.
```
maurelian, line 253: I think this opens up the following attack: 
1. The system config is reinitialized, which causes it to submit a large amount of data to L2 (we don't yet know how much L2 gas will be required, but it could be that in the future it grows to be significant). 
2. An attacker observes this transaction, and frontruns it to fill up the resource limit with a large deposit tx.
3. The attacker's transaction results in a deposit block larger than the L2 block gas limit.

protocol/standard-l2-genesis.md
```diff
@@ -153,8 +185,33 @@ The additional deposit gas is the only additional resource usage and its covered
...
+optionally generate the L2 state from the current commit as it currently does using
+`vm.getDeployedCode()`, or retrieve the code from the specified library.
+
+### Upgrade Process
+
+Note that this design modifies how L2 upgrade auth is managed (moving that management into a single
+storage slot on L1), but does not change how upgrades to predeploy contracts are performed.
+Predeploy upgrades can still be done either via a `TransactionDeposited()` event, or a [network
+upgrade automation
+transaction](https://github.com/ethereum-optimism/specs/blob/9f7226be064be0c87f90cbc6be6b0a4b4f58656a/specs/protocol/derivation.md#network-upgrade-automation-transactions).
```
mds1, line 214: We should have a preference here. My understanding is the preference is default to network upgrade automation
transactions, and only use deposit transactions for incident response?

protocol/standard-l2-genesis.md
maurelian: Consensus: 
1. Bytecode goes into the library when a release is created.
  - needs automation to generate and CI to validate that the libraries bytecode always corresponds to the correct tag (possibly with a natspec tag). 
  - versioning.md will required an update, it will serve as a spec for the tooling.
2. The L2Genesis script can optionally pull from the artifacts

maurelian: - TODO: create a section outlining release considerations.
- TODO: get clarity about how to package up L2 contracts releases. How will alternative clients consume the L2 Genesis?

protocol/standard-l2-genesis.md
maurelian: We will limit the scope here to how the L1/L2 contract releases will correlate.

protocol/standard-l2-genesis.md
maurelian: TODO: timeboxed product investigation into the implications of this change.

protocol/standard-l2-genesis.md
maurelian: Todo: bring this up with @smartcontracts.

protocol/standard-l2-genesis.md
maurelian: TODO: Elaborate on this. Include recco about using fresh deposit accounts to get same impl addresses on all chains.

protocol/standard-l2-genesis.md
maurelian: Note: this value is currently define in both the SystemConfig and the `rollup.json` file. Can it be removed entirely from the `SystemConfig`?

protocol/standard-l2-genesis.md
maurelian: This is tracked in another thread below, resolving.

protocol/standard-l2-genesis.md
maurelian: Nice, I prefer that. Am tracking it in a TODO comment.

protocol/standard-l2-genesis.md
maurelian: I think the fix is to just add a require statement into `useGas()`. If config deposits are being DoS'd, we can just submit via flashbots

protocol/standard-l2-genesis.md
maurelian: 567e27e24151

protocol/standard-l2-genesis.md
maurelian: Removed multicall mention.

protocol/standard-l2-genesis.md
maurelian: > so that the chain operator can increase the deposit gas limit as part of the same initialize call

In the context of the superchain though, it is the L1ProxyAdmin owner, not a chain operator, who will call `initialize` during an upgrade. AFAICT, no action would be required by a chain operator.

protocol/standard-l2-genesis.md
tynes: There is a buffer between the L1 deposit gas limit and the L2 gas limit. As long as the initializer deposits don't consume more than that buffer, then we are safe. Right now I believe the buffer is 9million and will increase when we increase the L2 block gas limit. With the deposit queue, this problem goes away since it will result in deposits being "dripped" into L2

tynes: We could put a require statement inside of `useGas` that takes into account the buffer, this method would not be dos'able via front run since calling `useGas` can only be done by a chain operator/governance

tynes:
Noticed something here: https://github.com/ethereum-optimism/optimism/blob/dcdf2b7693192f5bca0353bf22729f26c6240ea9/op-node/rollup/types.go#L43

The `op-node` starts deriving on the L1 block *after* what is in its config. This means we need `SystemConfig.startBlock() - 1` rather than the start block itself. Two options on how to handle this:
- Have the genesis generation script fill in `start block - 1` into the rollup config
- Make the starting L1 block be optional in the rollup config and use the `startBlock()` value as the starting point if its not present

I lean towards the second, it removes config overhead

protocol/standard-l2-genesis.md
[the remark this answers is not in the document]
maurelian: Also discussed below, resolving.

protocol/standard-l2-genesis.md
maurelian: Addressed with this commit:

9b8cb76f9ac0

protocol/standard-l2-genesis.md
maurelian: Tracking this in the doc:
7cc8562dedeb

protocol/standard-l2-genesis.md
maurelian: 0cf08e8f7b0d

protocol/standard-l2-genesis.md
maurelian: Agree the deposit-queue is the right long term fix, but for now we just need to work with what we've got to find a good enough solution.

> We could put a require statement inside of `useGas` that takes into account the buffer, this method would not be dos'able via front run since calling `useGas` can only be done by a chain operator/governance

Since `useGas()` and `_metered` both modify `prevBoughtGas`, I had envisioned reusing the [existing check](https://github.com/ethereum-optimism/optimism/blob/f99424ded3917ddc0c4ef14355d61e50a38d4d0d/packages/contracts-bedrock/src/L1/ResourceMetering.sol#L156):

```
if (int256(uint256(params.prevBoughtGas)) > int256(uint256(config.maxResourceLimit))) {
    revert OutOfGas();
}
```

This would indeed be dos'able by front-running the chain-operator with a large deposit tx. 

> There is a buffer between the L1 deposit gas limit and the L2 gas limit. As long as the initializer deposits don't consume more than that buffer, then we are safe. Right now I believe the buffer is 9million and will increase when we increase the L2 block gas limit. With the deposit queue, this problem goes away since it will result in deposits being "dripped" into L2. 

There are two calls to `useGas()` to consider:
2. in `OptimismPortal.setConfig()`: this uses 200_000 (`SYSTEM_DEPOSIT_GAS_LIMIT`) gas per call
1. in `OptimismPortal.upgrade()`: these calls will also be fairly cheap as they just need to overwrite the implementation slot.

Having outlined that, there appears to be very little danger of exceeding the buffer, so I am in favor of not adding the check, but I will look for an off-chain testing approach to ensure we don't exceed the buffer in the future.

protocol/standard-l2-genesis.md
maurelian: I think this is sufficient clarification: dc28731d4ec4

protocol/standard-l2-genesis.md
maurelian: Tracking here: https://github.com/ethereum-optimism/optimism/issues/12578#issue-2606361806

maurelian: Tracked in this issue: https://github.com/ethereum-optimism/optimism/issues/12578

protocol/standard-l2-genesis.md
maurelian: Tracked: https://github.com/ethereum-optimism/optimism/issues/12580

protocol/standard-l2-genesis.md
maurelian: Tracked here: https://github.com/ethereum-optimism/optimism/issues/12579

protocol/standard-l2-genesis.md
maurelian: I am confident that these changes do not impact on the `l1InfoGasOverhead` although it does need to be taken into account as it determines the remaining 'buffer' which has been discussed elsewhere. That buffer discussion is tracked here: https://github.com/ethereum-optimism/optimism/issues/12578

ajsutton:
I think the second is doable so long as op-node can load it from the latest L1 block reliably. Otherwise it doesn't know which block the system config first existed in.  In most cases (probably not all) op-node walks along L1 from the start block forwards loading data from the block its up to so it has a consistent view. This would need to be a case where it always goes to the latest value.  The start block has to be immutable so that seems reasonable to me.

protocol/standard-l2-genesis.md
tessr: I synced up with @maurelian on this over slack. Chiming in with the product perspective: 

This seems strictly positive. We don't really want the SC to be pausing chains that they're not also responsible for upgrading, and participation in the pause is a benefit that chains get when they join the Superchain ecosystem.

maurelian: Great, thanks, I added a not to reflect this: 9b3732bdd8daf5834b7e46ea271d401e1b80ecac

maurelian:
Tracking the startblock config topic here: https://github.com/ethereum-optimism/optimism/issues/12579

tynes:
@maurelian Can I get an approval on this so we can merge?

mds1 (approved):

PR #101, opened 2024-10-02
feat: solidity versioning process

smartcontracts:
Proposes a process for Solidity version upgrades

### DIFF
--- a/protocol/solidity-versioning.md
+++ b/protocol/solidity-versioning.md
@@ -1,0 +1,57 @@
+# Solidity Versioning
+
+## Context
+
+Solidity undergoes regular updates. These updates often introduce useful new features,
+optimizations, and bug fixes that can enhance the functionality, security, and efficiency of OP
+Stack smart contracts. OP Labs has been planning to propose an update of all existing OP Stack
+contracts to Solidity 0.8.25 to take advantage of some of these improvements.
+
+## Problem Statement
+
+As we approach this update and consider future upgrades, we need to establish a coherent strategy
+for managing Solidity version updates in our repository. This strategy should address the following
+key points:
+
+1. Ensure a consistent approach to each upgrade across the OP Stack codebase.
+2. Facilitate easy review and approval processes for proposed updates.
+3. Minimize the risk associated with adopting new compiler versions.
+4. Protect the OP Stack from compiler bugs that could impact contract functionality or security.
+
+## Proposed Solution
+
+To address the challenges outlined in the problem statement, we propose the following structured
+approach for Solidity version updates:
+
+### Establish a Minimum Delay Period
+
+- Implement a mandatory waiting period after a new Solidity release before it can be considered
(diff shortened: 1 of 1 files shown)


protocol/solidity-versioning.md
```diff
@@ -0,0 +1,46 @@
...
+  - The proposal must include the following components:
+    - The proposed Solidity version number.
+    - The release date of the proposed version.
+    - Specific feature(s) in the new version that would benefit the OP Stack codebase, with a
+      clear explanation of their potential impact and usefulness.
+    - A comprehensive summary of all significant features, changes, and bug fixes introduced
+      between the version currently used and the proposed upgrade.
+    - Highlights of any high-risk features or notable bug fixes that require special attention
+      during the review process.
+    - A link to a Git diff comparing the current and proposed Solidity versions.
```
mds1, line 42: This might be noisy and not too valuable. I think requiring the relevant portion of the official solidity changelog might be a useful addition: https://github.com/ethereum/solidity/blob/develop/Changelog.md

protocol/solidity-versioning.md
```diff
@@ -0,0 +1,46 @@
...
+3. Review and Approval Process:
+  - Establish a review panel.
+  - Conduct a thorough assessment of the upgrade proposal.
+  - Require unanimous approval from the review panel before proceeding with the upgrade.
```
mds1, line 46: This part feels a bit vague:
- Who should be on the review panel?
- What does the thorough assessment look like? Is it just read the proposal and make a judgement?

smartcontracts: Yeah I agree it's too vague. I think the review panel should be established ahead of time. Not really sure what heuristics the panel should be using.

smartcontracts: Added some color here

protocol/solidity-versioning.md
```diff
@@ -0,0 +1,57 @@
...
+
+To address the challenges outlined in the problem statement, we propose the following structured
+approach for Solidity version updates:
+
+### Establish a Minimum Delay Period
+
+- Implement a mandatory waiting period after a new Solidity release before it can be considered
+  for adoption (6 months).
+- The duration of this delay will be based on an analysis of the time typically required to
+  identify and fix critical bugs in previous Solidity releases.
```
mds1, line 31: We can link to the analysis we have. Let's also make sure that it lives somewhere publicly viewable, I'd suggest making it an appendix to this doc

protocol/solidity-versioning.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Formal Upgrade Proposal Process
```
mds1, line 33: Maybe we call them "Design Doc Proposals" to distinguish from governance proposals, and open a design doc PR as the way to do this process?
```suggestion
### Formal Upgrade Design Doc Proposal Process
```

PR #102, opened 2024-10-02
feat: Add FMA superchainerc20 standard

Joxess:
**Description**

This PR introduces the FMA for the `SuperchainERC20` standard. 

**Additional context**

It will allow the use of the `SuperchainERC20` standard without requiring interop functionalities to be enabled at that time.

### DIFF
--- a/security/fma-superchainerc20-standard.md
+++ b/security/fma-superchainerc20-standard.md
@@ -1,0 +1,54 @@
+## **SuperchainERC20 standard-only FMA (Failure Modes and Recovery Path Analysis)**
+
+| Author | Ng, Joxes |
+| --- | --- |
+| Created at | 2024-10-02 |
+| Needs Approval From | Mark Tyneway, Matt Solomon, and 0age
+ |
+| Other Reviewers |  |
+| Status | Review |
+
+## Introduction
+
+---
+
+This document is intended to be shared publicly for reviews and visibility purposes. It covers the Introduction of the `SuperchainERC20` standard, including the implementation and interface.
+
+Below are references for this project:
+
+- [Token standard specs](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/token-bridging.md).
+- [[DRAFT] Implementation](https://github.com/defi-wonderland/optimism/pull/73).
+
+It does not intend to cover related contracts such as `SuperchainERC20Bridge` or those involving migrated liquidity.
+
+## Failure Modes and Recovery Paths
+
+---
+
+### Unauthorized Access to `__superchainMint` & `__superchainBurn` Functions
+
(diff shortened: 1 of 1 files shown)


tynes:
There is another possible failure mode of different token addresses between chains. The fix for this would be to have good deterministic deployment tooling around battle tested create2 deployers

Joxess:
The FMA has been updated to reflect the latest changes made in the [specs](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/token-bridging.md) and to include a the case suggested by @tynes.

security/fma-superchainerc20-standard.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Different Token Addresses Across Chains
+
+- Description: For the `SuperchainERC20Bridge` to validate mints and burns correctly, the `SuperchainERC20` must be deployed at the same address across all interoperable chains. If different addresses are used, the bridge will be unable to successfully finalize the cross-chain transfer.
+- Risk Assessment: Medium
+    - Potential impact: High. Inconsistent token addresses will disable interoperability functionalities for the contract.
+    - Likelihood: Very Low. The interoperable set of chains follows the same opcode behavior and ensures identical availability of deployer contracts, such as `create2Deployer`.
+- **Mitigation**: For developers, ensure to employ the appropiate deterministic deployment tools, such as the one at `create2Deployer`.
+- **Detection**: Verify contract addresses upon deployment.
+- **Recovery Path(s)**: Redeploy token contracts correctly.
```
Ethnical, line 41: As all interop chain need to use the **same** address. 
Is that the same recovery path if we burn the correct nonce for the deployment to the *correct* deterministic address?
If not, could you details me how we should operate in this case?

Ethnical (approved):
This make sense with the current
https://github.com/ethereum-optimism/design-docs/blob/main/protocol/superchainERC20/superc20Redesign.md
- I would like to know also what happen if people are sending token directly to the `SuperchainERC20Bridge` do we have method to rescue them? 
Can this put the bridge in a inconsistent state (like if this is using `SuperTokenERC20.balanceOf(this);`).  This probably bit out of the scope but I am interested in it. 

Otherwise looks go to me 🚀

security/fma-superchainerc20-standard.md
Joxess: Developers can use any method, but as is the case in most scenarios where the same address deployment is required, the use of `CREATE` can be more problematic. You're right; there is a risk of burning the nonce on a chain. So, `create2Deployer` is the best option right now. I will make that clarification!

Joxess:
Sure, @Ethnical. Currently, there is no plan to add a rescue method yet; for now, we prefer to stick with minimal logic around the contract. Regarding the possibility of an inconsistent state, the bridge doesn’t hold state or care about its token balance, so it shouldn’t be a problem. Here is a [reference implementation](https://github.com/defi-wonderland/optimism/blob/sc-feat/crosschain-erc20/packages/contracts-bedrock/src/L2/SuperchainERC20Bridge.sol).

tynes (approved):
This looks good to me, cannot think of any other real risks

PR #103, opened 2024-10-02
feat: Integrate op-node with op-signer for block payload signing

mininny:
**Description**

This PR introduces how we can integrate op-node with op-signer when signing the block payload for sequencer gossiping.

### DIFF
--- a/security/op-node-op-signer-integration.md
+++ b/security/op-node-op-signer-integration.md
@@ -1,0 +1,128 @@
+# Integrating op-node with op-signer
+
+# Purpose
+
+Integrate op-node with op-signer, for more modular structure of a deployed op-stack. 
+
+## Summary
+
+In order to run a sequencer on a OP Stack, you must have the sequencer op-node communicate and gossip the new block payload with other nodes in the network via p2p. This new block payload is signed with a correct private key to prove the validity of the payload before being gossiped to the peers.  
+
+Previously, the operator of the sequencer node and the signer of the payload was equal, and this private key is directly injected as a run flag of op-node. There are few improvements to make: 
+
+1. Direct injection of key is dangerous and can be compromised. With a separate key management system, node operators do not need access to the private keys, which is safer. With this proposal, we can apply the same level of security to the sequencer nodes as well.
+2. We can abstract away such key management to introduce a more modular system. This allows us to have separate operators of the sequencer node and the signer of the payload, or have the secret key management into a more isolated service. 
+
+# Context
+
+Fortunately, there already exists `op-signer`, which serves a similar purpose in the op stack system, specifically the op-batcher and op-proposer. 
+
+`op-batcher` also has to sign transactions prior to sending the batched transactions to L1. Instead of injecting the private key directly into the run flag of op-batcher, you can input the rpc endpoint for handling the key management. This includes configurations like the signer endpoint, mtls certificate for communication, etc. 
+
+Users of op-batcher can choose from: 
+
+1. private-key
+2. mnemonic + hd path
+3. remote signer configuration
+
+to sign the new transactions. 
+
+However, op-node uses different signing scheme than the op-batcher, requiring the introduction of new RPC: 
+
+The signing of block payload in op-node is a custom signing structure where the signing hash is: `[32]byte keccak256Hash([32]byte domain, [32]byte chainId, [32]byte keccak256(payloadBytes))` (see op-node/p2p/signer.go). This signing hash is signed with the private key, to produce a ECDSA signature of 65 bytes. 
+
+# Proposed Solution
+
+Above design can be applied to op-node to allow for remote signing of new block payloads in a sequencer. 
+
+For this, we need changes in the following modules: 
+
+1. op-node to import corresponding `opsigner` module from the op-service
+
+This module allows the user to choose from private-key / mnemonic / remote signer and create a single Signer from the factory. This involves importing `op-service/signer` and `op-service/crypto` into op-node.
+
+2. Modify `op-service/signer` to handle signing of block payload, in addition to tx signing. 
+
+Currently, `op-service` only has `SignTransactions` function which calls `eth_signTranasction`. To abstractify the signing method for different use cases, we might add different signing strategy types, or add new functions like `SignBlockPayload`. 
+
+1. Modify `op-signer` to serve different signing requests.
+
+Currently, `op-signer` only handles `eth_signTrasnaction`. We need to introduce a new rpc endpoint to handle signing of block payload
+
+The diagram illustrates what the result op-stack service looks like:
+
+```mermaid
+sequenceDiagram
+    participant PublicPeers as public peers
+    participant Sequencer as sequencer op-node
+    participant OpSigner as op-signer
+    participant GCP as GCP HSM (or any other service)
+
+    Sequencer ->> OpSigner: block data payload
+    OpSigner -->> Sequencer: mTLS verify
+    Sequencer -->> OpSigner: mTLS verify
+    OpSigner ->> GCP: Request Signing of payload
+    GCP ->> OpSigner: Return Signature
+    OpSigner ->> Sequencer: Return signature
+    Sequencer ->> PublicPeers: Propagate block w/ signature
+
+```
+
+## New RPC: eth_signBlockPayload
+
+new eth-domain RPC (or under some other namespace)
+
+```go
+func eth_signBlockPayload(signingHash [32]byte) (sig *[65]byte, err error) 
+// where signingHash is keccak256Hash([32]byte domain, [32]byte chainId, [32]byte keccak256(payloadBytes))
+```
+
+This new RPC will be added to:
+
+1. optimism/op-service/signer/client.go: new SignBlockPayload function which calls, the eth_signBlockPayload RPC. This method is called by the op-node’s p2p/signer.go
+2. optimism-infra/op-signer/service/service.go: new SignBlockPayload, which is the receiver counterpart of the RPC. 
+
+Most implementation of this RPC will be decoupling the signing logic from op-node, and moving code around different parts of the monorepo/op-signer. 
+
+### Considerations
+
+Above signing structure is a custom scheme, and does not align with other more official signing schemes like the ones introduced by EIP-712 or ERC-191, so we can’t use something like `eth_sign` or `eth_signTypedData`. 
+
+We could migrate the signing scheme of the p2p gossiped payload to the other methods of signing like the ones mentioned above. However, we need to preserve backwards compatibility of the signing scheme for the existing nodes in the network to continue working after this change. 
+
+## Security
+
+In such service where op-node and op-signer are located in a separate system and need to communicate, we need secure channel, where both parties can trust each other. For this, we use mutual TLS (mTLS). 
+
+With mTLS, we can prove that both the client and the server are verified with their own respective certificate. This allows op-node to trust the op-signer, and op-signer to sign only the requests from valid op-node sequencers. 
+
+Thankfully, op-signer already has mTLS support, as integrated in op-batcher and op-proposer. 
+
+Here, users can specify
+
+- tls.ca
+- tls.cert
+- tls.key
+
+and the mTLS is handled in op-service/tls. We have to set up the correct CA and so that both the op-signer and op-node can verify the certificates from the TLS handshake. 
+
+There has already been discussions about integrating op-signer with op-node, such as
+
+- https://github.com/ethereum-optimism/optimism/pull/4357#discussion_r1045342439
+- https://github.com/ethereum-optimism/optimism/pull/4669#discussion_r1067409473
+
+# Alternatives Considered
+- When we send the Block Payload from op-node to op-signer, we could send the full block payload arguments like the following: 
+```go
+type BlockPayloadArgs struct {
+	domain [32]byte `json:"domain"`
+	chainId *big.Int `json:"chainId"`
+	payloadHash [32]byte `json:"payloadHash"`
+}
+```
+This means we are sending more data to the op-signer, but may be more secure since the op-signer controls all of the logic for signing the payload. 
+
+# Risks & Uncertainties
+
+1. Because op-signer will sign any payloads sent to its RPC, we must ensure that only trusted entities can access the RPC of op-signer. Otherwise, an attacker might request op-signer to sign malicious block payloads, and use its signature to propagate invalid block data to peers in the network. To prevent this, we must ensure that we have safe mechanisms in place to only allow the connection between trusted entities. 
+2. As we create secure connection between the op-signer and op-node (whether by mTLS or other methods), we must ensure that the time for signing the payload is quick enough. It must be always faster than the block build time, and quick enough for good gossiping across the network.
--- a/security/op-node-op-signer-integration.md
+++ b/security/op-node-op-signer-integration.md
@@ -29,7 +29,7 @@ to sign the new transactions.
 
 However, op-node uses different signing scheme than the op-batcher, requiring the introduction of new RPC: 
 
-The signing of block payload in op-node is a custom signing structure where the signing hash is: `[32]byte keccak256Hash([32]byte domain, [32]byte chainId, [32]byte keccak256(payloadBytes))` (see op-node/p2p/signer.go). This signing hash is signed with the private key, to produce a ECDSA signature of 65 bytes. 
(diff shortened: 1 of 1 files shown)


tynes:
This is awesome! Great design doc

security/op-node-op-signer-integration.md
```diff
@@ -0,0 +1,145 @@
...
+	chainId *big.Int `json:"chainId"`
+	payloadHash [32]byte `json:"payloadHash"`
+}
+```
+This means we are sending more data to the op-signer, but may be more secure since the op-signer controls all of the logic for signing the payload. 
+
+# Risks & Uncertainties
+
+1. Because op-signer will sign any payloads sent to its RPC, we must ensure that only trusted entities can access the RPC of op-signer. Otherwise, an attacker might request op-signer to sign malicious block payloads, and use its signature to propagate invalid block data to peers in the network. To prevent this, we must ensure that we have safe mechanisms in place to only allow the connection between trusted entities. 
+2. As we create secure connection between the op-signer and op-node (whether by mTLS or other methods), we must ensure that the time for signing the payload is quick enough. It must be always faster than the block build time, and quick enough for good gossiping across the network.
```
zhwrd, line 145: Think we need to double check a few other things related to performance, capacity and availability since the signing frequency will be increased a lot from what we do with batcher, proposer & challenger today. Also none of those services have as strict uptime requirements as a sequencer does.

* What is the downstream rate limit / quota for signing requests to HSM backed key in KMS. Do we expect to be within that quota?
* What is the impact to the sequencing if we have an op-signer or KMS outage. Should we have a fallback mechanism?
* What is the acceptable max latency for signing requests?

security/op-node-op-signer-integration.md
```diff
@@ -0,0 +1,145 @@
...
+This new RPC will be added to:
+
+1. optimism/op-service/signer/client.go: new SignBlockPayload function which calls, the eth_signBlockPayload RPC. This method is called by the op-node’s p2p/signer.go
+2. optimism-infra/op-signer/service/service.go: new SignBlockPayload, which is the receiver counterpart of the RPC. 
+
+Most implementation of this RPC will be decoupling the signing logic from op-node, and moving code around different parts of the monorepo/op-signer. 
+
+### Considerations
+
+Above signing structure is a custom scheme, and does not align with other more official signing schemes like the ones introduced by EIP-712 or ERC-191, so we can’t use something like `eth_sign` or `eth_signTypedData`.
```
alfonso-op, line 89: It seems desirable to try and align on a more standard signing scheme. Provided it doesn't add considerable complexity / additional engineering effort, I recommend we implement against a standard and preserve backwards compatibility for older versions.

security/op-node-op-signer-integration.md
```diff
@@ -0,0 +1,145 @@
...
+```go
+type BlockPayloadArgs struct {
+	domain [32]byte `json:"domain"`
+	chainId *big.Int `json:"chainId"`
+	payloadHash [32]byte `json:"payloadHash"`
+}
+```
+This means we are sending more data to the op-signer, but may be more secure since the op-signer controls all of the logic for signing the payload. 
+
+# Risks & Uncertainties
```
alfonso-op, line 142: Another potential risk to keep in mind is ensuring we enforce the invariant that only the active sequencer is able to successfully sign blocks without odd race conditions (e.g., during sequencer leadership transfer).

To prevent such issues, does the design require any further integration / communication with op-conductor, or will this be guaranteed with the existing op-node implementation?

security/op-node-op-signer-integration.md
```diff
@@ -0,0 +1,145 @@
...
+  - name: op-node.primary.mainnet.prod.oplabs.cloud
+    key: projects/oplabs-prod-mainnet/locations/eur6/keyRings/testop-signer/cryptoKeys/op-node/cryptoKeyVersions/1
+// ... other existing keys
+  - name: sequencer-batcher.primary.mainnet.prod.oplabs.cloud
+    key: projects/oplabs-prod-mainnet/locations/eur6/keyRings/op-signer/cryptoKeys/batcher/cryptoKeyVersions/1
+```
+for each signing request, op-signer will look at the requesting server's domain, and look at this yaml file for a matching config. Then, it will look for the gcp kms key that matches the `key` name.
+
+
+## Security
```
alfonso-op, line 110: Just a note that we'll want to conduct [an FMA](https://github.com/ethereum-optimism/design-docs/blob/main/assets/fma-template.md) as part of this project.

security/op-node-op-signer-integration.md
```diff
@@ -0,0 +1,145 @@
...
+2. mnemonic + hd path
+3. remote signer configuration
+
+to sign the new transactions. 
+
+However, op-node uses different signing scheme than the op-batcher, requiring the introduction of new RPC: 
+
+The signing of block payload in op-node is a custom signing structure where the signing hash is: `[32]byte keccak256Hash([32]byte domain, [32]byte chainId, [32]byte keccak256(payloadBytes))` (see op-node/p2p/signer.go). This signing hash is signed with the private key, to produce a ECDSA signature of 65 bytes. 
+
+# Proposed Solution
```
alfonso-op, line 34: Given the security-sensitive changes involved, these changes will require governance approval. We'll want to factor this into planning and timelines cc @taempark

alfonso-op (commented):
Thanks @mininny, great doc. I left some comments.

I agree with Zach there are a few other performance and availability considerations we'll want to expand on.

security/op-node-op-signer-integration.md
```diff
@@ -0,0 +1,145 @@
...
+	domain [32]byte `json:"domain"`
+	chainId *big.Int `json:"chainId"`
+	payloadHash [32]byte `json:"payloadHash"`
+}
+```
+This means we are sending more data to the op-signer, but may be more secure since the op-signer controls all of the logic for signing the payload. 
+
+# Risks & Uncertainties
+
+1. Because op-signer will sign any payloads sent to its RPC, we must ensure that only trusted entities can access the RPC of op-signer. Otherwise, an attacker might request op-signer to sign malicious block payloads, and use its signature to propagate invalid block data to peers in the network. To prevent this, we must ensure that we have safe mechanisms in place to only allow the connection between trusted entities.
```
raffaele-oplabs, line 144: I have read the code of op-signer and it technically does not manage authorization as far as I see.
Any authenticated user ( the one that have a valid client certificate) will be able to access the key with the same name as the one embedded in the certificate. 
This may work in the assumption that any client own the key with the same name, and any client can perform all operations with that key.
This may be a valid assumption, I just want to make sure this is something we are aware in the moment we manage the certificates signing. As I was not able to see if there is a revocation list, the whole system security depends on how well the KMS for the client certificate is managed, and that there is an explicit written procedure that spell out the assumption that the authorization works with "DNSNames[0]" = access to a specific key+ all the functionalities of op-signer.
On top of that op-signer at the moment does not log network informations for auditing purposes like origin ip, or certificate signature as far as I can see.

security/op-node-op-signer-integration.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Security
+
+In such service where op-node and op-signer are located in a separate system and need to communicate, we need secure channel, where both parties can trust each other. For this, we use mutual TLS (mTLS). 
+
+With mTLS, we can prove that both the client and the server are verified with their own respective certificate. This allows op-node to trust the op-signer, and op-signer to sign only the requests from valid op-node sequencers. 
+
+Thankfully, op-signer already has mTLS support, as integrated in op-batcher and op-proposer. 
+
+Here, users can specify
```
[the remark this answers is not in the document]
mininny: Yes, the tls certificate has to be set in both the op-signer / op-node service. Perhaps the wording `users can` is little misleading, as it's more of a `must`

security/op-node-op-signer-integration.md
mininny: Yes, I agree that we have an assumption that client with the specific DNS name + correct mTLS certificate will get access to all of the functionalities of op-signer using that specific key. 

I'm all in for adding logs to op-signer for log network information and incoming request information. 

Given that we safely set the mTLS certificate pair on the op-node and op-signer, what other security measures do you think we need to apply here?

security/op-node-op-signer-integration.md
mininny: I agree on using standard signing scheme and adding backwards compatibility to existing scheme when possible. I did look at other schemes like `eth_sign` and `eth_signTypedData`, but they require more payload and parsing as they contain ethereum-related information necessary for signing. 

I think this rpc is specific to the op-node's p2p signing behavior, and will most likely just be handled by the op-signer(or may beother optimism-related clients in the future), rather than being used in general clients like geth, etc. It'd be best to minimize the payload size to keep the signature latency as minimal possible, and that would require optimism-specific rpc. 

Rather, what do you think about changing the rpc name from `eth_signBlockPayload` to something like `optimism_signBlockPayload` to remove the impression that this operation is related to ethereum when it's not?

security/op-node-op-signer-integration.md
mininny: Of course! I will write it before we deploy it :) thanks

security/op-node-op-signer-integration.md
```diff
@@ -0,0 +1,145 @@
...
+
+1. private-key
+2. mnemonic + hd path
+3. remote signer configuration
+
+to sign the new transactions. 
+
+However, op-node uses different signing scheme than the op-batcher, requiring the introduction of new RPC: 
+
+The signing of block payload in op-node is a custom signing structure where the signing hash is: `[32]byte keccak256Hash([32]byte domain, [32]byte chainId, [32]byte keccak256(payloadBytes))` (see op-node/p2p/signer.go). This signing hash is signed with the private key, to produce a ECDSA signature of 65 bytes.
```
protolambda, line 32: And specified here: https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/rollup-node-p2p.md#block-signatures

Including the specs link, in addition to op-node, may be useful. While op-node is common, we want alt-clients to build on the spec.

security/op-node-op-signer-integration.md
```diff
@@ -0,0 +1,145 @@
...
+```
+
+## New RPC: eth_signBlockPayload
+
+new eth-domain RPC (or under some other namespace)
+
+```go
+func eth_signBlockPayload(signingHash [32]byte) (sig *[65]byte, err error) 
+// where signingHash is keccak256Hash([32]byte domain, [32]byte chainId, [32]byte keccak256(payloadBytes))
+```
```
protolambda, line 78: I think it's better if the 3 attributes are communicated over RPC fully. That way the op-signer server side can log the request, like it does with transaction signing requests, to clarify what is being signed.

Additionally, the chainID should be checked, as well as the signingHash, to not accidentally sign things not intended to be signed. E.g. when reusing the same sequencer key for multiple chains.

In a later protocol upgrade we might also want to structure it even more, to bind the block-number to the domain, or add one more layer of hashing to the `payloadBytes` part (like a head and a payload-body). That way the signing request can be more expressive, while summarizing the body content, to not pass through the full payload content which might otherwise slow things down.

security/op-node-op-signer-integration.md
protolambda: +1 one introducing it into the `optimism` RPC namespace, or a new `opsigner` namespace, rather than the ethereum namespace.

Signing with `eth_signTypedData` over typed data would make sense, but that API also has a few flaws (more complicated schema decoding and communication, on every signing request). We can look into unifying that in a future upgrade, if desired. Perhaps `optimism_signBlockPayload` can become a front-end, wrapping `eth_signTypedData` and applying the schema, type checks and logging needed for block signing.

security/op-node-op-signer-integration.md
```diff
@@ -0,0 +1,145 @@
...
+# Alternatives Considered
+- When we send the Block Payload from op-node to op-signer, we could send the full block payload arguments like the following: 
+```go
+type BlockPayloadArgs struct {
+	domain [32]byte `json:"domain"`
+	chainId *big.Int `json:"chainId"`
+	payloadHash [32]byte `json:"payloadHash"`
+}
+```
+This means we are sending more data to the op-signer, but may be more secure since the op-signer controls all of the logic for signing the payload.
```
protolambda, line 140: I think `op-signer` being configured with expected parameters and checking the RPC method arguments match is a better trade-off. That way you can detect misuse of the API, rather than producing a signature for the assumed domain and finding it's invalid much later when first doing a verification of the produced signature.

Perhaps the RPC should also include the "sender address", so we can assert that matches the expected address. And then it can be used as an intent, in case this same op-signer RPC server can sign for multiple different p2p signer keys.

security/op-node-op-signer-integration.md
taempark: Good callout, @alfonso-op, I will mention this in the FMA of SiaB project!

security/op-node-op-signer-integration.md
taempark: I think the scope this design doc is just adding more functionality to the op-signer. So I'm not sure if we'd need the governance approval for this feature. But, I think mentioning this to the FMA seem to be necessary.

One point I've thought about the governance approval is relevant is when we hand over the sequencer fully. I briefly discussed this with @mslipper, but that seem to not need the governance approval as well. But the decision of the conversation was let's double-check with @ben-chain just in case.

security/op-node-op-signer-integration.md
mininny: Thanks for the comment! then I'll modify the design doc and code to send the `domain`, `chainId`, `payloadBytesHash` and `senderAddress`, and add validation for these parameters. Then we could add configurable parameter for chainId and senderAddress when spinning up the op-signer, and validate the incoming rpc requests to match these configurations :)

protolambda: And maybe we should bundle them in a struct, rather than specify them as individual function arguments. This will make it easier to extend the input in the future, by adding optional struct fields.

security/op-node-op-signer-integration.md
mininny: I updated the design doc and the implementation to use the new `opsigner` namespace: https://github.com/ethereum-optimism/infra/pull/59/commits/dda416c364ab20f97c945845c901106c3df4d771. And left unifying the signing scheme to a possible future work in the docs

security/op-node-op-signer-integration.md
mininny: This is updated in the design doc as well as the op-node pr here! https://github.com/ethereum-optimism/optimism/pull/12325/commits/12c4820e35fbba947223748fac13a77e91557e8b

ImTei: Is the purpose of the sender address to verify if the request is valid in a expected restricted environment? I mean, if it has the security purpose, attackers can easily change the address of a request.

security/op-node-op-signer-integration.md
raffaele-oplabs: Before proceeding I think we should look at this more carefully. Or add a blocker to this before we are ready to deploy. At the moment, for how this is implemented, is possible for anyone able to deploy in the same kluster, to obtain a valid certificate with the characteristic mentioned.
@alfonso-op how could we manage this? As part of the MFA mentioned above?

alfonso-op: @raffaele-oplabs yeah I suggest we cover as part of the FMA. We should allocate some extra time, since we know there is some additional security hardening we will have to implement here.

We could start with request logs and alerts, but should brainstorm other potential approaches.

security/op-node-op-signer-integration.md
trianglesphere: +1 to not adding to the `eth` namespace. It doesn't look like neither https://github.com/ethereum-optimism/infra/pull/59 in the signer client nor https://github.com/ethereum-optimism/optimism/pull/12325 docs were updated to the new name.

mininny: @trianglesphere Oh you're right, the code were updated but not the descriptions. I just updated the description for those PRs to the new `opsigner_signBlockPayload`

security/op-node-op-signer-integration.md
protolambda: The address should not be trusted, but it should help specify which account is meant to be used, if there are multiple options. The eth signing API supports a from-address, so it makes sense to do the same here.

protolambda (approved):

PR #132, opened 2024-10-11
feat: Stage 1.4: Kona + Asterisc FMA

pauldowman:
### Description

A failure modes and recovery paths analysis for Stage 1.4: Kona and Asterisc.

*Status:* **In Review**

### DIFF
--- a/security/fma-kona-asterisc.md
+++ b/security/fma-kona-asterisc.md
@@ -96,7 +96,7 @@ Below are references for this project:
 
 - **Mitigations:**
   Action tests covering the derivation pipeline are run on every pull request in the [kona] repository.
-  VM Runner exectutes the derivation pipeline as part of the kona fault proof program running ontop of asterisc.
+  VM Runner executes the derivation pipeline as part of the kona fault proof program running on top of asterisc.
   [Extensive test coverage](https://app.codecov.io/gh/anton-rs/kona/tree/main/crates) of kona.
 
 - **Detection:**
--- a/security/fma-kona-asterisc.md
+++ b/security/fma-kona-asterisc.md
@@ -93,19 +93,31 @@ Below are references for this project:
   Low likelihood, medium impact.
   The contract upgrades are tested and very simple.
   Contracts in the hot path are audited and have been in production so are unlikely to have critical bugs.
-  // TODO: impact
+  Medium impact since an upgrade can impact code placed in the hot path.
 
 - **Mitigations:** 
+  Mitigations are done in the `superchain-ops` repository to validate upgrades prior to their execution.
+  These mitigations include:
+  - Simulate the job before execution.
+  - Run in tenderly
+    - Compare post-state diff with expectations (all signers do this)
+    - Run state diff checks, ensuring the upgrade went smoothly with static assertions.
+  - Run in superchain-ops CI until the upgrade to ensure it’s still working until the day of the upgrade.
 
 - **Detection:**
+  An invalid upgrade causing an issue in the hot path would be detected by the `op-dispute-mon`.
+  If a misconfiguration for `kona` + `asterisc` in the cold path is done, the `op-dispute-mon`
+  and `op-challenger` would detect an issue if the `kona` + `asterisc` game is placed in the hot path.
 
 - **Recovery Path(s):**
+  If an invalid upgrade somehow bricks both the `op-program` + `cannon` dispute game and the `kona` + `asterisc`
+  dispute game, the recovery path would be to activate the permissioned fallback.
 
 
 ### Divergent Kona Derivation Pipeline
 
 - **Description:**
-  The Kona derivation pipeline diverged in behavior from the op-program's derivation pipeline.
+  The `kona` derivation pipeline diverged in behavior from the `op-program`'s derivation pipeline.
 
 - **Risk Assessment:**
   Low likelihood, low impact.
@@ -114,7 +126,7 @@ Below are references for this project:
   [The "Kona and Asterisc Hot Path Failure"](#-Kona-and-Asterisc-Hot-Path-Failure) above covers the failure mode for `kona` + `asterisc` when placed in the hot path.
 
 - **Mitigations:**
-  Frequent action tests covering the derivation pipeline.
+  Action tests covering the derivation pipeline are run on every pull request in the [kona] repository.
   VM Runner exectutes the derivation pipeline as part of the kona fault proof program running ontop of asterisc.
   [Extensive test coverage](https://app.codecov.io/gh/anton-rs/kona/tree/main/crates) of kona.
 
@@ -132,8 +144,8 @@ Below are references for this project:
 ### Kona Execution Diverges from op-program
 
 - **Description:**
-  The Kona fault proof program executes blocks in a stateless L2 block executor that uses optimism-revm as a backend.
-  If optimism-revm or the L2 block executor diverge in behavior from the op-program, the kona + asterisc proof system will be invalid.
+  The Kona Fault Proof Program executes blocks in a stateless L2 block executor that uses optimism-revm as a backend.
+  If optimism-revm or the L2 block executor diverge in behavior from the `op-program`, the `kona` + `asterisc` proof system will be invalid.
 
 - **Risk Assessment:**
   Low likelihood, low impact.
@@ -142,17 +154,18 @@ Below are references for this project:
 
 - **Mitigations:**
   State testing against op-revm.
-  Frequent action tests against kona, which covers the L2 block executor.
-  `op-reth` which uses the same op-execution environment as kona.
+  Action tests covering the L2 block executor are run on every pull request in the [kona] repository.
+  `op-reth` which uses the same op-execution environment as `kona`.
 
 - **Detection:**
-  Since kona + asterisc is in the coldpath by default, a divergence would only be seen by action tests or local testing.
-
+  Running `op-reth` so it's synced and following tip would provide a way to watch for chain splits.
+  A chain split occuring means `kona`'s execution (the same as `op-reth`) diverges from the `op-program`.
+  
 - **Recovery Path(s):**
-  If kona + asterisc is in the hotpath, the game would need to be invalidated and op-program + cannon placed in the hotpath while kona + asterisc is fixed.
-  Once the execution fixes are released and deployed, kona + asterisc could be placed back in the hotpath.
-  If the game is in the coldpath, it would need to be fixed and re-deployed.
-
+  If `kona` + `asterisc` is in the coldpath, it would need to be fixed and re-deployed.
+  If `op-program` + `cannon` cannot be placed in the hotpath, and `kona` + `asterisc` fails, the permissioned fallback must be engaged.
+  [The "Kona and Asterisc Hot Path Failure"](#-Kona-and-Asterisc-Hot-Path-Failure) above covers the failure mode for `kona` + `asterisc` when placed in the hot path.
+  
 
 ### Breaking Backwards Compatibility for the op-challenger
 
@@ -160,21 +173,22 @@ Below are references for this project:
   The `op-challenger` has been updated to support the Kona + Asterisc Fault Dispute Game type ([Game Type 3]).
 
 - **Risk Assessment:**
-  Low likelihood as the action tests that are run in CI very frequently run the `op-challenger` with the kona + asterisc fault dispute game.
+  Low likelihood as the action tests that are run in CI very frequently run the `op-challenger` with the `kona` + `asterisc` fault dispute game.
   Medium impact since the off-chain `op-challenger` service can be updated with enough buffer time due to the large game time and nature of the chess clock.
 
 - **Mitigations:**
   The `op-challenger` is architected in such a way as to isolate game types and moreso the inididual game players.
   To allow the challenger to scale and play many different game types at once, a list of game types to play is passed into the `op-challenger` as a CLI flag.
   When games are detected on-chain, the `op-challenger` will create a "game player" for this game type.
-  Players are run in individual threads so a breaking kona-asterisc game player in the challenger will not cause a liveliness issue for the `op-challenger` to play other game types.
+  Players are run in individual threads so a breaking `kona` + `asterisc` game player in the challenger will not cause a liveliness issue for the `op-challenger` to play other game types.
   Another key mitigation is the VM runner. This is an offline simulation that runs asterisc + kona with the op-challenger. It's been running since around August 2024.
 
 - **Detection:**
-  The Dispute Monitor ([op-dispute-mon] service provides an isolated monitoring service that will detect if a game is not being played by the honest `op-challenger`. It will alert in this case.
+  The Dispute Monitor ([op-dispute-mon] service provides an isolated monitoring service that will detect if a game is not being played by the honest `op-challenger`.
+  It will alert in this case.
 
 - **Recovery Path(s):**
-  If the `op-challenger` fails to play Kona + Asterisc fault dispute games, a fix will need to be made to the `op-challenger`.
+  If the `op-challenger` fails to play `kona` + `asterisc` fault dispute games, a fix will need to be made to the `op-challenger`.
   It does not require governance, but would need a new release of the `op-challenger` to be published and rolled out to infrastructure.
 
 [op-dispute-mon]: https://github.com/ethereum-optimism/optimism/tree/develop/op-dispute-mon
@@ -204,7 +218,8 @@ Below are references for this project:
 - **Recovery Path(s):**
   If a divergence between `RISCV.sol` and asterisc is found, either or both would need to be updated, and re-released.
   Since the kona + asterisc Fault Dispute Game would need to use the updated `RISCV.sol`, if it was fixed, an upgrade would need to be performed.
-  If the kona + asterisc fault dispute game sits in the hotpath, the security council would need to fallback to the op-program + cannon fault dispute game, and only switch back once the fix has been released and deployed.
+  If the kona + asterisc fault dispute game sits in the hotpath, the security council would need to fallback to the op-program + cannon fault dispute
+  game, and only switch back once the fix has been released and deployed.
 
 
 ### Generic items we need to take into account:
--- a/security/fma-kona-asterisc.md
+++ b/security/fma-kona-asterisc.md
@@ -11,6 +11,7 @@
     - [Divergence between RISCV.sol and Asterisc](#divergence-between-riscvsol-and-asterisc)
     - [Generic items we need to take into account:](#generic-items-we-need-to-take-into-account)
   - [Audit Requirements](#audit-requirements)
+  - [Action Items](#action-items)
   - [Appendix](#appendix)
     - [Appendix A: op-challenger](#appendix-a-op-challenger)
 
--- a/security/fma-kona-asterisc.md
+++ b/security/fma-kona-asterisc.md
@@ -22,15 +22,6 @@
 | Need Approval From | Ben Clabby, Mark Tyneway |
 | Status | Draft |
 
-> [!TIP]
-> Guidelines for writing a good analysis, and what the reviewer will look for:
->
-> - Show your work: Include steps and tools for each conclusion.
-> - Completeness of risks considered.
-> - Include both implementation and operational failure modes
-> - Provide references to support the reviewer.
-> - The size of the document will likely be proportional to the project's complexity.
-> - The ultimate goal of this document is to identify action items to improve the security of the  project. The FMA review process can be accelerated by proactively identifying action items during the writing process.
 
 ## Introduction
 
@@ -62,13 +53,18 @@ Below are references for this project:
 - **Recovery Path(s):** If the `op-challenger` fails to play Kona + Asterisc fault dispute games, a fix will need to be made to the `op-challenger`. It does not require governance, but would need a new release of the `op-challenger` to be published and rolled out to infrastructure.
 
 
-### [Name of Failure Mode 2]
+### Divergence in the behavior between RISCV.sol and Asterisc
 
-- **Description:** *Details of the failure mode go here. What the causes and effects of this failure?*
-- **Risk Assessment:** *Simple low/medium/high rating of impact (severity) + likelihood.*
-- **Mitigations:** *What mitigations are in place, or what should we add, to reduce the chance of this occurring?*
-- **Detection:** *How do we detect if this occurs?*
-- **Recovery Path(s)**: *How do we resolve this? Is it a simple, quick recovery or a big effort? Would recovery require a governance vote or a hard fork?*
+- **Description:** Asterisc is a "Fault Proof VM". It is an onchain RISCV instruction emulator with the on-chain component of the kona + asterisc fault dispute game being the `RISCV.sol` contract. `RISCV.sol` runs a single RISCV instruction on chain with merkleized VM memory. Effectively, the behavior between `RISCV.sol` and asterisc cannot diverge.
+
+- **Risk Assessment:** Low likelihood, but high impact.
+
+- **Mitigation:** Heavily test `RISCV.sol`. Audit `RISCV.sol`. Asterisc already runs through action tests frequently in CI validiting that it produces the expected output.
+
+- **Detection:** Since `op-dispute-mon` uses the same asterisc backend that the `op-challenger` does, there may be divergent behaviour between asterisc and `RISCV.sol`. In the worst case, if a silent issue is exploited in `RISCV.sol`, a fault dispute game may resolve incorrectly. But since the kona + asterisc fault dispute game lives outside the hotpath, this game result is not used anywhere by the portal and can be discarded.
+
+- **Recovery Path(s):** If a divergence between `RISCV.sol` and asterisc is found, either or both would need to be updated, and re-released. Since the kona + asterisc Fault Dispute Game would need to use the updated `RISCV.sol`, if it was fixed, an upgrade would need to be performed. If the kona + asterisc fault dispute game sits in the hotpath, the security council would need to fallback to the op-program + cannon fault dispute game, and only switch back once the fix has been released and deployed.
+
 
 ### Generic items we need to take into account:
 
@@ -78,6 +74,7 @@ Below are references for this project:
 ## Audit Requirements
 
 *Given the failure modes and action items, will this project require an audit? See [OP Labs Audit Framework: When to get external security review and how to prepare for it](https://gov.optimism.io/t/op-labs-audit-framework-when-to-get-external-security-review-and-how-to-prepare-for-it/6864) for a reference decision making framework. Please explain your reasoning.*
+
 
 ## Appendix
 
--- a/security/fma-kona-asterisc.md
+++ b/security/fma-kona-asterisc.md
@@ -24,8 +24,8 @@
 | Author | Andreas Bigger |
 | Created at | 2024-10-17 |
 | Initial Reviewers | Matt Solomon, Ethnical |
-| Need Approval From | Ben Clabby, Mark Tyneway |
-| Status | In Review |
+| Need Approval From | Ben Clabby, Mark Tyneway, Matt Solomon |
+| Status | Implementing Actions |
 
 
 ## Introduction
--- a/security/fma-kona-asterisc.md
+++ b/security/fma-kona-asterisc.md
@@ -31,8 +31,6 @@ Together, Kona and Asterisc form an alternative proof stack to the [op-program](
 
 Since two main components are being shipped in Stage 1.4 there are two main places the system can break: the kona fault proof program, and the asterisc VM. The below Failure Modes and Recovery Paths go over those situations.
 
-Notably, 
-
 Below are references for this project:
 
 - ["Stage 1.4" internal project document](https://www.notion.so/oplabs/Stage-1-4-Partway-a6f57ad777b148dda01488f2646cff17)
--- a/security/fma-kona-asterisc.md
+++ b/security/fma-kona-asterisc.md
@@ -1,5 +1,3 @@
-# [Project Name]: Failure Modes and Recovery Path Analysis
-
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
@@ -14,25 +12,15 @@
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
-*Italics are used to indicate things that need to be replaced.*
+# Stage 1.4: Kona and Asterisc FMA (Failure Modes and Recovery Path Analysis)
 
 | | |
 |--------|--------------|
-| Author | *Author Name* |
-| Created at | *YYYY-MM-DD* |
+| Author | Andreas Bigger |
+| Created at | 2024-10-17 |
 | Initial Reviewers | *Reviewer Name 1, Reviewer Name 2* |
-| Need Approval From | *Security Reviewer Name* |
-| Status | *Draft / In Review / Implementing Actions / Final* |
-
-> [!NOTE]
-> 📢 Remember:
->
-> - The single approver in the “Need Approval From” must be from the Security team. 
-> - Maintain the “Status” property accordingly. An FMA document can have the following statuses:
->   - **Draft 📝:** Doc is created but not yet ready for review.
->   - **In Review 🔎:** Security is reviewing, and Engineering is iterating on the design. A checklist of action items will be created during this phase.
->   - **Implementing Actions 🛫:** Security has signed off on the content of the document, including the resulting action items. Engineering is responsible for implementing the the action items, and updating the checklist.
->   - **Final 👍:** Security will transition the status of the document to Final once all action items are completed.
+| Need Approval From | Ben Clabby, Mark Tyneway |
+| Status | Draft |
 
 > [!TIP]
 > Guidelines for writing a good analysis, and what the reviewer will look for:
@@ -48,9 +36,14 @@
 
 This document covers the initial deployment of [the Asterisc fault proof VM](https://github.com/ethereum-optimism/asterisc) with [the Kona fault proof program](https://github.com/anton-rs/kona).
 
+Together, Kona and Asterisc form an alternative proof stack to the [op-program]() and [cannon](). A secondary dispute game will be created that uses kona and asterisc, with the `op-challenger` supporting playing _both_ game types - `op-program` + `cannon` as well as `kona` + `asterisc`.
+
 Below are references for this project:
 
 - ["Stage 1.4" internal project document](https://www.notion.so/oplabs/Stage-1-4-Partway-a6f57ad777b148dda01488f2646cff17)
+- [Kona Documentation](https://anton-rs.github.io/kona/).
+- [Kona Repository](https://github.com/anton-rs/kona).
+- [Asterisc Repository](https://github.com/ethereum-optimism/asterisc)
 
 
 ## Failure Modes and Recovery Paths
--- a/security/fma-kona-asterisc.md
+++ b/security/fma-kona-asterisc.md
@@ -149,7 +149,7 @@ Below are references for this project:
   The `op-challenger` is architected in such a way as to isolate game types and more so the individual game players.
   To allow the challenger to scale and play many different game types at once, a list of game types to play is passed into the `op-challenger` as a CLI flag.
   When games are detected on-chain, the `op-challenger` will create a "game player" for this game type.
-  Players are run in individual threads so a breaking `kona` + `asterisc` game player in the challenger will not cause a liveliness issue for the `op-challenger` to play other game types.
+  Players are run in individual threads so a breaking `kona` + `asterisc` game player in the challenger will not cause a liveness issue for the `op-challenger` to play other game types.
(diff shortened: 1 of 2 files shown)


pauldowman:
We can share this branch to make edits as we think of things. (No rebasing and force pushing until it's done.)

refcell:
I've taken this on and updated. The initial draft is ready for review.

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,111 @@
...
+
+
+### Breaking Backwards Compatibility for the op-challenger
+
+- **Description:** The `op-challenger` has been updated to support the Kona + Asterisc Fault Dispute Game type ([Game Type 3](https://github.com/ethereum-optimism/optimism/blob/develop/op-challenger/game/fault/types/types.go#L30)).
+
+- **Risk Assessment:** Low likelihood as the action tests that are run in CI very frequently run the `op-challenger` with the kona + asterisc fault dispute game. Medium impact since the off-chain `op-challenger` service can be updated with enough buffer time due to the large game time and nature of the chess clock.
+
+- **Mitigations:** The `op-challenger` is architected in such a way as to isolate game types and moreso the inididual game players. To allow the challenger to scale and play many different game types at once, a list of game types to play is passed into the `op-challenger` as a CLI flag. When games are detected on-chain, the `op-challenger` will create a "game player" for this game type. Players are run in individual threads so a breaking kona-asterisc game player in the challenger will not cause a liveliness issue for the `op-challenger` to play other game types.
```
clabby, line 77: Another big mitigation here is the VM runner. We've been running kona + asterisc with op-challenger for months now in k8s.

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,111 @@
...
+| Status | Draft |
+
+
+## Introduction
+
+This document covers the initial deployment of [the Asterisc fault proof VM](https://github.com/ethereum-optimism/asterisc) with [the Kona fault proof program](https://github.com/anton-rs/kona).
+
+Together, Kona and Asterisc form an alternative proof stack to the [op-program](https://github.com/ethereum-optimism/optimism/tree/develop/op-program) and [cannon](https://github.com/ethereum-optimism/optimism/tree/develop/cannon). A secondary dispute game will be created that uses kona and asterisc, with the `op-challenger` supporting playing _both_ game types - `op-program` + `cannon` as well as `kona` + `asterisc`.
+
+Since two main components are being shipped in Stage 1.4 there are two main places the system can break: the kona fault proof program, and the asterisc VM. The below Failure Modes and Recovery Paths go over those situations.
```
clabby, line 32: I think the `op-challenger` integration is notable enough to mention as a primary failure mode.

clabby (commented):
This looks great so far, thanks for getting it started

A couple notes:
1. On the mitigation sections, it's nice to be able to link to some of the tests for security to quickly review. For kona, we have codecov, action tests, etc., and for asterisc, we have the [test suite](https://github.com/ethereum-optimism/asterisc/tree/master/tests)
2. I think it's also worth highlighting the failure mode if kona / asterisc is being used as the active dispute game. i.e., the situation where we fall back to kona/asterisc due to a failure in `op-program`, and then `kona` or `asterisc` fails too. This should look really similar to the permissioned game fallback runbook.
3. Mostly a templated failure mode in all other FMAs with contract upgrades, but should probably add a section for the contract upgrade risk.

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,111 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Introduction](#introduction)
```
tynes, line 4: nit: the toc hasn't been updated

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,111 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Stage 1.4: Kona and Asterisc FMA (Failure Modes and Recovery Path Analysis)
+
+| | |
+|--------|--------------|
+| Author | Andreas Bigger |
+| Created at | 2024-10-17 |
+| Initial Reviewers | *Reviewer Name 1, Reviewer Name 2* |
```
tynes, line 21: nit: initial reviewers are not filled in, they are likely the same as need approval from

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,111 @@
...
+## Failure Modes and Recovery Paths
+
+
+### Divergent Kona Derivation Pipeline
+
+- **Description:** The Kona derivation pipeline diverged in behavior from the op-program's derivation pipeline.
+
+- **Risk Assessment:** Low likelihood, low impact. A small divergence in derivation can cause invalid game results. But since kona + asterisc is in the coldpath, it doesn't have impact on anything in the production system.
+
+- **Mitigations:** Frequent action tests covering the derivation pipeline.
```
tynes, line 51: What exactly do you mean by frequent action tests? My understanding of action tests are the tests that we write to catch certain protocol behaviors/edge cases, they aren't fuzz tests that could catch net new behaviors

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Kona Execution Diverges from op-program
+
+- **Description:** The Kona fault proof program executes blocks in a stateless L2 block executor that uses optimism-revm as a backend. If optimism-revm or the L2 block executor diverge in behavior from the op-program, the kona + asterisc proof system will be invalid.
+
+- **Risk Assessment:** Low likelihood, low impact. A small divergence in execution behavior can cause invalid game results. But since kona + asterisc is in the coldpath, it doesn't have impact on anything in the production system.
+
+- **Mitigations:** State testing against op-revm. Frequent action tests against kona, which covers the L2 block executor. op-reth which uses the same op-execution environment as kona.
+
+- **Detection:** Since kona + asterisc is in the coldpath by default, a divergence would only be seen by action tests or local testing.
```
tynes, line 66: Syncing `op-reth` and watching for chain splits could be a way to detect this?

tynes (commented):
This document often uses the fact that asterisc/kona will not be in the hot path as a justification for low risk. My understanding of the goal of the project is to have a fallback for cannon/op-program that maintains the permissionless property. Therefore it is assumed that it may be part of the hot path, we should be planning for that case. What happens if it fails when its in the hot path? Line 94 touches on this, but not on the case there is a failure for both op-program/cannon and kona/asterisc

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,227 @@
...
+  Low likelihood, medium impact.
+  The contract upgrades are tested and very simple.
+  Contracts in the hot path are audited and have been in production so are unlikely to have critical bugs.
+  // TODO: impact
+
+- **Mitigations:** 
+
+- **Detection:**
+
+- **Recovery Path(s):**
```
refcell, line 102: @clabby I've added the contract upgrade section but am less familiar with superchain-ops. Do you think you could provide a bit more context here? Basically need to gauge what's the worst failure mode here, considering there could be a bug in the upgrade scripts, but **not** the contracts like the AnchorStateRegistry or the DisputeGameFactory. Is that a good consideration?

refcell:
> This document often uses the fact that asterisc/kona will not be in the hot path as a justification for low risk. My understanding of the goal of the project is to have a fallback for cannon/op-program that maintains the permissionless property. Therefore it is assumed that it may be part of the hot path, we should be planning for that case. What happens if it fails when its in the hot path? Line 94 touches on this, but not on the case there is a failure for both op-program/cannon and kona/asterisc

Good point. I've updated to add a section to `Failure Modes and Recovery Paths Analysis` detailing this situation. I believe with this, your point should be covered but let me know if there's any other holes you can poke!

security/fma-kona-asterisc.md
clabby: For this, I think it's generally fine to write it off as a bullet point. There's smart contract upgrade risk in every upgrade we do - don't think we need to get too granular.

Here's a good example: https://github.com/ethereum-optimism/design-docs/blob/main/security/fma-generic-hardfork.md#activation-failure-without-chain-halt-l1-transactions

security/fma-kona-asterisc.md
refcell: I left them out since I wasn't sure who would pick up from security and didn't want to forcefully perscribe it to someone, but [it looks like Matt Solomon and Ethnical are the assigned folks now](https://github.com/ethereum-optimism/security-pod/issues/177) so will update.

security/fma-kona-asterisc.md
refcell: Frequent here refers to them being frequently run in the kona repository since they are run on every pull request. I will update to be more specific.

pauldowman:
> This document often uses the fact that asterisc/kona will not be in the hot path as a justification for low risk. My understanding of the goal of the project is to have a fallback for cannon/op-program that maintains the permissionless property. Therefore it is assumed that it may be part of the hot path, we should be planning for that case. What happens if it fails when its in the hot path? Line 94 touches on this, but not on the case there is a failure for both op-program/cannon and kona/asterisc

Yes, and I think "hot path" isn't really a meaningful concept to talk about in here, because any of these failures will _only_ occur if we have switched the active game type to use Kona + Asterisc, i.e. it's "in the hot path", right? It's not going to be exercised if nobody is creating dispute games with it.

So it seems to me that "Kona and Asterisc Hot Path Failure" doesn't really add anything that's not covered in the other sections.

refcell:
> > This document often uses the fact that asterisc/kona will not be in the hot path as a justification for low risk. My understanding of the goal of the project is to have a fallback for cannon/op-program that maintains the permissionless property. Therefore it is assumed that it may be part of the hot path, we should be planning for that case. What happens if it fails when its in the hot path? Line 94 touches on this, but not on the case there is a failure for both op-program/cannon and kona/asterisc
> 
> Yes, and I think "hot path" isn't really a meaningful concept to talk about in here, because any of these failures will _only_ occur if we have switched the active game type to use Kona + Asterisc, i.e. it's "in the hot path", right? It's not going to be exercised if nobody is creating dispute games with it.
> 
> So it seems to me that "Kona and Asterisc Hot Path Failure" doesn't really add anything that's not covered in the other sections.

It sounds like your suggesting to remove the "Hot Path" section I added in response to the above is that correct? If so, then Mark's comment would be unsatisfied.

refcell:
This is a blocker for this FMA: https://github.com/ethereum-optimism/design-docs/pull/140

We need to reference `fma-generic-contracts.md` from this doc.

tynes (approved):

security/fma-kona-asterisc.md
mds1, line 1: This comment is a reminder for us to revisit https://github.com/ethereum-optimism/design-docs/pull/140#discussion_r1817428557 and https://github.com/ethereum-optimism/design-docs/pull/140#discussion_r1817421106 to make sure those are addressed

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+### Divergent Kona Derivation Pipeline
+
+- **Description:**
+  The `kona` derivation pipeline diverged in behavior from the `op-program`'s derivation pipeline.
+
+- **Risk Assessment:**
+  Low likelihood, low impact.
+  Low likelihood as the derivation pipeline is extensively unit tested, run in the VM Runner, and is action tested.
+  Low impact since by default `kona` + `asterisc` are placed in the cold path.
+  [The "Kona and Asterisc Hot Path Failure"](#-Kona-and-Asterisc-Hot-Path-Failure) above covers the failure mode for `kona` + `asterisc` when placed in the hot path.
```
Ethnical, line 95: Need to remove the - to fix the link.
```suggestion
  [The "Kona and Asterisc Hot Path Failure"](#Kona-and-Asterisc-Hot-Path-Failure) above covers the failure mode for `kona` + `asterisc` when placed in the hot path.
```

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+
+- **Detection:**
+  The VM runner would pick up an issue.
+  The action tests or local testing may pick up an issue.
+  Use of the derivation elsewhere, for example in alternative consensus clients like `op-rs`, would pick up an issue.
+
+- **Recovery Path(s):**
+  Fix the derivation pipeline divergence, release and deploy the updated kona + asterisc when in the coldpath.
+  If `kona` + `asterisc` is placed in the hotpath, either the `op-program` + `cannon` would need to be falled back on otherwise fallback to the permissed dispute game.
+  [The "Kona and Asterisc Hot Path Failure"](#-Kona-and-Asterisc-Hot-Path-Failure) above covers the failure mode for `kona` + `asterisc` when placed in the hot path.
```
Ethnical, line 110: Need to remove the - to fix the link.
```suggestion
  [The "Kona and Asterisc Hot Path Failure"](#Kona-and-Asterisc-Hot-Path-Failure) above covers the failure mode for `kona` + `asterisc` when placed in the hot path.
```

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+| Need Approval From | Ben Clabby, Mark Tyneway |
+| Status | In Review |
+
+
+## Introduction
+
+This document covers the initial deployment of [the Asterisc Fault Proof VM][asterisc] with [the Kona Fault Proof Program][kona].
+
+Together, Kona and Asterisc form an alternative proof stack to the [op-program] and [cannon].
+A secondary dispute game will be created that uses kona and asterisc.
```
Ethnical, line 35: >A secondary dispute game will be created that uses kona and asterisc.

Does this means that each time we call `create()` on the DGF this will also create a second game? 
Looks like gas heavy on L1 to me. Also,  hard to code something that manage both games in the same times.

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+- **Recovery Path(s):**
+  If `kona` + `asterisc` is in the coldpath, it would need to be fixed and re-deployed.
+  If `op-program` + `cannon` cannot be placed in the hotpath, and `kona` + `asterisc` fails, the permissioned fallback must be engaged.
+  [The "Kona and Asterisc Hot Path Failure"](#-Kona-and-Asterisc-Hot-Path-Failure) above covers the failure mode for `kona` + `asterisc` when placed in the hot path.
+
+
+### Breaking Backwards Compatibility for the op-challenger
+
+- **Description:**
+  The `op-challenger` has been updated to support the Kona + Asterisc Fault Dispute Game type ([Game Type 3]).
```
Ethnical, line 142: For personal knowledge what is the difference between `GameType 3` and `GameType 2` described here contains Asterisc in the name causing me to be confuse compare to cannon only having one gameType: 
https://github.com/ethereum-optimism/optimism/blob/39dc07942350899bc93b413a1688efd17048982f/op-challenger/game/fault/types/types.go#L29-L30

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+
+- **Risk Assessment:**
+  Low likelihood as the action tests that are run in CI very frequently run the `op-challenger` with the `kona` + `asterisc` fault dispute game.
+  Medium impact since the off-chain `op-challenger` service can be updated with enough buffer time due to the large game time and nature of the chess clock.
+
+- **Mitigations:**
+  The `op-challenger` is architected in such a way as to isolate game types and moreso the inididual game players.
```
Ethnical, line 149:
```suggestion
  The `op-challenger` is architected in such a way as to isolate game types and more so the individual game players.
```

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+### Breaking Backwards Compatibility for the op-challenger
+
+- **Description:**
+  The `op-challenger` has been updated to support the Kona + Asterisc Fault Dispute Game type ([Game Type 3]).
+
+- **Risk Assessment:**
+  Low likelihood as the action tests that are run in CI very frequently run the `op-challenger` with the `kona` + `asterisc` fault dispute game.
+  Medium impact since the off-chain `op-challenger` service can be updated with enough buffer time due to the large game time and nature of the chess clock.
```
Ethnical, line 146: Could you also confirm, that if the `op-challenger` is broken through bug. Honest other actors will attack the malicious game to get the free money (bond). 
So even if our challenger is down, and there is sufficient honest actor in the wild this is not suppose to be an issue correct?

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+
+- **Mitigations:**
+  The `op-challenger` is architected in such a way as to isolate game types and moreso the inididual game players.
+  To allow the challenger to scale and play many different game types at once, a list of game types to play is passed into the `op-challenger` as a CLI flag.
+  When games are detected on-chain, the `op-challenger` will create a "game player" for this game type.
+  Players are run in individual threads so a breaking `kona` + `asterisc` game player in the challenger will not cause a liveliness issue for the `op-challenger` to play other game types.
+  Another key mitigation is the VM runner. This is an offline simulation that runs asterisc + kona with the op-challenger. It's been running since around August 2024.
+
+- **Detection:**
+  The Dispute Monitor ([op-dispute-mon] service provides an isolated monitoring service that will detect if a game is not being played by the honest `op-challenger`.
```
Ethnical, line 156: > a game is not being played by the honest op-challenger

What does that means exactly? I thought that the `op-challenger` will challenge the invalid game by attacking them as the `CHALLENGER`. 
If this is true what do you mean by **played by the honest `op-challenger`** sounds to me that only the game that are played and created by the `op-challenger` would be considered as honest. 
But this is not possible except if this similar to a PermissionedGame and only the `challenger` can create and play the game?

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+  Asterisc already runs through action tests frequently in CI validiting that it produces the expected output.
+
+- **Detection:**
+  Since `op-dispute-mon` uses the same asterisc backend that the `op-challenger` does, there may be divergent behaviour between asterisc and `RISCV.sol`.
+  In the worst case, if a silent issue is exploited in `RISCV.sol`, a fault dispute game may resolve incorrectly.
+  But since the kona + asterisc fault dispute game lives outside the hotpath, this game result is not used anywhere by the portal and can be discarded.
+
+- **Recovery Path(s):**
+  If a divergence between `RISCV.sol` and asterisc is found, either or both would need to be updated, and re-released.
+  Since the kona + asterisc Fault Dispute Game would need to use the updated `RISCV.sol`, if it was fixed, an upgrade would need to be performed.
```
Ethnical, line 189: This will invalidate all the current games currently opened within the window right? 
If **yes**, maybe we can mention it here. As the recovery path, is cancelling game of users the users have to resubmit withdrawals once the `gameType` changed/fallback.

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+  [Extensive test coverage](https://app.codecov.io/gh/anton-rs/kona/tree/main/crates) of kona.
+
+- **Detection:**
+  The VM runner would pick up an issue.
+  The action tests or local testing may pick up an issue.
+  Use of the derivation elsewhere, for example in alternative consensus clients like `op-rs`, would pick up an issue.
+
+- **Recovery Path(s):**
+  Fix the derivation pipeline divergence, release and deploy the updated kona + asterisc when in the coldpath.
+  If `kona` + `asterisc` is placed in the hotpath, either the `op-program` + `cannon` would need to be falled back on otherwise fallback to the permissed dispute game.
```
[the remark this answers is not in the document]
clabby: Not sure this is right either. I think the original intention was to say that if kona/asterisc fails in the hot-path, we can either fall back to `op-program`/`cannon` or fall back to the permissioned dispute game.

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+[op-program]: https://github.com/ethereum-optimism/optimism/tree/develop/op-program
+
+
+## Failure Modes and Recovery Paths
+
+### Kona and Asterisc Hot Path Failure
+
+- **Description:**
+  The `op-program` + `cannon` dispute game breaks and the `kona` + `asterisc` fault dispute game is placed in the hot path.
+  If `kona` + `asterisc` also fail, the permissioned fallback will need to be activated.
```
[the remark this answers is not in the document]
clabby: This is described within the detection sub-section of this section. `op-challenger` and `op-dispute-mon` would notify us.

The hot-path switch is _not_ automatic. It requires a manual intervention, exactly like the permissioned game fallback.

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Divergent Kona Derivation Pipeline
+
+- **Description:**
+  The `kona` derivation pipeline diverged in behavior from the `op-program`'s derivation pipeline.
+
+- **Risk Assessment:**
+  Low likelihood, low impact.
+  Low likelihood as the derivation pipeline is extensively unit tested, run in the VM Runner, and is action tested.
+  Low impact since by default `kona` + `asterisc` are placed in the cold path.
```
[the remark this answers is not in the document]
clabby: The document discusses this, and separates the failure modes depending on the active game - is it unclear?

TL;DR, the hot path is the active game that the bridge contract is listening to (i.e., `respectedGameType`.) The cold path is a latent state, like the `PermissionedDisputeGame` sits in most of the time. They're registered and available to use, but not directly listened to unless activated by the FND via. `DeputyGuardianModule.setRespectedGameType`.

security/fma-kona-asterisc.md
clabby: No. This release does not enable multi-proofs, kona/asterisc will exist as a fully separate dispute game, like the `PermissionedDisputeGame`.

security/fma-kona-asterisc.md
clabby: `AsteriscGameType` = `2` = `op-program` + `asterisc`
`AsteriscKonaGameType` = `3` = `kona` + `asterisc`

security/fma-kona-asterisc.md
clabby: Yes. This is the exact same model as the `op-program` + `cannon`, if the challenger goes down.

security/fma-kona-asterisc.md
clabby: Same as the `op-program` + `cannon` dispute game, if the honest actor does not participate, the game can be finalized incorrectly. The fault proof system relies on the presence of the honest actor.

> PS: Do we also plan to run a canary to make sure the liveness of the dispute-mon is correct with Asterisc FP program?

Yes. We will do the same thing that we do for the existing game. Also, we run the challenger, kona, and asterisc in the [VM Runner](https://optimistic.grafana.net/d/bdtk5xud0cg00e/vm-runner?from=now-2d&to=now&timezone=browser&var-namespace=$__all), which continuously integrates the components.

security/fma-kona-asterisc.md
clabby: Not necessarily. If the bug hasn't been exploited, we can upgrade the implementation and allow for a roll-over. However, we could also fall back to the permissioned game, fix the bug, and then go back to the permissionless proofs, and that would indeed invalidate all ongoing games.

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+
+- **Mitigations:**
+  Action tests covering the derivation pipeline are run on every pull request in the [kona] repository.
+  VM Runner exectutes the derivation pipeline as part of the kona fault proof program running ontop of asterisc.
+  [Extensive test coverage](https://app.codecov.io/gh/anton-rs/kona/tree/main/crates) of kona.
+
+- **Detection:**
+  The VM runner would pick up an issue.
+  The action tests or local testing may pick up an issue.
+  Use of the derivation elsewhere, for example in alternative consensus clients like `op-rs`, would pick up an issue.
```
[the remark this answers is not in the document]
clabby: The client doesn't exist yet. This is going to be a future component that adds more coverage to the derivation pipeline used within `kona`.

security/fma-kona-asterisc.md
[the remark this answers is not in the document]
Ethnical: Thanks confirming here!
> - is it unclear?

It was the beginning of the review for me and I was bit confuse but just to make sure this is the desired concept I prefer to double ask.

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+
+- **Recovery Path(s):**
+  If the `op-challenger` fails to play `kona` + `asterisc` fault dispute games, a fix will need to be made to the `op-challenger`.
+  It does not require governance, but would need a new release of the `op-challenger` to be published and rolled out to infrastructure.
+
+[op-dispute-mon]: https://github.com/ethereum-optimism/optimism/tree/develop/op-dispute-mon
+[Game Type 3]: https://github.com/ethereum-optimism/optimism/blob/develop/op-challenger/game/fault/types/types.go#L30
+
+
+### Divergence between RISCV.sol and Asterisc
```
Ethnical, line 167: NIT: It would be nice to add the link the current repository, that are doing the differential fuzzing and testing to ensure there is no difference between the computation.

Ethnical (approved):
Overall looks great to me!
can't wait  having *multi-proofs* on L1 on day 🚀

security/fma-kona-asterisc.md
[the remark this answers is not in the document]
mds1: This makes sense and is similar to our existing detection and recovery, so will resolve this

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+
+- **Description:**
+  The `op-program` + `cannon` dispute game breaks and the `kona` + `asterisc` fault dispute game is placed in the hot path.
+  If `kona` + `asterisc` also fail, the permissioned fallback will need to be activated.
+
+- **Risk Assessment:**
+  Low likelihood, high impact.
+  Low likelihood since `op-program` + `cannon` are already running in production and have meaningful improvements on the way that will increase resilience.
+  If `kona` + `asterisc` fails while in the hotpath, the impact is high since the permissioned fallback will need to be triggered.
+  This engages the security council, requires comms, and causes the chain to fallback to a weaker security model.
```
mds1, line 69: A small clarification here is that the deputy guardian (foundation) can change the respected game type without involving the security council, so they are not necessarily involved

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+
+- **Mitigations:**
+  Offline testing with the VM Runner for both `kona` + `asterisc` as well as `op-program` + `cannon`.
+  The VM Runner also fully tests the integration of the `op-challenger` with these dispute games ([covered in a later section](#Breaking-Backwards-Compatibility-for-the-op-challenger)).
+  Action tests ensuring both proof systems follow spec.
```
mds1, line 74: What are action tests? I've heard them mentioned a lot but am not familiar with them

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+| Status | In Review |
+
+
+## Introduction
+
+This document covers the initial deployment of [the Asterisc Fault Proof VM][asterisc] with [the Kona Fault Proof Program][kona].
+
+Together, Kona and Asterisc form an alternative proof stack to the [op-program] and [cannon].
+A secondary dispute game will be created that uses kona and asterisc.
+The [op-challenger] supports playing _both_ game types - `op-program` + `cannon` as well as `kona` + `asterisc`.
```
mds1, line 36: For my own understanding: Is `op-program` coupled to `cannon` and `kona` coupled to `asterisc`, or can they be mixed, e.g. run `op-program with `asteristc`? If coupled, what makes them coupled and do we plan to decouple at some point, or do they have to be coupled for some reason?

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+  At this point, the permissioned fallback would need to be engaged since both proof systems are in repair.
+
+- **Recovery Path(s):**
+  Likely the only viable recovery path would be the permissioned fallback unless the `op-program` + `cannon` are fixed and able to be placed back into the hotpath.
+
+
+### Divergent Kona Derivation Pipeline
+
+- **Description:**
+  The `kona` derivation pipeline diverged in behavior from the `op-program`'s derivation pipeline.
```
mds1, line 89: When I hear "derivation pipeline" I think of deriving the L2 state from the data posted to L1, but it sounds like this is a different meaning of derivation pipeline that I'm not familiar with?

security/fma-kona-asterisc.md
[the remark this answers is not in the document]
mds1: Let's clarify that in the FMA, since this is a future detection method and not a current one

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+
+- **Risk Assessment:**
+  Low likelihood, low impact.
+  Low likelihood as the derivation pipeline is extensively unit tested, run in the VM Runner, and is action tested.
+  Low impact since by default `kona` + `asterisc` are placed in the cold path.
+  [The "Kona and Asterisc Hot Path Failure"](#Kona-and-Asterisc-Hot-Path-Failure) above covers the failure mode for `kona` + `asterisc` when placed in the hot path.
+
+- **Mitigations:**
+  Action tests covering the derivation pipeline are run on every pull request in the [kona] repository.
+  VM Runner executes the derivation pipeline as part of the kona fault proof program running on top of asterisc.
```
mds1, line 99: What does this mean exactly? Is the VM Runner something we're running off chain to verify equivalence against the existing op-program + cannon?

security/fma-kona-asterisc.md
mds1, line 89: Is the implication here that op-program's pipeline is the spec? Or is it possible they diverge because kona is correct and op-program has a bug or doesn't properly implement the spec?

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+  Fix the derivation pipeline divergence, release and deploy the updated kona + asterisc when in the coldpath.
+  If `kona` + `asterisc` is placed in the hotpath, the system operator would either need to fall back to `op-program` and `cannon`, or fall back to the permissioned dispute game.
+  [The "Kona and Asterisc Hot Path Failure"](#Kona-and-Asterisc-Hot-Path-Failure) above covers the failure mode for `kona` + `asterisc` when placed in the hot path.
+
+
+### Kona Execution Diverges from op-program
+
+- **Description:**
+  The Kona Fault Proof Program executes blocks in a stateless L2 block executor that uses optimism-revm as a backend.
+  If optimism-revm or the L2 block executor diverge in behavior from the `op-program`, the `kona` + `asterisc` proof system will be invalid.
```
mds1, line 117: Similar question to https://github.com/ethereum-optimism/design-docs/pull/132/files#r1829865714

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+
+- **Risk Assessment:**
+  Low likelihood, low impact.
+  A small divergence in execution behavior can cause invalid game results.
+  But since kona + asterisc is in the coldpath, it doesn't have impact on anything in the production system.
+
+- **Mitigations:**
+  State testing against op-revm.
```
mds1, line 125: For the various detections and mitigations listed in this doc, "state testing", "action tests", op-rs, monitoring, etc, we should either:
- Link out to, or name, the specific tests used, or
- Track the implementation of that mitigation in the Action Items section

It looks like currently there are no action items which would imply all detections and mitigations are implemented

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+  Since the kona + asterisc Fault Dispute Game would need to use the updated `RISCV.sol`, if it was fixed, an upgrade would need to be performed.
+  If the kona + asterisc fault dispute game sits in the hotpath, the security council would need to fallback to the op-program + cannon fault dispute
+  game, and only switch back once the fix has been released and deployed.
+
+
+### Generic items we need to take into account:
+
+- [X] [`DisputeGameFactory.setImplementation`](./fma-generic-hardfork.md#invalid-disputegamefactorysetimplementation-execution) (in [./fma-generic-hardfork.md](./fma-generic-hardfork.md))
+- [X] [./failure-modes-analysis.md](./failure-modes-analysis.md).
+- [X] [./fma-generic-contracts.md](./fma-generic-contracts.md)
```
mds1, line 198: Copying from https://github.com/ethereum-optimism/design-docs/pull/140#discussion_r1817428557:

> A lot of the sentences here [in `./fma-generic-contracts.md`] are very terse, it would be great if we can be a little more detailed to help simplify the work FMA authors will have to do per https://github.com/ethereum-optimism/design-docs/pull/140/files?diff=unified&w=0#r1817421106. For example, what tools might break, what kinds of users might report this, how can we validate ABI diffs, etc?

Let's make sure to flesh out the generic contracts FMA section a bit more per that comment. Since you are the first FMA to use this unfortunately that burden fell to you, but I do appreciate you adding it!

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,213 @@
...
+  But since the kona + asterisc fault dispute game lives outside the hotpath, this game result is not used anywhere by the portal and can be discarded.
+
+- **Recovery Path(s):**
+  If a divergence between `RISCV.sol` and asterisc is found, either or both would need to be updated, and re-released.
+  Since the kona + asterisc Fault Dispute Game would need to use the updated `RISCV.sol`, if it was fixed, an upgrade would need to be performed.
+  If the kona + asterisc fault dispute game sits in the hotpath, the security council would need to fallback to the op-program + cannon fault dispute
+  game, and only switch back once the fix has been released and deployed.
+
+
+### Generic items we need to take into account:
```
mds1, line 194: From @clabby's suggestion https://github.com/ethereum-optimism/design-docs/pull/140#discussion_r1817421106, is there somewhere in here that discusses the FMA-specific mitigations for the generic items listed here?

security/fma-kona-asterisc.md
clabby: Right now, they are coupled. There's not much that enforces this - we could run `op-program` on `asterisc` and `kona` on `cannon`. However, the two languages require much different things. `kona` needs a much less fully-featured VM than `op-program`, and so we'll likely continue in the direction of keeping the VMs coupled.

This may be revisited, but for now, we should treat them as coupled.

security/fma-kona-asterisc.md
clabby: This is exactly what it is. The `op-program` uses `op-node`'s derivation pipeline, and `kona` uses its own.

security/fma-kona-asterisc.md
clabby: VM Runner is a service we execute in our infra that runs the fault proof programs on the fault proof VMs end to end using live chain data, and checks expected results. By running the programs, it adds coverage to the derivation code.

security/fma-kona-asterisc.md
clabby: Yes, it is the "spec." The derivation specs are mostly incomplete on https://specs.optimism.io, so it is treated as the reference.

`op-program` can diverge from the written spec, but at least at the moment, that would become canonical. This has already happened, actually, with a receipts root bug in `op-geth` in the Regolith hardfork.

security/fma-kona-asterisc.md
clabby: Responded there.

security/fma-kona-asterisc.md
clabby: Action tests are a shared set of tests we run on both the `op-program` and `kona` to check for equivalent behavior in edge cases. They're really useful because they allow for us to control the batcher, chain progression, etc. This allows us to do things like test channel timeouts, precise batch inclusion, etc.

security/fma-kona-asterisc.md
mds1: Links have been added, let's just added the standard Action Items header and initial item that way we know we considered actions even if we have none because they are already done

security/fma-kona-asterisc.md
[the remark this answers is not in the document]
clabby: Done

security/fma-kona-asterisc.md
clabby: Done

security/fma-kona-asterisc.md
clabby: Done

security/fma-kona-asterisc.md
clabby: Done

security/fma-kona-asterisc.md
```diff
@@ -0,0 +1,272 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Stage 1.4: Kona and Asterisc FMA (Failure Modes and Recovery Path Analysis)](#stage-14-kona-and-asterisc-fma-failure-modes-and-recovery-path-analysis)
```
mds1, line 4: Looks like the TOC is out of date

mds1 (approved):
Approved pending table of contents update, and we can change the status to Implementing Actions

PR #138, opened 2024-10-23
feat: only allow challenger to play permissioned game

smartcontracts:
Proposes a change to the PermissionedDisputeGame such that the Proposer will no longer be allowed to play the game. Instead, only the Challenger will be allowed to play the game. This more clearly mirrors the behavior of the L2OutputOracle.

### DIFF
--- a/protocol/proofs/pdg-no-proposer-play.md
+++ b/protocol/proofs/pdg-no-proposer-play.md
@@ -1,0 +1,51 @@
+# Limit PermissionedDisputeGame Play to Challenger
+
+## Context
+
+The `PermissionedDisputeGame` is a version of the `FaultDisputeGame` that limits the parties that
+are allowed to execute moves within the game. The `PermissionedDisputeGame` is designed to act as
+either a fallback for the `FaultDisputeGame` or an independent game type if a chain is not using
+the permissionless `FaultDisputeGame` model. The `PermissionedDisputeGame` is intended to mirror
+the behavior of the legacy `L2OutputOracle` whereby an authorized `Proposer` can create proposals
+and an authorized `Challenger` can remove those proposals.
+
+## Problem Statement
+
+The `PermissionedDisputeGame` currently allows *both* the `Proposer` and the `Challenger` to
+participate in the game and create moves. The original intent of this behavior was to have the
+`PermissionedDisputeGame` act as a fully-fledged allowlisted dispute game. However, a number of
+changes to product needs and security modeling have popped up.
+
+- The concept of an allowlisted dispute game seems less important as the permissionless game
+  becomes increasingly battle-tested. Current product need for an allowlisted game is unclear and
+  the need for a such a game where only two participants can play is even less clear.
+- The `Proposer` tends to be a hot wallet which is easier to compromise compared to the
+  `Challenger` which is usually a multisig. In the case that the `PermissionedDisputeGame` is being
+  used as a fallback mechanism, the game is likely being used because the `FaultDisputeGame` has a
+  known bug. If the `Proposer` is compromised then it could be used to try to exploit this known
+  bug, which ultimately defeats the purpose of the fallback.
+
+We therefore propose to modify the `PermissionedDisputeGame` to resolve these issues.
+
+## Proposed Solution
+
+Given the above developments, this design doc proposes to remove the ability for the `Proposer` to
+submit moves. This would have the effect of aligning the `PermissionedDisputeGame` with the
+behavior of the original `L2OutputOracle` where a challenge by the `Challenger` role has ultimate
+authority.
+
+If future product research demonstrates a clear need for an allowlisted dispute game, we would
+recommend that an entirely new game type be written (as opposed to repurposing the
+`PermissionedDisputeGame`).
+
+## Risks & Uncertainties
+
+### Security Modeling
+
+This proposal slightly changes the security model of the `PermissionedDisputeGame` so that the
+`Proposer` can no longer counteract a malicious `Challenger`. However, this is exactly the security
+model of the `L2OutputOracle` and therefore more accurately mirrors what existing chain operators
+think that a permissioned proof model would look like.
+
+As previously stated, if there is clear product need for an allowlisted and fully fledged dispute
+game, we would recommend creating a new game type that clearly supports this behavior.


ajsutton (approved):
Makes sense to me.  Worth noting that [the spec](https://specs.optimism.io/fault-proof/stage-one/bridge-integration.html#permissioned-faultdisputegame) currently indicates that only the challenger can call attack, defend, step, resolve, resolveClaim and addLocalData.  As part of this we should update the spec to indicate that attack, defend and step are restricted to the CHALLENGER and resolve, resolveClaim remains permissionless (important so that op-challenger can call it and resolve games created by op-proposer).  Not sure if addLocalData should be permissioned on not, it's currently unrestricted which is probably fine since it only matters if you call step anyway.

protocol/proofs/pdg-no-proposer-play.md
```diff
@@ -0,0 +1,51 @@
...
+  bug, which ultimately defeats the purpose of the fallback.
+
+We therefore propose to modify the `PermissionedDisputeGame` to resolve these issues.
+
+## Proposed Solution
+
+Given the above developments, this design doc proposes to remove the ability for the `Proposer` to
+submit moves. This would have the effect of aligning the `PermissionedDisputeGame` with the
+behavior of the original `L2OutputOracle` where a challenge by the `Challenger` role has ultimate
+authority.
```
maurelian, line 35: What I find confusing is that the Permissioned game still has the full FDG interface and logic, just with different auth. 

If I understand correctly, the challenger still needs to play out the full game, which would be quite difficult for a multisig. Should we also enable the `Challenger` to invalidate the game with a single call, which was the case with the `L2OutputOracle`?

protocol/proofs/pdg-no-proposer-play.md
```diff
@@ -0,0 +1,51 @@
...
+the permissionless `FaultDisputeGame` model. The `PermissionedDisputeGame` is intended to mirror
+the behavior of the legacy `L2OutputOracle` whereby an authorized `Proposer` can create proposals
+and an authorized `Challenger` can remove those proposals.
+
+## Problem Statement
+
+The `PermissionedDisputeGame` currently allows *both* the `Proposer` and the `Challenger` to
+participate in the game and create moves. The original intent of this behavior was to have the
+`PermissionedDisputeGame` act as a fully-fledged allowlisted dispute game. However, a number of
+changes to product needs and security modeling have popped up.
```
maurelian, line 17: By allowlisted, I assume you mean that the set of allowed players can be arbitrarily expanded, rather than fixed at 2.

protocol/proofs/pdg-no-proposer-play.md
maurelian: OK, looking again, assuming that we change `onlyAuthorized` to accept calls from only the `CHALLENGER`, then the `PROPOSER` cannot respond to a `move()` from the `CHALLENGER`, and the game can never resolve, effectively making it invalid. Is that right?

maurelian (commented):
Left some comments to confirm my understanding, but this generally LGTM.

protocol/proofs/pdg-no-proposer-play.md
smartcontracts: Challenger can execute literally any attack on the root claim and proposer would not be able to respond, so claim would resolve in favor of the challenger.

protocol/proofs/pdg-no-proposer-play.md
smartcontracts: Correct

protocol/proofs/pdg-no-proposer-play.md
maurelian: Got it. It still feels confusing that there is so much unnecessary logic in here. It requires a lot of knowledge of how the entire system works to recognize that anny `move()` by the `CHALLENGER` immediately invalidates the game.   

IMO the `PermissionedGame` would be much easier to reason about if we simply override  to `move()` function to immediately set the state to `CHALLENGER_WINS`.

ajsutton: I think if we're going to change the game semantics like that we should just replace the PermissionedGame with something completely custom. It could probably auto-resolve itself after 3.5 days without a challenge as well (just change what it returns from the `status()` method based on timestamp rather than needing an additional transaction).

That said, it's important to note that to understand any fault dispute game you need to realise that it doesn't have to be played out to the max depth. Every proposal depends on that because every proposal creates a game with a single root claim and if it goes unchallenged it wins. It's not actually expected to ever play a game out to max depth because the incentives are set to discourage making invalid claims.

smartcontracts: Yeah like @ajsutton said, the main benefit here is that we don't need to add a bunch of custom code. We're reusing the fault dispute game under the hood. I'd really prefer not to add a completely custom dispute game (yet).

protocol/proofs/pdg-no-proposer-play.md
```diff
@@ -0,0 +1,51 @@
...
+  used as a fallback mechanism, the game is likely being used because the `FaultDisputeGame` has a
+  known bug. If the `Proposer` is compromised then it could be used to try to exploit this known
+  bug, which ultimately defeats the purpose of the fallback.
+
+We therefore propose to modify the `PermissionedDisputeGame` to resolve these issues.
+
+## Proposed Solution
+
+Given the above developments, this design doc proposes to remove the ability for the `Proposer` to
+submit moves. This would have the effect of aligning the `PermissionedDisputeGame` with the
```
mds1, line 33: So the proposer can still `step`, and the diff is to remove the `onlyAuthorized` modifier from `move` and replace it with `if (msg.sender != CHALLENGER)) revert BadAuth()`? Or is the diff to change the `onlyAuthorized` modifier so the auth change applies to `step` also? Whichever is the answer, what is the rationale?

protocol/proofs/pdg-no-proposer-play.md
mds1: > I'd really prefer not to add a completely custom dispute game (yet).

Can you expand on the "yet" here? How much bigger is the lift to trim down the permissioned game instead of having it inherit from the permissionless game? The permissioned game sounds like it would be pretty simple, so I'm wondering about the tradeoff here. What makes us lean in favor of this stopgap solution instead of the (seemingly) preferred custom solution?

protocol/proofs/pdg-no-proposer-play.md
ajsutton: Both `step` and `move` should be restricted to the CHALLENGER.  The proposer should be able to create games but not participate in them.  `step` and `move` are essentially the same concept of "counter this claim" just with implementation differences based on whether we need to actually execute the VM on chain to determine who's right or if we let the game continue playing out.

So only the proposer can create games, only the challenger can counter claims.

clabby (approved):
High-level idea lgtm here. Needs an amendment to the spec, but should be pretty simple.

protocol/proofs/pdg-no-proposer-play.md
smartcontracts: Basically any new contract is going to need a design doc, spec, implementation, probably an audit, etc etc. If we just make a tiny tweak to the existing contract we can get the exact same behavior immediately without doing any of that work and it ends up being a 1 line change in an audit. Just trying to reduce the scope of all of this work because we already have a lot of stuff involved in the incident response improvements project.

tynes:
Merging because this has 2 approvals

PR #143, opened 2024-10-28
Standardize on Custom errors

ControlCplusControlV:
**Description**

Set a clear standard on how to handle errors for OP Smart Contracts, using only custom errors and reverts.

**Tests**

No tests needed

**Additional context**

Carried over from discord, but the idea is to stop bikeshedding around using both require with error strings and custom errors and just standardized on custom errors, which I lay out in the doc. There was some concern around using require in scripts + tests, but after some discussion on discord we agreed `assertEq` should be able to cover that in most cases

**Metadata**
[optimism issue #12478](https://github.com/ethereum-optimism/optimism/issues/12478)

### DIFF
--- a/solidity/revert-strings-to-custom-error-migration.md
+++ b/solidity/revert-strings-to-custom-error-migration.md
@@ -37,3 +37,5 @@ would now become
         revert MIPS2_EmptyRightThreadStackError();
     }
 ```
+
+Additionally, all errors should live inside of the files they are used. This would mean the removal of `*Errors.sol` files, removing the need for files like `L1BlockErrors.sol`.
--- a/solidity/revert-strings-to-custom-error-migration.md
+++ b/solidity/revert-strings-to-custom-error-migration.md
@@ -1,0 +1,25 @@
+# Purpose
+
+The purpose of this document is to standardize all error messages / handling for the Optimism Smart Contracts, which current use a mix of error strings, and mix both `require` and `revert`.
+
+# Summary
+
+All error messages going forward should use the prefix of the `contract` name, and the suffix of `*Error`. Additionally, all errors should be returned via `revert` rather than through the use of `require`. This is both for visual clarity, and due to the efficiency concerns around `require` in the case of errors with arguments.
+
+# Problem Statement + Context
+
+The current problem is the Optimism smart contracts mix both `require` and `revert` statements inconsistently with error strings, and Solidity custom errors. For the sake of a consistent abi the Optimism smart contracts should converge on a single method for reverting.
+
+# Alternatives Considered
+
+## Require and Error Strings
+
+All errors could be handled using `require` statements with error strings, however the main issue with this approach is that error strings take up significantly more bytecode size, and given the small margins of the contract size for staying within the Spurious Dragon bytecode size limit, it would be very useful to cut down on contract size in favor of more optimizer runs.
+
+## Require and Custom Errors
+
+As of Solidity 0.8.24, `require` supports using custom errors, however due the limitation that `require` does not short-circuit (ie if there is a custom error which has a value passed in, and you pass in the result of some function call to that argument, the function call will be executed regardless of whether or not the error is returned) meaning that in all cases it is inefficient, and in some cases even dangerous.
+
+# Proposed Solution - Revert and Custom Errors
+
+All errors returned from Optimism Smart Contracts should use `revert` with custom errors for both efficiency and safety. In the case of scripts and tests which currently use `require` frequently, they should be instead replaced with `assertEq` or similar `assert*` where appropriate, including a string as a revert message. This will allow for semgrep rules to more easily enforce the no `require` constraint. Custom errors should also include the contract name as a prefix, and `Error` as a suffix for clarity and consistency.
--- a/solidity/revert-strings-to-custom-error-migration.md
+++ b/solidity/revert-strings-to-custom-error-migration.md
@@ -4,7 +4,7 @@ The purpose of this document is to standardize all error messages/handling for t
 
 # Summary
 
-Going forward, all error messages should use the prefix of the `contract` name and the suffix of `*Error`. Additionally, all errors should be returned via `revert` rather than `require`. This is both for visual clarity and due to the efficiency concerns around `require` in the case of errors with arguments.
+Going forward, all error messages should use the prefix of the `contract_` name and the suffix of `*Error`. Additionally, all errors should be returned via `revert` rather than `require`. This is both for visual clarity and due to the efficiency concerns around `require` in the case of errors with arguments.
 
 # Problem Statement + Context
 
@@ -22,7 +22,7 @@ As of Solidity 0.8.24, `require` supports using custom errors, however, due to t
 
 # Proposed Solution - Revert and Custom Errors
 
-All errors returned from Optimism Smart Contracts should use `revert` with custom errors for both efficiency and safety. In the case of scripts and tests which currently use `require` frequently, they should be instead replaced with `assertEq` or similar `assert*` where appropriate, including a string as a revert message. This will allow for semgrep rules to more easily enforce the no `require` constraint. Custom errors should also include the contract name as a prefix, and `Error` as a suffix for clarity and consistency.
+All errors returned from Optimism Smart Contracts should use `revert` with custom errors for both efficiency and safety. In the case of scripts and tests which currently use `require` frequently, they should be instead replaced with `assertEq` or similar `assert*` where appropriate, including a string as a revert message. This will allow for semgrep rules to more easily enforce the no `require` constraint. Custom errors should also include the contract name as a prefix, followed by an underscore, and `Error` as a suffix for clarity and consistency.
 
 For example, under this new policy, the following statement inside of MIPS2.sol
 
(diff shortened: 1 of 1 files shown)


solidity/revert-strings-to-custom-error-migration.md
```diff
@@ -0,0 +1,25 @@
+# Purpose
+
+The purpose of this document is to standardize all error messages/handling for the Optimism Smart Contracts, which currently use a mix of error strings, and mix both `require` and `revert`.
+
+# Summary
+
+Going forward, all error messages should use the prefix of the `contract` name and the suffix of `*Error`. Additionally, all errors should be returned via `revert` rather than `require`. This is both for visual clarity and due to the efficiency concerns around `require` in the case of errors with arguments.
```
smartcontracts, line 7: Can you provide an example?

solidity/revert-strings-to-custom-error-migration.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Require and Custom Errors
+
+As of Solidity 0.8.24, `require` supports using custom errors, however, due to the limitation that `require` does not short-circuit (ie if there is a custom error which has a value passed in, and you pass in the result of some function call to that argument, the function call will be executed regardless of whether or not the error is returned) meaning that in all cases it is inefficient, and in some cases even dangerous.
+
+# Proposed Solution - Revert and Custom Errors
+
+All errors returned from Optimism Smart Contracts should use `revert` with custom errors for both efficiency and safety. In the case of scripts and tests which currently use `require` frequently, they should be instead replaced with `assertEq` or similar `assert*` where appropriate, including a string as a revert message. This will allow for semgrep rules to more easily enforce the no `require` constraint. Custom errors should also include the contract name as a prefix, and `Error` as a suffix for clarity and consistency.
```
smartcontracts, line 25: Likely worth including a "risks and uncertainties" section as errors are part of the abi and changing errors may mean breaking external contracts or clients, unfortunately

smartcontracts: I would like to get a public policy doc merged that explains that revert data may break at any time

solidity/revert-strings-to-custom-error-migration.md
ControlCplusControlV: Added in a clear example [here](https://github.com/ethereum-optimism/design-docs/pull/143/commits/8f67890726d1fa5cd4848107a51d9d5b0b7a88b2)

solidity/revert-strings-to-custom-error-migration.md
ControlCplusControlV: Would you want this as a meta eg `POLICY.md` in something like `/meta` as a part of [this move](https://github.com/ethereum-optimism/optimism/issues/12670) ?

solidity/revert-strings-to-custom-error-migration.md
```diff
@@ -0,0 +1,39 @@
...
+
+```solidity
+            require(_state.rightThreadStack != EMPTY_THREAD_ROOT, "empty right thread stack");
+```
+
+would now become
+
+```solidity
+    if (_state.rightThreadStack == EMPTY_THREAD_ROOT) {
+        revert MIPS2EmptyRightThreadStackError();
```
tynes, line 37: Thoughts on adding a single underscore between the name of the contract and the error? ie `MIPS2_EmptyRightThreadStackError` ?

tynes:
There will be many opinions on this, it is important to not bikeshed so not going to die on a hill over this but I do think there is value in common error types like `Unauthorized`, under this we will get `SystemConfigUnauthorized` and `OptimismPortalUnauthorized` which is a different 4 byte selector for the same thing, I suppose that is fine given that the ABIs are uploaded to 4byte directories.

I also prefer to have a `_` in between the contract name and the error name, making it a bit easier to read. I do not think that we have any conditional logic based on returndata so there should be no internal backwards compatibility concerns. We have changed the returndata so many times at this point and nobody has complained, so i doubt that people are currently consuming the revert errors. The only time that I have every observed a revert personally was a deposit oog, with no good error message.

smartcontracts:
> I do think there is value in common error types like Unauthorized

One thing to note is that providing the contract name in the error means it's immediately obvious *where* an error was thrown.

> I also prefer to have a _ in between the contract name and the error name, making it a bit easier to read.

Yep totally down with this, easier to read.

solidity/revert-strings-to-custom-error-migration.md
smartcontracts: Yep

solidity/revert-strings-to-custom-error-migration.md
smartcontracts: +1

ControlCplusControlV: I like this a lot more, added to the doc

ControlCplusControlV:
I will pushback against common errors solely because it is nice to not have to run everything with `-vvv` in order to determine where an occurred, since a contract name in 90% of cases will make it immediately obvious where failure occurred

tynes:
Ok your arguments against common errors are good and I am on board with this standard

solidity/revert-strings-to-custom-error-migration.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Require and Custom Errors
+
+As of Solidity 0.8.24, `require` supports using custom errors, however, due to the limitation that `require` does not short-circuit (ie if there is a custom error which has a value passed in, and you pass in the result of some function call to that argument, the function call will be executed regardless of whether or not the error is returned) meaning that in all cases it is inefficient, and in some cases even dangerous.
+
+# Proposed Solution - Revert and Custom Errors
+
+All errors returned from Optimism Smart Contracts should use `revert` with custom errors for both efficiency and safety. In the case of scripts and tests which currently use `require` frequently, they should be instead replaced with `assertEq` or similar `assert*` where appropriate, including a string as a revert message. This will allow for semgrep rules to more easily enforce the no `require` constraint. Custom errors should also include the contract name as a prefix, followed by an underscore, and `Error` as a suffix for clarity and consistency.
```
mds1, line 25: Have we verified there are no forge issues with `assertEq` when used in `setUp()` or constructors?

solidity/revert-strings-to-custom-error-migration.md
```diff
@@ -0,0 +1,39 @@
...
+
+```solidity
+            require(_state.rightThreadStack != EMPTY_THREAD_ROOT, "empty right thread stack");
+```
+
+would now become
+
+```solidity
+    if (_state.rightThreadStack == EMPTY_THREAD_ROOT) {
+        revert MIPS2_EmptyRightThreadStackError();
```
mds1, line 37: What's the rationale for the `Error` suffix? My one nit would be to remove that to minimize verbosity, since we already know it's an error

solidity/revert-strings-to-custom-error-migration.md
ControlCplusControlV: I looked through open foundry issues and didn't see anything? Are there anything specific concerns you had in mind?

solidity/revert-strings-to-custom-error-migration.md
ControlCplusControlV: Mostly just clarity and to avoid any potential 4byte collision with similarly named functions, its not super important but I think its a nice addition on top for annoying cases like that

solidity/revert-strings-to-custom-error-migration.md
mds1: This was the original issue where assertions didn't work in `setUp` and I think in constructors too: https://github.com/foundry-rs/foundry/issues/1291

That was a long time ago, but I'd suggest setting up a quick test to verify that the `assertEq` methods still cause failures when hit in both `setUp` and constructors, just to be safe

solidity/revert-strings-to-custom-error-migration.md
mds1: Given that we prefix the errors with the contract name and an underscore, I wouldn't expect any name collisions. But I don't feel strongly and will not block on this so if people prefer the suffix we can keep it

smartcontracts: I'm ok with removing the suffix

ControlCplusControlV: I'll remove it for the sake of simplicity, I would hate to wake up one day at Optimism and realize I'm writing Java-level verbosity code

solidity/revert-strings-to-custom-error-migration.md
ControlCplusControlV: Re-ran your POC test
```
Ran 1 test suite in 124.97ms (414.42µs CPU time): 0 tests passed, 1 failed, 0 skipped (1 total tests)

Failing tests:
Encountered 1 failing test in test/Counter.t.sol:SetUpFailureTest
[FAIL. Reason: setup failed: assertion failed] setUp() (gas: 0)
```

Test now fails in the setUp process which is as expect and doesn't run the other tests, LGTM?

ControlCplusControlV:
Discussed in meeting #150 , consensus is approved but PR's should be done file by file with prior discussion with the protocol team

smartcontracts (approved):

PR #144, opened 2024-10-29
design doc for 'Always using _disableInitializers()  in constructors'

AmadiMichael:
This design doc proposes always using `_disableInitializers()` in proxied contracts constructors rather than the prevalent approach of calling the `initialize()` function from the constructor.

Issue: https://github.com/ethereum-optimism/optimism/issues/12634

### DIFF
--- a/protocol/disableInitializers-in-constructor.md
+++ b/protocol/disableInitializers-in-constructor.md
@@ -46,4 +46,6 @@ No other alternatives considered.
 
 ## Risks, Uncertainties, and Considerations
 
-No risks, uncertainties or considerations found.
+### Consideration
+
+- Another rule that might help this proposal is asserting via Semgrep that all `initialize()` functions have an `external` modifier and there's no occurence of `this.initialize()`. This would help in enforcing this rule. Is there any scenario where `initialize()` functions need to be called from within a contract or an inheriting contract?


smartcontracts (commented):
I'm biased but I really like this idea. I would want to get some feedback from @maurelian / @Ethnical  / @mds1 for any potential security considerations

maurelian:
Fully support this. It is currently being implemented in #12057 anyways, so it'll get in for isthmus :)

protocol/disableInitializers-in-constructor.md
```diff
@@ -0,0 +1,51 @@
...
+## Context
+
+Currently, most OP Stack proxied contracts prevent direct calls to their `initialize` functions on the implementation contract by calling it with default parameter values in the constructor. Additionally, these contracts do not expose a public `reinitialize` function to prevent reinitialization attempts.
+
+## Problem Statement
+
+So far this does the job but has a few drawbacks, namely:
+
+- It's a manual process (per contract) given that `initialize` function's input arguments are different for each contract. Also, a change in the `initialize` function's arguments will require updating the call to `initialize` manually.
+- If a public `reinitialize` function is added in the future, calling the `initialize` function in the constructor will not prevent reinitialization.
```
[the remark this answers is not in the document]
maurelian: Resolving in favor of discussion below.

protocol/disableInitializers-in-constructor.md
```diff
@@ -0,0 +1,51 @@
...
+- It can get verbose for initializers with many arguments. Here's an [example](https://github.com/ethereum-optimism/optimism/blob/36b41baea45f07e8644ff778a1ffd81fb0ad9e77/packages/contracts-bedrock/src/L1/SystemConfig.sol#L152).
+
+## Proposed Solution
+
+A proposed solution is to always use `_disableInitializers()` in the constructor of the implementation contract instead. This will, like the current approach, prevent anyone from calling the `initialize` function directly on the implementation contract but also solves the drawbacks mentioned above.
+
+- It has the benefit of being a one-time manual change for each contract regardless of how many arguments the `initialize` function expects or if it changes in the future.
+- If a public `reinitialize` function is used in the future, `_disableInitializers()`, already called in the constructor, prevents reinitialization attempts on the implementation contract.
```
maurelian, line 20: It's not clear to my why the current approach of calling `initialize()` (as long as it has the `initializer` modifier on it) is different from calling `_disableInitializers()`?

AmadiMichael: The major benefit is that `_disableInitializers()` is less verbose and does not need modification if initialize parameters change. 

But then also, there's a scenario where if `reinitializer` modifier is used in the future for some reason, the call to `initialize` in the constructor does not stop the implementation contract from being reinitialized after deployment. This is because `initialize` sets the initialized storage value to 1 which makes it reinitializable if the public function to do so exists but `_disableInitializers()` sets this value to `type(uint8).max`, preventing this entirely.

mds1 (approved):
Very supportive I've this, I've been lightly pushing for this for a while 

Semgrep rules for enforcement is a great idea, we should make sure that, for contracts with `/// @custom:proxied true`, that either:
- there is no initializer
- there is an initializer, and the constructor does not call it and instead calls `_disableInitializers`

protocol/disableInitializers-in-constructor.md
Ethnical: > This is because initialize sets the initialized storage value to 1 which makes it reinitializable if the public function to do so exists but _disableInitializers() sets this value to type(uint8).max, preventing this entirely.
What do you mean by this, I don't get it. 
Does this means that `reinitializer()` should check that the slot value is `1` only and should revert if the slot is set to `type(uint8).max` ? 
I don't see it here:  https://github.com/OpenZeppelin/openzeppelin-contracts/blob/448efeea6640bbbc09373f03fbc9c88e280147ba/contracts/proxy/utils/Initializable.sol#L152

Ethnical (approved):
Beside the only comment regarding the `reinitializer()` and `type(uint8).max`. 
Agreed for sake of verbosity! 
Too much verbosity could lead to miss an parameter during the initialization process.. So strong yes!

protocol/disableInitializers-in-constructor.md
AmadiMichael: It works if the slot is 1 but it can't be nested. For the max, it reverts if the version is > `type(uint64*).max` (which can be set manually via `reinitialize` or via `_disableInitializers()`

smartcontracts (approved):
I'm in 👍

AmadiMichael:
Discussed in #150. Conclusion was to implement it alongside adding the following semgrep rules:
- Checks that all initialize functions have `external` and `initializer` modifiers.
- Checks that all proxied contracts have a constructor and that it calls `_disableInitializers()`

PR #145, opened 2024-10-29
Proposal to move types and errors files to dedicated folders

AmadiMichael:
This design doc proposes moving types and errors files out of `*/libraries/` and `*/lib/` files and into dedicated folders named `*/types/` and `*/errors/`.

Issues:
- https://github.com/ethereum-optimism/optimism/issues/12642
- https://github.com/ethereum-optimism/optimism/issues/12645

### DIFF
--- a/protocol/types-and-error-folders.md
+++ b/protocol/types-and-error-folders.md
@@ -14,7 +14,7 @@ The proposed solution is to move these error and types files to dedicated folder
 
 ## Alternatives Considered
 
-No alternatives considered.
+An alternative is to make errors follow [#143](https://github.com/ethereum-optimism/design-docs/pull/143) which proposes an error naming convention of `ContractName_ErrorName(..)`. This would mean that errors would be unique for each contracts (hence defined within them) and so we would not need dedicated folders for them. However this does not address types and so we might still need dedicated folders for them.
 
 ## Risks, Uncertainties and Considerations
 
--- a/protocol/types-and-error-folders.md
+++ b/protocol/types-and-error-folders.md
@@ -1,0 +1,23 @@
+# Proposal to move types and errors files to dedicated folders
+
+## Context
+
+The OP Stack contracts in some instances define errors and custom types in a separate file from the contract(s) that use them. This is common among contracts that are part of a common folder e.g L1/, L2/ etc which might have a common error or/and types file where they import the needed errors and types from. This has the advantage of re-usage of errors and types across contracts without duplicating code, ease of modifiability and extensibility and makes it easy to compare errors and types used by different contracts.
+
+## Problem Statement
+
+These error and types files are most times located in a libraries or lib folder e.g `libraries/` `L1/libraries/`, `L2/libraries/` but these are not actual libraries and so do not belong in the libraries folder. This can also make it hard to find at first, especially to a developer newly exploring the codebase.
+
+## Proposed Solution
+
+The proposed solution is to move these error and types files to dedicated folders e.g `L1/errors/`, `L1/types/`, `L2/errors/`, `L2/types/` etc.
+
+## Alternatives Considered
+
+No alternatives considered.
+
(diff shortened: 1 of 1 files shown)


protocol/types-and-error-folders.md
```diff
@@ -0,0 +1,23 @@
...
+
+The OP Stack contracts in some instances define errors and custom types in a separate file from the contract(s) that use them. This is common among contracts that are part of a common folder e.g L1/, L2/ etc which might have a common error or/and types file where they import the needed errors and types from. This has the advantage of re-usage of errors and types across contracts without duplicating code, ease of modifiability and extensibility and makes it easy to compare errors and types used by different contracts.
+
+## Problem Statement
+
+These error and types files are most times located in a libraries or lib folder e.g `libraries/` `L1/libraries/`, `L2/libraries/` but these are not actual libraries and so do not belong in the libraries folder. This can also make it hard to find at first, especially to a developer newly exploring the codebase.
+
+## Proposed Solution
+
+The proposed solution is to move these error and types files to dedicated folders e.g `L1/errors/`, `L1/types/`, `L2/errors/`, `L2/types/` etc.
```
smartcontracts, line 13: We should consider how this relates to #143 

I think there may be a future where errors are always defined in the contracts themselves, especially since #143 suggests having the error format be `ContractName_SomeError`.

protocol/types-and-error-folders.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Alternatives Considered
+
+No alternatives considered.
```
smartcontracts, line 17: Regarding my comment above, I think the alternative to consider is that maybe we shouldn't have these type/error files at all. I'm more convinced of this for errors but less convinced of this for types.

AmadiMichael: Yes same here, with #143 i'm okay with errors being defined in the contract file itself. I'm also okay with leaving types as is especially with types that have dedicated libraries for working with them, these are more of libraries and so it should be okay to leave them as is. But re-defining types like structs in each contract bloats the file and makes it harder to navigate.

AmadiMichael:
Here's a recap and conclusion of the discussion between @smartcontracts, @ControlCplusControlV and I on discord

1.	Error Files: We agreed to remove error files altogether, embedding errors directly in contracts and possibly following #143 if the proposal is accepted. This has the advantage of also simplifying imports.
2.	Type Files: For now, types files will remain as is, until the discussion for moving dispute/cannon folders into the L1 folder commences. [#Issue](https://github.com/ethereum-optimism/optimism/issues/12643)

protocol/types-and-error-folders.md
maurelian: Agree with the principle here. There's a reason we wanted error strings to be prefixed with the contract name, and I find generic `Unauthorized` errors less helpful unless you're printing the trace (which requires more careful re

That said, I do like the idea of consolidating errors into a few locations, so what we could do is:

```solidity
error Unauthorized(string location);

// and called as: 
revert Unauthorized(type(SomeContract).name);

// this would be nice to, but for some reason is failing for me in `chisel`: 
revert Unauthorized(type(this).name);
```

smartcontracts: Interesting, can you bring this up on #143?

ControlCplusControlV: For how many errors do you think we could actually consolidate like that? I do like this syntax but if its just like 3 errors or so I think it might be fine to duplicate, but will continue discussion on #143

smartcontracts: I generally agree here. If it's just a few errors then I don't think it's worth it. Having a single unified standard is valuable.

AmadiMichael:
Discussed in #150. Conclusion was
- For errors: it will follow #143 
- For types: it will remain as is (for now)

PR #146, opened 2024-10-29
interop: interoperable ether transfers design doc

tremarkley:
## Description

This PR introduces a design for simplifying `ETH` transfers between two interoperable L2 chains by reducing the process from four transactions to two. The new approach leverages the `SuperchainWETH` contract, which now includes two new functions: `sendETH` and `relayETH`. 

- **`sendETH`**: Deposits `ETH` in the `ETHLiquidity` contract and sends a message to the destination chain, encoding the relay details.
- **`relayETH`**: Withdraws the specified amount of `ETH` from `ETHLiquidity` on the destination chain and transfers it to the recipient.

This update streamlines L2-to-L2 `ETH` transfers, bypassing the need for separate wrapping and unwrapping steps. Notably, custom gas token chains are excluded from this simplification to maintain compatibility and reduce risk.

## Additional context

- **Pros**: Simplifies developer and user experience by enabling direct ETH transfers.
- **Cons**: Adds a secondary entry point for transfers, increasing complexity slightly for developers handling asset transfers.

Any feedback on optimizing `sendETH` and `relayETH` for further efficiency or insights into handling custom gas token chains in the future is welcome.

### DIFF
--- a/protocol/interoperable-ether-transfers.md
+++ b/protocol/interoperable-ether-transfers.md
@@ -1,0 +1,261 @@
+# Purpose
+
+This document exists to align on a design for simplifying the process for sending `ETH` between two interoperable chains.
+
+# Summary
+
+New functionality is introduced to the `ETHLiquidity` contract that enables it to facilitate cross-chain `ETH` transfers to a specified recipient.
+
+# Problem Statement + Context
+
+Currently, L2-to-L2 `ETH` transfers between two interoperable chains require four separate transactions:
+
+1. Wrap `ETH` to `SuperchainWETH`.
+2. Call `SuperchainTokenBridge#SendERC20` to burn `SuperchainWETH` on source chain and relay a message to the destination chain that mints `SuperchainWETH` to the recipient.
+3. Execute the message on the destination chain, triggering `SuperchainTokenBridge#RelayERC20` to mint `SuperchainWETH` to the recipient.
+4. Unwrap the received `SuperchainWETH` to `ETH`.
+
+The goal is to reduce the transaction count from four to two, enabling users to send `ETH` to a destination chain directly:
+
+1. Burn `ETH` on source chain and relay a message to destination chain that mints `ETH` to recipient on destination chain.
+2. Execute the relayed message on the destination chain that mints `ETH` to the recipient.
+
+# Proposed Solution
+
+Introduce `SendETH` and `RelayETH` functions to the `ETHLiquidity` contract. By adding these functions directly to `ETHLiquidity`, the contract retains its original purpose of centralizing the minting and burning of `ETH` for cross-chain transfers, ensuring that all liquidity management occurs within a single, dedicated contract.
+
+### `SendETH` function
+
+The `SendETH` function combines the first two transactions as follows:
+
+1. Burns `ETH` within the `ETHLiquidity` contract equivalent to the `ETH` sent.
+2. Sends a message to the destination chain encoding a call to `RelayETH`.
+
+### `RelayETH` function
+
+The `RelayETH` function combines the last two transactions as follows:
+
+1. Mints the specified `_amount` of `ETH` from the `ETHLiquidity` contract.
+2. Sends the minted `ETH` to the specified recipient.
+
+### Contract changes
+
+Add an internal `_burn` function that will be called by `ETHLiquidity#SendETH` and update the external `burn` function to call `_burn`:
+
+```solidity
+/// @notice Allows an address to lock ETH liquidity into this contract.
+function burn() external payable {
+    if (msg.sender != Predeploys.SUPERCHAIN_WETH) revert Unauthorized();
+    _burn(msg.sender, msg.value);
+}
+
+/// @notice Allows an address to lock ETH liquidity into this contract.
+/// @param _sender Address that sent the ETH to be locked.
+/// @param _amount The amount of liquidity to burn.
+function _burn(address _sender, uint256 _amount) internal {
+    if (IL1Block(Predeploys.L1_BLOCK_ATTRIBUTES).isCustomGasToken()) revert NotCustomGasToken();
+    emit LiquidityBurned(_sender, _amount);
+}
+```
+
+Add an internal `_mint` function that will be called by `ETHLiquidity#RelayETH` and update the external `mint` function to call `_mint`:
+
+```solidity
+/// @notice Allows an address to unlock ETH liquidity from this contract.
+/// @param _amount The amount of liquidity to unlock.
+function mint(uint256 _amount) external {
+    if (msg.sender != Predeploys.SUPERCHAIN_WETH) revert Unauthorized();
+    _mint(msg.sender, _amount);
+}
+
+/// @notice Allows an address to unlock ETH liquidity from this contract.
+/// @param _recipient Address to send the unlocked ETH to.
+/// @param _amount    The amount of ETH liquidity to unlock.
+function _mint(address _recipient, uint256 _amount) internal {
+    if (IL1Block(Predeploys.L1_BLOCK_ATTRIBUTES).isCustomGasToken()) revert NotCustomGasToken();
+    new SafeSend{ value: _amount }(payable(_recipient));
+    emit LiquidityMinted(_recipient, _amount);
+}
+```
+
+Add `SendETH` and `RelayETH` function to `ETHLiquidity`:
+
+```solidity
+/// @notice Sends ETH to some target address on another chain.
+/// @param _to       Address to send ETH to.
+/// @param _chainId  Chain ID of the destination chain.
+function sendETH(address _to, uint256 _chainId) public payable {
+    if (IL1Block(Predeploys.L1_BLOCK_ATTRIBUTES).isCustomGasToken()) {
+        revert NotCustomGasToken();
+    }
+
+    _burn(msg.sender, msg.value);
+
+    // Send message to other chain.
+    IL2ToL2CrossDomainMessenger(Predeploys.L2_TO_L2_CROSS_DOMAIN_MESSENGER).sendMessage({
+        _destination: _chainId,
+        _target: address(this),
+        _message: abi.encodeCall(this.relayETH, (msg.sender, _to, msg.value))
+    });
+    emit SendETH(msg.sender, _to, msg.value, _chainId);
+}
+
+/// @notice Relays ETH received from another chain.
+/// @param _from    Address of the msg.sender of sendETH on the source chain.
+/// @param _to      Address to relay ETH to.
+/// @param _amount  Amount of ETH to relay.
+function relayETH(address _from, address _to, uint256 _amount) external {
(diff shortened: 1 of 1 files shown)


tremarkley:
* **#146** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/design-docs/146?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @tremarkley and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

protocol/interoperable-ether-transfers.md
```diff
@@ -0,0 +1,154 @@
...
+
+# Proposed Solution
+
+Add two new functions to the `SuperchainWETH` contract: `SendETH` and `RelayETH`.
+
+### `SendETH`
+
+The `SendETH` function combines the first two transactions as follows:
+
+1. Burns `ETH` within the `ETHLiquidity` contract equivalent to the `ETH` sent.
```
hamdiallam, line 31: Since we're not burning WETH, how about we add these functions to the `ETHLiquidity` contract?

I think it makes intent a bit clearer and keeps `WETH` separate as an ERC20?

tremarkley: it definitely simplifies things and creates better separation of concerns to leave WETH as-is and move this into `ETHLiquidity` I'm onboard with that

protocol/interoperable-ether-transfers.md
```diff
@@ -0,0 +1,244 @@
...
+    this.mint(_to, _amount);
+    emit RelayETH(_from, _to, _amount, messenger.crossDomainMessageSource());
+}
+```
+
+# Considerations
+
+## Custom Gas Token Chains
+
+To simplify the solution, custom gas token chains will not be supported and must follow the original four-transaction flow, which includes wrapping and unwrapping `SuperchainWETH`.
```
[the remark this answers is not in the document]
tremarkley: yeah, that does bring up an issue with this, there is not a good way to check at the smart contract level whether the destination is a custom gas token chain or if it is in the dependency set.

protocol/interoperable-ether-transfers.md
```diff
@@ -0,0 +1,248 @@
...
+/// @notice Sends ETH to some target address on another chain.
+/// @param _to      Address to send ETH to.
+/// @param _chainId  Chain ID of the destination chain.
+function sendETH(address _to, uint256 _chainId) public payable {
+    if (IL1Block(Predeploys.L1_BLOCK_ATTRIBUTES).isCustomGasToken()) {
+        revert NotCustomGasToken();
+    }
+
+    // Burn to ETHLiquidity contract.
+    this.burn{ value: msg.value }();
```
skeletor-spaceman, line 79: could we have both `mint` and `burn` functions be internal? and have the two external functions call the internal one? as to not have access control changes and weird calls to self that just waste gas and make things a bit spaghetti-like

tremarkley: good suggestion, updated!

protocol/interoperable-ether-transfers.md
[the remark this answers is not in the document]
tynes: There is no guarantee that standard config will always require ether for gas paying and the system was designed to be friendly to that changing

protocol/interoperable-ether-transfers.md
tynes: Since `ETHLiquidity` holds so much ether, we have to be really careful with what functionality is added to it. It should be callable in the most minimal amount of ways. I originally designed it so that there could be only a single useful caller. We need to be very careful with any modifications to it

tremarkley: @tynes are you onboard with the suggested solution or do you think it is too risky of a change?

protocol/interoperable-ether-transfers.md
```diff
@@ -0,0 +1,261 @@
...
+
+This solution does not eliminate the `SuperchainWETH` contract and `SuperchainWETH` transfers would still go through the `SuperchainTokenBridge`.
+
+## Custom Gas Token Chains
+
+To simplify the solution, custom gas token chains will not be supported and must follow the original four-transaction flow, which includes wrapping and unwrapping `SuperchainWETH`.
+
+# Open Questions
+- **Long-term improvements**: Could this functionality eventually extend to custom gas token chains?
+- **Rollbacks**: How would rollbacks be handled in this implementation?
```
nicholaspai, line 132: I assume rollbacks wouldn't be an issue as long as relayETH can only be called following the origin chain's finality period has passed?

protocol/interoperable-ether-transfers.md
```diff
@@ -0,0 +1,261 @@
...
+
+    new SafeSend{ value: _amount }(payable(_to));
+
+    emit RelayERC20(Predeploys.SUPERCHAIN_WETH, _from, _to, _amount, source);
+}
+```
+
+### Advantages
+
+The advantage of this solution is that `SuperchainTokenBridge`would handle both `ETH` transfers and `SuperchainERC20` transfers, simplifying developer integrations.
```
nicholaspai, line 257: > `SuperchainTokenBridge`would handle both `ETH` transfers and `SuperchainERC20` transfers

And it would handle sending messages. Sending messages/erc20's/eth from a single contract is something developer's are used to, though its not at all a deal breaker

protocol/interoperable-ether-transfers.md
hamdiallam: definitely see the concern on additions to `ETHLiquidity`. But want to note that even if this is in `SuperchainWETH`, a vulnerability there would also mean a vulnerability in `ETHLiquidity` since the relay mechanism pulls ETH from this contract

protocol/interoperable-ether-transfers.md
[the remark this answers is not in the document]
hamdiallam: +1 on no guarantee. Just noting the state of the world today but still requires addressing this since the protocol is agnostic to the superchain

protocol/interoperable-ether-transfers.md
tremarkley: by rollbacks here I am talking about ETH that is sent to a remote chain, but for some reason the relayETH message on the remote chain expires or is not able to be executed and we would like to "rollback" the ETH that was sent so the user can get their funds back

protocol/interoperable-ether-transfers.md
```diff
@@ -0,0 +1,261 @@
+# Purpose
+
+This document exists to align on a design for simplifying the process for sending `ETH` between two interoperable chains.
+
+# Summary
+
+New functionality is introduced to the `ETHLiquidity` contract that enables it to facilitate cross-chain `ETH` transfers to a specified recipient.
```
tynes, line 7: I don't love adding new functionality to the `ETHLiquidity` contract as it has a ton of virtual ether and makes reasoning about it more difficult. Its previous design only allowed one possible `msg.sender`. I definitely prefer to put the logic in the bridge contract to keep the liquidity contract more simple

skeletor-spaceman: +1 let's do this on the Bridge as originally planned :)

protocol/interoperable-ether-transfers.md
tynes: Reorgs are contingent between chains so there is no finality period. The concept of rollback is only relevant if you send to a chain where the call reverts on the remote chain and you cannot replay it or it always reverts. I think people also wanted to solve the problem of sending to a non existent chain using rollbacks, but I don't think that is possible. Now that we require the call on the remote chain to succeed, I think that the rollback design is going to be more difficult. We should think about rollbacks in a different way to solve that problem

tynes:
Generally much more in favor of putting the logic in the [SuperchainTokenBridge](https://github.com/ethereum-optimism/optimism/blob/bc6f8de543608be28590949259155f5ad163c986/packages/contracts-bedrock/src/L2/SuperchainTokenBridge.sol) given that its the cross chain bridging contract. Generally seems intuitive to put it there and mirrors our current design where the `StandardBridge` has ERC20 + ETH functionality

protocol/interoperable-ether-transfers.md
tremarkley: update the proposed solution in the design doc to be adding this functionality to the `SuperchainWETH` contract.

protocol/interoperable-ether-transfers.md
tynes: I recommend we use the following design for rollbacks: https://github.com/ethereum-optimism/specs/issues/460

I think there is confusion around what problem we are solving for, there should be no way that the cross chain token send should revert, so we don't need a rollback solution persay, we need to "prevent send to wrong chain" solution

protocol/interoperable-ether-transfers.md
```diff
@@ -153,62 +147,88 @@ The `RelayETH` function combines the last two transactions as follows:
...
+
+/// @notice Allows an address to unlock ETH liquidity from this contract.
+/// @param _recipient Address to send the unlocked ETH to.
+/// @param _amount    The amount of ETH liquidity to unlock.
+function _mint(address _recipient, uint256 _amount) internal {
+    if (IL1Block(Predeploys.L1_BLOCK_ATTRIBUTES).isCustomGasToken()) revert NotCustomGasToken();
+    new SafeSend{ value: _amount }(payable(_recipient));
+    emit LiquidityMinted(_recipient, _amount);
+}
+```
```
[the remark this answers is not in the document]
tremarkley: this is up to date. this design is under the `Alternatives Considered` section and is meant to illustrate a design that was considered for adding this functionality to the `ETHLiquidity` contract

tremarkley:
@tynes updated the design to handle sending ether to a remote chain that is custom gas token

skeletor-spaceman (changes requested):
nice doc! very few nit picks on proactively pushing for a better naming from the design doc onward.

protocol/interoperable-ether-transfers.md
```diff
@@ -0,0 +1,283 @@
...
+
+## Integrate ETH transfers into `ETHLiquidity` contract
+
+Introduce `SendETH` and `RelayETH` functions to the `ETHLiquidity` contract. By adding these functions directly to `ETHLiquidity`, the contract retains its original purpose of centralizing the minting and burning of `ETH` for cross-chain transfers, ensuring that all liquidity management occurs within a single, dedicated contract.
+
+### `SendETH` function
+
+The `SendETH` function combines the first two transactions as follows:
+
+1. Burns `ETH` within the `ETHLiquidity` contract equivalent to the `ETH` sent.
```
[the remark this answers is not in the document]
tremarkley: done.

tynes (approved):

skeletor-spaceman (approved):

PR #155, opened 2024-11-05
feat: created entrypoint-contracts design doc

0xParticle:
**Description**
This design document introduces `Entrypoint` contracts as a new primitive that allows anyone to add custom logic on top of the `L2ToL2CrossDomainMessenger`. It generalizes the `L2ToL2CrossDomainMessenger` design and unlocks other interop primitives such as message batching and expiring. To do so, it adds a parameter in the `L2ToL2CrossDomainMessenger` that binds the relaying to a particular address, where custom logic can live.

### DIFF
--- a/protocol/messaging/entrypoint-contracts.md
+++ b/protocol/messaging/entrypoint-contracts.md
@@ -64,20 +64,22 @@ An alternative design could be to emit an `EntrypointContext` struct with explic
 Notice that, in the intuition example, the `SwapperEntrypoint` is used as both pre and post-hook for the message relay. The `EntrypointContext` is used here in the post-action, but it would be needed beforehand in other cases (like batching).
 
 ### Binding
+The `EntrypointContext` is not inherently bound to its connected initiating message. This means that, in theory, anyone could execute a message alongside any other valid `EntrypointContext`. Therefore, each `Entrypoint` and initiating message contract handles its authentication logic.
 
-`EntrypointContext` is not binding to its connected initiating message. This means that, in theory, anyone could execute the message alongside any other valid `EntrypointContext`. 
-Each `Entrypoint` and initiating message contract will be responsible for handling authentication logic.
+`Entrypoints` can adopt the model used by the `L2ToL2CrossDomainMessenger`, which involves decoding additional parameters. This approach would require the `EntrypointContext` event to encode the necessary information.
 
-`Entrypoints` can follow the model of the `L2ToL2CrossDomainMessenger` and decode additional parameters.
-Of course, this would require the `EntrypointContext` event to encode this information.
-Some parameters that can be used for binding are
-- `L2ToL2CrossDomainMessenger` current `nonce`. The `Entrypoint` can then check that the `EntrypointContext` has the same `nonce` than the `SentMessage`.
-- Identifier's `origin` from `EntrypointContext` event matches `sender` from `SentMessage`. This could show that the context was created on the same contract than the connected message.
-- Identifier's `chainId` from `EntrypointContext` event matches `source` from `SentMessage`. Should match with the `SentMessage`.
+We strongly recommend that an `Entrypoint` only accepts `EntrypointContext` events from authorized sources. This implies that the `Entrypoint` should perform the following checks:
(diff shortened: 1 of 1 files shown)


protocol/messaging/entrypoint-contracts.md
```diff
@@ -0,0 +1,134 @@
...
+
+Anyone can deploy `Entrypoint` contracts with any logic. Users can indicate, explicitly or through a contract, which `Entrypoint` contract can process their messages in the destination chain. `relayMessage` will check if the message has an `Entrypoint` associated, and if it does, it will validate that this contract is the `msg.sender`; otherwise, it will revert.
+
+The key modification is the `msg.sender` binding feature on the `relayMessage` call. If `msg.sender` has logic, the design is equivalent to binding `relayMessage` to include a pre-hook with that logic.
+
+Notice that a pre/post-hook design is already possible with the current design.
+
+### Summary of required changes
+
+The following changes are required:
```
tynes, line 23: I believe that we agreed to overload `sendMessage`? Would that be possible with the diff required to `relayMessage`

tynes:
Big fan of this design

protocol/messaging/entrypoint-contracts.md
0xParticle: Good point. This would be possible, but we would need to prevent `relayMessage` to be successful. 
- One way could be to keep the selector check [here](https://github.com/ethereum-optimism/optimism/blob/bc9b6cd588588c9c4167c926a1782c658e5df921/packages/contracts-bedrock/src/L2/L2ToL2CrossDomainMessenger.sol#L225), which I think is going to be deprecated as per this [comment](https://github.com/ethereum-optimism/optimism/blob/bc9b6cd588588c9c4167c926a1782c658e5df921/packages/contracts-bedrock/src/L2/L2ToL2CrossDomainMessenger.sol#L60).
- The other is to encode the `entrypoint` first in the list, so that decoding succeeds, but `relayMessage` will fail on [this check](https://github.com/ethereum-optimism/optimism/blob/bc9b6cd588588c9c4167c926a1782c658e5df921/packages/contracts-bedrock/src/L2/L2ToL2CrossDomainMessenger.sol#L169).

I feel like the selector is more explicit, as it will throw the correct error. Do you have any context on the removal of it?

0xParticle: @tynes an alternative here is to rewrite `sendMessage` with the `entrypoint` parameter, an do not revert in `relayMessage` if entrypoint is `address(0)`.

0xParticle:
@tynes pls do not merge this PR yet, we might find a way to improve the security of EntrypointContext binding. Will update soon.

0xParticle:
> @tynes pls do not merge this PR yet, we might find a way to improve the security of EntrypointContext binding. Will update soon.

ready now

0xParticle:
Lost push access to this PR, so I created a new branch with updates here: https://github.com/ethereum-optimism/design-docs/pull/163

nambrot:
I might miss a bit of context, @skeletor-spaceman mostly shared this with me for purposes of looking into relaying semantics. I'm trying to understand the implications of the CDM API change. For that, I was wondering where "middlewares" (that would not require any API change) would break. 

For example for the expiring message use case, could one have an origin chain middleware contract that actually calls the CDM with the destination chain middleware contract as the recipient? Was the intention to keep having the CDM by the `callee` on the origin chain?

skeletor-spaceman:
> For example for the expiring message use case, could one have an origin chain middleware contract that actually calls the CDM with the destination chain middleware contract as the recipient? Was the intention to keep having the CDM be the `callee` on the origin chain?

hey @nambrot , I think the answer is yes, the destination middleware is not necessarely the recipient but the expected caller for the CDM.
and yes, the intention is to keep the CDM as the `callee`, since it's the invariant that contracts, such as the bridge, expect to be called by the CDM.
does that answer your question? else, could you add a specific example on the flow you are thinking about?

0xParticle:
This work has been moved here: https://github.com/ethereum-optimism/design-docs/pull/163

PR #156, opened 2024-11-05
opcm-redesign: Rearchitecting OPCM to handle only one L1 smart version release

blmalone:
**Description**
OPCM requires a redesign based on recent conversations highlighting the need for upgrade functionality. This design doc is preemtively making some changes to make introducing upgradability features to OPCM easier.

**Additional context**

Ongoing discussions happening in Optimism R&D [discord](https://discord.com/channels/1244729134312198194/1273519241076670526).

### DIFF
--- a/protocol/op-contracts-manager-single-release-redesign.md
+++ b/protocol/op-contracts-manager-single-release-redesign.md
@@ -35,7 +35,7 @@ Continue with the approach of adding additional conditional logic for every new 
 Having one OPCM per *L1 smart contract release version*[^1] means that we can remove the proxy pattern that currently exists with OPCM. The high level architecture changes can be seen below:
 
 ### Previous OPCM Deploys
-Before OPCM was proxied so that it was upgradable. e.g. [`0x18CeC91779995AD14c880e4095456B9147160790`](https://etherscan.io/address/0x18CeC91779995AD14c880e4095456B9147160790)
+Before, OPCM was proxied so that it was upgradable. e.g. [`0x18CeC91779995AD14c880e4095456B9147160790`](https://etherscan.io/address/0x18CeC91779995AD14c880e4095456B9147160790)
 ```mermaid
 flowchart LR
     A[Any Account] -->|call| C[Proxy]
@@ -43,7 +43,7 @@ flowchart LR
 ```
 
 ### New OPCM Deploys
-Now OPCM is not proxied and therefore not upgradable. This is by design though as we want to release a new OPCM for each new L1 smart contract release.
+Now, OPCM is not proxied and therefore not upgradable. This is by design though as we want to release a new OPCM for each new L1 smart contract release.
 ```mermaid
 flowchart LR
     A[Any Account] -->|call| B[OPCM - op-contracts/v1.3.0]
--- a/protocol/op-contracts-manager-single-release-redesign.md
+++ b/protocol/op-contracts-manager-single-release-redesign.md
@@ -2,7 +2,7 @@
 
 The original OPContractsManager (OPCM) design, outlined [here](./op-contracts-manager-arch.md), defined a singleton proxied contract that deployed the L1 contracts for standard OP Stack chains in a single transaction.
 
-Implementing upgrade functionality for OP Stack chains via OPCM is now a key goal. As we approach this feature, we've identified several interim adjustments to OPCM that will that will not only enable future upgradability but also simplify its overall design. Therefore, the purpose of this design document is to streamline OPCM’s architecture, making it easier to build and integrate upgradability features.
+Implementing upgrade functionality for OP Stack chains via OPCM is now a key goal. As we approach this feature, we've identified several interim adjustments to OPCM that will not only enable future upgradability but also simplify its overall design. Therefore, the purpose of this design document is to streamline OPCM’s architecture, making it easier to build and integrate upgradability features.
 
 # Summary
 
--- a/protocol/op-contracts-manager-single-release-redesign.md
+++ b/protocol/op-contracts-manager-single-release-redesign.md
@@ -20,9 +20,13 @@ OPCM will transition from its current singleton architecture to a multi-deployme
 We’ve identified the need to implement an upgradability feature for OP Stack chains. This will involve two core components: [op-deployer](../ecosystem/op-deployer.md) and [OPCM](./op-contracts-manager-arch.md). As of November 6, 2024, both op-deployer and OPCM support only the deployment of new chains at fixed versions.
 Each new chain version requires additional conditional logic in both op-deployer and the OPCM singleton. This creates logic that is difficult to maintain and reason about. This approach will become increasingly unwieldy as we integrate OP Stack upgrade functionality into op-deployer and OPCM.
 
-## Examples of Unwanted Conditional Logic 
+## Example of Unwanted Conditional Logic in OPCM
 - Deciding how to initialize the `SystemConfig` contract correctly via OPCM - [code](https://github.com/ethereum-optimism/optimism/blob/28283a927e3124fa0b2cf8d47d1a734e95478215/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L457-L462). This branching logic exists because OPCM needed to be able to deploy chains at older tags (e.g. op-contracts/v1.6.0) as well as later chains. The primary challenge here is that the `develop` branch in the monorepo naturally evolves, requiring deployment (OPCM and op-deployer) code to evolve with it, all while continuing to support older deployments.
-- `// TODO: Add a link to an example in op-deployer that contains unwanted conditional logic.`
+
+## Example of Unwanted Complexity in op-deployer
+- Each `op-deployer` releases currently bundles standard version TOML files from the superchain-registry. These files contain the addresses corresponding to each L1 smart contract release. Developers must [manually sync](https://github.com/ethereum-optimism/optimism/blob/bc9b6cd588588c9c4167c926a1782c658e5df921/op-chain-ops/Makefile#L50-L52) the TOML files before cutting a new release. Since these release addresses are tightly coupled to op-deployer, they need to be injected into OPCM. Without this injection, OPCM is unaware of the implementation addresses of any L1 smart contract releases. A system in which OPCM is 'release aware' simplifies the logic within op-deployer. In this setup, op-deployer would maintain a list of OPCM addresses and manage interactions with each address on a case-by-case basis.
+
+## Optional: Authentication
 
 Authentication was not originally implemented in OPCM because it solely supported deployments, with the rationale being that there was no immediate need to restrict access to the deploy function. However, with the introduction of upgrade functionality, it’s essential to establish a robust authentication mechanism within OPCM. We now need to determine appropriate access controls and define roles for authorized users who will be permitted to perform upgrades.
 
--- a/protocol/op-contracts-manager-single-release-redesign.md
+++ b/protocol/op-contracts-manager-single-release-redesign.md
@@ -6,9 +6,8 @@ Implementing upgrade functionality for OP Stack chains via OPCM is now a key goa
 
 # Summary
 
-The summary of the changes to OPCM is as follows:
+OPCM will transition from its current singleton architecture to a multi-deployment model, with each deployment focused on a specific L1 smart contract release version. To achieve this, the following changes need to be implemented:
 
-- Move away from singleton OPCM approach in favor of one OPCM deployment per L1 smart contract release.
 - Remove the [proxy](https://github.com/ethereum-optimism/optimism/blob/4c015e3a36f8910e2cf8b447d62ab4c44b944cca/packages/contracts-bedrock/scripts/deploy/DeployImplementations.s.sol#L545) from OPCM.
 - Remove [initialize](https://github.com/ethereum-optimism/optimism/blob/28283a927e3124fa0b2cf8d47d1a734e95478215/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L210) functionality from OPCM and introduce standard constructor initialization.
 - Add versioning to OPCM to expose which L1 contracts are deployed for the specific OPCM deployment e.g. OPCM address `0x1234...` deploys `op-contracts/v1.6.0`.
@@ -18,7 +17,7 @@ The summary of the changes to OPCM is as follows:
 
 # Problem Statement + Context
 
-We’ve identified the need to implement an upgradability feature for OP Stack chains. This will involve two core components: [op-deployer](../ecosystem/op-deployer.md) and [OPCM](./op-contracts-manager-arch.md). As of November 5, 2024, both op-deployer and OPCM support only the deployment of new chains at fixed versions.
+We’ve identified the need to implement an upgradability feature for OP Stack chains. This will involve two core components: [op-deployer](../ecosystem/op-deployer.md) and [OPCM](./op-contracts-manager-arch.md). As of November 6, 2024, both op-deployer and OPCM support only the deployment of new chains at fixed versions.
 Each new chain version requires additional conditional logic in both op-deployer and the OPCM singleton. This creates logic that is difficult to maintain and reason about. This approach will become increasingly unwieldy as we integrate OP Stack upgrade functionality into op-deployer and OPCM.
 
 ## Examples of Unwanted Conditional Logic 
--- a/protocol/op-contracts-manager-single-release-redesign.md
+++ b/protocol/op-contracts-manager-single-release-redesign.md
@@ -10,9 +10,8 @@ OPCM will transition from its current singleton architecture to a multi-deployme
 
 - Remove the [proxy](https://github.com/ethereum-optimism/optimism/blob/4c015e3a36f8910e2cf8b447d62ab4c44b944cca/packages/contracts-bedrock/scripts/deploy/DeployImplementations.s.sol#L545) from OPCM.
 - Remove [initialize](https://github.com/ethereum-optimism/optimism/blob/28283a927e3124fa0b2cf8d47d1a734e95478215/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L210) functionality from OPCM and introduce standard constructor initialization.
-- Add versioning to OPCM to expose which L1 contracts are deployed for the specific OPCM deployment e.g. OPCM address `0x1234...` deploys `op-contracts/v1.6.0`.
+- Add versioning to OPCM to expose which L1 contracts are deployed for the specific OPCM deployment e.g. `string constant l1ContractsVersion = "op-contracts/v1.6.0";` or something more nuanced that shows the relationship between the version string and the [standard release implementation addresses](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/standard/standard-versions-mainnet.toml#L9).
 - Update `op-deployer`, `DeploySuperchain.s.sol`, `DeployImplementations.s.sol` and `DeployOPChain.s.sol` to work with the new multi-deployment paradigm of one OPCM per L1 smart contract release.
-- **Optional**: Add authentication to OPCM in anticipation for upgrade features.
 
 
 # Problem Statement + Context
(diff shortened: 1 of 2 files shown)


protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,83 @@
+# Purpose
+
+The original OPContractsManager (OPCM) design, outlined [here](./op-contracts-manager-arch.md), defined a singleton proxied contract that deployed the L1 contracts for standard OP Stack chains in a single transaction.
```
mds1, line 3: This reminds me that we will also need to update the spec once this design doc has consensus: https://specs.optimism.io/experimental/op-contracts-manager.html

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,83 @@
...
+- Remove [initialize](https://github.com/ethereum-optimism/optimism/blob/28283a927e3124fa0b2cf8d47d1a734e95478215/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L210) functionality from OPCM and introduce standard constructor initialization.
+- Add versioning to OPCM to expose which L1 contracts are deployed for the specific OPCM deployment e.g. OPCM address `0x1234...` deploys `op-contracts/v1.6.0`.
+- Update `DeploySuperchain.s.sol`, `DeployImplementations.s.sol` and `DeployOPChain.s.sol` to work with the new paradigm of one OPCM per L1 smart contract release.
+- **Optional**: Add authentication to OPCM in anticipation for upgrade features.
+
+
+# Problem Statement + Context
+
+We’ve identified the need to implement an upgradability feature for OP Stack chains. This will involve two core components: [op-deployer](../ecosystem/op-deployer.md) and [OPCM](./op-contracts-manager-arch.md). As of November 6, 2024, both op-deployer and OPCM support only the deployment of new chains at fixed versions.
```
mds1, line 20: Even without the upgradeability feature, we would still want this design doc because it simplifies the architecture for deploying new chains to the latest version (for the same reasons of removing conditionals and simplifying devex for devs making contract changes), so let's make sure to mention that also

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Summary
+
+OPCM will transition from its current singleton architecture to a multi-deployment model, with each deployment focused on a specific L1 smart contract release version. To achieve this, the following changes need to be implemented:
+
+- Remove the [proxy](https://github.com/ethereum-optimism/optimism/blob/4c015e3a36f8910e2cf8b447d62ab4c44b944cca/packages/contracts-bedrock/scripts/deploy/DeployImplementations.s.sol#L545) from OPCM.
+- Remove [initialize](https://github.com/ethereum-optimism/optimism/blob/28283a927e3124fa0b2cf8d47d1a734e95478215/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L210) functionality from OPCM and introduce standard constructor initialization.
+- Add versioning to OPCM to expose which L1 contracts are deployed for the specific OPCM deployment e.g. OPCM address `0x1234...` deploys `op-contracts/v1.6.0`.
```
mds1, line 13: Is this basically saying we add a `string constant l1ContractsVersion = "op-contracts/v1.6.0"` for easy identification of what version a given OPCM deploys?

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,83 @@
...
+- Update `DeploySuperchain.s.sol`, `DeployImplementations.s.sol` and `DeployOPChain.s.sol` to work with the new paradigm of one OPCM per L1 smart contract release.
```
mds1, line 14: We likely will need changes to op-deployer here as well

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Example of Unwanted Complexity in op-deployer
+- Each `op-deployer` releases currently bundles standard version TOML files from the superchain-registry. These files contain the addresses corresponding to each L1 smart contract release. Developers must [manually sync](https://github.com/ethereum-optimism/optimism/blob/bc9b6cd588588c9c4167c926a1782c658e5df921/op-chain-ops/Makefile#L50-L52) the TOML files before cutting a new release. Since these release addresses are tightly coupled to op-deployer, they need to be injected into OPCM. Without this injection, OPCM is unaware of the implementation addresses of any L1 smart contract releases. A system in which OPCM is 'release aware' simplifies the logic within op-deployer. In this setup, op-deployer would maintain a list of OPCM addresses and manage interactions with each address on a case-by-case basis.
+
+## Authentication (Optional)
```
mds1, line 29: We can remove all auth section and info and leave this to @maurelian's design doc which will be upgrades focused and have to solve the auth problems. If anything we should mention that deployments are entirely permissionless

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Alternatives Considered
+
+Continue with the approach of adding additional conditional logic for every new chain version that is released. As noted, this leads to a system that is complex and hard to maintain.
+
+# Proposed Solution
+
+## OPCM Architecture Changes
+
+Having one OPCM per *L1 smart contract release version*** means that we can remove the proxy pattern that currently exists with OPCM. The high level architecture changes can be seen below:
```
mds1, line 41: nit, you can use footnotes! https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/
```suggestion
Having one OPCM per *L1 smart contract release version*[^1] means that we can remove the proxy pattern that currently exists with OPCM. The high level architecture changes can be seen below:
```

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,83 @@
...
+    A[Any Account] -->|call| B[OPCM - op-contracts/v1.3.0]
+    C[Any Account] -->|call| D[OPCM - op-contracts/v1.6.0]
+```
+
+## Versioning
+
+Now that each OPCM deploy is going to be tethered directly to an L1 smart contract release version, for improved UX, we should provide public getters to expose two pieces of information: 
+
+1. The **current L1 smart contracts** that are deployed via this version of OPCM e.g. `op-contracts/v1.6.0`. This means we will no longer need to pass the standard versions toml [file](https://github.com/ethereum-optimism/optimism/blob/4c015e3a36f8910e2cf8b447d62ab4c44b944cca/packages/contracts-bedrock/scripts/deploy/DeployImplementations.s.sol#L61) around like [this](https://github.com/ethereum-optimism/optimism/blob/4c015e3a36f8910e2cf8b447d62ab4c44b944cca/packages/contracts-bedrock/scripts/deploy/DeployImplementations.s.sol#L1060). The implementation addresses can be added to the OPCM contract as as immutable values. 
+2. The **address of the previous OPCM** contract.
```
mds1, line 64: Great idea, this makes sense to me

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,83 @@
...
+This creates a linked list whereby a user can find all prior OPCMs starting from the current OPCM contract. This will prove useful for implementing upgrade functionality because with this information an OPCM can easily expose the version of L1 smart contracts that it supports upgrading **from** and **to**.
+
+## Authentication (Optional)
+
+As we know that upgrades are coming to OPCM we can preemtively add authentication to OPCM as part of this architecture redesign. 
+
+It is our intention to keep the OPCM [deploy](https://github.com/ethereum-optimism/optimism/blob/28283a927e3124fa0b2cf8d47d1a734e95478215/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L226) function permissionless. 
+
+- `//TODO Determine WHO should perform these upgrades`
+- `//TODO Determine HOW these access controls should be implemented in OPCM`
```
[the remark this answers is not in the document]
mds1: See https://github.com/ethereum-optimism/design-docs/pull/156/files#r1831355583

sbvegan:
When this new implementation is added, we'll want to make sure the docs are updated to include the new contract addresses: https://docs.optimism.io/stack/opcm

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,83 @@
...
+
+
+# Risks & Uncertainties
+
+With more OPCM versions, we’ll need to write additional production Solidity code. While our engineering practices help minimize bugs, it is still worth considering whether increasing the surface area of onchain interactions is desirable.
```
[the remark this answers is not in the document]
sbvegan: Good call out, it should be updated here: https://docs.optimism.io/stack/opcm

protocol/op-contracts-manager-single-release-redesign.md
blmalone: Tracked in Figma.

protocol/op-contracts-manager-single-release-redesign.md
blmalone: Yes exactly. I can add example the code in here to make it super clear.

protocol/op-contracts-manager-single-release-redesign.md
blmalone: Added this extra detail in: https://github.com/ethereum-optimism/design-docs/pull/156/commits/ea3f67eb705508d5d25531d7fe0c120dfee0e073

protocol/op-contracts-manager-single-release-redesign.md
blmalone: Added op-deployer in: https://github.com/ethereum-optimism/design-docs/pull/156/commits/ea3f67eb705508d5d25531d7fe0c120dfee0e073

protocol/op-contracts-manager-single-release-redesign.md
blmalone: Awesome, I've added footnotes in: https://github.com/ethereum-optimism/design-docs/pull/156/commits/ea3f67eb705508d5d25531d7fe0c120dfee0e073

protocol/op-contracts-manager-single-release-redesign.md
blmalone: removed in: https://github.com/ethereum-optimism/design-docs/pull/156/commits/5be2ed2dc27726cd5c166838110417ceb463c810

protocol/op-contracts-manager-single-release-redesign.md
blmalone: Added in: https://github.com/ethereum-optimism/design-docs/pull/156/commits/5be2ed2dc27726cd5c166838110417ceb463c810

protocol/op-contracts-manager-single-release-redesign.md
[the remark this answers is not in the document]
blmalone: removed in: https://github.com/ethereum-optimism/design-docs/pull/156/commits/5be2ed2dc27726cd5c166838110417ceb463c810

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,67 @@
+# Purpose
+
+The original OPContractsManager (OPCM) design, outlined [here](./op-contracts-manager-arch.md), defined a singleton proxied contract that deployed the L1 contracts for standard OP Stack chains in a single transaction.
+
+Implementing upgrade functionality for OP Stack chains via OPCM is now a key goal. As we approach this feature, we've identified several interim adjustments to OPCM that will that will not only enable future upgradability but also simplify its overall design. Therefore, the purpose of this design document is to streamline OPCM’s architecture, making it easier to build and integrate upgradability features.
```
maurelian, line 5: typo
```suggestion
Implementing upgrade functionality for OP Stack chains via OPCM is now a key goal. As we approach this feature, we've identified several interim adjustments to OPCM that will not only enable future upgradability but also simplify its overall design. Therefore, the purpose of this design document is to streamline OPCM’s architecture, making it easier to build and integrate upgradability features.
```

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,67 @@
...
+
+# Summary
+
+OPCM will transition from its current singleton architecture to a multi-deployment model, with each deployment focused on a specific L1 smart contract release version. To achieve this, the following changes need to be implemented:
+
+- Remove the [proxy](https://github.com/ethereum-optimism/optimism/blob/4c015e3a36f8910e2cf8b447d62ab4c44b944cca/packages/contracts-bedrock/scripts/deploy/DeployImplementations.s.sol#L545) from OPCM.
+- Remove [initialize](https://github.com/ethereum-optimism/optimism/blob/28283a927e3124fa0b2cf8d47d1a734e95478215/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L210) functionality from OPCM and introduce standard constructor initialization.
+- Add versioning to OPCM to expose which L1 contracts are deployed for the specific OPCM deployment e.g. `string constant l1ContractsVersion = "op-contracts/v1.6.0";` or something more nuanced that shows the relationship between the version string and the [standard release implementation addresses](https://github.com/ethereum-optimism/superchain-registry/blob/main/validation/standard/standard-versions-mainnet.toml#L9).
```
maurelian, line 13: I think `"op-contracts/v1.6.0"` is fine. Best to stay in alignment with the release process.

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,67 @@
...
+### Previous OPCM Deploys
+Before OPCM was proxied so that it was upgradable. e.g. [`0x18CeC91779995AD14c880e4095456B9147160790`](https://etherscan.io/address/0x18CeC91779995AD14c880e4095456B9147160790)
+```mermaid
+flowchart LR
+    A[Any Account] -->|call| C[Proxy]
+    C -->|delegatecall| D[OPCM]
+```
+
+### New OPCM Deploys
+Now OPCM is not proxied and therefore not upgradable. This is by design though as we want to release a new OPCM for each new L1 smart contract release.
```
mslipper, line 46: Might be worth describing how we'd handle patches to old versions of the OPCM - i.e., just deploy a new one.

protocol/op-contracts-manager-single-release-redesign.md
mslipper: Love this, but how would we handle the case where we have to patch the old release? Would we patch the upstream OPCMs as well? Maybe it doesn't matter, but thought I'd call out this edge case nonetheless.

protocol/op-contracts-manager-single-release-redesign.md
```diff
@@ -0,0 +1,67 @@
...
+2. The **address of the previous OPCM** contract. 
+
+This creates a linked list whereby a user can find all prior OPCMs starting from the current OPCM contract. This will prove useful for implementing upgrade functionality because with this information an OPCM can easily expose the version of L1 smart contracts that it supports upgrading **from** and **to**.
+
+# Risks & Uncertainties
+
+With more OPCM versions, we’ll need to write additional production Solidity code. While our engineering practices help minimize bugs, it is still worth considering whether increasing the surface area of onchain interactions is desirable.
```
mslipper, line 64: I think it's true that we'll be writing more prod Solidity code, but we'd end up doing this anyway if we followed the existing deployment process.

protocol/op-contracts-manager-single-release-redesign.md
maurelian: I would also add that the [OPCM semver](https://github.com/ethereum-optimism/optimism/blob/feat/isthmus-contracts/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L133-L134) should change to just match the release version, ie. we'd have: 

```
string public constant version = "1.0.0-beta.21";

function l1ContractsVersion() public returns(string) {
  return string.concat("op-contracts/", version);
}
```

Also consider `release()` for the function name.

mslipper (approved):
Looks great! Left a few non-blocking comments.

maurelian (dismissed):

protocol/op-contracts-manager-single-release-redesign.md
mds1: Perhaps the patch just points to the prior release and indicates that it replaces a previous version, for example if the `prior` var points to the preceding release and `replaces` points to what it replaces:
- Start with: OPCM 1.7.0 with `prior = address(1.6.0)` and `replaces = null`
- Deploy OPCM 1.7.1 with `prior = address(1.6.0)` and `replaces = address(1.7.0)`

blmalone: One of the reasons I was suggesting to include the address of the previous OPCM contract was so that you technically only ever needed to store the latest address. In other words, you can always find the previous OPCM addresses by traversing the list by continually reading `.prior()`. 

I'm not quite following how we can keep this property if we want to patch an old version of an OPCM contract. For example, let's say we have 3 contracts. `1`, `2` and `3`. 

1. Contract `1.prior()` returns `null`.
2. Contract `2.prior()` returns `1`.
3. Contract `3.prior()` returns `2`. 

If we patch contract `2` such that we now have two contracts `2` and `2-patch`. We will want a call to `3.prior()` to return `2-patch`.
How do we do this? My immediate thought is to have an authenticated setter method on OPCM that allows us to `updatePrior(address patchedVersion)`. Or keep the proxy and swap the implementation address to not have to change any `prior()` values. 

@mds1 - We could implement your suggestion in the `2-patch` contract but how would `3` know to reference the new `2-patch` instead of the old `2` contract?

blmalone: On second thought, the scenario I described might not really happen in reality. 

I don't think it makes sense to make a patch on contract `2` if contract `3` exists.

mds1: That is good callout to make sure we're aligned, but yes I agree with your conclusion that if release `3` is out, it means `2` is deprecated and we wouldn't patch it

protocol/op-contracts-manager-single-release-redesign.md
blmalone: The versioning bullet point here is meant to be a separate version than the regular semver that exists on all out contracts. 

OPCM should continue to include the semver as normal. But additionally, it should include the `op-contracts/v*.*.*` L1 smart contract release that it deploys. The code you provided seems like it is mixing the versions. I was thinking something more along the lines of:

```solidity
/// @custom:semver 1.0.0-beta.21
string public constant version = "1.0.0-beta.21";

// @notice L1 smart contracts release deployed by this version of OPCM.
string constant l1ContractsVersion = "op-contracts/v1.6.0";
```

blmalone: I've clarified this in the following commit: https://github.com/ethereum-optimism/design-docs/pull/156/commits/2011fb5d2c4fa482ecee2127d59376465129142d

protocol/op-contracts-manager-single-release-redesign.md
blmalone: I've addressed this point in the following commit: https://github.com/ethereum-optimism/design-docs/pull/156/commits/2011fb5d2c4fa482ecee2127d59376465129142d

protocol/op-contracts-manager-single-release-redesign.md
blmalone: I've addressed this point in the following commit: https://github.com/ethereum-optimism/design-docs/pull/156/commits/2011fb5d2c4fa482ecee2127d59376465129142d

protocol/op-contracts-manager-single-release-redesign.md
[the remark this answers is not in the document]
blmalone: I've addressed this point in the following commit: https://github.com/ethereum-optimism/design-docs/pull/156/commits/2011fb5d2c4fa482ecee2127d59376465129142d

protocol/op-contracts-manager-single-release-redesign.md
blmalone: I've made your suggestion clear here in this commit: https://github.com/ethereum-optimism/design-docs/pull/156/commits/2011fb5d2c4fa482ecee2127d59376465129142d#diff-a9741395d54e86787fd591639a1e294e53c8ee37ddbbf26f8fb03a596e302b46R53

protocol/op-contracts-manager-single-release-redesign.md
blmalone: Closing this based on conversation in discord: https://discord.com/channels/1244729134312198194/1273519241076670526/1303831875852304455

blmalone: I outlined the discussion here: https://github.com/ethereum-optimism/design-docs/pull/156/commits/7f50c4f7bd700be7d8664306acc31885d2cb71c3

protocol/op-contracts-manager-single-release-redesign.md
blmalone: Unresolving this thread, we no longer want to keep a reference to the old contract. 

Patching is useful and having a reference to the previous opcm complicates things.

blmalone: addressed in this commit: https://github.com/ethereum-optimism/design-docs/pull/156/commits/5ee9be2e4d2301af2358a061883bc7f2ce84085f

Issue #158, opened 2024-11-06
Multi-deployment OPCM model


blmalone:
**Title of Meeting**
Multi-deployment OPCM model

**Date, Time and Duration**

Thursday, November 7 · 3:00 – 4:00pm
Time zone: America/New_York/EST

**Link to design doc**
https://github.com/ethereum-optimism/design-docs/pull/156

**Additional Context**
The [design doc](https://github.com/ethereum-optimism/design-docs/pull/156) should provide you with all the context you need.

**Requested Attendees**
- @mds1 
- @maurelian 
- @mslipper 
- @smartcontracts

op-aaron:
This meeting has been posted and all invitees should have permission to invite others or modify if needed.  You can find the meeting info below and on the [OP Stack Calendar](https://calendar.google.com/calendar/embed?src=c_e7b35eadabec39777b28192d371c45b6ef4177e01740517a234e7c768881fbfe%40group.calendar.google.com&ctz=America%2FLos_Angeles) event.  Please let me know if you have any questions!

**Design Review: Multi-deployment OPCM model**
Thursday, November 7 · 3:00 – 4:00pm
Time zone: America/New_York/EST
Google Meet joining info
Video call link: https://meet.google.com/ged-ywup-whx
Or dial: ‪(US) +1 484-713-9724‬ PIN: ‪740 826 505‬#
More phone numbers: https://tel.meet/ged-ywup-whx?pin=1158869783920

op-aaron:
Design review complete

Issue #154, opened 2024-11-05
Contract Upgrade Idea


tynes:
This is a quick sketch on what ideal contract upgrades could look like.

It is desirable to be able to roll forward and backward contract upgrades by versioning the storage layout and defining a migration between any arbitrary version. We have this concept of 2 step upgrades already where we simply zero out the initialize storage slot. We could make the 2 step upgrade a more first class citizen, where the first step of the 2 step upgrade is made smarter and knows about the storage layout schemas of the particular contract.

What this would look like in practice is breaking out the storage of a contract into its own implementation and composing together the storage contract and the implementation code. Example:

```solidity
contract OptimismPortal is OptimismPortalStorage ... {}
```

The migration contract would look something like:

```solidity
contract OptimismPortalMigration is OptimismPortalStorage {
    // magic slot for the storage schema
    bytes32 version slot = keccak256("opstack.storage-schema-version");

   function migrate(uint256 _version) {
     // replace the functionality of StorageSetter by setting the initialize slot to 0
     Storage.setUint(initializableSlot, 0);     

     uint256 currentVersion = Storage.readUint(slot);
     if (currentVersion == _version) return;
     if (_version > currentVersion) {
        // for each version between _version and currentVersion
        // call upgradeTo(i)
     }
     if (_version < currentVersion) {
       // for each version between currentVersion and _version
       // call downgradeTo(i)
     }
   }

   // When adding a new storage schema, the developer would add the logic here
   // We would preserve all possible storage migrations
   function upgradeTo(uint256 _version) internal {
     //
   }
}
```

This may not be the most simple abstractions for this idea, I think it can be improved, somebody needs to prototype this a bit more, but the idea is that we have the developer add the migration permanently to the migration contract and defines the forwards and backwards migration. 

The downside of this is a new migration contract has to be deployed each time there is a new migration for a particular contract. We could have 1 migration contract that defines the transition between 2 versions or we could have the latest version always support all previous migrations, i lean towards the second.

PR #159, opened 2024-11-12
OPCM L1 Upgrades

maurelian:
Describes a fully onchain upgrade process.

### DIFF
--- a/protocol/l1-upgrades.md
+++ b/protocol/l1-upgrades.md
@@ -69,7 +69,7 @@ Our upgrades have the following issues:
 Each new release will have its own OP Contracts Manager which can perform the following distinct
 actions:
 
-1. Deploy new proxies for a new OP Chain (this work is already covered by [link to Blaine's doc]())
+1. Deploy new proxies for a new OP Chain (this work is already covered by a recent [OPCM redesign](./op-contracts-manager-single-release-redesign.md)).
 1. Upgrade all existing OP Chains.
 1. Deploy any newly added superchain-shared contracts.
 1. Upgrade existing superchain-shared contracts.
@@ -77,14 +77,18 @@ actions:
 Note that the above functionality also makes for a good set of milestones, each of which can and
 should be implemented incrementally.
 
-## Declarative storage management
+## Contract specific upgrade paths
 
 Taking inspiration from Chugsplash, we wish to separate the concern of what bytecode to run, and
 what to write to storage. However we also wish to avoid using the current `StorageSetter` pattern,
 which simply provides logic to "write anything anywhere" in storage.
 
-In order to achieve this, each contract will have its `initializer()` function moved intoa
-release specific `__Populator` contract.
+In order to achieve this, each contract will:
+
+1. Use the OpenZeppelin v5 Initializable contract, which stores the `initialized` value in an
+   [unstructured storage location](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.0.2/contracts/proxy/utils/Initializable.sol#L77).
+2. Have a new `upgrade()` method added alongside the existing `initialize()` method, both of which
+   will have the `initializer` modifier
 
 By way of example, we take an initializable version of `SimpleStorage`:
 
@@ -111,56 +115,44 @@ contract SimpleStorage is Initializable {
 }
 ```
 
-This document proposes that the above contract should be refactored accordingly:
+The only change to this contract will be to add a new `upgrade()` function which also
+has the `initializer` modifier.
 
 ```solidity
-/// @notice The base contract with all runtime logic
-contract SimpleStorage {
-  uint public foo;
-  bytes32 public bar; // new variable added for the current release.
-}
-
-/// @notice an ephemeral contract which is temporarily set as the implementation during deployment
-///         or upgrade.
-contract SimpleStoragePopulator is SimpleStorage {
-  /// @notice Called to initialize a freshly deployed system.
-  function initialize(uint _foo, bytes32 _bar) public {
-    foo = _foo;
-    bar = _bar;
-  }
-
   /// @notice Called to upgrade a pre-existing system.
-  ///         This function is only necessary if new values are being added to the storage layout.
-  function upgrade(bytes32 _bar) public {
+  ///         If new values are not being added to the storage layout, this function can be empty
+  ///         and take no arguments.
+  function upgrade(bytes32 _bar) public initializer {
     bar = _bar;
   }
-}
 ```
 
-With this architecture, the OPCM contract (which will be a non-proxied singleton per release) would
-then either:
+The `StorageSetter` contract can then be replaced with a special purpose contract which
+directly reset the `initialized` slot.
 
-The new contract deployment flow used by the OPCM thus becomes:
+The new contract **deployment** flow used by the OPCM thus becomes:
 
-1.  Deploy a new proxy
-2.  If and only if the proxy has state variables:
-    1.  set its implementation to the `SimpleStoragePopulator`
-    1.  call `initialize()`
-3.  Finally set its implementation to `SimpleStorage`
+1. Deploy a new proxy.
+2. Set its implementation to `SimpleStorage`.
+3. Call `SimpleStorage.initialize()`.
 
-And the contract upgrade flow used by the OPCM would be:
+And the contract **upgrade** flow used by the OPCM would be:
 
-1. If and only if the proxy has NEWLY ADDED state variables:
-   1. set its implementation to the `SimpleStoragePopulator`
-   1. call `upgrade()`
-1. set its implementation to `SimpleStorage`
+1. set its implementation to an `InitializerResetter` contract (does what the name says).
+2. call `InitializerResetter.reset()`
+3. set its implementation to `SimpleStorage`.
+4. Call `SimpleStorage.initialize()`.
 
-This workflow has the benefit of removing the `initializer()` functions, which:
+**Altneratives considered:** Other approaches were considered including:
 
-- create a risk of reinit attacks
-- bloat the runtime bytecode size
-- are extremely verbose
-- require finesse to avoid modifying or resetting values which are not meant to be changed.
+1. Remove the two step upgrade approach. This would have required using a `reinitializer` like pattern
+   which was deemed more complex and error prone.
+2. Separating the `initializer()` and `upgrade()` methods into a separate contract used in the
+   first part of a two step upgrade. This was deemed to complex as it created more contracts to manage.
+3. Breaking out the storage layout into a separate contract, this would have created a complex refactor
+   and inheritance tree.
+4. Using `StorageSetter` directly udring the first step, in place of an `upgrade()` method. This approach
+   would not allow us to easily perform sanity checks on inputs without duplicating logic into the OPCM.
 
 ## A single `SuperchainProxyAdmin`
 
@@ -170,22 +162,30 @@ will have minimal storage, and will not internally track the type of Proxy which
 
 This contract will be owned by the [Upgrade
 Controller](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/stage-1.md#L30), and
-will be a simple pass through contract with a minimal interface.
+will have the following minimal interface:
 
 ```solidity
-/// @notice forwards calldata to the specified address.
-///   Reverts unless called by owner.
-function forward(address, bytes) external;
+function upgrade(address _proxy, address _implementation) external;
+function upgradeAndCall(address _proxy, address _implementation, bytes memory _data) external;
 
-/// @notice forwards calldata to the specified addresses.
-///   Reverts unless called by owner.
-function forward(address[], bytes[]) external;
+function upgradeResolved(address _addressManager, address _implementation) external;
+function upgradeAndCallResolved(address _addressManager, address _implementation, address _proxy, bytes memory _data) external;
+
+function upgradeChugsplash(address _proxy, address _implementation) external;
+function upgradeAndCallChugsplash(address _proxy, address _implementation, bytes memory _data) external;
 ```
 
 This architecture makes it possible to manage upgrade authorization for the whole superchain through
-a single storage value in a single contract (`SuperchainProxyAdmin.owner`), rather than having a separate contract per OP Chain.
+a single storage value in a single contract (ie. `SuperchainProxyAdmin.owner`), rather than having a separate contract per OP Chain.
 
-It also avoids the need for an `sload` to read the `proxyType` mapping for each contract.
+By having the caller identify the proxy type, it also avoids the need for an `sload` to read the
+`proxyType` mapping for each contract.
+
+**Altneratives considered:** Other approaches were considered including:
+
+1. Keeping many `ProxyAdmin`s: this is complicated and expensive.
+2. Making the `SuperchainProxyAdmin` an even thinner wrapper which effectively just forwards
+   calldat. The approach chosen will reduce the amount of boilerplate in the OPCM.
 
 ## Identifying OP Chains by the SystemConfig
 
@@ -204,7 +204,7 @@ compatibility, the existing `version()` getters will return the value from `Syst
 The `DeployImplementations` script will write the version into the OPCM, which writes it into the
 `SystemConfig`.
 
-The `rc` tag will be included on the OPCM's initial version. It will be programmatically removed
+The `rc` tag will be included on the OPCM upon deployment. It will be programmatically removed
 when `upgrade()` is called by the Upgrade Controller multisig, which signifies governance approval.
 
 ## Proof system considerations
@@ -212,14 +212,23 @@ when `upgrade()` is called by the Upgrade Controller multisig, which signifies g
 ### Adding the PermissionlessDisputeGame to a chain
 
 Given that not all chains will support the `PermissionlessDisputeGame` upon deployment, an
-`OPCM.addGameType()` method will be added which will handle this.
+`OPCM.addGameType()` method will be added which will orchestrate the actions required to add a
+new game type.
 
-WIP
+This method will:
 
-1. ideally we update op-program to remove the chain config, so that it is identical across chains
-   then it reads the config at run time.
-   But we won't block isthmus on it, and will allow the `FaultDisputeGame` and `PermissionedDisputeGame` to be non-compliant.
-2. inputs for each hardfork: one prestatehash per chain.
+1. deploy the `FaultDisputeGame` contract
+2. setup the `DelayedWethProxy` for the new game
+3. Reinitialize the `AnchorStateRegistry` to add the new game type.
+
+### Making `op-program` shared
+
+`op-program` will be updated by [other work](https://github.com/ethereum-optimism/design-docs/pull/161) to remove the chain config, so that it can be shared across chains.
+
+### Short term allowances for non-compliant contracts
+
+In order to avoid blocking this work, if the `FaultDisputeGame` and `PermissionedDisputeGame`
+have not become MCP-L1 compatible in time, the OPCM upgrade path will deploy chain specific instances.
 
 ## Release process
 
@@ -229,12 +238,9 @@ The Upgrade Controller (which MUST be a Safe), will `DELEGATECALL` the `OPCM.upg
 providing the address of the `SuperchainProxyAdmin` and a list of the `SystemConfig` contracts for the
 OP Chains to be upgraded.
 
-Importantly,
-
 Thus the high level logic for upgrading a contract should be roughly as follows:
 
-```js
-
+```solidity
 // Some upgrades will
 struct NewChainConfig {
   address newValues;
@@ -250,14 +256,10 @@ function upgrade(SuperchainProxyAdmin _admin, ISystemConfig[] _systemConfigs, Ne
   for(uint i=0; i< systemConfigs.length; i++) {
     // Read the `Addresses` struct from each `SystemConfig` in the `systemConfigs` mapping.
     // For each entry in the `Addresses` struct:
-      // 1. Call the `SuperchainProxyAdmin` to update the implementation to the `Populator` contract
-      //    TODO: OR update the implementation to
-      // 2. call
-      // 2. Update the implementation to the final implementation contract.
+      // 1. Call the appropriate `SuperchainProxyAdmin.upgradeAndCall()` function to reset the initialized slot.
+      // 2. Call the appropriate `SuperchainProxyAdmin.upgradeAndCall()` function to update the
+      //    implementation and call `upgrade()` on the contract.
   }
-  // Return ownership of the `SuperchainProxyAdmin` to the Upgrade Controller (`msg.sender`)
-      //    `SuperchainProxyAdmin`. The unique upgrade call encoding required for the non-standard proxies
-      //    should be handled here.
 }
 ```
 
@@ -266,18 +268,29 @@ function upgrade(SuperchainProxyAdmin _admin, ISystemConfig[] _systemConfigs, Ne
 - A new getter should be added to the `SystemConfig` to retrieve the `Addresses` in a single call.
 - The `AddressManager` for each chain must be used to upgrade the `L1CrossDomainMessenger`, therefore it should be added to the `Addresses` struct.
 
+Importantly, this design does not provide special affordances to enable systems with different
+upgrade controllers to upgrade together atomically, although this could be achieved with additional
+coordination.
+
+## Upgrading multiple versions
+
+All systems MUST undergo the same upgrade path, using each OPCM in order.
+
+For example, if the latest version is `2.2.0`, and a system is on `2.0.0`, then it will first need to
+be upgraded to `2.1.0` using the OPCM for that version.
+
+The address of the OPCM corresponding to each upgrade will be stored in the `superchain-registry`,
+and `op-deployer` will ensure that upgrades conform to the specified ordering.
+
 ## Development and Testing considerations
 
 This solution should be implemented such that it both enables and _requires_ developers to implement
 the upgrade path from the most recent release to the one curently being developed. This means it
 will be necessary to have a deployment of the previous release available on the current commit, in
-order to test the upgrade path.
+order to test the upgrade path. This is achievable using `op-deployer`.
 
-This can be achieved by storing an L1 Genesis representation of the latest release in the monorepo.
-This genesis file can be generated by from a state dump of the currently deployed test system. The
-test suite should then be modified so that each test runs twice, one against a system freshly
-deployed by the OPCM, and another which was upgraded from the previous release. In CI these two
-major test groups can be run in parallel.
+Testing should also be done to ensure that, given the same configuration as inputs, a fresh chain
+has the same resulting configuration as an upgraded chain.
 
 # Resource Usage
 
@@ -289,10 +302,12 @@ A rough per-chain estimate (accounting for lookups but not memory) based on the 
 - `CALL` the `SuperchainProxyAdmin`: 2600
   - Do the following 6 times:
     - `CALL` the Proxy: 2600
-    - Cold `SSTORE` the temporary `Populator` implementation address over a non-zero slot: 5,000
-    - Cold `SSTORE` new values: 20,000 each assuming empty slots.
+    - Cold `SSTORE` the temporary `InitializerResetter` implementation address over a non-zero slot: 5,000
+    - Cold `SSTORE` a zero value: 2900
     - Warm `CALL` the Proxy again: 100
     - Warm `SSTORE` the final implementation address over a non-zero slot: 100
+    - Warm `CALL` the Proxy again: 100
+    - Cold `SSTORE` new values: 20,000 each
 
 Aggregating the above slightly:
 
@@ -310,20 +325,8 @@ We will likely need to handle that limitation at some point in the future, but c
 
 # Alternatives Considered
 
-1. Keeping the existing L1 Proxy Admin was rejected as being overly redundant and inefficient.
-2.
-3. Keeping initializers, and the StorageSetter was rejected as more over-prone, though possible simpler for testing purposes.
-4.
+Where relevant, alternatives considered are documented above in each section.
 
 # Risks & Uncertainties
 
 TODO
-
----
-
-6. We will track the correct addresses of the OPCMs in the registry. If you are upgrading by more than one version, you'll
-   get the list from there.
-
-function upgrade() {
-require(systemConfig.releaseVersion() < )
-}
--- a/protocol/l1-upgrades.md
+++ b/protocol/l1-upgrades.md
@@ -121,12 +121,12 @@ The only change to this contract will be to:
 2. add a new `upgrade()` function which also has the `initializer` modifier.
 
 ```solidity
-  /// @notice Called to upgrade a pre-existing system.
-  ///         If new values are not being added to the storage layout, this function can be empty
-  ///         and take no arguments.
-  function upgrade(bytes32 _bar) public initializer {
-    bar = _bar;
-  }
+/// @notice Called to upgrade a pre-existing system.
+///         If new values are not being added to the storage layout, this function can be empty
+///         and take no arguments.
+function upgrade(bytes32 _bar) public initializer {
+  bar = _bar;
+}
 ```
 
 The `StorageSetter` contract can then be replaced with a special purpose contract which
@@ -262,6 +262,7 @@ Thus the high level logic for upgrading a contract should be roughly as follows:
 // Some upgrades will
 struct NewChainConfig {
   address newValues;
+  address disputeGameBlueprint;
 }
 
 Addresses immutable implementations;
@@ -279,6 +280,7 @@ function upgrade(SuperchainProxyAdmin _admin, ISystemConfig[] _systemConfigs, Ne
       // 1. Call the appropriate `SuperchainProxyAdmin.upgradeAndCall()` function to reset the initialized slot.
       // 2. Call the appropriate `SuperchainProxyAdmin.upgradeAndCall()` function to update the
       //    implementation and call `upgrade()` on the contract.
+      // 3. For non-MCP compliant dispute game contracts, call `setImplementation()` and update the AnchorStateRegistry.
   }
   // run safety assertions to validate the upgrade
 }
@@ -291,7 +293,9 @@ To enumerate the full flow:
    1. immutables pointing to all new implementation contracts and shared config (e.g. a new contract).
 1. For each new implementation an upgrade call will be executed via the `SuperchainProxyAdmin`
 1. A Safe will `DELEGATECALL` to the OPCM.upgrade(...) method, which MUST be `pure` for safety purposes.
-1. A two step upgrade is used where the first upgrade is to an `UpgradePreparer` which resets the `initialized` value, then the implementation is updated to the final address and `upgrade()` is called.
+1. A two step upgrade is used where the first upgrade is to an `InitializerResetter` which resets
+   the `initialized` value, then the implementation is updated to the final address and `upgrade()`
+   is called.
 
    **Notes:**
 
@@ -319,10 +323,13 @@ the upgrade path from the most recent release to the one curently being develope
 will be necessary to have a deployment of the previous release available on the current commit, in
 order to test the upgrade path. This is achievable using `op-deployer`.
 
-Testing should also be done to ensure that, given the same configuration as inputs, a fresh chain
-has the same resulting configuration as an upgraded chain.
+All tests should be executed against both a freshly deployed and upgraded system.
 
 **Alternatives considered:**
+
+- Rather than duplicated tests, new tests could ensure that given the same configuration as inputs,
+  a fresh chain has the same resulting configuration as an upgraded chain. Duplicating tests was
+  deemed acceptable as they are parallelizable.
 
 # Resource Usage
 
--- a/protocol/l1-upgrades.md
+++ b/protocol/l1-upgrades.md
@@ -36,9 +36,8 @@ This document is not concerned with:
 1. modifications to the ownership system, ie, it does not attempt to modify Safes, modules or
    guards.
 
-More over, it does not present a plan for managing contracts which are not proxied, and not
-Multichain Prepared (MCP-L1) contracts (ie. FaultDisputeGame, AnchorStateRegistry). This
-work will either include, or else depend on, all L1 contracts becoming either:
+With the exception of dispute game implementations this work will either include, or else depend on,
+all L1 contracts becoming either:
 
 1. Proxied and MCP ready
 2. Superchain-shared contracts (proxied or not)
@@ -259,7 +258,7 @@ OP Chains to be upgraded.
 Thus the high level logic for upgrading a contract should be roughly as follows:
 
 ```solidity
-// Some upgrades will
+// Some upgrades will not require any new values
 struct NewChainConfig {
   address newValues;
   address disputeGameBlueprint;
@@ -280,7 +279,7 @@ function upgrade(SuperchainProxyAdmin _admin, ISystemConfig[] _systemConfigs, Ne
       // 1. Call the appropriate `SuperchainProxyAdmin.upgradeAndCall()` function to reset the initialized slot.
       // 2. Call the appropriate `SuperchainProxyAdmin.upgradeAndCall()` function to update the
       //    implementation and call `upgrade()` on the contract.
-      // 3. For non-MCP compliant dispute game contracts, call `setImplementation()` and update the AnchorStateRegistry.
+      // 3. For non-MCP compliant dispute game contracts, call `setImplementation()` and update the `AnchorStateRegistry`.
   }
   // run safety assertions to validate the upgrade
 }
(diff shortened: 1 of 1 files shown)


protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+
+This `SimpleStorage` will become:
+
+```solidity
+abstract contract SimpleStorageLayout {
+  uint public foo;
+  bytes32 public bar;
+}
+
+contract SimpleStoragePopulator is SimpleStorageLayout {
```
mds1, line 121: What would a more complex example look like, where `SimpleStorage` inherits from 2 or more base contracts that have their own storage layouts, epecially when they are deps like OZ where we might not have control over the storage layout? I suspect this gets a bit messer and perhaps not doable.

The first alternative that comes to mind is to remove `SimpleStorageLayout` and do `SimpleStoragePopulator is SimpleStorage` and add the populator methods. The populator has a lot of extra code but that is mainly harmless

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+
+The existing L1 `ProxyAdmin` (of which there is currently one per OP Chain), will be replaced with a
+single L1 `SuperchainProxyAdmin`. The `SuperchainProxyAdmin`, in contrast with the L1 `ProxyAdmin`,
+will have minimal storage, and will not internally track the type of Proxy which each contract is.
+
+This contract will be owned by the [Upgrade
+Controller](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/stage-1.md#L30), and
+will be a simple pass through contract with a minimal interface.
+
+```
```
mds1, line 167: nit
```suggestion
```solidity
```

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+
+- create a risk of reinit attacks
+- bloat the runtime bytecode size
+- are extremely verbose
+- require finesse to avoid modifying or resetting values which are not meant to be changed.
+
+## A single `SuperchainProxyAdmin`
+
+The existing L1 `ProxyAdmin` (of which there is currently one per OP Chain), will be replaced with a
+single L1 `SuperchainProxyAdmin`. The `SuperchainProxyAdmin`, in contrast with the L1 `ProxyAdmin`,
```
mds1, line 160: Just to avoid any confusion:
- Technically there are 2 roles that have a proxy admin: The SuperchainConfig and ProtocolVersions proxies, and the `Ownable` parts of contracts like the SystemConfig and DisputeGameFactory (DGF)
- Pre-OPCM, for the standard config, we use the same exact proxy admin for both of those, conflating the two roles
- OPCM split these out into two distinct proxy admins: a `superchainProxyAdmin` and an `opChainProxyAdmin`

Therefore I just want to avoid confusion on what `SuperchainProxyAdmin` means here since it's basically talking about a new version of the `ProxyAdmin` contract

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+///   Reverts unless called by owner.
+function forward(address, bytes) external;
+
+/// @notice forwards calldata to the specified addresses.
+///   Reverts unless called by owner.
+function forward(address[], bytes[]) external;
+```
+
+This architecture makes it possible to manage upgrade authorization for the whole superchain through
+a single storage value in a single contract, rather than having a separate contract per OP Chain.
```
mds1, line 178: Referring to the `owner` in the SuperchainProxyAdmin?

Another benefit is we can change the single proxy admin in the future to add new capabilities, such as a timelock to delay transactions

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+
+## The upgrade process
+
+As part of the release process, the associated implementation contracts and a new OPCM, will be deployed.
+
+The Upgrade Controller MUST be a Safe in order to atomically:
+
+1. transfer ownership of the `SuperchainProxyAdmin` to the new OPCM.
+2. call `OPCM.upgrade()` which will return ownership to the Safe immediately upon completing the
+   upgrade.
```
mds1, line 199: This is pretty critical, how can we enforce onchain this transfer back to the original owner occurred and revert if it didn't? Perhaps we want a guard in the 2/2 safe that leverages the `checkAfterExecution` hook for this

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+
+Thus the high level logic for upgrading a contract should be roughly as follows:
+
+```js
```
mds1, line 203:
```suggestion
```solidity
```

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+    systemConfigs.push(_systemConfigs[i])
+  }
+}
+
+function upgrade() public {
+  for(uint i=0; i< systemConfigs.length; i++) {
+    // Read the `Addresses` struct from each `SystemConfig` in the `systemConfigs` mapping.
+    // For each entry in the `Addresses` struct:
+      // 1. Call the `SuperchainProxyAdmin` to update the implementation to the `Populator` contract and call `Populator.upgrade()`.
+      // 2. Update the implementation to the final implementation contract.
```
mds1, line 221: We talked about this a bit, another option to consider is the double-delegatecall approach: do a 1 time change of all proxy implementations to a beacon contract, which looks up the implementation. Then this step moves outside the loop

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+```
+
+**Note:**A new getter should be added to the `SystemConfig` to perform this in a single call.
+
+## L1ChugsplashProxy Considerations
+
+The L1 Chugsplash proxy is much more expensive to upgrade, as it is done by calling
+[`setCode()`](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/legacy/L1ChugSplashProxy.sol#L92)
+which creates a new implementation contract. In this sense the L1StandardBridge is not truly MCP-L1
+ready.
```
mds1, line 236: What the ProxyAdmin currently does for chugsplash proxies is call `setStorage` on the proxy to directly overwrite the implementation address

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+This solution should be implemented such that it both enables and _requires_ developers to implement
+the upgrade path from the most recent release to the one curently being developed. This means it
+will be necessary to have a deployment of the previous release available on the current commit, in
+order to test the upgrade path.
+
+This can be achieved by storing an L1 Genesis representation of the latest release in the monorepo.
+This genesis file can be generated by from a state dump of the currently deployed test system. The
+test suite should then be modified so that each test runs twice, one against a system freshly
+deployed by the OPCM, and another which was upgraded from the previous release. In CI these two
+major test groups can be run in parallel.
```
mds1, line 268: Having to run the full test suite twice is a big downside that we don't have for the alternate approach where we keep the `initialize` method. Something to discuss in the "Alternatives considered" section

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+  // Whatever runtime logic is needed
+}
+```
+
+With this architecture, the OPCM contract (which will be a non-proxied singleton per release) would
+then either:
+
+1. Deploy a new contract by:
+   1. Deploying a new proxy
+   1. setting its implementation to the `SimpleStoragePopulator`
```
mds1, line 142: What if a given upgrade requires no changes to the initialization logic and therefore no populator is needed, do we still do the step as a no-op or does it get skipped? Let's cover that case in this doc also

protocol/l1-upgrades.md
mds1: Actually, is there a reason we need to do the ownership transfers at all? We can just have the safe delegatecall to `OPCM.upgrade()`

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+1. They are divorced from the implementation, meaning the work of writing the upgrade script occurs
+   in `superchain-ops` apart from the work of writing the updated code in the `optimism` monorepo.
+2. The act of writing the upgrade script is onerous, requiring:
+   1. A bespoke script with custom assertions.
+   2. A bespoke Validations.md document
+3. The work required to perform the upgrade is difficult for signers, as they must work through the
+   Validations.md document. This work has taken up to an hour in the past.
+4. The effort described above will scale linearly with the number of OP Chains.
+
+# Proposed Solution
```
mds1, line 74: We should have an onchain way to enforce sequencing between OPCMs when upgrading multiple versions. For example if a chain is on 2.0.0, and 2.1.0 and 2.2.0 have since been released, we want to prevent a footgun where the chain uses OPCM 2.2.0 to upgrade straight to the latest. They should be forced to apply the 2.1.0 and 2.2.0 upgrades (nice to have would be to do both upgrades atomically, but separate transactions seems ok if it's simpler)

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+(`ResolvedDelegateProxy`).
+
+## The upgrade process
+
+As part of the release process, the associated implementation contracts and a new OPCM, will be deployed.
+
+The Upgrade Controller MUST be a Safe in order to atomically:
+
+1. transfer ownership of the `SuperchainProxyAdmin` to the new OPCM.
+2. call `OPCM.upgrade()` which will return ownership to the Safe immediately upon completing the
```
mds1, line 198: How is chain specific config passed? Is it calldata? It could also be in storage which feels easier for atomic execution. For example to atomically upgrade all standard chains (some owned by a 2/2 and some owned by a 3/3) we could have a pattern where chains register their config first, then upgrade reads it from storage. For example:

```solidity
contract OPContractsManager {
  mapping (address systemConfig => uint256) params;

  function set(address _systemConfig, uint256 _param) external {
    // This check is not feasble onchain so we'd need a workaround, see https://github.com/ethereum-optimism/client-pod/issues/699
    require(msg.sender == _systemConfig.admin().owner());

    param[_systemConfig] = param;
  }

  function upgrade(address _systemConfig) external { 
    uint256 param = params[_systemConfig];
    // rest of logic here
  }
}
```

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+
+constructor(Addresses _implementations, ISystemConfigs[] _systemConfigs) {
+  implementations = _implementations;
+  for(uint i=0; i < _systemConfigs.length; i++){
+    systemConfigs.push(_systemConfigs[i])
+  }
+}
+
+function upgrade() public {
+  for(uint i=0; i< systemConfigs.length; i++) {
```
mds1, line 217: We can simplify logic by taking the system config addrs as calldata. This way any chain (regardless of their L1 ProxyAdmin owner) can use this contract to upgrade their L1 contracts.

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+
+# Proposed Solution
+
+Each new release will have its own OP Contracts Manager which can perform the following distinct
+actions:
+
+1. Deploy new proxies for a new OP Chain (this work is already covered by [link to Blaine's doc]())
+1. Upgrade all existing OP Chains.
+1. Deploy any newly added superchain-shared contracts.
+1. Upgrade existing superchain-shared contracts.
```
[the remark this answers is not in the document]
mds1: What is the absolute prestate a function of? I'd like to better understand why the above requirements are the case

> having the absolute prestate as an immutable in FaultDisputeGame seems like a pretty perfect fit for what we need, other than that it doesn't fit the typical MCP approach.

Could it be a regular storage variable with no setter method? That way we reuse the FaultDisputeGame implementation, and just put a proxy in front to have different storage values per chain

protocol/l1-upgrades.md
maurelian: > What would a more complex example look like, where SimpleStorage inherits from 2 or more base contracts that have their own storage layouts, especially when they are deps like OZ where we might not have control over the storage layout? I suspect this gets a bit messer and perhaps not doable.

I expect they would look a lot like the state var declaration section of our existing contracts. Unless I've missed something, the only external deps our L1 contracts currently inherit from are `Initializable` and `OwnableUpgradable`. This proposals removes the former, and we should also move away from using the Ownable pattern. Gaps and spacers can be used as needed. 

> The first alternative that comes to mind is to remove `SimpleStorageLayout` and do `SimpleStoragePopulator` is `SimpleStorage` and add the populator methods.

The benefit of the proposed approach is that we don't get the additional buffer below the contract size limit, that we get form moving the `initializer` into a contract with no other logic. However that is a 'nice to have' beyond the goals, so I like this as a simplification.

protocol/l1-upgrades.md
maurelian: Yes, and agree!

protocol/l1-upgrades.md
maurelian: Yeah, DELEGATECALL is easier, lets just do that.

protocol/l1-upgrades.md
maurelian: Ahhh, good catch! 

So then what we could do for the `L1StandardBridge` is have the make that storage write at the end `upgrade()`/`initialize()`. 

And thinking about it now, perhaps we should do that for all of the contracts so that: 
1. they all have a consistent upgrade flow
2. we have two fewer CALLs per contract being upgraded
wdyt?

protocol/l1-upgrades.md
maurelian: Instead of duplicating the tests, I think it should be sufficient to prove that the two paths result in contracts with identical config.

protocol/l1-upgrades.md
maurelian: Good callout. 

I think given we want to use the superchain-registry as the source of truth for OPCM addresses, the approach that comes to my mind is: 

1. Add an `l1ContractsVersion()` getter to the `SystemConfig`. The same getter being added to the OPCM by #158.
2. Add a `previousL1ContractsVersion` value to the OPCM. Have the OPCM check that the current `SystemConfig.l1ContractsVersion() == OPCM.previousL1ContractsVersion` before performing the upgrade.

protocol/l1-upgrades.md
maurelian: I had imagined it being in calldata, but I like this pattern a lot for how it enables auditing and clear auth around setting new config values before an `upgrade()`. I'd call it `newParams` just to be clear.

protocol/l1-upgrades.md
maurelian: If we're indifferent to gas costs them I'm down to do this.

protocol/l1-upgrades.md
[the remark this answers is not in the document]
maurelian: I imagine it relates to the use of CWIA to deploy the game instances.

protocol/l1-upgrades.md
maurelian: agreed we won't do the double proxy thing.

protocol/l1-upgrades.md
maurelian: We won't have a Populator, so in this case we just `upgrade` instead of `upgradeAndCall`

protocol/l1-upgrades.md
maurelian: Decision: do config params in calldata.
For chains with a different upgrade controller, it's a separate transaction.

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+    foo = 1;
+    bar = 2_000
+  }
+
+  function upgrade() public {
+    bar = 2_000;
+  }
+}
+
+contract SimpleStorage is SimpleStorageLayout {
```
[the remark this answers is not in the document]
maurelian: we decided: 
1. no Populator
2. add an `upgrade()` function

protocol/l1-upgrades.md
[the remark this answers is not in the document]
ajsutton: The absolute prestate is basically the hash of `op-program` (converted to a cannon state snapshot). It effectively embeds the consensus rules (ie current hard fork implementation) and the chain configuration.  It can embed the chain configuration for multiple chains in the same version, but chain config changes happen at a whole bunch of times, not just when op-mainnet activates a hard fork (new chains, chains activating forks at different times etc).

We can't really use a proxy per chain because the `DisputeGameFactory` needs to be able to create new proxies when it creates a new game. I don't think the use of CWIA vs some other proxy really matters here, it's just that it's not a one-off thing, we create new instances all the time. Theoretically the `DisputeGameFactory` could store the absolute prestate for each game type but that's way outside what `DisputeGameFactory` should know about and bleeds the specific details of `FaultDisputeGame` into the otherwise pretty generic `DisputeGameFactory` (which really should have a more generic name since the "games" it creates could just as easily be for ZK proofs and wouldn't have an absolute prestate at all).

My best idea for solving this is to introduce a `ChainConfigRegistry`: https://github.com/ethereum-optimism/design-docs/pull/161 but there's details still to work out there and it hasn't been scheduled so is unlikely to be done prior to Isthmus.

We do have a good superchain-ops for the `setImplementation` calls required to upgrade fault proofs (https://github.com/ethereum-optimism/superchain-ops/pull/363) but still need something to deploy the contracts (and because of the potential for chain configs to change, that may need to happen outside of hard forks or releases of op-contracts) but it would only differ by the absolute prestate.

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+   governance approved bytecode.
+
+Further gas cost analysis and product discussion will be required to make a final decision.
+
+## Development and Testing considerations
+
+This solution should be implemented such that it both enables and _requires_ developers to implement
+the upgrade path from the most recent release to the one curently being developed. This means it
+will be necessary to have a deployment of the previous release available on the current commit, in
+order to test the upgrade path.
```
clabby, line 262: This will probably not work for proofs in the short-medium term.

We have `kona` + `asterisc` shipping soon (likely before Ithsmus on OP mainnet), which will be the first of several alternative proof systems deployed on the superchain. `asterisc` for example has a dependency on the monorepo, and has its FPVM contract (`RISCV.sol`) in that repository. We keep this separation to not bloat the monorepo, as the monorepo's contracts provide the interface for these alternative proof systems. Once we start moving towards multi-proofs, we'll likely have more examples of this, for example for a dispute game backed by a ZKP.

It would be great if OPCM can support contracts that do not exist within the monorepo to support the extensibility of the proof system (and other parts of the protocol) without requiring that everything must be upstreamed into the monorepo.

mds1: If these are becoming part of the core protocol, it seems they should live in the monorepo? Or could the dependency graph be flipped so the monorepo pulls them in as a dependency? Having them in the monorepo feels significantly simpler:
- Fits into the OPCM upgrades process
- All of our existing solidity and CI checks get run for free
- Easy to add new e2e tests with these proof systems in the monorepo
- No additional changes needed to our, development, release, and governance processes

Having key parts of the protocol in a separate repo feels like it opens a lot of questions, what is the plan for addressing all of the above?

mds1: The decision was made to vendor the `RISCV.sol` contract and this has been done in https://github.com/ethereum-optimism/optimism/pull/12994, so I will resolve this thread

protocol/l1-upgrades.md
maurelian: Agree.

protocol/l1-upgrades.md
maurelian: I removed this section as no longer relevant.

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,338 @@
...
+While the length will likely be proportional to the length of the full document,
+the summary should be as succinct as possible. -->
+
+The proposed solution will perform upgrades by transferring upgrade authorization of all L1
+contracts to the release specific OPCM. Upon completion of the upgrade, this authorization will
+be returned to the Security Council.
+
+Rather than the curent `StorageSetter` based approach which simply deletes the `_initialized` slot,
+allowing for `initialize()` to be called, we take this two-step upgrade further, with a short-lived
+`Populator` contract being temporarily set as the implementation in order to set or modify any
```
tynes, line 57: I do not buy the benefits listed here as the benefits of the `Populator` contract. The `Populator` contract is basically `StorageSetter` but with awareness of the storage layout of the particular contract. You can do everything you would with `Populator` with `StorageSetter`. It could be the case that its just more simple to continue to use `StorageSetter` (and improve its API if necessary with things like `setBool(bool _value, uint256 _offset)`)

The benefits listed are benefits, but have nothing to do with `Populator` specifically

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,338 @@
...
+```solidity
+/// @notice The base contract with all runtime logic
+contract SimpleStorage {
+  uint public foo;
+  bytes32 public bar; // new variable added for the current release.
+}
+
+/// @notice an ephemeral contract which is temporarily set as the implementation during deployment
+///         or upgrae.
+contract SimpleStoragePopulator is SimpleStorage {
```
tynes, line 134: This isn't guaranteed to work due to codesize limits. This will not work with the `OptimismPortal` today as any new code will push it over the limit. So it would mean that we would have to pay off tech debt before being able to do this sometimes

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,338 @@
...
+
+By way of example, we take an initializable version of `SimpleStorage`:
+
+- the current version of the contract has only a `uint256 foo` variable.
+- the new version of the contract introduces a `bytes32 bar` variable.
+
+The status quo implementation of the new Simple Storage would look like this:
+
+```solidity
+contract SimpleStorage is Initializable {
```
tynes, line 106: How much more difficult would it be to maintain using `StorageSetter` instead of using the `Populator` pattern? There are 2 paths where this code needs to be used. Once with new chains and once with upgrading chains.

For new chains, it would be set all storage and set the storage version to latest

For upgrading chains, it would be check storage version is 1 less than latest, add/move/remove any slots, then set the storage version to latest

There are 2 chunks of logic which may be different no matter what

tynes: The 2 chunks of logic are represented in `upgrade` vs `initialize`, this makes sense

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,338 @@
...
+
+We will use the `SystemConfig` as the unique identifier for each OPChain, and get the contract
+addresses from it. The custom logic required for the two non-standard proxy types will be handled in
+the OPCM contract, and does not require tracking the proxy type (as currently done in the
+`ProxyAdmin`), since it is known already for each contract, ie. the only non-standard Proxies are
+the `L1StandardBridge` (`L1ChugsplashProxy`) and `L1CrossDomainMessenger` (`ResolvedDelegateProxy`).
+
+## Contract versioning
+
+As part of this work, we will cease to version each L1 contract, and instead move to a global L1
```
tynes, line 209: This is interesting, I do think this will simplify things. Out of scope for this document, but we should od the same thing for the L2 contracts and put the source of truth in the `L1Block`. Right now the `SystemConfig` of L2 predeploys is the `L1Block` contract

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,338 @@
...
+## Proof system considerations
+
+### Adding the PermissionlessDisputeGame to a chain
+
+Given that not all chains will support the `PermissionlessDisputeGame` upon deployment, an
+`OPCM.addGameType()` method will be added which will handle this.
+
+WIP
+
+1. ideally we update op-program to remove the chain config, so that it is identical across chains
```
tynes, line 228: Design doc for this: https://github.com/ethereum-optimism/design-docs/pull/161

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+
+# Summary
+
+<!-- Most (if not all) documents should have a summary.
+While the length will likely be proportional to the length of the full document,
+the summary should be as succinct as possible. -->
+
+The proposed solution will perform upgrades by transferring upgrade authorization of all L1
+contracts to the release specific OPCM. Upon completion of the upgrade, this authorization will
+be returned to the Security Council.
```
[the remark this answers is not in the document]
maurelian: Ah, this section is outdated. I'll rewrite it once all the details are ironed out, but  FYI the current direction is now to: 
1. implement the upgrade orchestration logic in `OPCM.upgrade()` 
2. have the Upgrade Controller (SC+FND Safe) `DELEGATECALL` to `OPCM.upgrade()`, so that no ownership transfer is necessary.

protocol/l1-upgrades.md
maurelian: > How much more difficult would it be to maintain using StorageSetter instead of using the Populator pattern? There are 2 paths where this code needs to be used. Once with new chains and once with upgrading chains.

I believe it would be very difficult to replace [`SystemConig.initialize()`](https://github.com/ethereum-optimism/optimism/blob/maur/repeat-isthmus-contracts/packages/contracts-bedrock/src/L1/SystemConfig.sol#L185-L229) (which accepts 20 values) with the `StorageSetter`. Writing all the values is unergononomic but doable, however the many sanity checks on those values would need to be moved into the OPCM which would be difficult to maintain and reason about.

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,338 @@
...
+1. Upgrade existing superchain-shared contracts.
+
+Note that the above functionality also makes for a good set of milestones, each of which can and
+should be implemented incrementally.
+
+## Declarative storage management
+
+Taking inspiration from Chugsplash, we wish to separate the concern of what bytecode to run, and
+what to write to storage. However we also wish to avoid using the current `StorageSetter` pattern,
+which simply provides logic to "write anything anywhere" in storage.
```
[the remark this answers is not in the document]
maurelian: > What is the specific value prop of the Populator pattern? This will result in 2x contracts being deployed. 

IMO this pattern is similar to the `OptimismPortalMigration` you describe [here](https://github.com/ethereum-optimism/design-docs/issues/154), and I'd have used the `Migration` name had I seen that issue before writing this doc. 

The specific value props are: 
1. Storage layout awareness, meaning we can use vanilla solidity to set storage variables. ie. `foo = _foo` rather than `StorageSetter.setUint(bytes32(0), _foo)`, where `0` is a magic number, or you use assembly to get `foo.slot`. 
2. Ability to implement more complex logic (ie. if-branches, assertions and external calls), examples:
	1. an [if-branch](https://github.com/ethereum-optimism/optimism/blob/maur/repeat-isthmus-contracts/packages/contracts-bedrock/src/L1/ResourceMetering.sol#L168-L172) called within `OptimismPortal2.initialize()`.
	2. [safety](https://github.com/ethereum-optimism/optimism/blob/maur/repeat-isthmus-contracts/packages/contracts-bedrock/src/L1/SystemConfig.sol#L548-L564) [assertions](https://github.com/ethereum-optimism/optimism/blob/maur/repeat-isthmus-contracts/packages/contracts-bedrock/src/L1/SystemConfig.sol#L492-L493) called in `SystemConfig.intialize()`. 
	3. an [external call](https://github.com/ethereum-optimism/optimism/blob/maur/repeat-isthmus-contracts/packages/contracts-bedrock/src/L1/SystemConfig.sol#L483-L486) in `SystemConfig.initialize()`.

> We could just use StorageSetter and call it in a particular way for each contract 

I don't think we can due to my [response here](https://github.com/ethereum-optimism/design-docs/pull/159/files/34679b374bbf155a64182e5527b58f56a6c07faf#r1850696184) and above.

protocol/l1-upgrades.md
maurelian: I listed the benefits [here](https://github.com/ethereum-optimism/design-docs/pull/159/files/34679b374bbf155a64182e5527b58f56a6c07faf#r1850718398). 

Question for you: 

IMO this pattern is similar to the `OptimismPortalMigration` you describe [here](https://github.com/ethereum-optimism/design-docs/issues/154), and I'd have used the `Migration` rather than `Populator` name had I seen that issue before writing this doc. 

Could you please enumerate the benefits you think my proposal is missing from yours, so that I can try to address them in identifying the final design?

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+1. upgrading predeploys or other L2 contracts.
+1. modifications to the ownership system, ie, it does not attempt to modify Safes, modules or
+   guards.
+
+More over, it does not present a plan for managing contracts which are not proxied, and not
+Multichain Prepared (MCP-L1) contracts (ie. FaultDisputeGame, AnchorStateRegistry). This
+work will either include, or else depend on, all L1 contracts becoming either:
+
+1. Proxied and MCP ready
+2. Superchain-shared contracts (proxied or not)
```
mds1, line 44: I think we said we will have to make an exception for this, and the OPCM will support 3 main actions: deploy a new chain, upgrade an existing chain, and deploy/set FP contracts with a new absolute prestate hash for an existing chain

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,305 @@
...
+3. The work required to perform the upgrade is difficult for signers, as they must work through the
+   Validations.md document. This work has taken up to an hour in the past.
+4. The effort described above will scale linearly with the number of OP Chains.
+
+# Proposed Solution
+
+Each new release will have its own OP Contracts Manager which can perform the following distinct
+actions:
+
+1. Deploy new proxies for a new OP Chain (this work is already covered by [link to Blaine's doc]())
```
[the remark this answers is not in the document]
maurelian: TY

protocol/l1-upgrades.md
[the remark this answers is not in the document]
maurelian: Resolving as we're no long proposing the `Populator` and just adding `upgrade()` to the impl.

protocol/l1-upgrades.md
maurelian: Resolving as we're no long proposing the `Populator` and just adding `upgrade()` to the impl.

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+- The `AddressManager` for each chain must be used to upgrade the `L1CrossDomainMessenger`, therefore it should be added to the `Addresses` struct.
+
+## Development and Testing considerations
+
+This solution should be implemented such that it both enables and _requires_ developers to implement
+the upgrade path from the most recent release to the one curently being developed. This means it
+will be necessary to have a deployment of the previous release available on the current commit, in
+order to test the upgrade path.
+
+This can be achieved by storing an L1 Genesis representation of the latest release in the monorepo.
```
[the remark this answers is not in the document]
maurelian: OK, yes, we can just do that then.

protocol/l1-upgrades.md
tynes: We should make sure we are building in the same direction, re: deploy/set FP contracts, that seems related to the MCP the FP contracts line of work. Some of the work for this is described in https://github.com/ethereum-optimism/design-docs/pull/161

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+1. Proxied and MCP ready
+2. Superchain-shared contracts (proxied or not)
+
+# Summary
+
+<!-- Most (if not all) documents should have a summary.
+While the length will likely be proportional to the length of the full document,
+the summary should be as succinct as possible. -->
+
+TODO
```
tynes, line 52: This todo seems like a dead comment

protocol/l1-upgrades.md
tynes: I think we should have an unstructured storage slot that is set in every contract and read that and assert on its value when performing the migration rather than having a new getter on the system config

protocol/l1-upgrades.md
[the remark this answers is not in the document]
tynes: This was resolved but not addressed

protocol/l1-upgrades.md
tynes: This is a very lazy approach in my opinion where we opt towards higher deploy time/compile time costs under the guise of simplicity. This introduces tech debt that we will need to clean up in the future, we should remove an intermediate step and just go towards a more simple approach first.

Thinking of these options as better approaches:
- use storage setter, we already have tests for the constructor/initializer working as expected
- refactor the storage layout to be its own contract so that the populator doesn't need the implementation details

protocol/l1-upgrades.md
tynes: Two things:
- Base contract with storage layout inherited by the impl and the migrator rather than extending the impl with `initialize()`
- I now believe its less work to just use `StorageSetter` and hand write setting the storage slots in the impl of OPCM. This is very easy to cover with testing and reduces a ton of boilerplate that needs to be tested (every populator contract having a test) as well as boilerplate that needs to go on chain (why does populator have the impl details of the contract?)

protocol/l1-upgrades.md
karlfloersch: I'd be wary of using a storage setter - feels very error prone. Storage layout being in its own contract & inheriting from that is probably a bit better

protocol/l1-upgrades.md
karlfloersch: Hmm intuitively we want to test both fresh deployments & upgrades so I am not super duper concerned about this, especially because they can run in parallel

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+      // 1. Call the `SuperchainProxyAdmin` to update the implementation to the `Populator` contract
+      //    TODO: OR update the implementation to
+      // 2. call
+      // 2. Update the implementation to the final implementation contract.
+  }
+  // Return ownership of the `SuperchainProxyAdmin` to the Upgrade Controller (`msg.sender`)
+      //    `SuperchainProxyAdmin`. The unique upgrade call encoding required for the non-standard proxies
+      //    should be handled here.
+}
+```
```
karlfloersch, line 264: Am I correct in understanding that the upgrade function called with a DELEGATECALL in the safe is mostly just passed a list of addresses for the currently deployed contracts & the upgrad populator contracts? And all these contracts / all the upgrade function calldata is stored somewhere else?

Assuming my understanding is roughly correct, my main concern with this approach is that it seems extremely stateful. I really don't want there to be a footgun where we can deploy populator contracts or set upgrade values and then somehow those contracts change by the time we call upgrade and there's a diff in how the upgrade is executed.

It would be nice if the `upgrade(..)` function that the safe DELEGATECALLs to included as input data all information required for the upgrade. That way we just have one upgrade payload that we need to audit that contains all information utilized in the upgrade. I just don't want to have to look in a bunch of different places to find all the info I need to audit an upgrade if that makes sense

maurelian: Good callout, I share your concern.


> It would be nice if the upgrade(..) function that the safe DELEGATECALLs to included as input data all information required for the upgrade. 

Agree. What we can do is use the `pure` modifier on `upgrade()` to ensure that no reads or writes occur in the Safe.

mds1: Overall flow I am envisioning based on the most recent conversation I had with @maurelian:
- Deploy new `OPCM` contract, which contains:
  - `deploy` and `upgrade` methods to deploy new chains and ugprade existing chains
  - It also contains immutables pointing to all new implementation contracts (e.g. a new SystemConfig contract)
  - For each new implementation it knows how to to encode it's e.g. `SystemConfig.upgrade` call, and any chain-specific arguments to that call are passed into the root OPCM `upgrade` method
  - Sample `OPCM.upgrade` sig is `upgrade(systemConfig, arg1, arg2)`
- A safe delegate calls to the `OPCM.upgrade(...)` method
- The upgrade flow that OPCM does for each new implementation contract is a two step flow where first we set the implementation to a contract like this:
```solidity
contract UpgradePreparer {
  // Same initializer slot used by OpenZeppelin v5
  // keccak256(abi.encode(uint256(keccak256("openzeppelin.storage.Initializable")) - 1)) & ~bytes32(uint256(0xff))
  bytes32 private constant INITIALIZABLE_STORAGE = 0xf0c57e16840df040f15088dc2f81fe391c3923bec73e23a9662efc9c229c6a00;
  function prepare() external {

    // Clear the initializer slot so we can call `upgrade` on the new implementation
    assembly {
      sstore(INITIALIZABLE_STORAGE, 0)
    }
  }
}
```
- We call `prepare` to clear the slot
- Then we put the new `SystemConfig` contract in place and call it's `upgrade()` method. This sets the `INITIALIZABLE_STORAGE` slot to 1, and runs all applicable logic
- I am in favor of https://github.com/ethereum-optimism/design-docs/pull/159#discussion_r1852291740 which suggests always doing this two step upgrade flow even when `upgrade` is a no-op, as it simplifies the number of paths we have to worry about and reduces the chance of us forgetting an `upgrade` call

protocol/l1-upgrades.md
maurelian: Resolving. Please leave feedback on that plan on the "Modfications to L1 contracts" section.

protocol/l1-upgrades.md
maurelian: Outdated

protocol/l1-upgrades.md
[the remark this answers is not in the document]
maurelian: Fixed locally but wasn't pushed. https://github.com/ethereum-optimism/design-docs/blob/L1-upgrades/protocol/l1-upgrades.md#L72

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+about 250,000 per chain upgraded, we can fit 120 OP Chain upgrades
+inside an L1 block.
+
+We will likely need to handle that limitation at some point in the future, but can defer that for the time being.
+
+# Alternatives Considered
+
+1. Keeping the existing L1 Proxy Admin was rejected as being overly redundant and inefficient.
+2.
+3. Keeping initializers, and the StorageSetter was rejected as more over-prone, though possible simpler for testing purposes.
```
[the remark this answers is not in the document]
maurelian: This content is outdated, resolving.

protocol/l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+
+1.  Deploy a new proxy
+2.  If and only if the proxy has state variables:
+    1.  set its implementation to the `SimpleStoragePopulator`
+    1.  call `initialize()`
+3.  Finally set its implementation to `SimpleStorage`
+
+And the contract upgrade flow used by the OPCM would be:
+
+1. If and only if the proxy has NEWLY ADDED state variables:
```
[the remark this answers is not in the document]
maurelian: We are doing that now :)

protocol/l1-upgrades.md
maurelian:
```suggestion
TODO write this summary when the details are finalized.
```

protocol/l1-upgrades.md
maurelian: We are no longer using Populator, so I won't bother to defend it here. 
Touching on other topics: 

- Using a base storage layout contract: I don't think this is as simple or low overhead as it sounds. For example we would have the following inheritance graph: `L1CrossDomainMessenger -> CrossDomainMessenger  -> CrossDomainMessengerLayout`. This would mean that the `L2CrossDomainMessenger` can no longer inherit from `CrossDomainMessenger `, which is something we should strive to maintain.
- Using `StorageSetter`: I think we can use this pattern quite safely by: 
  1. Placing the `initialized` value in the same unstructured storage location for all contracts. 
  2. Using a custom contract (in place of `StorageSetter`) to reset that slot prior to each upgrade.

protocol/l1-upgrades.md
maurelian: I've ensured that the flow above is outlined in sufficient detail in the contents of the doc. 

Keeping this open as a place to discuss the use of DELEGATECALL in particular.

protocol/l1-upgrades.md
maurelian: > Hmm intuitively we want to test both fresh deployments & upgrades so I am not super duper concerned about this, especially because they can run in parallel

OK, I've included this in the document: 

```
All tests should be executed against both a freshly deployed and upgraded system.

**Alternatives considered:**

- Rather than duplicated tests, new tests could ensure that given the same configuration as inputs,
  a fresh chain has the same resulting configuration as an upgraded chain. Duplicating tests was
  deemed acceptable as they are parallelizable.
```

protocol/l1-upgrades.md
maurelian: I agree with @blmalone that we should not reverse the previous decision to enforce upgrade ordering off-chain only.

Keeping this open specifically to discuss @tynes comment further.

protocol/l1-upgrades.md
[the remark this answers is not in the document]
maurelian: For the purpose of this document, I'd summarize this as: 
1. significant work is required to make the dispute games MCP compatible, which we should not block this on. 
2. Therefore this work needs to either: 
 1. consider dispute game contracts as out of scope and needing to be upgraded via other means until they are MCP compatible.
 2. identify a workaround.
 
 
I prefer 2, using a blueprint based approach.

protocol/l1-upgrades.md
maurelian: I have added some code to the Release process section indicating a `blueprint` based deployment process for non-MCP compliant contracts.

mds1 (approved):
love this

PR #162, opened 2024-11-18
feat: propose DeputyPauseModule

smartcontracts:
Proposes the DeputyPauseModule to be installed into the Foundation Safe.

### DIFF
--- a/protocol/deputy-pause-module.md
+++ b/protocol/deputy-pause-module.md
@@ -1,0 +1,108 @@
+# Deputy Pause Module
+
+## Context
+
+The Foundation Safe is given the ability to act as the Security Council Safe for a limited number
+of safety-net actions on OP Mainnet. For instance, the Foundation Safe can act as the Security
+Council Safe to trigger the Superchain-wide pause function. These powers are given to the
+Foundation Safe through the `DeputyGuardianModule` installed in the Security Council Safe.
+
+## Problem Statement
+
+The Foundation Safe pre-signs certain transactions to reduce response time in case of an emergency.
+Because pre-signed Foundation Safe transactions are dependent on the nonces of that Safe, these
+pre-signed transactions become invalidated any time the Foundation Safe is required to sign
+literally any other transaction. In an attempt to reduce the overhead from this state of affairs,
+a *second* Foundation Safe sometimes referred to as the "Foundation Operations Safe" with the exact
+same configuration as the original Foundation Safe was made to be the Deputy Guardian.
+
+Even with this second Foundation Safe, pre-signed pauses are invalidated on a regular basis
+whenever an upgrade touches the `DeputyGuardianModule`. Gripes with the pre-signed pause system
+could fill a whole role of toilet paper and are not just limited to the issues noted above.
+
+## Proposed Solution
+
+We propose eliminating the pre-signed pause system entirely by introducing a `DeputyPauseModule` to
+be installed into the primary Foundation Safe. The `DeputyPauseModule` would allow a deputy account
+to act as the Foundation Safe to trigger the Superchain-wide pause functionality. The deputy
+account would not have any other capabilities.
+
+The `DeputyPauseModule` solves a litany of problems. Because the deputy account could be made to be
+an externally owned account, we gain the benefit of the pre-signed pause (immediate pause
+execution) without all of the many downsides of a pre-signed transaction. We could also remove the
+Foundation Operations Safe entirely, generally simplifying our multisig setup.
+
+### EOA vs Smart Contract
+
+We propose using an Externally Owned Account (EOA) instead of a smart contract as the deputy here.
+Using an EOA is simpler and easier to reason about. The private key for this EOA can be stored
+securely and made accessible to a limited set of security personel.
+
+### Single Account vs Mapping
+
+We propose having the `DeputyPauseModule` use a single account instead of a mapping of accounts
+that are able to act as the deputy. A single account is easier to keep track of and having multiple
+accounts does not decrease the risk involved with this module, it simply spreads it across more
+private keys. Having multiple keys be able to act as the deputy here might have some slighty
+benefits but this begins to scope creep beyond the original intention of replacing the pre-signed
+pause functionality.
+
+### Exact Proposed Modifications
+
+We currently have two Safe accounts for the Foundation:
+
+1. `0x847B5c174615B1B7fDF770882256e2D3E95b9D92` - known as the "Foundation Upgrade Safe" was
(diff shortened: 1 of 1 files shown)


protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,80 @@
...
+
+The Foundation Safe is given the ability to act as the Security Council Safe for a limited number
+of safety-net actions on OP Mainnet. For instance, the Foundation Safe can act as the Security
+Council Safe to trigger the Superchain-wide pause function. These powers are given to the
+Foundation Safe through the `DeputyGuardianModule` installed in the Security Council Safe.
+
+## Problem Statement
+
+The Foundation Safe pre-signs certain transactions to reduce response time in case of an emergency.
+Because pre-signed Safe transactions are dependent on the nonces of that Safe, these pre-signed
```
raffaele-oplabs, line 13: As we are talking about two different Safes, we should make it clear/disambiguate which safe we refer at the end of the sentence

Because pre-signed Safe transactions are dependent on the nonces of that **Safe**

raffaele-oplabs, line 13:
```suggestion
Because pre-signed Foundation Safe transactions are dependent on the nonces of that Foundation Safe, these pre-signed
```

raffaele-oplabs: Disambiguating which safe we refer to, as there are two safes.

protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,80 @@
...
+configuration as the original Foundation Safe was made to be the Deputy Guardian.
+
+Even with this second Foundation Safe, pre-signed pauses are invalidated on a regular basis
+whenever an upgrade touches the `DeputyGuardianModule`. Gripes with the pre-signed pause system
+could fill a whole role of toilet paper and are not just limited to the issues noted above.
+
+## Proposed Solution
+
+We propose eliminating the pre-signed pause system entirely by introducing a `DeputyPauseModule` to
+be installed into the primary Foundation Safe. The `DeputyPauseModule` would allow a deputy account
```
raffaele-oplabs, line 26: **"deputy account"** Shall we specify if this is EOA or Smart Contract?

raffaele-oplabs (approved):

protocol/deputy-pause-module.md
mds1: There's also two foundation safes :) 
```suggestion
Because pre-signed Foundation Operations Safe transactions are dependent on the nonces of the Foundation Operations Safe, these pre-signed
```

mds1: Ah this is mentioned in the next paragraph, I'm indifferent to how this thread is resolved

protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,80 @@
...
+to act as the Foundation Safe to trigger the Superchain-wide pause functionality. The deputy
+account would not have any other capabilities.
```
mds1, line 28: Let's get a bit more specific on the Safe changes. Currently we have:
- `0x847B5c174615B1B7fDF770882256e2D3E95b9D92`: Known as "Foundation Upgrades Safe", deployed on 2024-02-08, this is used as one owner on the 2/2 upgrader (L1 ProxyAdmin Owner) Safe with the security council.
- `0x9BA6e03D8B90dE867373Db8cF1A58d2F7F006b3A`: Known as "Foundation Operations Safe", deployed on 2021-01-16, this is used for presigned pauses, and is also one signer on Base's 2/2 Safe

Because 0x9B is older, I assume this is the primary foundation safe being referred to? So we'd install the module on this 0x9B safe, change the 2/2 with the security council to use this Safe, and deprecate the 0x84 safe? We should ask the data team to help find any usage of that safe that we might be forgetting (i.e. any contract with `847B5c174615B1B7fDF770882256e2D3E95b9D92`, or it's alias, somewhere in storage)

protocol/deputy-pause-module.md
mds1: Let's also specify if the module will allow exactly one deputy pauser, or a mapping of various allowed pausers

lucadonnoh (approved):
lgtm

protocol/deputy-pause-module.md
```diff
@@ -0,0 +1,108 @@
...
+
+Please note that step (2) requires action by the Security Council and governance communication. If
+going through the whole process of replacing one of the 2/2 signers is too much of a pain, we can
+also consider the following alternative:
+
+1. `DeputyPauseModule` will be installed into `0x9BA6`.
+1. `DeputyPauseModule` will be installed into `0x847B`.
+1. Security Council sets `0x847B` as the Deputy Guardian.
+1. Base replaces `0x9BA6` with `0x847B` in its 2/2.
+1. `0x9BA6` is deprecated generally.
```
mds1, line 75: The top flow feels cleaner to me. The big risk with either of these flows is changing one owner on the 2/2, as we can brick it if we transfer to the wrong owner. The options there are either:
1. A single `swapOwner` call. Risk is messing up the `swapOwner` call and bricking the safe
2. Add `0x9BA6` as an owner so it's a 2/3, ensure we can execute a tx, then remove the original `0x847B`. Downside is 3 transactions we need to execute, and there is a period where Foundation can unilaterally upgrade

mds1 (approved):

protocol/deputy-pause-module.md
smartcontracts: Tagged in Ben. Personally I feel like changing the Deputy Guardian is much safer than touching the 2/2 but open to be challenged on that.

mds1: In either scenario (step 2 of the top flow, or step 4 of the bottom alternate flow) we do have to change owners on a 2/2 and will have the same decision between `swapOwner` and a three step process where the Foundation temporarily has unilateral control. The difference is just which 2/2 it is

protocol/deputy-pause-module.md
maurelian, line 1: Some open questions:
1. how is the deputy address managed, can the deputy or some other account change it? 
2. Is the DPM upgradeable?

protocol/deputy-pause-module.md
smartcontracts: Yeah I guess you're correct on that point!

protocol/deputy-pause-module.md
smartcontracts: 1. Deputy address would probably be hard-coded into the `DeputyPauseModule` and could only be modified by having the Foundation Safe remove the module and replace it with another module.
2. No, I would propose that the `DeputyPauseModule` not be upgradeable and simply be replaced with a new module if we need to update it.

maurelian (approved):

PR #163, opened 2024-11-18
feat: added entrypoints.md

0xParticle:
**Description**

Continues https://github.com/ethereum-optimism/design-docs/pull/155

This design document introduces Entrypoint contracts as a new primitive that allows anyone to add custom logic on top of the `L2ToL2CrossDomainMessenger`. It generalizes the `L2ToL2CrossDomainMessenger `design and unlocks other interop primitives such as message batching and expiring. To do so, it adds a parameter in the `L2ToL2CrossDomainMessenger` that binds the relaying to a particular address, where custom logic can live.

### DIFF
--- a/protocol/messaging/entrypoints.md
+++ b/protocol/messaging/entrypoints.md
@@ -1,0 +1,156 @@
+## Summary
+
+This design document introduces `Entrypoint` contracts as a new primitive that allows anyone to add custom logic on top of the `L2ToL2CrossDomainMessenger`.
+It generalizes the `L2ToL2CrossDomainMessenger` design and unlocks other interop primitives such as message batching and expiring.
+To do so, the `L2ToL2CrossDomainMessenger` will allow to authorize a single address to relay a message. The authorized address can then add custom logic before executing the call.
+
+## Problem Statement + Context
+
+Building functionality on top of the `L2ToL2CrossDomainMessenger` is not simple. For example, to allow message recovery, the contract would need to track expired/failed messages to prevent someone from relaying them after recovery. This would require adding code on the `L2ToL2CrossDomainMessenger`. If another feature were required, we would need to add more code, which doesn’t scale.
+
+The `Entrypoint` contract primitive solves this issue and unlocks new powerful cross-chain actions. The integration with `Entrypoint` contracts is simple, and the code difference is small.
+
+## `Entrypoint` Contracts
+
+### Main idea
+
+Anyone can deploy `Entrypoint` contracts with any logic. Users can indicate, explicitly or through a contract, which `Entrypoint` contract can process their messages in the destination chain. `relayMessage` will check if the message has an `Entrypoint` associated, and if it does, it will validate that this contract is the `msg.sender`; otherwise, it will revert.
+
+The key modification is the `msg.sender` binding feature on the `relayMessage` call. If `msg.sender` has logic, the design is equivalent to binding `relayMessage` to include a pre-hook with that logic.
+
+Notice that post-hook design is already possible with the current design using logic on `target`.
+
+The single authorized address design can also work to permission relaying for an EOA. There is no check that the calling address is a contract or not.
+
+### Intuition Example
+
+This section will describe a simple example to showcase the power of `Entrypoint` contracts. Let’s consider the situation where Alice wants to send `AliceSuperToken` from Chain A to Chain B and then swap them.
+
+1. Alice (or anyone) deploys a `SwapperEntrypoint` as an `Entrypoint` contract in Chain B that can:
+    - Receive `AliceSuperToken` tokens.
+    - Swap for a target token.
+2. Alice (or anyone) deploys a `CrosschainSwapper` contract in chain A that will have a `sendAndSwap()` function.
+    - We assume the `CrosschainSwapper` is allowed to call `crosschainBurn()` on `AliceSuperToken`.
+3. Alice calls `sendAndSwap()` . The `CrosschainSwapper` burns tokens on behalf of Alice, and calls `sendMessage()` in `L2ToL2CrossDomainMessenger` with the `SwapperEntrypoint` in destination as `entrypoint`.
+4. `Alice` (or anyone) goes to the `SwapperEntrypoint` in Chain B, and calls `relaySendAndSwap()` with her message. The contract will a. Call `relayMessage()` in the `L2ToL2CrossDomainMessenger` and process her message, which relays the `AliceSuperToken` to the `SwapperEntrypoint` contract. b. Perform a swap in a Uniswap pool, for a target token. c. Optionally, the `SwapperEntrypoint` could bridge the target token back to Alice on chain A.
+
+The message cannot be relayed by any other `msg.sender` besides the `SwapperEntrypoint`, or it will revert. This is what we mean by binding.
+
+This will allow Alice to perform complex but safe cross-chain actions with a simple interaction in the source chain.
+
+## `entrypointContext`
+
+### Main idea
+
+Following the intuition example above, we could notice a problem in 4b: the `SwapperEntrypoint` doesn’t know what pool to use, how much of the total tokens given to it to swap, or who to give the tokens to after the swap. Information is missing. `SwapperEntrypoint` could ask for these values in the `swapTokens` function, but anyone could call it with Alice’s event and provide bad values.
+
+To solve this, `Entrypoint` contracts can define a decoding method and expect a second set of data that we will call `entrypointContext`. The context is binded into the message, meaning that its not possible to frontrun the `SentMessage` event with a different context.
+
+Entrypoints that wish to use `entrypointContext` will need to implement the `onRelayMessage` function. The `L2ToL2CrossDomainMessenger` will perform a callback to `onRelayMessage` inside the `relayMessage` call and before calling the target.  
+
+`Entrypoint` can use the context however they want. It’s important to notice the `relayMessage()` is non reentrant, so the callback cannot call it back. Some possible ways to use the context are the following:
+
+**Optimistic**
+
+Context is assumed to be true, but revert on callback if not. 
+
+1. In origin, Alice will encode the `entrypointContext` as a single `bytes` parameter into the `sendMessage` call. 
+2. On destination, the `Entrypoint` will ask for the context as a user input, in addition to the message and id. `Entrypoint` will assume this input context is true (optimistic).
+    1. `Entrypoint` will include an `onRelayMessage` callback.
+3. The `Entrypoint` will check that the context is valid.
+4. If so, it will call `relayMessage`, which will 
+    1. validate usual properties of the message (domain binding, non-replayability).
+    2. check `entrypoint == msg.sender`.
+    3. do a callback to the `Entrypoint` for `onRelayMessage`, passing the encoded `entrypointContext`. `onRelayMessage` will check that the context that was inputed matches the `entrypointContext`, and revert if not.
+    4. finally, call `target` with the `message`.
+
+**Store and use later**
+
+Some applications might not require the context at the time of the message (expire messages for instance). In that case, the `Entrypoint` can ignore the context parameter and just store it on the callback for later use.
+
+1. In origin, Alice will encode the `entrypointContext` as a single `bytes` parameter into the `sendMessage` call.
+2. On destination, the `Entrypoint` will ask for the message and id.
+    1. `Entrypoint` will include an `onRelayMessage` callback.
+3. `Entrypoint` will call `relayMessage`, which will
+    1. validate usual properties of the message (domain binding, non-replayability).
+    2. check `entrypoint == msg.sender`.
+    3. do a callback to the `Entrypoint` for `onRelayMessage`, passing the encoded `entrypointContext`. `onRelayMessage` the `Entrypoint` will store the context for later use.
+    4. finally, call `target` with the `message`.
+
+## Summary of required changes
+
+The following changes are required:
+
+- Overload `sendMessage` to
+    - include `address entrypoint` and `bytes entrypointContext`.
+    - modify `SentMessage` event to include `address entrypoint` and `bytes entrypointContext`.
+- Overload `relayMessage` to
+    - check if `msg.sender == entrypoint` and reverts if not.
+    - checks if `entrypointContext.length > 0` , and if so, it performs a callback to the `entrypoint` address with `entrypointContext` data.
+- `_decodeSentMessagePayload` will decode the `entrypoint` and `entrypointContext`.
+- `hashL2toL2CrossDomainMessage` will include `entrypoint` and `entrypointContext`.
+
+## Full example: Expire messages
+
+Here’s an example of how `Entrypoint` contracts with `entrypointContext` can enable powerful features like the expiration of messages:
+
+Suppose an `ExpirableTokenBridge` contract exists, that uses an `Entrypoint` with the added functionality of expiring a failed message in destination. `ExpirableTokenBridge` has minting and burning rights over `SuperchainERC20`.
+
+- Alice will transfer `AliceSuperToken` from Chain A to Chain B through the `ExpirableTokenBridge`, alongside some optional expiring parameters that will get encoded in the `entrypointContext` (these could be expire time window for instance).
+- If `AliceSuperToken` has not yet been deployed, relaying the message from the `Entrypoint` will fail. 
+    - If the message has `entrypointContext`, `Entrypoint` can store it for later use.
+- The user will then use the `Entrypoint` functionality to expire the message in chain B, making it non-processable by adding it to a mapping.  This will emit a message that the contract in Chain A can consume to handle the expired message. This call might optionally check the stored context as well for time windows or allowed senders.
+Any call to the `Entrypoint` to relay the expired message will revert. As the `Entrypoint` is the only contract that can process this message in the `L2ToL2CrossDomainMessenger`, the expired message is effectively non-processable in Chain B.
+
+The sequence diagrams will be divided in three:
+
+1. Shows the initiating interaction of the user with the `ExpirableTokenBridge` to send the message to Chain B.
+2. Shows the interaction of the user in Chain B with the `Entrypoint` to process the expirable message, and shows how it fails to be relayed, and how the `Entrypoint` expires it.
+3. Shows the interaction of the user back with the `ExpirableTokenBridge` in Chain A to recover the lost tokens.
+
+```mermaid
+sequenceDiagram
+
+title Message creation (Chain A)
+
+Alice->>ExpirableTokenBridge_A: sendERC20(to)
+ExpirableTokenBridge_A->>L2ToL2CrossDomainMessenger: sendMessage(..., entrypointAddr, entrypointContext)
+L2ToL2CrossDomainMessenger-->L2ToL2CrossDomainMessenger: emit SentMessage(...params, entrypointAddr, message, entrypointContext)
+
+```
+
+```mermaid
+sequenceDiagram
+
+title Message failed and expired (Chain B)
+
+Anyone->>Entrypoint: processMessage(Identifier messageId, Message message, Identifier contextId, Message context)
+Entrypoint->>L2ToL2CrossDomainMessenger: relayMessage(id, message)
+L2ToL2CrossDomainMessenger ->> Entrypoint: onRelayMessage(entrypointContext)
+L2ToL2CrossDomainMessenger->>ExpirableTokenBridge_B: relayERC2O(to, amount)
+ExpirableTokenBridge_B->>SuperchainERC20_B: FAILS
+Anyone->>Entrypoint: expireMessage(messageHash)
+Entrypoint-->Entrypoint: checks if the deadline of the message is due
+Entrypoint-->L2ToL2CrossDomainMessenger: checks if the message has been successfully relayed before
+Entrypoint-->Entrypoint: adds it to expiredMessages mapping to avoid future processing
+Entrypoint-->Entrypoint: emit ExpiredMessage(destination, messageHash, receiverOfExpiredMessage)
+
+```
+
+```mermaid
+sequenceDiagram
+
+title Expired message handling and funds recovery (Chain A)
+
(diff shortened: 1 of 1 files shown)


tynes:
Regarding expiring messages: https://github.com/ethereum-optimism/specs/issues/460

0xParticle:
> Regarding expiring messages: [ethereum-optimism/specs#460](https://github.com/ethereum-optimism/specs/issues/460)

Perfect. I feel like these are two separate types of problems. 
- The one presented in this PR is about messages that fail when `relayMessage` gets called. 
- The one you mention is for messages that can't even reach that stage. 
Both are important to address. We might find it convenient to use different names.

protocol/messaging/entrypoints.md
```diff
@@ -0,0 +1,195 @@
+## Summary
+
+This design document introduces `Entrypoint` contracts as a new primitive that allows anyone to add custom logic on top of the `L2ToL2CrossDomainMessenger`.
+It generalizes the `L2ToL2CrossDomainMessenger` design and unlocks other interop primitives such as message batching and expiring.
+To do so, the `L2ToL2CrossDomainMessenger` will allow to authorize a single address to relay a message. The authorized address can then add custom logic before executing the call.
```
mds1, line 5: From our call: let's clarify the text a bit to make it more clear this is a diff to the L2ToL2CrossDomainMessenger that lets you require the call be relayed by the `entrypoint` address

protocol/messaging/entrypoints.md
```diff
@@ -0,0 +1,195 @@
...
+
+Building functionality on top of the `L2ToL2CrossDomainMessenger` is not simple. For example, to allow message recovery, the contract would need to track expired/failed messages to prevent someone from relaying them after recovery. This would require adding code on the `L2ToL2CrossDomainMessenger`. If another feature were required, we would need to add more code, which doesn’t scale.
+
+The `Entrypoint` contract primitive solves this issue and unlocks new powerful cross-chain actions. The integration with `Entrypoint` contracts is simple, and the code difference is small.
+
+## `Entrypoint` Contracts
+
+### Main idea
+
+Anyone can deploy `Entrypoint` contracts with any logic. Users can indicate, explicitly or through a contract, which `Entrypoint` contract can process their messages in the destination chain. `relayMessage` will check if the message has an `Entrypoint` associated, and if it does, it will validate that this contract is the `msg.sender`; otherwise, it will revert.
```
mds1, line 17:
```suggestion
Anyone can deploy `Entrypoint` contracts with any logic. Users can indicate, explicitly or through a contract, which `Entrypoint` contract can process their messages in the destination chain. `relayMessage` will check if the message has an `Entrypoint` associated, and if it does, it will validate that this contract is the `msg.sender`; otherwise, it will revert. Any address can be an `Entrypoint`, including an EOA.
```

protocol/messaging/entrypoints.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Intuition Example
+
+This section will describe a simple example to showcase the power of `Entrypoint` contracts. Let’s consider the situation where Alice wants to send `AliceSuperToken` from Chain A to Chain B and then swap them.
+
+1. Alice (or anyone) deploys a `SwapperEntrypoint` as an `Entrypoint` contract in Chain B that can:
+    - Receive `AliceSuperToken` tokens.
+    - Swap for a target token.
+2. Alice (or anyone) deploys a `CrosschainSwapper` contract in chain A that will have a `sendAndSwap()` function.
+    - We assume the `CrosschainSwapper` is allowed to call `crosschainBurn()` on `AliceSuperToken`.
```
mds1, line 33: I thought with SuperchainERC20 we'd want only the `SUPERCHAIN_TOKEN_BRIDGE` to call `crosschainBurn()` on this token?

protocol/messaging/entrypoints.md
```diff
@@ -0,0 +1,195 @@
...
+- Identifier's `chainId` from `EntrypointContext` event matches `source` from `SentMessage`. Should match with the `SentMessage`.
+
+However, the most efficient way of binding both events is to check who emitted the `EntrypointContext` event against an allowed-list. This might seem restricting, but will be more than enough for many applications.
+
+### Summary of required changes
+
+The following changes are required:
+
+- `sendMessage` includes `address entrypoint`.
+- `SentMessage` event includes `address entrypoint`.
```
mds1, line 76: Let's check with @MSilb7 about which event params should be indexed

sergeyshemyakov:
Hi guys, I'm Sergey, research scientist at Lisk. I met Skeletor on Devcon and he directed me to these specs since I was very curious about making OP Interop more user friendly. 

I think this design is a step in the correct direction of making Interop more usable by actual applications. However I have a couple of more practical and some more abstract questions about this design. Maybe I will start with very particular ones:

1. I am curious how the swap example will work with the actual `SuperchianTokenBridge`? This goes in the direction of the question above by @mds1. Namely, which messages (if any) will be emitted by the bridge itself? Will the bridge smart contract have to be modified to know about the entrypoint? 
2. How could trustless incentivized relaying work with entrypoints? (I have message `m` and tokens `t` on chain A, I want `m` to be relayed to chain B for the fee of `t`) Would all relayers use the same entrypoint? How many transactions / messages will have to be sent on chain A and on chain B?

In general, I see that entrypoints enable all the nice usecases but I am not convinced that sending several messages that go through the entrypoint and also target the entrypoint is the simplest design. However maybe we cannot do much better because sending tokens and doing an action (swap) has to happen in two different cross-chain messages...

protocol/messaging/entrypoints.md
0xParticle: Just made a large refactor to the design doc. It should be more clear now, let me know.

protocol/messaging/entrypoints.md
0xParticle: You are correct. This example was wrong, so I changed in the latest commit.

protocol/messaging/entrypoints.md
mds1, line 1: We should make sure to add the "Alternatives Considered" and "Risks & Uncertainties" sections
- Do app layer interop products have a similar problem to solve, and what are their approaches? Why or why not would their designs work for us?
- What risks might this have, for example what if an entrypoint a user specifies is unable to complete the relay? Any other risks and uncertainties to consider?

sergeyshemyakov:
The updated version is much clearer, especially without the swap example. I think it was a bit confusing because ultimately this logic can be easier implemented with a `target` smart contract and not an entrypoint.

However now I feel that entrypoint design is skewed in the direction of expiring messages a bit too much. I think the discussion on the context that you had before was important, I have an example of ordered relays to illustrate it:

In order to implement message batching / guaranteed ordered execution you need the destination chain to know the message dependencies or to which batch does the message belong. A natural solution would be for the entrypoint to check this ordering condition, but the entrypoint needs to know some additional info except for the message itself (e.g. the list of msg hashes to be relayed before this message). So this context should be passed to the entrypoint somehow.

#### Alternative proposal
I have an idea of an alternative entrypoint approach, but for now I struggle to see whether it solves all the problems that your design solves. 

1. A cross-chain entrypoint is deployed both on chains A,B with the same code and the same address.
2. `CCEntrypoint` on chain A calls `sendMessage` with a wrapped call to the `CCEntrypoint` on the other side. Note that this message can contain both a context for `CCEntrypoint` on the other side and a wrapped calldata for some other contract on chain B (a bit like internet protocol layering). So any user cross-chain message A -> B that needs the entrypoint functionality must go via the entrypoint on chain A.
3. `L2ToL2CDM` on chain B is still the first stop for the message processing, as in the current interop design. `relayMessage` will unpack and make the corresponding call to `CCEntrypoint` on chain B.
4. `CCEntrypoint` on chain B can guarantee that the message originated from the other `CCEntrypoint` on chain A via `crossDomainMessageSender` (so some static validation can even be done only on chain A). It can also contain arbitrary validation and execution logic.

Pros of this design:
- No change to current interop protocol.
- Can pack entrypoint context into the cross chain message together with the original message to be executed.

Cons:
- Bridging transactions could not be routed via an endpoint since `SuperchainTokenBridge` directly uses `L2ToL2CDM` to pass bridging messages. However I think your design also requires some modifications to the bridging protocol.
- The cross-chain entrypoint will have to know and probably verify some details of cross chain messaging that are handled by `L2ToL2CDM`, e.g. how to execute a message. Introduces code duplication, somehow not very satisfying design.

Generally I am curious to hear what you think about this idea. I hope our exchange will make OP Interop as well-designed as possible!

0xParticle:
@sergeyshemyakov I've given your suggestion further thought, and here's my perspective:

While it's technically possible to achieve most of the use cases outlined in the design document by encoding additional information in an Initiator contract (or manually) on the origin chain and then sending messages to a dedicated Receiver contract on the destination chain for decoding and execution, this method significantly complicates the developer experience.

Advantages of Using `entrypoint`:
- Simplified Logic Delegation: Instead of embedding complex conditions into the message or the target contract, users can delegate this logic to the `entrypoint`. This separation of concerns keeps the message payload and target contract focused on their core functionalities, enhancing simplicity and composability.
- Improved Developer Experience: Developers don't need to handle intricate encoding and decoding processes or manage intermediate contracts that handle user funds and approvals.
- Composability: By offloading additional logic to the `entrypoint`, it's easier to compose and reuse components across different cross-chain interactions. With encodings, this becomes harder.

For message bundles, `entrypoints` offer a much better user flow. While it's possible to process bundles using `Receiver` contracts without `entrypoints`, this approach requires:
- Receiving each relayed message individually.
- Waiting for the entire bundle to be received and stored.
- Executing a final transaction to process all messages collectively (resulting in 𝑁+1 relay transactions instead of just one).

This not only adds complexity but also increases the potential for errors and delays.

## Example: two `SuperchainERC20` transfers
### With `entrypoint`
The `entrypoint` has a single function to relay that takes three messages as inputs: two messages corresponding to a crosschain transfer (a call to the `SuperchainTokenBridge`) and a Context event. The Context event encodes information linking the two transfers, preventing someone from passing a wrong pair. The `entrypoint` checks that the Context was emmited by an approved address in destination. 

This approach ensures the token transfers cannot be relayed independently. The Context could have been even emitted later on, as long as each transfer referenced the `entrypoint`.

### Without `entrypoint`
It is possible to achieve the same end result by using dedicated encodings and decodings in each chain, but the flow is quite complex and requires that these intermediate contracts handle user funds and approvals.
An example encoder contract in origin would look like this

```solidity
solidity
function sendTokens(
    address token1,
    address to1,
    uint256 amount1,
    uint256 chain1,
    address token2,
    address to2,
    uint256 amount2,
    uint256 chain2
) public {
    // Transfer token1 and 2 from msg.sender to this contract
    IERC20(token1).transferFrom(msg.sender, address(this), amount1),
    IERC20(token2).transferFrom(msg.sender, address(this), amount2),
    
    // Send token1 using SuperchainTokenBridge
    bytes32 msgHash1 = superchainTokenBridge.sendERC20(token1, receiverAddr, amount1, chain1);

    // Send token2 using SuperchainTokenBridge
    bytes32 msgHash2 = superchainTokenBridge.sendERC20(token2, receiverAddr, amount2, chain2);

    // Emit Context event with original msg.sender and msgHashes
    emit Context(msg.sender, token1, to1, amount1, token2, to2, amount2, msgHash1, msgHash2);
}
```

Then, in destination, each call can be relayed independently to the `receiverAddr`, that will need to store completed `msgHash` and have a dedicated function to do the final relay to the recipients. 

Notice how much complex this second flow is. Entrypoints are not unblocking new features, but they are making it way easier for devs and users.

protocol/messaging/entrypoints.md
0xParticle: I added both sections. 
- I searched for a similar design but couldn't find any. The difference with other third-party bridges here is the lack of a centralized relay that already orders messages. Most bridges focus on tokens only and not so much on messages, so custom logic checks, use of storage, and so on might be a bit of overkill for them. We are stepping into unexplored territory here. This has the potential to establish some standards as well. 
- Addressed this in the Risks section. It's mainly bad implementations of the entry point (gas, bugs, malicious) and compatibility with existing protocols (only if part of the flow uses the entry point and the other part doesn't).

You might find this comment interesting as well, as using the target directly can be considered an Alternative: https://github.com/ethereum-optimism/design-docs/pull/163#issuecomment-2501739374

sergeyshemyakov:
@0xParticle Thanks for having such a deep dive into my question, appreciate your comments!

So what I am taking out of this discussion:

1. Entrypoint design provides simpler composability and batching logic since you have control over how you call `relayMessage` on `L2ToL2CDM`. Thinking about my design with receiver contracts I realized that I imagine relayers somehow batching message execution anyway, so it is nice to provide a standard solution to this.
2. Entrypoints can interact with `SuperchainTokenBridge` much more easily, because they live in different design spaces. My idea of receiver contract lives on the same layer as the bridge in the sense that they both are built on top of `L2ToL2CDM` and are expected to be called directly from the messenger, thus making interactions complicated.
3. For the two reasons above, entrypoints are a superoir design.

Still I have two comments from my side:

1. I am still not completely convinced that the composability will be easily implementable for entrypoints. Imagine if you have an entrypoint that allows message batching and another entrypoint that makes sure to reward the relayer for passing a message, I don't see how you could easily get a solution to batch and incentivize without rewriting code. Although I see that entrypoints are reusable in different contexts (e.g. batching messages for DEX swaps or batching messages for voting in DAO).
2. The entrypoint context is paramount for almost all interesting applications, so I urge you to already embed it into this proposal by defining a context-passing event and a clear interface how to manage the context on the receiving chain.

#### Appendix: Further thoughts on my design

Now I am rethinking my receiver contract idea more in terms of cross-chain wallet, where a user "owns" a particular sender / receiver on several different chains and only they can move messages via the receiver (can be enforced e.g. by checking a signature on the sending side). Then it is possible to implement batching and relayer incentivization in the following way (which I have implemented [here](https://github.com/sergeyshemyakov/contracts/blob/master/CCBatcher/cross_chain_batcher/src/CrossChainWallet.sol)).

1. Receiver "cross-chain wallet" implements sequencing of messages, where every message can come with an array of `msgHash`es that must be successfully relayed before the message (as defined by `successfulMessages` on `L2ToL2CDM`).
2. User sends a transaction on chain A that a) transfers 100 USDC tokens to the "cross-chain wallet" receiver on chain B (`msg1`) b) sends a message to "cross-chain walllet" receiver on chain B to swap USDC tokens into OP once `msg1` is processed (`msg2`) c) sends a message to "cross-chain walllet" receiver on chain B to deploy all but 1 OP tokens into AAVE pool once `msg2` is processed (`msg3`) d) sends a message to "cross-chain walllet" receiver on chain B to pay 1 OP to `tx.origin` once `msg3` is processed. Note that `msgHash` can be computed immediately.
3. Relayer sees this sequence of messages and realizes that they can relay all 4 messages to chain B for a reward of 1 OP. They batch relaying of all messages because they have a financial incentive to do so (if they relay messages by 4 different transactions, someone can relay the final message and steal the whole reward). 

As an outcome, the user bridged tokens A->B, swapped, provided liquidity to staking protocol and paid the relayer all within one transaction. However it is important that user actually controls the "cross-chain wallet" receiver on chain B in the same way as they could control any other smart contract wallet.

0xParticle:
> @0xParticle Thanks for having such a deep dive into my question, appreciate your comments!
> 
> So what I am taking out of this discussion:
> 
> 1. Entrypoint design provides simpler composability and batching logic since you have control over how you call `relayMessage` on `L2ToL2CDM`. Thinking about my design with receiver contracts I realized that I imagine relayers somehow batching message execution anyway, so it is nice to provide a standard solution to this.
> 2. Entrypoints can interact with `SuperchainTokenBridge` much more easily, because they live in different design spaces. My idea of receiver contract lives on the same layer as the bridge in the sense that they both are built on top of `L2ToL2CDM` and are expected to be called directly from the messenger, thus making interactions complicated.
> 3. For the two reasons above, entrypoints are a superoir design.
> 
> Still I have two comments from my side:
> 
> 1. I am still not completely convinced that the composability will be easily implementable for entrypoints. Imagine if you have an entrypoint that allows message batching and another entrypoint that makes sure to reward the relayer for passing a message, I don't see how you could easily get a solution to batch and incentivize without rewriting code. Although I see that entrypoints are reusable in different contexts (e.g. batching messages for DEX swaps or batching messages for voting in DAO).
> 2. The entrypoint context is paramount for almost all interesting applications, so I urge you to already embed it into this proposal by defining a context-passing event and a clear interface how to manage the context on the receiving chain.
> 
> #### Appendix: Further thoughts on my design
> Now I am rethinking my receiver contract idea more in terms of cross-chain wallet, where a user "owns" a particular sender / receiver on several different chains and only they can move messages via the receiver (can be enforced e.g. by checking a signature on the sending side). Then it is possible to implement batching and relayer incentivization in the following way (which I have implemented [here](https://github.com/sergeyshemyakov/contracts/blob/master/CCBatcher/cross_chain_batcher/src/CrossChainWallet.sol)).
> 
> 1. Receiver "cross-chain wallet" implements sequencing of messages, where every message can come with an array of `msgHash`es that must be successfully relayed before the message (as defined by `successfulMessages` on `L2ToL2CDM`).
> 2. User sends a transaction on chain A that a) transfers 100 USDC tokens to the "cross-chain wallet" receiver on chain B (`msg1`) b) sends a message to "cross-chain walllet" receiver on chain B to swap USDC tokens into OP once `msg1` is processed (`msg2`) c) sends a message to "cross-chain walllet" receiver on chain B to deploy all but 1 OP tokens into AAVE pool once `msg2` is processed (`msg3`) d) sends a message to "cross-chain walllet" receiver on chain B to pay 1 OP to `tx.origin` once `msg3` is processed. Note that `msgHash` can be computed immediately.
> 3. Relayer sees this sequence of messages and realizes that they can relay all 4 messages to chain B for a reward of 1 OP. They batch relaying of all messages because they have a financial incentive to do so (if they relay messages by 4 different transactions, someone can relay the final message and steal the whole reward).
> 
> As an outcome, the user bridged tokens A->B, swapped, provided liquidity to staking protocol and paid the relayer all within one transaction. However it is important that user actually controls the "cross-chain wallet" receiver on chain B in the same way as they could control any other smart contract wallet.

## Entrypoint Composability
I agree that composability isn't straightforward in all cases.

In the scenario you mentioned—having an entrypoint that allows message batching and another that rewards relayers—I was thinking on using the Context event to encode a "chain of callers." Specifically, we can pass the address of the `RelayerFeeEntrypoint` as part of the Context to the `BatchingEntrypoint`. Then, the `BatchingEntrypoint` checks that `msg.sender` matches the decoded address of the `RelayerFeeEntrypoint`. This approach ensures that the only valid flow for relaying is:

`User → RelayerFeeEntrypoint → BatchingEntrypoint → L2ToL2CDM`

It would also be possible to encode it the other way around, as long as the check is in both entrypoints.

I believe this design is relatively simple and can be standardized—it's akin to a "meta-entrypoint" pattern. It allows us to stack entrypoints by having each one verify the caller based on the context, thus enabling composability without extensive code rewriting.
In contrast, using Initiator/Receiver patterns would require the decoding method to be aware of the execution order

## Context enshrinement
We also think that eventually, the context should be enshrined within the protocol. However, for now, we'll start with the entrypoint alone since it's upgradeable, and we want to see how these contracts are used in practice.

Enshrining the context is indeed the best way to securely bind a message to its context, but it requires the context to be known upfront at the source—it can't be emitted later or via a call to a contract that isn't context-aware. For that reason, we'll begin without enshrining the context and have the entrypoint trust a sender for the Context event.

We want to wait a bit and see how entrypoints and context are utilized before modifying the `L2ToL2CrossDomainMessenger` further.

## Message batching
Your approach is very clean and a clever use of the `successfulMessages` mapping. 
There are two invariants we had in mind for batching that might not be possible with this approach:
- Messages SHOULD NOT be aware that they are part of a batch when they're sent from the source chain. 
	- In particular, users SHOULD be able to create a batch from a Batcher contract that calls intermediate contracts, where each of them calls to `sendMessage` and is not aware that is being called by a Batcher (a call to `sendERC20` in the `SuperchainTokenBridge` for instance).
- It SHOULD be possible to execute an arbitrary subset of the batch (if some predefined conditions are met). 
	- As I see it, a batch is just a set of related transactions such that individual messages cannot be processed alone (without taking the other messages into account). That relationship can be used for enforced ordered execution as a particular case, but it could also happen that they have a parenthood relationship (batches inside batches) or another kind of dependancy. 

We've been discussing two possible designs internally that modify the `L2ToL2CrossDomainMessenger` to support these functionalities (still a work in progress). You can check it [here](https://gist.github.com/0xParticle/5119cfa271a6fba6ae8b7649e0bba5db). 

That said, if these two points aren't critical for your use case, your design is excellent, especially since it doesn't require any modifications to the `L2ToL2CrossDomainMessenger`. It provides an elegant solution within the existing framework.

We still need to gather feedback on the key invariants for batching. 

Also, feel free to reach out via Telegram @parti0x :)

gotzenx:
We continue our work on the [specs](https://github.com/ethereum-optimism/specs/pull/484) and the [implementation](https://github.com/ethereum-optimism/optimism/pull/13456) since we had green like from Matt [on discord](https://discord.com/channels/1244729134312198194/1303461365221949460/1311134840598565005) without having the formal approval from him at that time.

nambrot:
I'm a bit late to the party but upon first read, I share @sergeyshemyakov 's concerns around whether entrypoint composability is actually easily done or not. It also seems like message batching is a big priority/use case for `entrypoints`, but I think a definition of the problem would be useful. IIUC, I would maybe describe it more like "conditional relaying", and I'm actually not sure if that composes well with let's say strict ordering entrypoints that would be used by a "subbatcher".

@0xParticle I'd actually love to see the `entrypoint` implementation for the `two SuperchainERC20 transfers` example, I can't quite visualize it. Specifically, how would the entrypoint handle the case where one of the transfers fails?

tynes:
Regarding entrypoint composability, it is possible to create a single entrypoint that allows for composable hooks. I don't think that we need to arbitrarily compose entrypoints that were not designed to be composable, that sounds like it will add complexity/security risk

My main concern around entrypoints is their compatibility with a relayer, we would likely want to define some top level standard interface for an entrypoint so that relayers can just work by passing through data from an event on the source chain to the destination chain. Obviously an entrypoint doesn't need to follow this interface, but then it would be extra integration to just work with relayers

PR #166, opened 2024-11-21
feat: propose expanded challenger set

smartcontracts:
Proposes replacing the address of the CHALLENGER for OP Mainnet with a new contract that allows the Guardian or any member of the Security Council Safe to act as the CHALLENGER.

### DIFF
--- a/protocol/proofs/expanded-challenger.md
+++ b/protocol/proofs/expanded-challenger.md
@@ -1,0 +1,101 @@
+# Expanded Permissioned Challenger
+
+## Context
+
+The OP Stack currently officially provides two dispute game types, the standard `FaultDisputeGame`
+contract and the alternative `PermissionedDisputeGame`. The `PermissionedDisputeGame` is meant to
+act either as the default game for chains that, for whatever reason, do not want to run a
+permissionless game or as a fallback game type if bugs in the `FaultDisputeGame` are discovered.
+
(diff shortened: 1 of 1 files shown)


smartcontracts:
We had a long conversation about this on an impromptu design review meeting today, I'm going to think more on this to see if there are better simplifications we can make here.

protocol/proofs/expanded-challenger.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+We would additionally need to implement the proposal in
+[Design Doc 138](https://github.com/ethereum-optimism/design-docs/pull/138) to guarantee that the
+proposer cannot also play in the dispute game, which would potentially allow the proposer to
+exploit bugs in the dispute game if they exist. Given that the `PermissionedDisputeGame` fallback
+is primarily meant to be used in the case that a bug in the `FaultDisputeGame` is discovered, it
+is very realistic that such a bug would exist if the `PermissionedDisputeGame` is being used.
+
+## Alternatives Considered
```
smartcontracts, line 69: Alternative that we may propose is to replace the `DeputyGuardianModule` with a simplified module.

smartcontracts:
I've opened a different design doc as an alternative to this proposal. I believe this new design doc is much simpler and more generic: https://github.com/ethereum-optimism/design-docs/pull/167

PR #167, opened 2024-11-21
feat: propose DeputyGuardianModule improvements

smartcontracts:
Proposes a number of improvements to the DeputyGuardianModule. These improvements would supercede the previous proposal to introduce a DeputyPauseModule into the Foundation Safe and would simplify the OP Stack security model.


protocol/deputy-guardian-module-improvements.md
```diff
@@ -0,0 +1,125 @@
...
+
+The `DeputyGuardianModule` would be permitted to:
+
+1. Trigger the Superchain-wide pause.
+1. Blacklist an individual dispute game.
+
+The `DeputyGuardianModule` would **no longer** be permitted to:
+
+1. Undo the Superchain-wide pause ("unpause").
+1. Change the respected game type.
```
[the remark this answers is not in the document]
smartcontracts: Does your opinion here change if we add the `invalidateExistingGames()` function to the `AnchorStateRegistry` and allow the deputy guardian to call that function?

protocol/deputy-guardian-module-improvements.md
```diff
@@ -0,0 +1,125 @@
...
+1. Set the anchor state in the `AnchorStateRegistry`.
```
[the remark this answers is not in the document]
smartcontracts: For the sake of being able to roll this out before the `AnchorStateRegistry` changes (which may take a while), I would accept a temporary version of this where the deputy guardian is still able to do all of these actions (except for unpausing the system).

Once the `AnchorStateRegistry` changes are made, we'd roll out a version with more restrictions. Thoughts?

protocol/deputy-guardian-module-improvements.md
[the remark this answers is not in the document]
smartcontracts: I think I could be convinced of allowing the deputy guardian to switch the respected game type, need to think about it a bit more. My primary concern is that if we're making the deputy guardian more permissive then giving it the ability to change the respected game type becomes a bit more dangerous. A malicious/leaked deputy key would be able to swap the respected game type back to the permissionless game which would probably force you to trigger the pause to avoid an endless back-and-forth.

That said I think the likelihood of this happening is relatively low. My short-term proposal would then be that we make the deputy guardian more permissive and give it access to everything except for `unpause`.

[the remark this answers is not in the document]
smartcontracts: I was going to come up with a broader proposal but I think the reality is that with Stage 1.4 + interop it's possible that the landscape around security responses looks very different in 6 months, so I'd be happy with a simplified/permissive deputy guardian for now, even if it's not ideal.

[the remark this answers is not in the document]
smartcontracts: Yeah I'm also trying to simplify the security model at the same time. I think it's far too many nested permissions and different roles and it's a nightmare to keep track of. I'd really like to be able to keep it down to System Owner + Guardian + Deputy Guardian and nothing else.

> I'd be ok though with the EOA deputy guardian being able to do anything except unpause because it doesn't seem unreasonable to me to pause the superchain if a deputy guardian key was compromised and started being used by an attacker.

This was my thinking too. Compromised key is easy to detect (we can easily monitor for any transactions going through the `DeputyGuardianModule`). It's annoying but it mostly just messes with withdrawals temporarily. We're essentially favoring safety/simplicity over liveness.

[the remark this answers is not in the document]
smartcontracts: We could also potentially allow the deputy guardian module to switch the respected game type to permissioned but not allow it to switch back to permissionless.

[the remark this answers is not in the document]
ajsutton: That's actually a pretty good compromise.  When kona is deployed we would typically want to switch to it instead of permissioned if we can so need to think through if that should be allowed by the deputy guardian or if we're ok with requiring the security council for that.  e.g. we could allow the deputy guardian to change the respected game arbitrarily _unless it is already the permissioned game_.

protocol/deputy-guardian-module-improvements.md
```diff
@@ -0,0 +1,125 @@
...
+We propose resolving the above problems by modifying the existing `DeputyGuardianModule` to be more
+restrictive in the actions it can carry out while also being more permissive in who can carry out
+these actions.
+
+The `DeputyGuardianModule` would be permitted to:
+
+1. Trigger the Superchain-wide pause.
+1. Blacklist an individual dispute game.
+1. Change the respected game type to the `PermissionedDisputeGame`.
+1. Trigger the `setAnchorState` function in the `AnchorStateRegistry`.
```
maurelian, line 49: IMO allowing the Guardian to trigger the "fallback" to the permissioned game goes beyond liveness considerations and is what muddies the waters. Would prefer to have only the first two items in this list available.

protocol/deputy-guardian-module-improvements.md
```diff
@@ -0,0 +1,125 @@
...
+pre-signed transactions become invalidated any time the Foundation Safe is required to sign
+literally any other transaction. In an attempt to reduce the overhead from this state of affairs,
+a *second* Foundation Safe sometimes referred to as the "Foundation Operations Safe" with the exact
+same configuration as the original Foundation Safe was made to be the Deputy Guardian.
+
+Even with this second Foundation Safe, pre-signed pauses are invalidated on a regular basis
+whenever an upgrade touches the `DeputyGuardianModule`. Gripes with the pre-signed pause system
+could fill a whole role of toilet paper and are not just limited to the issues noted above.
+
+### Unclear Responsibilities
```
maurelian, line 29: I'm worried about changing what the DGM can do, but not also changing what that Guardian role is capable of doing. IMO it is much cleaner to maintain the property that "Anything the Guardian can do, the DGM can do", so I'd prefer to see us move any safety impacting actions (ie. unpausing, fallback activation) to the Upgrade Controller (2of2 Safe).

protocol/deputy-guardian-module-improvements.md
smartcontracts: I generally agree with you, this was my original proposal but see conversation with Adrian above for why I changed it

protocol/deputy-guardian-module-improvements.md
smartcontracts: Generally speaking, my goal with this proposal is to solve a lot of problems in the short-term and then come back to this problem in ~6 months when the landscape has changed a bit with Stage 1.4

maurelian: > This would violate the condition that L2Beat wants that the Security Council alone can unpause the system.

Actually, what I'm advocating or is to also update the auth in the system.

It should not be too much of a lift once the Isthmus contract changes are in, as they introduce a new `upgrader` role that is stored in the `SuperchainConfig`, so all you'd need to do is update the auth check in [`unpause()`](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L1/SuperchainConfig.sol#L95) and  [`setRespectedGameType()`](https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts-bedrock/src/L1/OptimismPortal2.sol#L668).

smartcontracts: Yes I agree with this in principle, my goal with this proposal is to be able to ship something into prod ASAP and then work on these more in-depth changes later

smartcontracts:
Sigh as much as I want to see this proposal be approved, I am sufficiently concerned that this allows a minority of the security council to brick everything that I will likely retract this proposal until we have a liveness guard in place on the security council.

protocol/deputy-guardian-module-improvements.md
```diff
@@ -0,0 +1,125 @@
...
+
+The `DeputyGuardianModule` would **no longer** be permitted to:
+
+1. Undo the Superchain-wide pause ("unpause").
+1. Change the respected game type to the `FaultDisputeGame`.
+
+The `DeputyGuardianModule` would be accessible to:
+
+1. Any member of the Security Council Safe.
+1. An EOA operated by the Optimism Foundation.
```
ben-chain, line 59: Can we make this scalable to `N` EOAs, so that ability to pause can be delegated to other security-facing orgs?

protocol/deputy-guardian-module-improvements.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Compromised Deputy
+
+A motivated attacker with access to a deputy's private key could try to drain the deputy's wallet
+constantly to prevent it from being used to trigger transactions. Although this could be
+side-stepped with private RPC/bundling tools, it would not be considered good practice to include
+live third-party infrastructure in the hot path for critical security actions. It therefore seems
+prudent that the deputy be able to act via signature instead of directly via `msg.sender` check.
+
+If we allow the deputy to act via signature then we must also make sure that the signature includes
```
ben-chain, line 109: How different is this from just using 1/1 SAFEs instead of raw EOA addresses?

smartcontracts: If 1/1 safe is compromised then owner can be changed and you lose access to pause. Using raw EOA makes this impossible.

maurelian:
> Sigh as much as I want to see this proposal be approved, I am sufficiently concerned that this allows a minority of the security council to brick everything that I will likely retract this proposal until we have a liveness guard in place on the security council.

@smartcontracts Is the design review meeting scheduled for today going to be looking at the design as is?

PR #170, opened 2024-11-25
interop: expired token transfers

0xParticle:
**Description**
This design document introduces a solution to recover stuck `SuperchainERC20` cross-chain transfers due to a failed relay. It builds on top of the `Entrypoint` vertical to minimize new logic on the bridge and optimize for upgradability. It introduces a method for expiring messages and sending them back for the original sender to recover its burn balance in origin.


tynes:
Given https://github.com/ethereum-optimism/specs/issues/460, this is not necessary and only adds complexity. The token transfer either fails local side or succeeds

PR #172, opened 2024-11-27
Upgrade test path

maurelian:


solidity/20241126-upgrade-path-testing.md
```diff
@@ -0,0 +1,121 @@
...
+    A["CommonTest.setUp()"] --> B["Setup.L1()"] -->  C["Upgrade.run()"]
+    C --> D["DeployImplementations.run()"]
+    C --> E["OPCM.upgrade()"]
+```
+
+A description of what new components would do is:
+
+- **`Upgrade.run()`:**
+  - Calls `op-deployer bootstrap superchain` to deploy new superchain contracts (`SuperchainConfig`
+    and `ProtocolVersions`), corresponding to the previous release.
```
maurelian, line 91: > corresponding to the previous release.

@mslipper This might be a gotcha I've miseed. ie. since `op-deployer` now only deploys a single release version, we'd need to use the previous release of `op-deployer`, not the one that exists on `develop`. Can we easily install and run a previous version?

mslipper: As long as a tagged release exists that can deploy all the prerequisite contracts, then yes.

solidity/20241126-upgrade-path-testing.md
```diff
@@ -0,0 +1,119 @@
...
+graph LR
+    A["CommonTest.setUp()"] --> B["Setup.L1()"] -->  C["Upgrade.run()"]
+    C --> D["DeployImplementations.run()"]
+    C --> E["OPCM.upgrade()"]
+```
+
+A description of what new components would do is:
+
+- **`Upgrade.run()`:**
+  - Calls `op-deployer bootstrap superchain` to deploy new superchain contracts (`SuperchainConfig`
```
mds1, line 88: Is executing this command analogous to running both `DeploySuperchain.s.sol` and `DeployImplementations.s.sol`, but only having those scripts deploy contracts that changed from the last release? Don't need to get into implementation here, but one idea I've had for doing this is have those scripts use create2, precompute deploy addresses, and skip when code already exists. This requires a one-time redeploy of all contracts using this method, which we have to do for isthmus anyway since we're bumping to 0.8.25

mds1, line 88: Does this introduce a circular dependency issue at all?

solidity/20241126-upgrade-path-testing.md
```diff
@@ -0,0 +1,119 @@
...
+    and `ProtocolVersions`), corresponding to the previous release.
+  - Calls `op-deployer bootstrap opcm` to deploy release OPCM, corresponding to the previous release.
```
mds1, line 90: Should this be merged with `op-deployer bootstrap superchain`? It seems like any time we want to deploy new superchain implementation contracts we'd also want to deploy a corresponding new OPCM

solidity/20241126-upgrade-path-testing.md
```diff
@@ -0,0 +1,119 @@
...
+  - Upgrades proxies to new implementation contracts and bespoke singleton contracts as necessary for a new OP Chain.
+  - This flow is descibed in detail in the [L1 Upgrades design](../protocol/l1-upgrades.md#release-process).
+
+This testing setup route would be indicated with a new `useUpgradedSystem` flag in `CommonTest`. The
+new flag could only be enabled when other flags (`useAltDAOverride`, `useLegacyContracts`,
+`useInteropOverride`, `customGasToken`) are disabled.
+
+Note that this testing would need to be run against an `anvil` node, as `op-deployer bootstrap`
+requires an RPC endpoint. I am not experienced with running `forge test` against an anvil node,
+so appreciate any gotchas I might be missing.
```
mds1, line 104: Am I correct that the flow here would be for our forge tests to invoke `op-deployer` against an anvil RPC URL, then after op-deployer runs it would `vm.createSelectFork(anvilRpcUrl)`?

maurelian: > invoke `op-deployer` against an anvil RPC URL 

Yes. 

> then after op-deployer runs it would `vm.createSelectFork(anvilRpcUrl)`?

I had envisioned calling `forge test --fork-url anvilRpcUrl`. Creating the fork in the script seems less error prone so I like this, are there any other differences I should be aware of between those two approaches?

solidity/20241126-upgrade-path-testing.md
maurelian: I don't have a strong opinion here, maybe @mslipper does?

For the purposes of this design, I'm happy to make two `op-deployer bootstrap` calls or just the one if we do combine then.

solidity/20241126-upgrade-path-testing.md
maurelian: Can you say more about what you think that might be?

mds1: Ah, and `op-deployer bootstrap opcm` deploys the new implementations before deploying OPCM?

mds1: Originally OPCM and it's solidity tests were self-contained, and op-deployer wrapped it. Now, OPCM requires op-deployer for it's own testing. I am not familiar enough with the architecture to know if there is actually a problem here, and I don't have a better solution that feels less circular

solidity/20241126-upgrade-path-testing.md
mds1: No strong opinion either, happy to defer

solidity/20241126-upgrade-path-testing.md
mds1: > I had envisioned calling `forge test --fork-url anvilRpcUrl`. Creating the fork in the script seems less error prone so I like this

I am not certain but it's possible the `forge test --fork-url anvilRpcUrl` approach might not work because forge may always make requests at whatever the initial block is at fork time, i.e. I don't know if it follows the chain live, if that makes sense. So starting our fork from within the test/script itself after anvil is properly configured seems safer anyway

I don't think there should be any other differences. There are some `createSelectFork` things to be aware of if you switch between forks in the tests, like `vm.makePersistent`

solidity/20241126-upgrade-path-testing.md
maurelian: hmmm, I think that this is OK because `op-deployer` is only being used to setup a system from a previous release, and then the testing will act directly on the OPCM and updated implementations which are on `develop`

PR #173, opened 2024-11-29
feat: Add OPCM upgrades FMA

maurelian:
FMA associated with [OPCM Upgrades](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/l1-upgrades.md).

### DIFF
--- a/security/fma-l1-upgrades.md
+++ b/security/fma-l1-upgrades.md
@@ -26,7 +26,7 @@ _Italics are used to indicate things that need to be replaced._
 | Created at         | 2024-03-26 |
 | Initial Reviewers  | @blmalone  |
 | Need Approval From | [TBD]      |
-| Status             | Draft      |
+| Status             | In Review  |
 
 ## Introduction
 
--- a/security/fma-l1-upgrades.md
+++ b/security/fma-l1-upgrades.md
@@ -1,0 +1,305 @@
+# [Project Name]: Failure Modes and Recovery Path Analysis
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Introduction](#introduction)
+- [Failure Modes and Recovery Paths](#failure-modes-and-recovery-paths)
+  - [[Name of Failure Mode 1]](#name-of-failure-mode-1)
+  - [[Name of Failure Mode 2]](#name-of-failure-mode-2)
+- [Audit Requirements](#audit-requirements)
+- [Action Items](#action-items)
+- [Appendix](#appendix)
+  - [Appendix A: This is a Placeholder Title](#appendix-a-this-is-a-placeholder-title)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+_Italics are used to indicate things that need to be replaced._
+
+|                    |            |
+| ------------------ | ---------- |
+| Author             | Maurelian  |
+| Created at         | 2024-03-26 |
+| Initial Reviewers  | [TBD]      |
+| Need Approval From | [TBD]      |
+| Status             | Draft      |
+
+## Introduction
+
+## Introduction
+
+This document covers the L1 Upgrades project, which aims to improve and standardize the approach to upgrading L1 contracts across the superchain. The project introduces a new upgrade system that consolidates ProxyAdmin contracts, implements a two-step upgrade flow, and provides a more deterministic and safer upgrade process.
+
+Below are references for this project:
+
+- [L1 Upgrades Design Doc](../protocol/l1-upgrades.md)
+
+### FM1: Contract Initializers
+
+#### **Description:**
+
+During the first step of the ugprade, the contracts will not yet have a shared
+storage location for the `initialized` value. Therefore each contract will require custom storage
+manipulation to properly reset the `initialized` value, without modifying values which may be
+packed into the same slot.
+
+Moreoever, contracts could be re-initialized if the initialization state is not properly managed
+during the two-step upgrade process.
+
+#### **Risk Assessment:**
+
+High impact, Medium likelihood
+
+#### **Mitigations:**
+
+1. In the first step to reset the initializer with varied layouts, a custom contract can be
+   created which hardcodes the exact modifications required for each contract with a unique
+   contracts specific function, such as `resetL1CrossDomainMessenger()`.
+2. Upgrading to use OpenZeppelin's unstructured storage for initialization state reduces the risk
+   of incorrectly reinitializing at the end of the upgrade, as it ensures that the `initializer`
+   modifier can operate on the same slot across contracts.
+3. After calling `OPCM.upgrade()` the `superchain-ops` script can use `vm.load()` and attempt to call `L1CrossDomainMessenger.upgrade()` (for example) to verify the intitialized status.
+4. Tenderly simulation should look for unexpected storage changes.
+
+#### **Detection:**
+
+Global event monitoring COULD be used on `monitorism` for the `Initialized` event, howevrer in the past similar monitoring
+has been deemed to noisy and not useful.
+
+#### **Recovery Path(s)**:
+
+An emergency upgrade would be required.
+
+### FM2: Storage Layout Corruption in an OP Chain Contract
+
+#### **Description:**
+
+During an upgrade, incorrect storage layout modifications could lead to
+corruption of contract state. This could happen if the upgrade process doesn't
+properly account for existing storage slots or if there are conflicts in the
+storage layout between versions.
+
+We disregard issues with the initializer as they are covered in FM1.
+
+#### **Risk Assessment:**
+
+High impact, Low likelihood
+
+#### **Mitigations:**
+
+Existing mitigations for this failure mode include:
+
+1. The use of spacer variables in L1 contracts to prevent storage collisions.
+2. The use of storage layout lock files tracked in git to detect breaking changes.
+3. The use of off-chain checks in the `superchain-ops` scripts to ensure that getters
+   continue to return the expected values.
+4. The use of state change validation prior to execution.
+
+Additionally, the [proposal](../protocol/l1-upgrades.md#modfications-to-l1-contracts) of adding a new minimally invasive
+`upgrade()` method to L1 contracts ensures minimal storage write operations during the upgrade process.
+
+#### **Detection:**
+
+Unexpected storage layout changes which are not caught by the mitigations above will be very difficult to detect, unless they
+immediately impact on the operation of the system. For example changing the value of `L1CrossDomainMessenger.otherMessenger()`
+will break both the messenger's `relayMessage()` and `sendMessage()` functions. Other values, such as an entry in a mapping or
+a rarely used address will be harder to detect.
+
+Regardless, strong emphasis should be placed on mitigation here.
+
+#### **Recovery Path(s)**:
+
+Emergency upgrade to fix the affected storage slots.
+
+### FM3: Storage Layout Corruption in the Upgrade Controller Safe
+
+#### **Description:**
+
+The decision was made to have the Upgrade Controller Safe DELEGATECALL
+to `OPCM.upgrade()`, as this reduces the risk needing to transfer ownership to the OPCM.
+
+However this creates the risk of modifying storage in the Upgrade Controller Safe.
+
+#### **Risk Assessment:**
+
+High impact, Low likelihood
+
+#### **Mitigations:**
+
+Similar to FM2, we have existing mitigation patterns in place which we can employ here:
+
+1. The use of off-chain checks in the `superchain-ops` scripts to ensure that getters
+   continue to return the expected values.
+2. The use of state change validation prior to execution.
+
+We can also add static analysis to ensure that no `SSTORE` operations occur
+within `OPCM.upgrade()`. Slither would likely be a good approach here, or
+another tool which operates on the IR.
+
+#### **Detection:**
+
+As with FM2, unless something breaks immediately, this will be very hard to
+detect. Strong emphasis should be placed on mitigation.
+
+#### **Recovery Path(s)**:
+
+It may be impossible to recover from this, depending on which storage changes have been made.
+
+Perhaps now is the time to consider adding the Recover Module.
+
+### FM4: Failure to follow upgrade path
+
+#### **Description:**
+
+For an OP Chain upgrading though multiple releases, there is a risk that the correct upgrade path is not followed.
+
+For example, skipping from v2.0.0 to v2.2.0 without going through v2.1.0.
+
+```mermaid
+graph LR
+    A[2.0.0] --> B[2.2.0]
+
+    %% Invisible elements to help center 2.1.0
+    A ~~~C[2.1.0]
+    C ~~~B
+```
+
+This is a particular risk we have chosen to accept as opposed to enforcing the upgrade path onchain,
+because doing so comes with operational complications.
+
+#### **Risk Assessment:**
+
+Medium impact, Medium likelihood
+
+#### **Mitigations:**
+
+1. Define the correct upgrade path in the `superchain-registry`.
+2. Enforce the correct upgrade path in `op-deployer` which is the official interface for upgrading OP Chains.
+
+#### **Detection:**
+
+**Montoring:** We could consider introducing monitoring which:
+
+1. Reads OPCM addresses from the `superchain-registry`
+2. Monitors those addresses for `Upgraded` events
+3. Detects when the version is not correctly incremented.
+
+However, such monitoring would likely be difficult to maintain, would only focus on the 'official' superchain, and is unlikely to be triggered as effective mitigations are not difficult to implement.
+
+#### **Recovery Path(s):**
+
+The recovery path is highly dependent on the nature of the failure resulting from an invalid upgrade path, which is very difficult to predict.
+
+### FM5: Patches result in a complex upgrade path
+
+#### **Description:**
+
+Note: this item is somewhat meta as it describes a complexity arising from a generic failure mode, but
+it felt like it was worth documenting here.
+
+Consider the following sequence of events involving two chains (`OPChainA` and `OPChainB`):
+
+- Both chains are deployed at `v2.0.0`
+- `OPChainA` upgrades to `v2.1.0`
+- `OPChainB` (for whatever reason) is not ready to upgrade and remains on `v2.0.0`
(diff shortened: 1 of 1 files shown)


blmalone (commented):
I'm working my way through this, got to FM1. Will continue soon.

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,306 @@
...
+- [OPCM spec](https://github.com/ethereum-optimism/specs/blob/opcm/upgrades/specs/experimental/op-contracts-manager.md)
+
+### FM1: Contract Initializers
+
+#### **Description:**
+
+During the first step of the ugprade, the contracts will not yet have a shared
+storage location for the `initialized` value. Therefore each contract will require custom storage
+manipulation to properly reset the `initialized` value, without modifying values which may be
+packed into the same slot.
```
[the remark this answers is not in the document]
blmalone: I was looking into this more and realized a few more things. Here's my summary below:

1. This will be a one time operation whereby we will move all contracts that are initializable to the new OpenZepplin Initializable [v5.1.0 contract](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.1.0/contracts/proxy/utils/Initializable.sol).
2. Some contracts use older versions of [Initializable.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.4/contracts/proxy/utils/Initializable.sol) which do not use unstructured storage. We want all contracts to use the unstructured storage because it's better to have a uniform way to reset initializer values across the entire system. 
3. When moving to the new  [v5.1.0 contract](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.1.0/contracts/proxy/utils/Initializable.sol), we need to be mindful about storage slot collisions. It'll be necessary to use spacers in the implementation contracts to ensure storage collisions are avoided.

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+#### **Mitigations:**
+
+Similar to FM2, we have existing mitigation patterns in place which we can employ here:
+
+1. The use of off-chain checks in the `superchain-ops` scripts to ensure that getters
+   continue to return the expected values.
+2. The use of state change validation prior to execution.
+
+We can also add static analysis to ensure that no `SSTORE` operations occur
+within `OPCM.upgrade()`. Slither would likely be a good approach here, or
```
blmalone, line 137: If we're still using the superchain-ops/tenderly approach. Tenderly allows you to see which opcodes were used as part of a transaction. Checking there as a sanity check would be something to add to the VALIDATIONS.md file.

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+graph LR
+    A[2.0.0] --> B[2.2.0]
+
+    %% Invisible elements to help center 2.1.0
+    A ~~~C[2.1.0]
+    C ~~~B
+```
+
+This is a particular risk we have chosen to accept as opposed to enforcing the upgrade path onchain,
+because doing so comes with operational complications.
```
blmalone, line 169:
```suggestion
because doing so comes with operational [complications](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/op-contracts-manager-single-release-redesign.md#managing-patches-for-opcm-deployments).
```

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+
+#### **Risk Assessment:**
+
+Medium impact, Medium likelihood
+
+#### **Mitigations:**
+
+1. Define the correct upgrade path in the `superchain-registry`.
+2. Enforce the correct upgrade path in `op-deployer` which is the official interface for upgrading OP Chains.
```
blmalone, line 178: I wonder if it would be useful to have a solidity script in superchain-ops that helps with double checking the OPCM address is correct for the current version by using a library like: https://github.com/memester-xyz/surl

blmalone, line 178: For my understanding, is the high level steps here something like:

1. op-deployer creates transaction data for signers (akin to `input.json`)
2. define superchain-ops task to allow signers to validate the upgrade
3. execute the upgrade as normal via superchain-ops

blmalone: In addition the the op-deployer check, this would add some defense in depth.

security/fma-l1-upgrades.md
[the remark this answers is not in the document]
maurelian: > For my own understanding, what's an example of a contract that we use that isn't using unstructured storage for the initialized value?

All of them. :)

ie. they are using [this](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v4.7.3/contracts/proxy/utils/Initializable.sol#L62) version of `Initializable`, which just puts `_initialized` in the [first slot](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v4.7.3/contracts/proxy/utils/Initializable.sol#L62).

The proposal is to use [this version](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/fa525310e45f91eb20a6d3baa2644be8e0adba31/contracts/proxy/utils/Initializable.sol#L77).

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+
+Existing mitigations for this failure mode include:
+
+1. The use of spacer variables in L1 contracts to prevent storage collisions.
+2. The use of storage layout lock files tracked in git to detect breaking changes.
+3. The use of off-chain checks in the `superchain-ops` scripts to ensure that getters
+   continue to return the expected values.
+4. The use of state change validation prior to execution.
+
+Additionally, the [proposal](../protocol/l1-upgrades.md#modfications-to-l1-contracts) of adding a new minimally invasive
```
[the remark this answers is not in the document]
maurelian: The spec describes it [here](https://github.com/ethereum-optimism/specs/blob/34e18e4de0d27362e13f735a0bb3df3d79c8fd80/specs/experimental/op-contracts-manager.md#L210).

Basically, the function will appear on all contracts, and look one of two ways: 

1. If nothing is added to the contract's `initializer()`, then it's empty: 
```solidity
function upgrade() external initializer{};
```

2. If a new arg/state var is being added to `initialize()` then `upgrade()` will set just that value: 
```solidity
function upgrade(SomeType _newValue) external initializer{
    newValue = _newValue;
    // or if there are some checks on newValue, then both initialize() and upgrade()
    // should use an internal setter function, ie.
    _setNewValue(_newValue);
};
```

security/fma-l1-upgrades.md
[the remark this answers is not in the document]
blmalone: Awesome, thanks for the clarification. Going to resolve this.

security/fma-l1-upgrades.md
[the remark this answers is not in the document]
blmalone: Gotcha thanks. Resolving.

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+
+#### **Recovery Path(s)**:
+
+Emergency upgrade to fix the affected storage slots.
+
+### FM3: Storage Layout Corruption in the Upgrade Controller Safe
+
+#### **Description:**
+
+The decision was made to have the Upgrade Controller Safe DELEGATECALL
```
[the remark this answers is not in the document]
maurelian: An ugly but effective approach would be to record the most important Safe storage values at the start of the call then reset them afterwards

It looks like all of the OPCM state variables are in structs, which unfortunately solidity does not allow to be made `immutable`.

> I wonder if we enforced all state variables in opcm to be held at unstructured storage locations. This would probably be super ugly though.

[EIP-7201](https://eips.ethereum.org/EIPS/eip-7201) gives a much nicer syntax for accessing structs with unstructured storage. If we used that pattern, then it might be easier to detect when storage is being accessed. 

Though I do think that static analysis of the IR can be pretty robust.

security/fma-l1-upgrades.md
maurelian: that's a good idea. I think we'll likely still use tenderly, just have it be much simpler.

security/fma-l1-upgrades.md
maurelian: Interesting idea. One of the main open gaps is to clarify how superchain-ops fits into the flow and where the line is between it and op-deployer.

maurelian: Yes, that's the high level.

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,329 @@
...
+This scenario would require two additional PATCH OPCMs, and an additional OCPM for upgrading from the earlier patched version to the newer patched version (ie. `v2.0.1` to `v2.1.1`), for a total of 5 OPCMs:
+
+1. `OPCM-v2.1.0`
+2. `OPCM-PATCH-v2.0.1`
+3. `OPCM-v2.1.1`
+4. `OPCM-PATCH-v2.1.1`
+5. `OPCM-v2.2.0`
+
+This is not necessarily dangerous, but is likely to complicate operations in various ways.
+to complicate operations in various ways.
```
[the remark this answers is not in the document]
maurelian:
```suggestion
```

security/fma-l1-upgrades.md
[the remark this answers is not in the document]
maurelian: oh, I didn't see your initial reply to self when I sent mine. 

> When moving to the new v5.1.0 contract, we need to be mindful about storage slot collisions. It'll be necessary to use spacers in the implementation contracts to ensure storage collisions are avoided.

Yes, FWIW I have recently dealt with this pretty extensively in the [Isthmus PR](https://github.com/ethereum-optimism/optimism/pull/12932/files#diff-7d99abb08e3cb939c50026beb0d42dc3b72c3e3b5c29e254ebd1e63c552e25bf) that was not included.

security/fma-l1-upgrades.md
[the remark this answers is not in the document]
blmalone: Make sense, resolving comment.

security/fma-l1-upgrades.md
blmalone: Gotcha, you mean slither right?

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,328 @@
...
+
+|                    |            |
+| ------------------ | ---------- |
+| Author             | Maurelian  |
+| Created at         | 2024-03-26 |
+| Initial Reviewers  | @blmalone      |
+| Need Approval From | [TBD]      |
+| Status             | Draft      |
+
+## Introduction
```
blmalone, line 27:
```suggestion
```

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,328 @@
...
+
+#### **Mitigations:**
+
+1. In the first step to reset the initializer with varied layouts, a custom, one-time use, contract can be
+   created which hardcodes the exact modifications required for each contract with a unique
+   contracts specific function, such as `resetL1CrossDomainMessenger()`.
+2. Upgrading to use OpenZeppelin's unstructured storage for initialization state reduces the risk
+   of incorrectly reinitializing at the end of the upgrade, as it ensures that the `initializer`
+   modifier can operate on the same slot across contracts.
+3. After calling `OPCM.upgrade()` the `superchain-ops` script can use `vm.load()` and attempt to call `L1CrossDomainMessenger.upgrade()` (for example) to verify the intitialized status.
```
blmalone, line 62:
```suggestion
3. After calling `OPCM.upgrade()` the `superchain-ops` script can use `vm.load()` and attempt to call `L1CrossDomainMessenger.upgrade()` (for example) to verify the initialized status.
```

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,328 @@
...
+Medium impact, Medium likelihood
+
+#### **Mitigations:**
+
+1. Define the correct upgrade path in the `superchain-registry`.
+2. Enforce the correct upgrade path in `op-deployer` which is the official interface for upgrading OP Chains.
+
+#### **Detection:**
+
+**Montoring:** We could consider introducing monitoring which:
```
blmalone, line 182:
```suggestion
**Monitoring:** We could consider introducing monitoring which:
```

security/fma-l1-upgrades.md
blmalone: Resolving.

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,328 @@
...
+#### **Mitigations:**
+
+This is not directly mitigable aside from ongoing efforts to avoid releasing buggy code.
+
+It also underscores the need for a machine readable format which can serve to describe multiple valid
+paths between two versions. That can be done during development however.
+
+#### **Detection:**
+
+This is detected whenever and however it is detected.
```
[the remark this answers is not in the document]
maurelian: Exactly. I think this might be too fuzzy of a failure mode and it's already covered by "FM4: Failure to follow upgrade path". How do you feel about removing it?

security/fma-l1-upgrades.md
maurelian: We'll use slither yes. 
But we should also use Tenderly in the tx submission process. My point is just that validating the state diff should be much simpler now. 
Although maybe that's not exactly true with the increase in the number of chains.

Still I think we should continue using Tenderly for at least the first upgrade to have extra assurance about it, and minimize the diff in process.

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,306 @@
...
+
+High impact, Low likelihood
+
+#### **Mitigations:**
+
+Existing mitigations for this failure mode include:
+
+1. The use of spacer variables in L1 contracts to prevent storage collisions.
+2. The use of storage layout lock files tracked in git to detect breaking changes.
+3. The use of off-chain checks in the `superchain-ops` scripts to ensure that getters
```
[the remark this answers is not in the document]
maurelian: > Are you talking about a check that looks something like [this](https://github.com/ethereum-optimism/superchain-ops/blob/main/tasks/eth/020-set-gas-target/SignFromJson.s.sol?rgh-link-date=2024-12-04T16%3A10%3A21Z#L36)?

Yes.

> Am I right in saying we will continue to use the same superchain-ops process once we upgrade to opcm?

I think this is an open question, and one that the upgrades team will need to make together. I know there are significant changes to the ops repo planned, so that will have an impact here.

security/fma-l1-upgrades.md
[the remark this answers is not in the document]
blmalone: I think it's fine to remove. Like you said, it's covered in FM4.

security/fma-l1-upgrades.md
blmalone: Resolving.

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,328 @@
...
+
+As the number of OP Chains grows, upgrades might exceed block gas limits, causing failed upgrades.
+
+#### **Risk Assessment:**
+
+Low impact, Medium likelihood
+
+#### **Mitigations:**
+
+Gas benchmarks should be created to determine the full cost of upgradings all OP Chains. This will require
```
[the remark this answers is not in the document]
blmalone: thanks resolving.

security/fma-l1-upgrades.md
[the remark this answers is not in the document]
maurelian: 6d5736eaf971b68cfc507e98ba55bad6c5ff2755

security/fma-l1-upgrades.md
blmalone: Still feels like an open question here. Going to resolve as I don't think this is a blocker for the FMA, more of a design decision.

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,274 @@
...
+
+This will be detected when an upgrade fails due to OOG. That's fine.
+
+#### **Recovery Path(s):**
+
+The upgrade will need to be gas optimized or rearchitected to break it up. We can cross that bridge when we get to it.
+
+#### Action items:
+
+- [ ] Create gas benchmarks for upgrading all OP Chains.
```
blmalone, line 233: Curious how you'd go about doing this? Have you a simple example to help me understand?

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,274 @@
...
+
+_Italics are used to indicate things that need to be replaced._
+
+|                    |            |
+| ------------------ | ---------- |
+| Author             | Maurelian  |
+| Created at         | 2024-03-26 |
+| Initial Reviewers  | @blmalone  |
+| Need Approval From | [TBD]      |
+| Status             | Draft      |
```
blmalone, line 29: Not sure if this should be 'In Review' or not.

blmalone (commented):
Did my second pass of this pr. Not approving just yet as I've left some question that will help me better understand some sections. 

Out of all the failure modes 'FM4: Failure to follow upgrade path' jumps out to me as the most likely to hit (mainly because humans are heavily involved). Making sure that the `superchain-registry` is properly updated is super important here. I think we may need to define a process around this to make sure it _always_ contains the correct list of OPCM addresses.

security/fma-l1-upgrades.md
```diff
@@ -0,0 +1,274 @@
...
+has been deemed too noisy and not useful.
+
+#### **Recovery Path(s)**:
+
+An emergency upgrade would be required.
+
+#### Action items:
+
+- [ ] Expose the upgraded state of the contracts being upgraded.
+- [ ] Add a check to the OPCM to verify the upgraded state of the contracts being upgraded.
```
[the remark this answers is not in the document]
maurelian: the current upgrade doesn't give us much that we can make assertions on, at least not without going into the nitty-gritty details (ie. checking all the getters, etc). 

We are not yet able to check init'd status on-chain. 
We _could_ check that semvers are set as expected?

At the least we should create an internal `_postCheck()` function in the OPCM which prompts developers to define whatever checks are appropriate while creating an upgrade.

security/fma-l1-upgrades.md
maurelian: We have `test_upgrade_succeeds` in the Monorepo, which currently has a cost of about 360k gas. We could record the gas costs of running it and revert if it gets beyond some value.

So, we know its pretty far off, but we can create an issue for future work to automatically simulate an upgrade of standard chains.

maurelian: Captured by this test: https://github.com/ethereum-optimism/optimism/issues/13835

blmalone (approved):
looks good!

PR #175, opened 2024-12-04
Cannot use pure keyword for OPCM upgrade function

blmalone:
**Description**

Clarifying that we cannot use the `pure` keyword for the OPCM upgrade function. Additionally, fixed some small nits.


maurelian (approved):
LGTM TYVM!

PR #190, opened 2025-01-27
protocol: test-sequencing service

protolambda:
**Description**

This design-doc introduces the idea of building a test service for sequencing, to help cover the many tx-inclusion and reorg cases in the interop test plan. Later, we can potentially build it out into a production sequencer, standalone from the op-node.

### DIFF
--- a/protocol/test-sequencing.md
+++ b/protocol/test-sequencing.md
@@ -40,9 +40,11 @@ For testing and experiments, the service should build blocks manually, with full
 For regular operation, and rbuilder-backed flashblocks,
 we can make the service fall back to only run the block-proposing part of sequencing: see "engine mode" section.
 
-## Increments
+## Milestones
 
-### Testing scope
+Note: milestones 1-4 are nice-to-have extensions. Milestone 0 has priority.
+
+### Milestone 0: Testing scope
 
 Test-functionality: Build the block without the engine API.
 
@@ -71,7 +73,7 @@ Building the testing functionality consists of:
 - Add a RPC method to op-node to insert the block into the canonical chain, and publish it to gossip, with a sequencer signature.
   Related problem: the existing op-node `admin_postUnsafePayload` (used by op-conductor) method doesn't
   include the block signature, so it cannot relay the block via gossip to other nodes.
-- Host new endpoints:
+- Host new endpoints (for usage in Kurtosis / e2e tests):
     - `test_open(opts: {l2Parent, l1Origin}) -> jobid`: start building a block
     - `test_includeBest(jobid)`: include best next tx from the tx-pool
     - `test_includeFirst(jobid, address)`: include best next tx from specific address from tx-pool
@@ -86,7 +88,17 @@ These RPC calls can then be used by tests to include the transactions in the des
 to create the right test scenario. E.g. coordinating a cyclic transaction dependency,
 or including an otherwise invalid transaction (invalid executing message).
 
-#### L1 mode
+Milestone 0 sub-tasks:
+- 0.0: core service definition
+- 0.1: basic empty block building, with signing, and op-node endpoints to apply block.
+- 0.2: implement essential RPC endpoints (`test_open`, `test_includeRaw`, `test_cancel`, `test_seal`)
+- 0.3: attach tx-pool
+- 0.4: implement remaining RPC endpoints
+
+Note: an RPC interface is chosen over a CLI tool, to make it more readily available in a devnet setting,
+and have a single source of truth and collected logs of block-building (test) actions.
+
+#### Milestone 1: L1 mode
 
 L1-block building shares a lot of code with L2, and can be done for testing as well.
 No system-transactions or formatted extra attributes like the Holocene `extraData` need to be inserted.
@@ -117,7 +129,7 @@ There is an `engine auto` command for automatic block-building of L1 blocks (no 
 on an interval using the engine-API.
 And it provides none of the testing or production sequencer functionality.
 
-### Production
+### Milestone 2: Production
 
 What if we want to make the test sequencer a production sequencer?
 
@@ -137,7 +149,7 @@ The following integration work is needed:
 - Support the DA-throttling RPC.
   Or better, implement proper multi-block DA congestion safe-guards.
 
-### Engine mode
+### Milestone 3: Engine mode
 
 To not only build blocks locally, we can make the service connect to a regular engine-API endpoint,
 and use that to build the blocks.
@@ -153,7 +165,7 @@ The service essentially becomes an interface for the op-conductor setup to work 
 [Flashblocks utilizes a sidecar design](https://github.com/ethereum-optimism/design-docs/pull/86),
 called rollup-boost, that enables flash-blocks to be flexibly attached between the services.
 
-### Interop experimentation
+### Milestone 4: Interop experimentation
 
 With some communication between the build processes, we can coordinate transaction-inclusion across chains.
 E.g.:
--- a/protocol/test-sequencing.md
+++ b/protocol/test-sequencing.md
@@ -68,7 +68,7 @@ Building the testing functionality consists of:
 - Connect to geth node, via debug RPC, for remote state-access:
     - We can seal the block with a `stateRoot` this way.
 - Reuse code from op-node to connect to op-signer, to sign blocks.
-- Add some RPC method to op-node to publish payloads to gossip, with a sequencer signature.
+- Add a RPC method to op-node to insert the block into the canonical chain, and publish it to gossip, with a sequencer signature.
   Related problem: the existing op-node `admin_postUnsafePayload` (used by op-conductor) method doesn't
   include the block signature, so it cannot relay the block via gossip to other nodes.
 - Host new endpoints:
@@ -98,11 +98,14 @@ Another way might be to intercept the engine-API `engine_forkchoiceUpdated(fcSta
 which can then queue the block-building work and complete it when there is a matching test block-building job.
 This is similar to the sidecar design in OP-Stack.
 However, this then makes the op-node lead, and prevents reorgs from being initiated by the test tool.
+The "mock" L1 mode will thus be prioritized.
 
 #### Compared to op-wheel
 
-`op-wheel` may be time to deprecate.
-It consists of `db` (patch geth DB contents) and `engine` utils.
+`op-wheel` may be time to deprecate. Full deprecation is out of scope for this design doc,
+but decision is made to not add additional things to `op-wheel`.
+
+The `op-wheel` CLI consists of `db` (patch geth DB contents) and `engine` utils.
 
 The forkchoice / reorg CLI tools are still used,
 but may be possible to move to op-node or op-geth.
@@ -138,6 +141,9 @@ The following integration work is needed:
 
 To not only build blocks locally, we can make the service connect to a regular engine-API endpoint,
 and use that to build the blocks.
+
+To keep the op-node in sync, forkchoice updates will need to be directed at a new op-node RPC,
+to keep the op-node forkchoice in sync, while passing through the calls to the actual engine API.
 
 This won't support the test capabilities, but does enable migration away from op-node sequencing.
 
@@ -192,12 +198,19 @@ The `op-wheel` service set up as a CLI tool, not as a service.
 `op-wheel` has a range of functions that can be deprecated.
 Expanding the tool now takes on debt of the tool, and makes it more difficult to deprecate parts.
 
-## Extending op-node
+## Extending op-node sequencer internals
 
 The op-node only has access to the building of the initial block template,
 and the start/stop timing of block building jobs.
 It does not have tx-pool access, ability to order transactions, or EVM or state access.
 The sequencer code-path in op-node is a critical-path, and must not break due to test changes.
+
+## Direct engine-API vs op-node RPC
+
+The service needs to connect to the op-node to apply forkchoice updates,
+since it otherwise can conflict with the verifier work.
+
+This means that the "engine mode" cannot use the engine-API directly, if there is a live op-node.
 
 ## As sidecar between op-node and op-geth
 
--- a/protocol/test-sequencing.md
+++ b/protocol/test-sequencing.md
@@ -1,0 +1,216 @@
+# Summary
+
+Interop testing requires completion of a long
+[checklist](https://github.com/ethereum-optimism/optimism/issues/13855) of particularly constructed blocks and reorgs.
+To implement these tests, we need new tooling that can include transactions and perform reorgs in exact ways.
+
+While implementing the test tooling, we have an opportunity to prepare separation of sequencing from the op-node,
+to reduce op-node complexity and encapsulate sequencing concerns better.
+
+# Problem Statement + Context
+
+Sequencing in the op-stack can be bucketed into different types:
+- Legacy sequencing (tx-pool, op-node with op-geth, recommit, priority fee)
+- Flashblocks (rollup-boost proxy between op-node and op-geth, mux to builder node, 250ms incremental block building)
+- Testing (action-tests, next-up interop work)
+
+The testing type of work is in need of change, to provide the level of block-building control
+that [action-tests](https://github.com/ethereum-optimism/optimism/tree/develop/op-e2e/actions)
+have to fully end-to-end tests like those backed by
+[Kurtosis](https://github.com/ethereum-optimism/optimism/tree/develop/kurtosis-devnet).
+
+Instead of building a limited testing tool, we can consider building a full sequencer service:
+- Op-node complexity can be reduced.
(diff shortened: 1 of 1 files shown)


protocol/test-sequencing.md
```diff
@@ -0,0 +1,216 @@
...
+**This includes rollup-boost**: without modifications to rollup-boost, we can migrate sequencing away from the op-node.
+[Flashblocks utilizes a sidecar design](https://github.com/ethereum-optimism/design-docs/pull/86),
+called rollup-boost, that enables flash-blocks to be flexibly attached between the services.
+
+### Interop experimentation
+
+With some communication between the build processes, we can coordinate transaction-inclusion across chains.
+E.g.:
+- Host multiple builders in the same Go process.
+- Connect multiple instances with streams.
```
axelKingsley, line 155: this would be huge ❤️

protocol/test-sequencing.md
```diff
@@ -0,0 +1,216 @@
...
+allowing sequencer-rotation to work the same as it does today.
+
+# Alternatives Considered
+
+## Extending op-geth
+
+The block-building described here conflict with that within the `miner` package of geth:
+- the `miner` package block-building recommits blocks.
+- the `miner` package pulls in transactions from the pool eagerly.
+- the `miner` package is a difficult diff to upstream geth, something we do not want to expand.
```
axelKingsley, line 187: +1, extending `op-geth` is not viable for new block building policies if we want it to be extensible and also minimal diff.

protocol/test-sequencing.md
```diff
@@ -0,0 +1,216 @@
...
+The sequencer code-path in op-node is a critical-path, and must not break due to test changes.
+
+## As sidecar between op-node and op-geth
+
+Inserting this as sidecar between op-node and op-geth is possible, but adds extra complexity, especially w.r.t. reorgs.
+See exploration in the "L1 mode" section about this.
+
+# Risks & Uncertainties
+
+Slow remote-state access can limit the usefulness in non-test settings.
```
axelKingsley, line 209: Since this doc focuses on testing, I think this is a lesser concern.

axelKingsley (commented):
Generally very supportive of this proposal. My primary feedback is:

- This doc is focused on test usecases, but also talks about production and other use cases. While it's good to demonstrate this work can extend beyond just tests, since the goal is to enable robust interop testing, I don't want to prioritize our decisions against these alternative use cases.

- In ares of this document where multiple choices are provided, let's just choose one formally in this document. I called this out in the `L1 Mode` section but there could be other spots too.

protocol/test-sequencing.md
```diff
@@ -0,0 +1,216 @@
...
+- Flashblocks (rollup-boost proxy between op-node and op-geth, mux to builder node, 250ms incremental block building)
+- Testing (action-tests, next-up interop work)
+
+The testing type of work is in need of change, to provide the level of block-building control
+that [action-tests](https://github.com/ethereum-optimism/optimism/tree/develop/op-e2e/actions)
+have to fully end-to-end tests like those backed by
+[Kurtosis](https://github.com/ethereum-optimism/optimism/tree/develop/kurtosis-devnet).
+
+Instead of building a limited testing tool, we can consider building a full sequencer service:
+- Op-node complexity can be reduced.
```
tynes, line 23: what is the difference between this tool and what `rbuilder` does?

protolambda: - ability to reorg, important for a lot of the reorg tests.
- ability to bypass validation rules when/if desired, e.g. for the invalid-message test cases.
- ability to build blocks when desired, rather than on the start/stop command of a real sequencer.
- ability to synchronously confirm txs with RPC calls, rather than submitting them to a tx-pool.

protocol/test-sequencing.md
```diff
@@ -0,0 +1,216 @@
...
+- Host new endpoints:
+    - `test_open(opts: {l2Parent, l1Origin}) -> jobid`: start building a block
+    - `test_includeBest(jobid)`: include best next tx from the tx-pool
+    - `test_includeFirst(jobid, address)`: include best next tx from specific address from tx-pool
+    - `test_includeByHash(jobid, txHash)`: include specific tx from tx-pool
+    - `test_includeRaw(jobid, rawTx)`: include specific given tx
+    - `test_cancel(jobid)`: stop building a block
+    - `test_seal(jobid)`: complete building a block
+    - `test_auto(active)`: automatically repeat block-building
+    - `test_inspect(jobid) -> partial block`: inspect thus-far built block
```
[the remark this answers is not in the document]
ajsutton: In general I'd agree with this and prefer to keep the API minimal then have the test code be explicit (likely aided by good helper methods which we'd need around all of these RPCs anyway).

For testing the block building code itself we may want to have a method to just fill the block automatically but then it may just be better to have a single RPC that builds a single block in one call with no need to separately open and seal the block.

I'm not particularly strong on that view, but I do find it a bit unclear with our current tests when they just "include next" transactions and always worry it may not wind up being the transaction I think it should be.

protocol/test-sequencing.md
```diff
@@ -0,0 +1,216 @@
...
+- Host the same sequencer admin RPC as op-node, to hande sequencer rotation:
+    - `admin_startSequencer`: start
+    - `admin_stopSequencer`: stop
+    - `admin_sequencerActive`: sequencer state
+    - `admin_postUnsafePayload`: publish a block, for failover
+    - `admin_overrideLeader`: used by op-conductor
+    - `admin_conductorEnabled`: used by op-conductor
+- Connect to geth P2P, to read/write the shared tx-pool.
+- Support the DA-throttling RPC.
+  Or better, implement proper multi-block DA congestion safe-guards.
```
ajsutton, line 135: Building and testing this out for production is a pretty significant scope - particularly the need to fully support op-conductor. I suspect we should focus on testing needs now - our test code should still be high quality and anything complex (like this) should have tests so we know our tests are doing what we expect anyway.  Then we can separately look at product priorities and determine if we should build a separate block builder service for production (and exactly what it should support).

That said, would probably be good to get product input on what block building improvements are being considered/asked for etc so we're at least aware. (cc @elihaims ?)

protocol/test-sequencing.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Engine mode
+
+To not only build blocks locally, we can make the service connect to a regular engine-API endpoint,
+and use that to build the blocks.
```
ajsutton, line 140: Wouldn't this either mean it has to derive the chain so it's EL is up to date and ready to build, or if it's sharing an engine API with an op-node we'd get a heap of potentially conflicting `engine_forkChoiceUpdated` calls? It might work out but it feels very complex to reason about.

protocol/test-sequencing.md
```diff
@@ -0,0 +1,216 @@
...
+
+## As sidecar between op-node and op-geth
+
+Inserting this as sidecar between op-node and op-geth is possible, but adds extra complexity, especially w.r.t. reorgs.
+See exploration in the "L1 mode" section about this.
+
+# Risks & Uncertainties
+
+Slow remote-state access can limit the usefulness in non-test settings.
+When the node is close, this may not be a problem.
```
ajsutton, line 210: Any extra latency in world state access is a real killer for performance. I _think_ the besu team has finally worked out something that's viable to provide shared state storage for Infura but it took years of research and quite possibly a database redesign.

I'd suggest we should find a way to test if remote state is a show stopper for performance before doing any extra work because we think this service will be useful for production.

protocol/test-sequencing.md
```diff
@@ -0,0 +1,216 @@
...
+By encapsulating the sequencer work, we can monitor and scale it independent of regular verifier work.
+
+Long-term this can open up more scalable block-building for interop.
+
+## Single Point of Failure and Multi Client Considerations
+
+If we decide to productionize this service after test-tooling, we can support the same set of admin RPCs,
+allowing sequencer-rotation to work the same as it does today.
+
+# Alternatives Considered
```
ajsutton, line 180: I think the missing alternative here is to do the simplest possible thing that enables our testing, which is to simply have the test code produce a block and post it to op-node (ie all op-node instances in the test set up are verifiers only and the sequencer p2p key is owned by the test).  The test would just use synchronous block building which we already do in a few unit tests to build the blocks. Since it's the only sequencer it would inherently have built the whole chain history so no need to run any kind of sync to follow the chain.  We'd still want to build utility methods to provide a really clean API for that block building but it wouldn't need to be a service, wouldn't need RPC etc - just the absolute minimal code to build a block and extend a chain.  We could either just use the existing `admin_postUnsafePayload` method and call it on all nodes individually or add a new RPC that accepts the block signature and also publishes the block to gossip.

The advantage of this is that it actually gives us the maximal flexibility as we can literally build any block we want and it's as simple as possible - wouldn't need any kind of tx pool even.

Admittedly we would have to write a wrapper if we wanted to use this in kurtosis, but I suspect the assumption that our wrapper is the only block builder would still be reasonable there and thus could still use the same basic approach.  And I think the wrapper could be extremely minimal  - potentially even just a CLI tool that takes a list of transactions and builds a new block that contains them then sends it off.

ajsutton (commented):
I like the general idea of separating block production but I'm concerned that we're expanding our scope here a fair bit because we _might_ be able to leverage the code in production later. I can see a lot of benefits to this in production, but I'm not sure if it's necessary and I don't think we've understood even the initial requirements and validated them yet so would be really concerned about increasing scope here.

I think if we build the test helper code well to provide a nice clean API for block building we could start out by just building the block in process and then if there is product demand for a separate block building service later simply replace the local back end with remote calls to the most complex/fully-featured service then.

protocol/test-sequencing.md
sigma: I find the idea intriguing.
Can you clarify the "wrapper for kurtosis" part? I'm not sure I understand what is kurtosis-specific here. Even with a test running outside of kurtosis, couldn't we handle an external sequencer?

protocol/test-sequencing.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Increments
+
+### Testing scope
+
+Test-functionality: Build the block without the engine API.
+
+The test-functionality is inspired a lot by the action-tests:
+having exact control over block-construction is very useful for testing.
+And by putting it in a service, we can use it more places than action-tests, like Kurtosis / devnets.
```
sigma, line 51: I understand there are other use-cases that justify the service approach, but for test-only cases, couldn't we also get away with just a library providing the block building capability?
I might be missing something

protocol/test-sequencing.md
ajsutton: Yeah it's not really kurtosis specific, just that's the tool we're using for more standalone devnet spin ups.  I just wouldn't really expect us to spin up a long lived devnet and use this testing block builder - we could, but typically by the time you're running a long lived devnet with nodes setup in k8s etc you probably want it to run like a real network and have a normal honest sequencer. The test sequencer wouldn't actually be doing derivation from L1 so it will get out of sync if the sequencer window ever elapses or the batcher does something stupid and causes a reorg - still useful for testing, not well suited for long lived devnets. Not handling those cases dramatically reduces the scope and complexity of the custom block builder though.

protocol/test-sequencing.md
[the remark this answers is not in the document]
protolambda: I'll prioritize the `includeByHash` and `includeRaw`. The other ones I think can still be useful

protocol/test-sequencing.md
protolambda: +1, agreed on priorities. After delivering the test scope however, I do strongly believe we need to revisit sequencing for production interop deployments, as part of this tool or not, given all the requirements and DoS risks that we discuss during testing calls.

protocol/test-sequencing.md
protolambda: You are right about the engine-API forkchoice RPC call conflicts. I think it will have to update the forkchoice via op-node, instead of the engine API. I will update the doc to reflect this.

protocol/test-sequencing.md
protolambda: Agreed, it's listed as risk for a reason. I think we can learn a lot from the test-scope first though.

protocol/test-sequencing.md
protolambda: I like the idea of posting the block to op-node. I think it combines with the forkchoice RPC conflict solution as well. Add an op-node RPC for block insertion, and make the test service just use that RPC to deliver the block.

protocol/test-sequencing.md
protolambda: A library can produce the block, but then the block still needs to be inserted into the canonical chain, including the act of reorging.

protocol/test-sequencing.md
```diff
@@ -0,0 +1,216 @@
...
+This can support shadow-forking (since L1 consensus is not followed), and avoid the need to run a beacon-node at all.
+
+Another way might be to intercept the engine-API `engine_forkchoiceUpdated(fcState, attributes)`,
+which can then queue the block-building work and complete it when there is a matching test block-building job.
+This is similar to the sidecar design in OP-Stack.
+However, this then makes the op-node lead, and prevents reorgs from being initiated by the test tool.
+
+#### Compared to op-wheel
+
+`op-wheel` may be time to deprecate.
```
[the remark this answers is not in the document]
protolambda: Yes, updated text to clarify

tynes:
Providing some milestones that break it down incrementally could be helpful in making sure there is consistent value provided and there are no long periods of time rushing to provide value

protocol/test-sequencing.md
```diff
@@ -0,0 +1,229 @@
...
+To keep the op-node in sync, forkchoice updates will need to be directed at a new op-node RPC,
+to keep the op-node forkchoice in sync, while passing through the calls to the actual engine API.
+
+This won't support the test capabilities, but does enable migration away from op-node sequencing.
+
+The service essentially becomes an interface for the op-conductor setup to work with any EL builder setup.
+
+**This includes rollup-boost**: without modifications to rollup-boost, we can migrate sequencing away from the op-node.
+[Flashblocks utilizes a sidecar design](https://github.com/ethereum-optimism/design-docs/pull/86),
+called rollup-boost, that enables flash-blocks to be flexibly attached between the services.
```
tynes, line 154: If you designed things from first principles, would it make sense to no longer run rollup boost as an intermediate proxy? I think its fine for now to keep things the way they are but concerned about tech debt coming from inability to coordinate an integration point

protocol/test-sequencing.md
```diff
@@ -0,0 +1,229 @@
...
+The "mock" L1 mode will thus be prioritized.
+
+#### Compared to op-wheel
+
+`op-wheel` may be time to deprecate. Full deprecation is out of scope for this design doc,
+but decision is made to not add additional things to `op-wheel`.
+
+The `op-wheel` CLI consists of `db` (patch geth DB contents) and `engine` utils.
+
+The forkchoice / reorg CLI tools are still used,
```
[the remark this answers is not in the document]
protolambda: I think making it a sub-command of op-node, and maybe making it accessible via admin-API, would be more useful. Since op-node is already distributed to the users who may need to access the forkchoice/reorg tools. And RPC commands can be a lot easier than invoking op-node with all the right parameters for engine API and such.

protocol/test-sequencing.md
```diff
@@ -0,0 +1,229 @@
...
+    - `eth_sendRawTransaction`: for tx flow. Op-geth nodes could configure this as sequencer-endpoint RPC to relay txs to.
+    - `eth_chainId`: maybe, for chain check.
+- Import the geth tx-pool. The tx-pool journal and pool maintenance code is already quite decoupled in geth.
+- Connect to geth node, via debug RPC, for remote state-access:
+    - We can seal the block with a `stateRoot` this way.
+- Reuse code from op-node to connect to op-signer, to sign blocks.
+- Add a RPC method to op-node to insert the block into the canonical chain, and publish it to gossip, with a sequencer signature.
+  Related problem: the existing op-node `admin_postUnsafePayload` (used by op-conductor) method doesn't
+  include the block signature, so it cannot relay the block via gossip to other nodes.
+- Host new endpoints:
```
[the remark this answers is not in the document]
protolambda: Added a note about that :+1:

protocol/test-sequencing.md
protolambda: +1 There are likely better ways to run rollup-boost on top of this if/when this reaches the production milestone, but I think for now that's out of scope. We can learn from prior milestones, and then adapt to rollup-boost/builder needs based on the learnings.

ajsutton (approved):
As mentioned in discord, I still think it makes more sense to build this out incrementally and just start with an in-process API used by action tests, then wrap that for kurtosis but the general idea makes a lot of sense to give us control over block building so happy to go ahead with this and leave implementation details to the person doing the work.

tynes (approved):

PR #180, opened 2024-12-20
feat: Add FMA Shared Lockbox

Joxess:
**Description**

This PR introduces the FMA for the Shared Lockbox design.

**Additional context**

Below are the references for this project:

- [Design doc](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/eth-shared-lockbox.md).
- [Specs PR](https://github.com/ethereum-optimism/specs/pull/465).
- [Implementation PR](https://github.com/ethereum-optimism/optimism/pull/13144).

PR #193, opened 2025-01-28
feat: Add new FMA Shared Lockbox

Joxess:
**Description**

This PR introduces the FMA for the new Shared Lockbox design.

**Additional context**

Below are the references for this project:

- [Specs PR](https://github.com/ethereum-optimism/specs/pull/465).
- [Implementation PR](https://github.com/ethereum-optimism/optimism/pull/13144)

It replaces the previous FMA, which can be found [here](https://github.com/ethereum-optimism/design-docs/pull/180) for reference.

### DIFF
--- a/security/fma-shared-lockbox.md
+++ b/security/fma-shared-lockbox.md
@@ -1,0 +1,147 @@
+# Shared Lockbox: Failure Modes and Recovery Path Analysis
+
+| Author | Joxes, Agusduha |
+| --- | --- |
+| Created at | 2025-01-28 |
+| Initial Reviewers | Pending |
+| Needs Approval From | Pending |
+| Status | In Review |
+
+## Introduction
+
+This document covers the changes introduced by the addition of the Shared Lockbox design, a singleton contract that stores all ETH liquidity given a set of interoperable chains. It addresses the [ETH withdrawals problem](https://github.com/ethereum-optimism/specs/issues/362) by the introduction of [SuperchainWETH](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/superchain-weth.md). The following components are:
+
+- **Contracts**:
+    - Introducing `SharedLockbox`: Stores all ETH liquidity for an interoperable graph. Only authorized `OptimismPortal` addresses can call `lockETH` and `unlockETH`.
+    - Introducing `DependencyManager`: A predeploy proxy and implementation contract. Manages the local chain’s dependency set. Triggers an L2 → L1 withdrawal to request adding a new chain to the dependency set on L1 in `SuperchainConfigInterop`.
+    - Introducing `SuperchainConfigInterop`: A proxy and implementation contract in L1. Processes add-dependency requests, track the official “mesh” of authorized chains and their portals, and authorizes each chain’s `OptimismPortal` in the `SharedLockbox`.
+    - Updates to the `OptimismPortal2`: Contain the `lockETH` and `unlockETH` functions, enabled when the `Sharedlockbox` is used.
+    - Updates to the `OptimismPortalInterop`: Extended functionality on top of `OptimismPortal2` to handle migrating its ETH liquidity to the `SharedLockbox` and locking/unlocking.
+
+Below are references for this project:
+
+- Specs PR: https://github.com/ethereum-optimism/specs/pull/465
+- Implementation: Refer to this branch https://github.com/defi-wonderland/optimism/tree/sc-feat/add-shared-lockbox
+
+Note that the inclusion of the dependency set in the fault-proof mechanism is relevant for the `SharedLockbox`, and it must function prior to the migration process of the contracts (analyzed here) to secure state transitions and thus withdrawals when a chain is added to any dependency set. The discussion around its design and implementation is still ongoing.
+
+> 🗂️
+> **For more context about the Interop project, refer to the following docs:**
+> 1. [Interop System Diagram](https://www.notion.so/16c8052fcbb24b93ad1a539b5f8db4c1?pvs=21)
+> 2. [Interop PID](https://www.notion.so/16c8052fcbb24b93ad1a539b5f8db4c1?pvs=21)
+> 3. [Interop Audit Request](https://docs.google.com/document/d/1Rcuzbsguh7koT2jFru5ft9T8zAvjBEzbt0zF5LNQQ08/edit?tab=t.0)
+
+## Failure Modes and Recovery Paths
+
+### FM1: Unauthorized access to `unlockETH` function
+
+- **Description:** If an attacker bypasses access controls for `unlockETH`, the deposit and withdrawal invariants can be broken, potentially resulting in the worst-case scenario of draining all ETH liquidity from the `SharedLockbox`.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. Unauthorized ETH unlocking would result in a direct loss of ETH from the `SharedLockbox`.
+    - Likelihood: Low. Access controls are specified in `if (!superchainConfig().authorizedPortals(msg.sender)) revert Unauthorized()`, but errors or misconfigurations remain possible.
+- **Mitigation:** There should be tests for `lockETH` and `unlockETH` checks. Access control must be strictly enforced, permitting only approved `OptimismPortal` contracts via the `authorizedPortals` mapping in `SuperchainConfigInterop`.
+- **Detection:** Monitor `ETHUnlocked` events to verify consistency with authorized portal addresses. Set up alerts for suspected unauthorized activity from non-approved addresses.
+- **Recovery Path(s):** Pause the system through `SuperchainConfig` to halt the `SharedLockbox` upon detection of unauthorized access.
+
+### FM2: Compromised or poorly implemented `clusterManager` on L1
+
+- **Description:** The `clusterManager` role in `SuperchainConfigInterop` can directly call `addDependency` on L1, bypassing the normal L2 → L1 flow. If compromised, an attacker could add malicious or arbitrary chain IDs and authorize their own portal.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. A malicious addition to the dependency set could allow an attacker to `unlockETH` from the `SharedLockbox`.
+    - Likelihood: Low. The `clusterManager` should be a highly secured role (e.g., a multi-sig or governance key).
+- **Mitigations:** Strictly secure the `clusterManager` key.
+- **Detection:** Track the `DependencyAdded` event for calls from `clusterManager`. Compare newly added chain IDs to an approved list.
+- **Recovery Path(s):** Pause the system through `SuperchainConfig`. Since removal functions are unavailable, an upgrade directly over **`SuperchainConfigInterop`** to clear storage slots will be required.
+
+### FM3: Unauthorized or incorrectly initialized `superchainConfig` address in `OptimismPortalInterop`.
+
+- **Description:** In both `SharedLockbox` and `OptimismPortalInterop`, the address of the `SuperchainConfig` is stored in a special storage slot. If it’s set incorrectly or to a malicious address, the entire access-control scheme breaks.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. Since the entire access-control scheme depends on `SuperchainConfig`, all funds could potentially be extracted from the **OptimismPortal**.
+    - Likelihood: Low. Storage slots and values are carefully set during protocol upgrades.
+- **Mitigations:** There should be tests for the initialization step, and ensuring that `superchainConfig` is only set once.
+- **Detection:** Verify the final `superchainConfig` address in newly deployed or upgraded contracts.
+- **Recovery Path(s):** Pause the system, to upgrade the contract to set `superchainConfig` correctly.
+
+### FM4: Solidity Compiler generates an exploitable Bytecode
+
+- **Description:** Smart contracts are typically audited based on their **source code**, assuming that the Solidity compiler (`solc`) produces a correct and verifiable output. Relying solely on source code audits, without verifying the actual bytecode, could introduce undetected vulnerabilities.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. If the Solidity compiler generates incorrect or exploitable bytecode, it could introduce a hidden attack vector.
+    - Likelihood: Low. The Solidity compiler is widely used and battle-tested, but compiler bugs have occurred in the past.
+- **Mitigations:** Use a widely audited Solidity version. ****Compare the Keccak-256 hash of the compiled bytecode against an expected reference. Track Solidity compiler CVEs to be aware of past and ongoing issues.
+- **Detection:** Verify deployed bytecode against expected hash values before proxy to target the implementation contract.
+- **Recovery Path(s):** Pause the system if a Solidity compiler bug is identified.
+
+### FM5: Buggy upgrade over the `SharedLockbox`
+
+- **Description:** The `SharedLockbox` is a proxied contract, which means its implementation can be upgraded. While this provides flexibility for fixes and improvements, a buggy or malicious upgrade could introduce vulnerabilities that compromise all ETH liquidity.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. If an upgrade introduces a bug or an exploitable backdoor, all ETH stored in the `SharedLockbox` could be drained.
+    - Likelihood: Low to Medium. Upgrades are controlled by Security Council, after governance approval. However, upgrade mistakes or compromised keys remain a risk.
+- **Mitigations:** Any new implementation must go through the audit and testing. Upgrades should not be regular in this contract and it should maintain minimal code. Upgrade procedures and keys must follows the proper security practices.
+- **Detection:** Verify proxy and implementation contracts
+- **Recovery Path(s):** Pause the system through `SuperchainConfig` to halt the `SharedLockbox`.
+
+### FM6: Unauthorized use of `DEPOSITOR_ACCOUNT` to call `addDepedency` in `DependencyManager`
+
+- **Description:** On L2, only the `DEPOSITOR_ACCOUNT` can call `addDependency` in `DependencyManager`. If attackers can spoof or impersonate that account, they can push arbitrary additions to the dependency set, eventually letting malicious portals into the `SharedLockbox`.
+- **Risk Assessment:** Medium.
+    - Potential Impact: High. Attackers could create fake or malicious chain entries that, if get accepted on L1, would authorize an attacker-controlled portal in `SharedLockbox`.
+    - Likelihood: Low. The `DEPOSITOR_ACCOUNT` is typically used by the derivation pipeline or otherwise restricted. Previously, there were specific deposit calls that used this account, but those are no longer in use. If the `DEPOSITOR_ACCOUNT` becomes callable in other ways under future upgrades, the attack surface would increase.
+- **Mitigations:** There should be tests that ensure `DEPOSITOR_ACCOUNT` only can be called as part of a coordinated upgrade. Review `DEPOSITOR_ACCOUNT` privileges if it is used for other features in the future. Ensure interop fault proof is properly set to reject states that add a non-expected chain.
+- **Detection:** Monitor `DependencyAdded` events to catch unexpected additions. Do periodical checks to `dependencySet` to review the existing dependency set recorded in the chain.
+- **Recovery Path(s):** Because removing a chain from the dependency set is currently irreversible, a hard fork will be the only way to revert an unauthorized addition. Patch any potential issue around `DEPOSITOR_ACCOUNT`.
+
+### FM7: Unable to add new Chains by Reaching the `uint8` limit in `addDependency`
+
+- **Description:** Both `DependencyManager` and `SuperchainConfigInterop` limit the number of dependencies to `type(uint8).max` (= 255). Once the set size reaches 255, any further `addDependency` call will revert.
+- **Risk Assessment:** Low.
+    - Potential Impact: Medium. Future expansions of the interoperable set are blocked.
+    - Likelihood: Low. 255 chains are already quite large, but could be a problem if the interoperable set wants to grow beyond that.
+- **Mitigations:** There should be a defined path to upgrade to increase the type from `uint8` to `uint16` or another approach. Current dependency set size should be monitored.
+- **Detection:** A simple periodic check of `dependencySetSize` in both `DependencyManager` and `SuperchainConfigInterop` should be sufficient to ensure the limit is not approaching. Reverts on calls to `addDependency` would indicate this issue.
+- **Recovery Path(s):** An upgrade over `SuperchainConfigInterop` and `DependencyManager` is needed to support a larger capacity.
+
+### FM8: `addDependency` completed in L2 but `SharedLockbox` not activated, causing withdrawal failures
+
+- **Description:** After an L2 call to `addDependency`, the message must be proven and finalized on L1 to “activate” the chain’s membership in the `SuperchainConfigInterop` and trigger `migrateLiquidity`. If users are allowed to send ETH to the new chain through interop and perform a withdrawal before the liquidity migration is complete, they might see reverts.
+- **Risk Assessment:** Low.
+    - Potential Impact: Low. ETH withdrawals may fail if they are executed before the migration is completed.
+    - Likelihood: Very Low. Withdrawals should be able to finalize in order, so the chance of an ETH withdrawal request being finalized before the migration is executed is quite minimal.
+- **Mitigations:** Migration procedures should ensure monitoring, proving, and finalizing the withdrawal transaction once it is possible to do so. Alternatively, chains may choose to delay finalizing cross-chain transfers until the migration is completed.
+- **Detection:** Up to the upgrade procedures.
+- **Recovery Path(s):** Prove and finalize the withdrawal that executes the migration.
+
+### FM9: `ETHMigrated` flag remains `false` in the `OptimismPortalInterop` after migration
+
+- **Description:** During the `migrateLiquidity` call, the `OptimismPortalInterop` sets an internal `migrated = true` flag. If that flag is never set, the portal incorrectly tries to handle ETH locally (even though liquidity is already in the `SharedLockbox`).
+- **Risk Assessment:** Low.
+    - Potential Impact: Medium. This situation will cause deposits and withdrawals to fail.
+    - Likelihood: Very Low. The logic is straightforward, so only an upgrade or a code bug might cause it.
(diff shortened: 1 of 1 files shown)


security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
+# Shared Lockbox: Failure Modes and Recovery Path Analysis
+
+| Author | Joxes, Agusduha |
+| --- | --- |
+| Created at | 2025-01-28 |
+| Initial Reviewers | Pending |
+| Needs Approval From | Pending |
```
mds1, line 7: @smartcontracts adding you as a second "needs approval from" reviewer here since the lockbox is a huge honeypot and very critical
```suggestion
| Initial Reviewers | Josep Bove |
| Needs Approval From | Matt Solomon, Kelvin Fichter |
```

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+| Status | In Review |
+
+## Introduction
+
+This document covers the changes introduced by the addition of the Shared Lockbox design, a singleton contract that stores all ETH liquidity given a set of interoperable chains. It addresses the [ETH withdrawals problem](https://github.com/ethereum-optimism/specs/issues/362) by the introduction of [SuperchainWETH](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/superchain-weth.md). The following components are:
```
mds1, line 12: typo?
```suggestion
This document covers the changes introduced by the addition of the Shared Lockbox design, a singleton contract that stores all ETH liquidity for a given set of interoperable chains. It addresses the [ETH withdrawals problem](https://github.com/ethereum-optimism/specs/issues/362) by the introduction of [SuperchainWETH](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/superchain-weth.md). The following components are:
```

tynes:
This looks pretty thorough to me

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+
+- **Description:** The `SharedLockbox` is a proxied contract, which means its implementation can be upgraded. While this provides flexibility for fixes and improvements, a buggy or malicious upgrade could introduce vulnerabilities that compromise all ETH liquidity.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. If an upgrade introduces a bug or an exploitable backdoor, all ETH stored in the `SharedLockbox` could be drained.
+    - Likelihood: Low to Medium. Upgrades are controlled by Security Council, after governance approval. However, upgrade mistakes or compromised keys remain a risk.
+- **Mitigations:** Any new implementation must go through the audit and testing. Upgrades should not be regular in this contract and it should maintain minimal code. Upgrade procedures and keys must follows the proper security practices.
+- **Detection:** Verify proxy and implementation contracts
+- **Recovery Path(s):** Pause the system through `SuperchainConfig` to halt the `SharedLockbox`.
+
+### FM6: Unauthorized use of `DEPOSITOR_ACCOUNT` to call `addDepedency` in `DependencyManager`
```
tynes, line 86: Even if a chain decides to modify their sequencer code to try to enable this, the bridge will block the withdrawal

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+
+- **Description:** During the `migrateLiquidity` call, the `OptimismPortalInterop` sets an internal `migrated = true` flag. If that flag is never set, the portal incorrectly tries to handle ETH locally (even though liquidity is already in the `SharedLockbox`).
+- **Risk Assessment:** Low.
+    - Potential Impact: Medium. This situation will cause deposits and withdrawals to fail.
+    - Likelihood: Very Low. The logic is straightforward, so only an upgrade or a code bug might cause it.
+- **Mitigations:** There should be tests for `migrateLiquidity` to set as `true`.
+- **Detection:** Check whether the `migrated` function is marked as `false` after `ETHMigrated` event is emitted. Watch for unexpected reverts on user withdrawals.
+- **Recovery Path(s):** Perform an upgrade to set the flag correctly.
+
+### FM9: `addDependency` completed in L2 but `SharedLockbox` not activated, causing withdrawal failures
```
tynes, line 116: This is my largest concern until we test it end to end

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+| Status | In Review |
+
+## Introduction
+
+This document covers the changes introduced by the addition of the Shared Lockbox design, a singleton contract that stores all ETH liquidity given a set of interoperable chains. It addresses the [ETH withdrawals problem](https://github.com/ethereum-optimism/specs/issues/362) by the introduction of [SuperchainWETH](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/superchain-weth.md). The following components are:
+
+- **Contracts**:
+    - Introducing `SharedLockbox`: Stores all ETH liquidity for an interoperable graph. Only authorized `OptimismPortal` addresses can call `lockETH` and `unlockETH`.
+    - Introducing `DependencyManager`: A predeploy proxy and implementation contract. Manages the local chain’s dependency set. Triggers an L2 → L1 withdrawal to request adding a new chain to the dependency set on L1 in `SuperchainConfigInterop`.
+    - Introducing `SuperchainConfigInterop`: A proxy and implementation contract in L1. Processes add-dependency requests, track the official “mesh” of authorized chains and their portals, and authorizes each chain’s `OptimismPortal` in the `SharedLockbox`.
```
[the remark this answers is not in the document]
mds1: For the first question, I now see them linked below: https://github.com/defi-wonderland/optimism/tree/sc-feat/add-shared-lockbox. Since these exist in a fork, have we been running the standard monorepo CI on this code? It looks like no, but we definitely should

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+    - Potential Impact: Critical. Unauthorized ETH unlocking would result in a direct loss of ETH from the `SharedLockbox`.
+    - Likelihood: Low. Access controls are specified in `if (!superchainConfig().authorizedPortals(msg.sender)) revert Unauthorized()`, but errors or misconfigurations remain possible.
+- **Mitigation:** There should be tests for `lockETH` and `unlockETH` checks. Access control must be strictly enforced, permitting only approved `OptimismPortal` contracts via the `authorizedPortals` mapping in `SuperchainConfigInterop`.
+- **Detection:** Monitor `ETHUnlocked` events to verify consistency with authorized portal addresses. Set up alerts for suspected unauthorized activity from non-approved addresses.
+- **Recovery Path(s):** Pause the system through `SuperchainConfig` to halt the `SharedLockbox` upon detection of unauthorized access.
+
+### FM2: Compromised or poorly implemented `clusterManager` on L1
+
+- **Description:** The `clusterManager` role in `SuperchainConfigInterop` can directly call `addDependency` on L1, bypassing the normal L2 → L1 flow. If compromised, an attacker could add malicious or arbitrary chain IDs and authorize their own portal.
```
mds1, line 48: Why do we have this flow in addition to the normal L2 -> L1 flow? When would each be used? I did a quick search in the specs but did not see this answered

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. A malicious addition to the dependency set could allow an attacker to `unlockETH` from the `SharedLockbox`.
+    - Likelihood: Low. The `clusterManager` should be a highly secured role (e.g., a multi-sig or governance key).
+- **Mitigations:** Strictly secure the `clusterManager` key.
```
mds1, line 52: The action items say to add tests for this, so just added a corresponding sentence here for that
```suggestion
- **Mitigations:** Tests to ensure `addDependency` cannot be called by unauthorized accounts. Strictly secure the `clusterManager` key.
```

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+- **Recovery Path(s):** Pause the system through `SuperchainConfig`. Since removal functions are unavailable, an upgrade directly over **`SuperchainConfigInterop`** to clear storage slots will be required.
+
+### FM3: Unauthorized or incorrectly initialized `superchainConfig` address in `OptimismPortalInterop`.
+
+- **Description:** In both `SharedLockbox` and `OptimismPortalInterop`, the address of the `SuperchainConfig` is stored in a special storage slot. If it’s set incorrectly or to a malicious address, the entire access-control scheme breaks.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. Since the entire access-control scheme depends on `SuperchainConfig`, all funds could potentially be extracted from the **OptimismPortal**.
+    - Likelihood: Low. Storage slots and values are carefully set during protocol upgrades.
+- **Mitigations:** There should be tests for the initialization step, and ensuring that `superchainConfig` is only set once.
+- **Detection:** Verify the final `superchainConfig` address in newly deployed or upgraded contracts.
```
mds1, line 63: This verification should be a check somewhere in the OPCM scripts

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+- **Recovery Path(s):** Pause the system, to upgrade the contract to set `superchainConfig` correctly.
+
+### FM4: Solidity Compiler generates an exploitable Bytecode
+
+- **Description:** Smart contracts are typically audited based on their **source code**, assuming that the Solidity compiler (`solc`) produces a correct and verifiable output. Relying solely on source code audits, without verifying the actual bytecode, could introduce undetected vulnerabilities.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. If the Solidity compiler generates incorrect or exploitable bytecode, it could introduce a hidden attack vector.
+    - Likelihood: Low. The Solidity compiler is widely used and battle-tested, but compiler bugs have occurred in the past.
+- **Mitigations:** Use a widely audited Solidity version. ****Compare the Keccak-256 hash of the compiled bytecode against an expected reference. Track Solidity compiler CVEs to be aware of past and ongoing issues.
```
mds1, line 72: nit
```suggestion
- **Mitigations:** Use a widely audited Solidity version. Compare the Keccak-256 hash of the compiled bytecode against an expected reference. Track Solidity compiler CVEs to be aware of past and ongoing issues.
```

mds1: Also, what benefit does the hash check give you? 

For tracking solidity compiler CVEs, @AmadiMichael has done some research into this in the past as part of deciding what solidity version to upgrade to. He probably has some references to share to help here

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+- **Detection:** Verify deployed bytecode against expected hash values before proxy to target the implementation contract.
```
mds1, line 73: Staying on top of new solidity bugs is the other detection here. Though in reality if something bad happened, solidity team would likely reach out to us as part of incident response efforts

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+- **Recovery Path(s):** Pause the system if a Solidity compiler bug is identified.
+
+### FM5: Buggy upgrade over the `SharedLockbox`
```
mds1, line 76: We have a detailed list of FMA failure modes for bad upgrades in https://github.com/ethereum-optimism/design-docs/blob/03cf124987420a73fb7695d718ce3b9aa5ca1a0c/security/fma-generic-contracts.md. I would suggest having a failure mode that links to that doc, and explains any lockbox-specific mitigations for each item

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+
+- **Description:** The `SharedLockbox` is a proxied contract, which means its implementation can be upgraded. While this provides flexibility for fixes and improvements, a buggy or malicious upgrade could introduce vulnerabilities that compromise all ETH liquidity.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. If an upgrade introduces a bug or an exploitable backdoor, all ETH stored in the `SharedLockbox` could be drained.
+    - Likelihood: Low to Medium. Upgrades are controlled by Security Council, after governance approval. However, upgrade mistakes or compromised keys remain a risk.
+- **Mitigations:** Any new implementation must go through the audit and testing. Upgrades should not be regular in this contract and it should maintain minimal code. Upgrade procedures and keys must follows the proper security practices.
+- **Detection:** Verify proxy and implementation contracts
```
mds1, line 83:
```suggestion
- **Detection:** Verify proxy and implementation contracts through automated checks in the superchain-ops task
```

security/fma-shared-lockbox.md
mds1: Can you expand on how the bridge would block it? If you are an authorized portal, what prevents you from adding an arbitrary withdraw transaction from your chain to pass [these checks](https://github.com/defi-wonderland/optimism/blob/f805e783f10991e19ebbea406dd8dedbc7f624da/packages/contracts-bedrock/src/L1/SuperchainConfigInterop.sol#L127-L130) and add a new chain? It seems any chain in the interop set has permission to change the dependency set. I believe the problem is that everyone has to run the same malicious code that adds the withdrawal to change the dependency set, so the malicious chain would have to sneak their code into the main node software? (Which could happen during e.g. big PRs)

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+    - Likelihood: Low. The `DEPOSITOR_ACCOUNT` is typically used by the derivation pipeline or otherwise restricted. Previously, there were specific deposit calls that used this account, but those are no longer in use. If the `DEPOSITOR_ACCOUNT` becomes callable in other ways under future upgrades, the attack surface would increase.
+- **Mitigations:** There should be tests that ensure `DEPOSITOR_ACCOUNT` only can be called as part of a coordinated upgrade. Review `DEPOSITOR_ACCOUNT` privileges if it is used for other features in the future. Ensure interop fault proof is properly set to reject states that add a non-expected chain.
+- **Detection:** Monitor `DependencyAdded` events to catch unexpected additions. Do periodical checks to `dependencySet` to review the existing dependency set recorded in the chain.
+- **Recovery Path(s):** Because removing a chain from the dependency set is currently irreversible, a hard fork will be the only way to revert an unauthorized addition. Patch any potential issue around `DEPOSITOR_ACCOUNT`.
```
mds1, line 94: typo?
```suggestion
- **Recovery Path(s):** Because removing a chain from the dependency set is currently unsupported, a hard fork will be the only way to revert an unauthorized addition. Patch any potential issue around `DEPOSITOR_ACCOUNT`.
```

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+
+### FM7: Unable to add new Chains by Reaching the `uint8` limit in `addDependency`
```
mds1, line 96: Why limit it to uint8? Can we just make it a uint64 or uint256 and remove this failure mode?

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+
+- **Description:** Both `DependencyManager` and `SuperchainConfigInterop` limit the number of dependencies to `type(uint8).max` (= 255). Once the set size reaches 255, any further `addDependency` call will revert.
+- **Risk Assessment:** Low.
+    - Potential Impact: Medium. Future expansions of the interoperable set are blocked.
+    - Likelihood: Low. 255 chains are already quite large, but could be a problem if the interoperable set wants to grow beyond that.
+- **Mitigations:** There should be a defined path to upgrade to increase the type from `uint8` to `uint16` or another approach. Current dependency set size should be monitored.
+- **Detection:** A simple periodic check of `dependencySetSize` in both `DependencyManager` and `SuperchainConfigInterop` should be sufficient to ensure the limit is not approaching. Reverts on calls to `addDependency` would indicate this issue.
+- **Recovery Path(s):** An upgrade over `SuperchainConfigInterop` and `DependencyManager` is needed to support a larger capacity.
+
+### FM8: `ETHMigrated` flag remains `false` in the `OptimismPortalInterop` after migration
```
mds1, line 106: Is the migration plan fleshed out more? Migration is a high-risk action and it would be good to see a concrete sequence of steps we plan to take to perform and validate ETH migration for all portals.

For example one way to mitigate issues here is to atomically set portal implementations to a migrator contract, and then once migration is complete, set it to a portal implementation with no migration features. This removes the need to carry around all migration code and `if (!s.migrated)` checks, and removes the need to remember to remove those in the future

(This may have all been discussed in the design doc phase, in which case feel free to link to those docs. I feel like we discussed this at some point in the past but don't recall where we left off)

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+- **Risk Assessment:** Low.
+    - Potential Impact: Medium. This situation will cause deposits and withdrawals to fail.
+    - Likelihood: Very Low. The logic is straightforward, so only an upgrade or a code bug might cause it.
+- **Mitigations:** There should be tests for `migrateLiquidity` to set as `true`.
+- **Detection:** Check whether the `migrated` function is marked as `false` after `ETHMigrated` event is emitted. Watch for unexpected reverts on user withdrawals.
+- **Recovery Path(s):** Perform an upgrade to set the flag correctly.
+
+### FM9: `addDependency` completed in L2 but `SharedLockbox` not activated, causing withdrawal failures
+
+- **Description:** After an L2 call to `addDependency`, the message must be proven and finalized on L1 to “activate” the chain’s membership in the `SuperchainConfigInterop` and trigger `migrateLiquidity`. If users are allowed to send ETH to the new chain through interop and perform a withdrawal before the liquidity migration is complete, they might see reverts.
```
mds1, line 118: Can you expand on the failure mode here? I don't fully follow the details here

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+### Generic items we need to take into account:
+
+See [fma-generic-contracts.md](https://github.com/ethereum-optimism/design-docs/blob/main/security/fma-generic-contracts.md).
+
+- [x]  Check this box to confirm that these items have been considered and updated if necessary.
+
+## Action Items
+
+- [ ]  Resolve all the comments.
+- [ ]  FM1: Provide tests.
```
mds1, line 135: may need to tweak some of these a bit based on the above, e.g. adding info about where to verify things in scripts, etc

security/fma-shared-lockbox.md
AmadiMichael: Ah yeah, you might find these helpful!

- https://soliditylang.org/blog/category/security-alerts/
- https://00xsev.github.io/solidityBugsByVersion/

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+- **Detection:** Track the `DependencyAdded` event for calls from `clusterManager`. Compare newly added chain IDs to an approved list.
+- **Recovery Path(s):** Pause the system through `SuperchainConfig`. Since removal functions are unavailable, an upgrade directly over **`SuperchainConfigInterop`** to clear storage slots will be required.
+
+### FM3: Unauthorized or incorrectly initialized `superchainConfig` address in `OptimismPortalInterop`.
+
+- **Description:** In both `SharedLockbox` and `OptimismPortalInterop`, the address of the `SuperchainConfig` is stored in a special storage slot. If it’s set incorrectly or to a malicious address, the entire access-control scheme breaks.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. Since the entire access-control scheme depends on `SuperchainConfig`, all funds could potentially be extracted from the **OptimismPortal**.
+    - Likelihood: Low. Storage slots and values are carefully set during protocol upgrades.
+- **Mitigations:** There should be tests for the initialization step, and ensuring that `superchainConfig` is only set once.
```
JosepBove, line 62: I would add here that the security team must validate "checklist style" that this storage slot is not changed unintentionally before any upgrade.

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+
+### FM1: Unauthorized access to `unlockETH` function
+
+- **Description:** If an attacker bypasses access controls for `unlockETH`, the deposit and withdrawal invariants can be broken, potentially resulting in the worst-case scenario of draining all ETH liquidity from the `SharedLockbox`.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. Unauthorized ETH unlocking would result in a direct loss of ETH from the `SharedLockbox`.
+    - Likelihood: Low. Access controls are specified in `if (!superchainConfig().authorizedPortals(msg.sender)) revert Unauthorized()`, but errors or misconfigurations remain possible.
+- **Mitigation:** There should be tests for `lockETH` and `unlockETH` checks. Access control must be strictly enforced, permitting only approved `OptimismPortal` contracts via the `authorizedPortals` mapping in `SuperchainConfigInterop`.
+- **Detection:** Monitor `ETHUnlocked` events to verify consistency with authorized portal addresses. Set up alerts for suspected unauthorized activity from non-approved addresses.
+- **Recovery Path(s):** Pause the system through `SuperchainConfig` to halt the `SharedLockbox` upon detection of unauthorized access.
```
JosepBove, line 44: Ideally an automated pause

security/fma-shared-lockbox.md
Joxess: This case refers to a situation where a user withdrawal with an amount greater than the ETH available in `OptimismPortal` is finalized before the Shared Lockbox is integrated. I believe this is possible if the chain accepts cross-chain messages before migrating to the Shared Lockbox, given that there is at least a seven-day gap between the `addDependency` calls in L2 and L1, respectively.

A risk scenario could involve accepting a `relayETH` transaction with an amount greater than the ETH available in `OptimismPortal`, followed by a withdrawal attempt before migrating to the lockbox. This is concerning because finalized withdrawals are not necessarily executed in the order in which they were initiated.

security/fma-shared-lockbox.md
Joxess: Currently, the `clusterManager` is used to add the first chain to the `SuperchainConfigInterop`. After that, its purpose may be limited to function as a fallback.

security/fma-shared-lockbox.md
Joxess: provide more details in mitigations on how this case is going to be prevented.

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,147 @@
...
+
+- **Description:** The `SharedLockbox` is a proxied contract, which means its implementation can be upgraded. While this provides flexibility for fixes and improvements, a buggy or malicious upgrade could introduce vulnerabilities that compromise all ETH liquidity.
+- **Risk Assessment:** High.
+    - Potential Impact: Critical. If an upgrade introduces a bug or an exploitable backdoor, all ETH stored in the `SharedLockbox` could be drained.
+    - Likelihood: Low to Medium. Upgrades are controlled by Security Council, after governance approval. However, upgrade mistakes or compromised keys remain a risk.
+- **Mitigations:** Any new implementation must go through the audit and testing. Upgrades should not be regular in this contract and it should maintain minimal code. Upgrade procedures and keys must follows the proper security practices.
+- **Detection:** Verify proxy and implementation contracts
+- **Recovery Path(s):** Pause the system through `SuperchainConfig` to halt the `SharedLockbox`.
+
+### FM6: Unauthorized use of `DEPOSITOR_ACCOUNT` to call `addDepedency` in `DependencyManager`
```
mds1, line 86: From FMA review call: We should have a dedicated design doc review for this approach to L1 upgrades and corresponding security considerations, since it circumvents our standard SDLC for L1 upgrades and is an entirely new way to make L1 changes. This approach to modifying upgrades can be decoupled from the lockbox and ship as its own feature

tynes: We wanted to avoid adding extra work to the bottleneck of multisig operations. Are you suggesting that we only do multisig operations to add chains?

security/fma-shared-lockbox.md
tynes: The original spec passed the dep set via L1 attributes, so it was limited to u8 to keep it small enough to pass on every block. The max size should actually be smaller, given it should not be possible to create an unprovable cluster

Joxess:
After reviewing the latest Shared Lockbox design, the decision has been made to remove the feature that allowed adding a new chain via an upgrade initiated from L2. This change has been reflected in the code, and I will update this doc as well.

security/fma-shared-lockbox.md
Joxess: Update: It is descouped from this work for now

security/fma-shared-lockbox.md
Joxess: Update: It is descouped from this work for now

Joxess: Update: It is decoupled from this work for now

security/fma-shared-lockbox.md
Joxess: Update: It is decoupled from this work for now

security/fma-shared-lockbox.md
Joxess: Update: It is decoupled from this work for now

security/fma-shared-lockbox.md
Joxess: Update: It is decoupled from this work for now

security/fma-shared-lockbox.md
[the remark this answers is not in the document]
Joxess: Added the official PR

security/fma-shared-lockbox.md
agusduha: Is this possible with the current guardian?

mds1: This is not something we currently have the infrastructure setup for. If you agree it's a good idea we can track this as an action/interop issue to make sure the idea doesn't get lost, but I don't think this would be a blocker

Joxess: Added into action items

security/fma-shared-lockbox.md
Joxess: I have added we should provide tests for FM1, FM2 and FM3 cases; and monitor solutions should be based on detection needs

security/fma-shared-lockbox.md
```diff
@@ -0,0 +1,128 @@
+# Shared Lockbox: Failure Modes and Recovery Path Analysis
+
+| Author | Joxes, Agusduha |
+| --- | --- |
+| Created at | 2025-01-28 |
+| Initial Reviewers | Josep Bove |
+| Needs Approval From | Matt Solomon, Kelvin Fichter |
+| Status | In Review |
```
mds1, line 8:
```suggestion
| Status | Implementing Actions |
```

mds1 (approved):
Also needs approval from @smartcontracts before merging

smartcontracts (approved):

PR #201, opened 2025-02-10
Pectra on Isthmus FMA

meyer9:
**Description**

Adds an FMA for Pectra on Isthmus.


governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Failure Modes and Recovery Paths
+
+### FM1: Chain halt due to upgrade transaction failing
+
+- **Description:** The chain could halt if a network upgrade transaction can't be included in the upgrade block.
+- **Risk Assessment:** High severity, low likelihood
+- **Mitigations:**
+  1. end-to-end tests created for network upgrade transactions
+  2. same upgrade path as previous predeploys
```
refcell, line 75: If possible, can we link to other predeploys here so we can be specific in what we're referring to?

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+
+### FM2: Chain halt due to request hash mismatch
+
+- **Description:** Requests hash was added to the block header. If it mismatches between geth and op-node, the chain could halt
+- **Risk Assessment:** High severity, low likelihood
+- **Mitigations:**
+  1. End-to-end tests
+  2. Simple validation rules - always must be an empty hash
+- **Detection:** Monitors should trigger an alert immediately.
+- **Recovery Path(s)**: This would require an emergency update if it occurred fixing the network upgrade transaction and updates to `op-node`.
```
refcell, line 87: This wasn't updated from prior FM

```suggestion
- **Recovery Path(s)**: This would require an emergency update if it occurred fixing requests hash logic in respective clients (e.g. `op-node`, `op-geth`, `op-reth`, ...)
```

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+
+### FM3: EIP-7702 transactions cannot be included on the chain
+
+- **Description:** SetCode (EIP-7702) transactions can't be included on chain.
+- **Risk Assessment:** Low impact, low likelihood
+- **Mitigations:** 
+	1. End-to-end tests of set code transactions before and after the fork
+- **Detection:** Sending set code transactions would fail; probably manual detection is most likely.
+- **Recovery Path(s)**: This would require another update and fork.
```
refcell, line 96: Is this detailing a bug where the specification logic isn't sufficient for EIP-7702 txs to land on chain?

I could also see a small RPC bug or something non-logic related to cause an EIP-7702 tx issue that wouldn't necessarily require a hardfork.

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+
+### FM5: BLS Precompiles could fail to execute in the FP program
+
+- **Description:** If implemented in the FP program, BLS precompiles could fail to execute. This could cause certain blocks to be unprovable.
+- **Risk Assessment:** High impact, low likelihood
+- **Mitigations:**
+	1. Fully mitigated by using an accelerated precompile that calls out to L1 instead of calculating inside the program. This means it should 
+  have the same performance characteristics as existing crypto precompiles like `ecrecover`.
+- **Detection:** op-program trace runner failure (possibly triggering an alert)
+- **Recovery Path(s)**: If this occurred, which it can't with the mitigations, we'd have to make op-program efficient enough to execute BLSprecompiles in sufficient time.
```
refcell, line 116: nit
```suggestion
- **Recovery Path(s)**: If this occurred, which it can't with the mitigations, we'd have to make op-program efficient enough to execute BLS precompiles in sufficient time.
```

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+
+### FM6: Increased call data cost affects network upgrade transactions
+
+- **Description:** Increased calldata cost could mean that network upgrade transactions fail.
+- **Risk Assessment:** High impact, low likelihood
+- **Mitigations:**
+	1. Tested that network upgrade transactions don't fail.
+- **Detection:** op-program trace runner failure (possibly triggering an alert)
+- **Recovery Path(s)**: If this occurred, which it can't with the mitigations, we'd have to make op-program efficient enough to execute BLSprecompiles in sufficient time.
```
refcell, line 125: This needs to be updated for this FM

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+
+### FM7: Early fork if batches containing EIP-7702 transactions could be posted before Pectra
```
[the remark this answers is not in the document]
meyer9: A malicious batcher could cause this to happen if we don't have a check to ensure type 4 (SetCode) txs can only appear in a batch after Isthmus is active. The specs do cover this, so this would only be caused by a batcher implementation diverging from the spec.

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+_Italics are used to indicate things that need to be replaced._
+
+|                    |                                                    |
+| ------------------ | -------------------------------------------------- |
+| Author             | Julian Meyer                                       |
+| Created at         | 2025-02-10                                         |
+| Initial Reviewers  |                  |
```
mds1, line 28: I see @refcell left some comments, any other non-security people that should review this? 
```suggestion
| Initial Reviewers  |  refcell                |
```

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+Included EIPs:
+
+-   EIP-7702: Set code transaction
+-   EIP-2537: BLS12-381 precompiles
+-   EIP-2935: Block hashes contract predeploy
+-   EIP-7251: Increase the MAX_EFFECTIVE_BALANCE
+-   EIP-7623: Increase calldata cost
+-   EIP-7002: Execution layer triggerable withdrawals
+-   EIP-6110: Supply validator deposits on chain
+-   EIP-7685: General purpose execution layer requests
```
mds1, line 65: Are there OP Stack-specific considerations, design docs, specs, etc we can link to? This makes it seem like we are not doing any bespoke work and are just incorporating what e.g. upstream geth implements, and I am unsure if that is the case

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+
+_Italics are used to indicate things that need to be replaced._
+
+|                    |                                                    |
+| ------------------ | -------------------------------------------------- |
+| Author             | Julian Meyer                                       |
+| Created at         | 2025-02-10                                         |
+| Initial Reviewers  |                  |
+| Need Approval From |                            |
+| Status             | Draft |
```
mds1, line 30: When do you expect this to be out of draft and ready for review?

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+
+### FM1: Chain halt due to upgrade transaction failing
+
+- **Description:** The chain could halt if a network upgrade transaction can't be included in the upgrade block.
+- **Risk Assessment:** High severity, low likelihood
+- **Mitigations:**
+  1. end-to-end tests created for network upgrade transactions
+  2. same upgrade path as previous predeploys
+- **Detection:** Monitors should trigger an alert immediately
+- **Recovery Path(s)**: This would require an emergency update if it occurred fixing the network upgrade transaction and updates to `op-node`.
```
mds1, line 77: For each failure mode, let be as specific as possible. Some examples based on the text in failure mode:
- When tests are a mitigation, let's link to those tests as evidence they exist, or create an action item to track writing those tests
- Like @refcell said, let's be explicit about what "same upgrade path" means
- When detection requires monitoring, what exactly are we monitoring? If that monitoring already exists we can link to it, if not there should be an action item to track that
- Does an emergency update mean an emergency hard fork, or just an update of node software with no hard fork needed?

governance/fma-isthmus-l2-pectra.md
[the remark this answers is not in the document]
mds1: @meyer9 General FMA comment since I'm not sure if this is documented anywhere—in general all github comments can be resolved once either: the FMA itself is updated with info from the comment thread (like clarifying this Failure Mode's description with this info), or a new action item to track whatever was in the comment. Therefore in this case we should implement the info from this comment thread directly into the FMA itself.

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+Below is what needs to be done before launch to reduce the chances of the above failure modes occurring, and to ensure they can be detected and recovered from:
+
+- [ ] Resolve all comments on this document and incorporate them into the document itself (Assignee: document author)
+- [ ] FM6: double check that call data cost doesn't break upgrade transactions
+
+## Audit Requirements
+
+_Given the failure modes and action items, will this project require an audit? See [OP Labs Audit Framework: When to get external security review and how to prepare for it](https://gov.optimism.io/t/op-labs-audit-framework-when-to-get-external-security-review-and-how-to-prepare-for-it/6864) for a reference decision making framework. Please explain your reasoning._
+
+**No.** All of the code is well-tested and has very low likelihood for any failure modes with reasonably high impact.
```
mds1, line 155: We should explain the rationale in the context of the audit framework linked above

governance/fma-isthmus-l2-pectra.md
meyer9: Should be sometime today, hopefully in a few hours - planning to go through all the comments and address.

governance/fma-isthmus-l2-pectra.md
meyer9: added links!

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,164 @@
...
+- **Risk Assessment:** High severity, low likelihood
+- **Mitigations:**
+  1. end-to-end tests created for network upgrade transactions
+  2. same upgrade path as previous predeploys
+- **Detection:** Monitors should trigger an alert immediately
+- **Recovery Path(s)**: This would require an emergency update if it occurred fixing the network upgrade transaction and updates to `op-node`.
+
+### FM2: Chain halt due to request hash mismatch
+
+- **Description:** Requests hash was added to the block header. If it mismatches between geth and op-node, the chain could halt
```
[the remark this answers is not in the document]
meyer9: updated wording here

governance/fma-isthmus-l2-pectra.md
meyer9: fixed!

governance/fma-isthmus-l2-pectra.md
meyer9: Updated this FMA with an explanation of what might happen. This would be a fork caused by a consensus bug in a client not allowing EIP-7702 txs after Isthmus is active.

governance/fma-isthmus-l2-pectra.md
meyer9: fixed

governance/fma-isthmus-l2-pectra.md
meyer9: fixed

governance/fma-isthmus-l2-pectra.md
meyer9: added refcell! Not sure about other non-security people, but I'll post in the upgrade channel now that it's ready for review

governance/fma-isthmus-l2-pectra.md
meyer9: I added an explanation for each. There's no central source listing each change currently. We were going a lot off of this document: https://gist.github.com/protolambda/59b534d804c23dc7b3592ad1c38048d4

I can add this doc somewhere, but I think the updated explanations in this doc should give a good overview.

governance/fma-isthmus-l2-pectra.md
meyer9: Clarified all of these items.

About monitoring, what would a good link be? Most of these are consensus bugs that will cause a chain halt (guessing there's a monitor for unsafe L2 liveness) or a fork (would trigger the same alert for a different client like `op-reth`). Or are there specific monitoring tasks we need to do before this is approved - like an open source repo for adding hard fork checks?

governance/fma-isthmus-l2-pectra.md
meyer9: Added to the end with an explanation on why we should rely on real-world testing rather than auditing.

governance/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,195 @@
+
+# Pectra Features on Isthmus
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Introduction](#introduction)
+- [Failure Modes and Recovery Paths](#failure-modes-and-recovery-paths)
```
mds1, line 8: One failure mode I don't see is that L1 Pectra means `msg.sender == tx.origin` checks no longer ensure the caller has no code. We have this check in the PreimageOracle, and we have `Address.isContract` checks in the ERC721Bridges and StandardBridges. The latter two are codesize based so I believe they should be unaffected. But we should have a failure mode for this, and explain why (hopefully) this change doesn't break anything

mds1: Asked in discord about this as well: https://discord.com/channels/1244729134312198194/1331301843829002313/1342155969400012860

meyer9: Good call out! Will add a section for that

mds1: Update on the plan for these checks is here: https://github.com/ethereum-optimism/design-docs/pull/209

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<!-- _Italics are used to indicate things that need to be replaced._ -->
+
+|                    |                                                    |
+| ------------------ | -------------------------------------------------- |
+| Author             | Julian Meyer                                       |
+| Created at         | 2025-02-10                                         |
+| Initial Reviewers  | refcell                                            |
+| Need Approval From |                                                    |
```
Ethnical, line 30:
```suggestion
| Need Approval From |                Tom ASSAS     (Michael Amadi shadowing                |
```

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+| Status             | In Review 🔎                                       |
```
Ethnical, line 31: We need to add this add the end of the current review please not accept this current suggestion now. 
I will accept this at the end myself. 
```suggestion
| Status             |  Implementing Action 🛫                                       |
```

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+> - The size of the document will likely be proportional to the project's complexity.
+> - The ultimate goal of this document is to identify action items to improve the security of the project. The FMA review process can be accelerated by proactively identifying action items during the writing process. -->
+
+## Introduction
+
+This document covers implementation of Pectra features in the OP stack. This requires a hard fork due to new transaction types like EIP-7702 transactions. This does not cover generic hardfork related security concerns, which will be addressed in a separate document.
+
+Below is a list of EIPs that are enabled with Pectra, and a quick description of spec changes needed to support the EIP.
+
+- EIP-7702: Set code transaction
```
Ethnical, line 59: For here, I would like to know if this is possible to make a table like this: 

https://github.com/ethereum-optimism/design-docs/blob/879b1c4d725bf0769c4a589faff7e9a4d468ff63/security/fma-isthmus-l2-pectra.md

And also add one column that is "Is that require custom code"

The idea is to identify if this required custom code to support this EIP or we just merge from UPstream-Geth

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+  - Not applicable to OP
+  - Requests sent to this contract are ignored, and the contract is not deployed by default
+- EIP-7685: General purpose execution layer requests
+  - No requests are currently applicable to OP, so the [execution requests array is always empty in Isthmus](https://github.com/ethereum-optimism/specs/blob/b5e0fa98881171f658f782597a46b641e8f3dfd0/specs/protocol/exec-engine.md#engine_newpayloadv4).
+
+## Failure Modes and Recovery Paths
+
+### FM1: Chain halt due to upgrade transaction failing
+
+- **Description:** The chain could halt if a network upgrade transaction can't be included in the upgrade block.
```
Ethnical, line 82: Which transaction upgrade are speaking here? Can we details more in the description?
How does look like an "upgrade transaction"? Do we have a doc or referrence to point to this?

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+### FM2: Chain halt due to request hash mismatch
+
+- **Description:** Requests hash was added to the block header. If it mismatches between the EL client and the rollup node, the chain could halt.
+- **Risk Assessment:** High severity, low likelihood
+- **Mitigations:**
+  1. End-to-end test ensuring no requests are included in blocks
+      - https://github.com/ethereum-optimism/optimism/pull/14253
+  2. Simplicity of validation rules - always must be an empty hash
+- **Detection:** Alert for L2 Unsafe liveness should be triggered
+- **Recovery Path(s)**: This would require an emergency update if it occurred fixing requests hash logic in respective clients (e.g. `op-node`, `op-geth`, `op-reth`, ...)
```
Ethnical, line 102: How we make sure that the proxyd, op-conductor, op-batcher are also not impacted with the `request_hash` field.

Another question, seems to be mismatch hash so there are both real hash or they are empty hash both side? Because here, this is written that the hash should be empty (`always must be an empty hash`). 

1. If this is the case, how we can make sure that through P2P when we share block from another op-reth/op-geth this one will have a valid `request_hash`. 
2. Do we have test that ensure that if there is an incorrect `request_hash` propagated across multiples components.

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+
+### FM3: EIP-7702 transactions cannot be included on the chain
+
+- **Description:** A client implementation could have a bug that disallows SetCode (EIP-7702) transactions from being included on chain. This could cause an unintended fork by the affected client.
+- **Risk Assessment:** High impact, low likelihood
+- **Mitigations:** 
+	1. End-to-end tests of set code transactions before and after the fork
+        - https://github.com/ethereum-optimism/optimism/pull/14197
+        - https://github.com/ethereum-optimism/optimism/pull/14288
+- **Detection:** Sending set code transactions would fail - manual detection is most likely.
```
Ethnical, line 112: Can you detail more about the *manual* detection? 
You mean the user will see his transaction reverting directly or *not supported*. 
But, would be a case to case stuff?

I wondering if would make sense to you guys, to craft a small `eth_call` with a type 4 that we can run against any EL on L2 to detect if this one is updated or not.

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+
+### FM5: BLS Precompiles could fail to execute in the FP program
+
+- **Description:** If implemented in the FP program, BLS precompiles could fail to execute. This could cause certain blocks to be unprovable.
+- **Risk Assessment:** High impact, low likelihood
+- **Mitigations:**
+	1. Fully mitigated by using an accelerated precompile that calls out to L1 instead of calculating inside the program. This means it should 
+  have the same performance characteristics as existing crypto precompiles like `ecrecover`.
+        - [Specs](https://github.com/ethereum-optimism/specs/blob/b5e0fa98881171f658f782597a46b641e8f3dfd0/specs/protocol/isthmus/exec-engine.md#evm-changes)
+- **Detection:** op-program trace runner failure (possibly triggering an alert)
```
Ethnical, line 134: Can we be more specific about the kind of alert that would be triggered by the `vm-runner` here? 
Probably this would be OOM, If this the performance, I assume?

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+- **Recovery Path(s)**: This would require an emergency client update of the implementation with the bug (e.g. `op-reth`, `op-geth`, ...).
+
+### FM4: BLS Precompiles could cause increased FP Program execution time
+
+- **Description:** If implemented in the FP program, BLS precompiles could greatly increase computation time.
+- **Risk Assessment:** High impact, low likelihood
+- **Mitigations:**
+	1. Fully mitigated by using an accelerated precompile that calls out to L1 instead of calculating inside the program.
+        - [Specs](https://github.com/ethereum-optimism/specs/blob/b5e0fa98881171f658f782597a46b641e8f3dfd0/specs/protocol/isthmus/exec-engine.md#evm-changes)
+- **Detection:** op-program trace runner computation time would increase significantly (should trigger an alert)
```
Ethnical, line 122: Can we be more specific about the kind of alert that would be triggered by the `vm-runner` here? 
Probably this would be OOM, If this the performance, I assume?

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+- **Recovery Path(s)**: If this occurred, which it can't with the mitigations, we'd have to make op-program efficient enough to execute
+BLS precompiles in sufficient time.
+
+### FM5: BLS Precompiles could fail to execute in the FP program
+
+- **Description:** If implemented in the FP program, BLS precompiles could fail to execute. This could cause certain blocks to be unprovable.
+- **Risk Assessment:** High impact, low likelihood
+- **Mitigations:**
+	1. Fully mitigated by using an accelerated precompile that calls out to L1 instead of calculating inside the program. This means it should
```
Ethnical, line 131: Where did we tests that the E2E  tests for transaction using BLS producing block and are submitted as expected by the batcher correctly?

security/fma-isthmus-l2-pectra.md
Ethnical: EDIT: On the specs I found this here we should link this to `request_hash`  + in the reference section. 
https://github.com/ethereum-optimism/specs/blob/b5e0fa98881171f658f782597a46b641e8f3dfd0/specs/protocol/isthmus/exec-engine.md#header-validity-rules

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+
+### FM6: Increased call data cost affects network upgrade transactions
+
+- **Description:** Increased calldata cost could mean that network upgrade transactions fail.
+- **Risk Assessment:** High impact, low likelihood
+- **Mitigations:**
+	1. Tested that network upgrade transactions don't fail.
+        - [Action test created here](https://github.com/ethereum-optimism/optimism/blob/01ddb2e6a09edf55a7cb2130e0a5b6acd0c2d2fa/op-e2e/actions/upgrades/isthmus_fork_test.go#L290)
+- **Detection:** Manual detection is likely after inspecting the fork block.
+- **Recovery Path(s)**: We would have to release an emergency rollup node (e.g. `op-node`) update with fixed network upgrade trnasactions.
```
geoknee, line 145:
```suggestion
- **Recovery Path(s)**: We would have to release an emergency rollup node (e.g. `op-node`) update with fixed network upgrade transactions.
```

Ethnical, line 145:
```suggestion
- **Recovery Path(s)**: We would have to release an emergency rollup node (e.g. `op-node`) update with fixed network upgrade transactions.
```

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+- **Mitigations:**
+	1. Fully mitigated by using an accelerated precompile that calls out to L1 instead of calculating inside the program. This means it should 
+  have the same performance characteristics as existing crypto precompiles like `ecrecover`.
+        - [Specs](https://github.com/ethereum-optimism/specs/blob/b5e0fa98881171f658f782597a46b641e8f3dfd0/specs/protocol/isthmus/exec-engine.md#evm-changes)
+- **Detection:** op-program trace runner failure (possibly triggering an alert)
+- **Recovery Path(s)**: If this occurred, which it can't with the mitigations, we'd have to make op-program efficient enough to execute BLS precompiles in sufficient time.
+
+### FM6: Increased call data cost affects network upgrade transactions
+
+- **Description:** Increased calldata cost could mean that network upgrade transactions fail.
```
Ethnical, line 139: here,  how the calldata cost can increase before the upgrade?
Are we referring to fees on L2 that can potentially increase making the cost of the transaction to high to execute? 
If not can we elaborate to help to understand here?

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+      - [Action test created here](https://github.com/ethereum-optimism/optimism/blob/01ddb2e6a09edf55a7cb2130e0a5b6acd0c2d2fa/op-e2e/actions/upgrades/isthmus_fork_test.go#L290)
+  2. same upgrade path as previous predeploys
+      - example: [EIP-4788 deployment transaction](https://github.com/ethereum-optimism/optimism/blob/b6b74290b5502f45daae57946f969327cdb2d383/op-node/rollup/derive/ecotone_upgrade_transactions.go#L130)
+    
+- **Detection:** Alert for L2 Unsafe liveness should be triggered
+- **Recovery Path(s)**: This would require an emergency update if it occurred fixing the network upgrade transaction in rollup node client software (like `op-node`).
+
+### FM2: Chain halt due to request hash mismatch
+
+- **Description:** Requests hash was added to the block header. If it mismatches between the EL client and the rollup node, the chain could halt.
```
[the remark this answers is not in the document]
meyer9: This can happen if these two clients calculate a different block hash for a given payload. So this would be a bug in `op-node` or an execution client that causes them to disagree on block hash due to request hash mismatch.

I'll clarify here!

security/fma-isthmus-l2-pectra.md
meyer9: We check geth here: https://github.com/ethereum-optimism/optimism/commit/e76409e18ee82b4765ae9d51991ffeae1e032671#diff-b708c9a61d73e393273eb3a63bf8538ab51214cd3d33fa2aff74588eddb45c2cR1101-R1103

But, I think we should also check that op-program validates a block correctly with BLS precompiles. I'll add that test and link it here.

security/fma-isthmus-l2-pectra.md
meyer9: Yeah OOM or a stall where it never succeeds. We (Base) have an alert that checks if the runner has run successfully in the past 8 hrs, but not sure about what alerts OP has.

I'll clarify here if someone can confirm that OP has an alert like this.

security/fma-isthmus-l2-pectra.md
meyer9: See below

security/fma-isthmus-l2-pectra.md
meyer9: Depends on the signer used I think. It can't happen until Isthmus is activated. Blocks and tx pool will reject until Isthmus is activated (at least for Geth): https://github.com/ethereum-optimism/op-geth/pull/535/files#diff-58204a02692cd9ef1cf0767b18dbc4df99eb678126c8982350e5b727af075b5eR74

If a client accidentally does accept them early and include them in a block, it would cause an L2 chain halt for the non-sequencer client.

If a non-sequencer client does not accept `type 4` txs after they should be supported, it would also cause an L2 chain halt for the non-sequencer client.

security/fma-isthmus-l2-pectra.md
meyer9: > Are we referring to fees on L2 that can potentially increase making the cost of the transaction to high to execute?

Yes, due to the calldata gas change EIP included in this update. This could cause network upgrade transactions to fail. But, we checked that they don't fail in an e2e test, so this should be fully tested.

security/fma-isthmus-l2-pectra.md
Ethnical: Should be nice to update and add this PR for tests  in case of incorrect `Request_hash` header.

security/fma-isthmus-l2-pectra.md
Ethnical: Thanks! In this test we should make sure to compute how much cost the l1deposit for a transaction and also include them in this test. 
It would be to avoid scenario that attacker can spend a reasonable gas on L1 for the deposit that can revert the upgrade transaction.

security/fma-isthmus-l2-pectra.md
meyer9: Added this here: https://github.com/ethereum-optimism/design-docs/pull/201/commits/5401b30af365b9ca4f9fcad0569dff2d3bf31616

@geoknee or @sebastianst - would be awesome if you could review the components changed for each of these! Not confident I know exactly what op-geth changes were required for each of these.

security/fma-isthmus-l2-pectra.md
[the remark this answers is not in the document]
meyer9: added! [f977675](https://github.com/ethereum-optimism/design-docs/pull/201/commits/f97767529c39a56f7e3123591b3e94a9250c5fdb)

Also added block header diff

security/fma-isthmus-l2-pectra.md
meyer9: Added those references. About the P2P issue, it reduces to peers following consensus rules. 

- If a malicious client sends a block header with requests root not set to `sha256('')`, the calculated requests root after EVM execution will not match the header and the block will be rejected.
- If a malicious client sends requests in the bodies, but still has an empty requests root, the requests in the bodies will not hash to the correct requests root.

I can add these to a separate FMA if you want, or include them here.

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,207 @@
...
+- **Description:** If implemented in the FP program, BLS precompiles could fail to execute. This could cause certain blocks to be unprovable.
+- **Risk Assessment:** High impact, low likelihood
+- **Mitigations:**
+	1. Fully mitigated by using an accelerated precompile that calls out to L1 instead of calculating inside the program. This means it should 
+  have the same performance characteristics as existing crypto precompiles like `ecrecover`.
+        - [Specs](https://github.com/ethereum-optimism/specs/blob/b5e0fa98881171f658f782597a46b641e8f3dfd0/specs/protocol/isthmus/exec-engine.md#evm-changes)
+- **Detection:** op-program trace runner failure (possibly triggering an alert)
+- **Recovery Path(s)**: If this occurred, which it can't with the mitigations, we'd have to make op-program efficient enough to execute BLS precompiles in sufficient time.
+
+### FM6: Increased call data cost affects network upgrade transactions
```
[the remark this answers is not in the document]
meyer9: [268551e](https://github.com/ethereum-optimism/design-docs/pull/201/commits/268551e25c0dfe43c01d16d1ca93296592d8f513)

security/fma-isthmus-l2-pectra.md
Ethnical: @meyer9 yes, I do think we should put here please and then we can resolve this!

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,268 @@
...
+<!-- See [generic hardfork failure modes](./fma-generic-hardfork.md) and [generic smart contract failure modes](./fma-generic-contracts.md).
+Incorporate any applicable failure modes with FMA-specific mitigations and detections directly into this document. -->
+
+- [x] Check this box to confirm that these items have been considered and updated if necessary.
+
+## Action Items
+
+Below is what needs to be done before launch to reduce the chances of the above failure modes occurring, and to ensure they can be detected and recovered from:
+
+- [ ] Resolve all comments on this document and incorporate them into the document itself (Assignee: document author)
```
Ethnical, line 193:
```suggestion
- [ ] Resolve all comments on this document and incorporate them into the document itself (Assignee: document author)
- [ ] (NON-BLOCKING)  Make a way to enforce the gas limit of a block can't be lower of 30M (cf this [discord thread](https://discord.com/channels/1244729134312198194/1334603521454247946/1351651773755949117) with @tynes & @sebastianst) 
```

security/fma-isthmus-l2-pectra.md
Ethnical: Thank you, I am pinging them to see how we can do that. 
I don't know if https://op-geth.optimism.io/ for now would be enough accurate. 
I will update the content when I got the answer probably tmw.

security/fma-isthmus-l2-pectra.md
Ethnical: @meyer9 So, we have clarify with the alerts `op-challenger` that we have currently within Grafana. 
However, most of the alerts are not public.
For this exact case we have the alert `challenger-runner-panic` that should trigger an alert in case of a challenger that panic from any issue (can be OOM or a panic from the VM).

security/fma-isthmus-l2-pectra.md
Ethnical: Thank you for clarifying!
Make sense here.

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,271 @@
...
+- [ ] (NON-BLOCKING)  Make a way to enforce the gas limit of a block can't be lower of 30M (cf this [discord thread](https://discord.com/channels/1244729134312198194/1334603521454247946/1351651773755949117) with @tynes & @sebastianst) 
+- [x] (NON-BLOCKING):  Add the description of the new block header with the new field `request_hash` in this FMA (@meyer9)
+- [ ] (BLOCKING): Add a test that Validate a block with BLS accelerated precompile (@meyer9 ref to this [comment](https://github.com/ethereum-optimism/design-docs/pull/201/files#r1985623601))
```
Ethnical, line 196:
```suggestion
- [ ] (BLOCKING): Add a test that Validate a block with BLS accelerated precompile (@meyer9 ref to this [comment](https://github.com/ethereum-optimism/design-docs/pull/201/files#r1985623601))
- [ ] (BLOCKING): Add a E2E test that Validate that minting to EOA that has code is not an issue
```

security/fma-isthmus-l2-pectra.md
sebastianst: Yes https://op-geth.optimism.io/ shows the up-to-date diff of op-geth to upstream geth.

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,271 @@
...
+
+### FM3: Increased call data cost affects network upgrade transactions
+
+- **Description:** Increased calldata cost as the result of [EIP-7623](https://eips.ethereum.org/EIPS/eip-7623) could mean that network upgrade transactions fail.
+- **Risk Assessment:** High impact, low likelihood
+- **Mitigations:**
+	1. Tested that network upgrade transactions don't fail.
+        - [Action test created here](https://github.com/ethereum-optimism/optimism/blob/01ddb2e6a09edf55a7cb2130e0a5b6acd0c2d2fa/op-e2e/actions/upgrades/isthmus_fork_test.go#L290)
+  2. The EIP results in the same gas behavior as long as most of the gas is not spent on call data.
+        - Most upgrade transactions create a new contract, and contract creation gas remains the same.
```
sebastianst, line 121: Also Kurtosis, the alpha/betanets and e2e tests all run through the upgrade block, so this would have been caught.

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,271 @@
...
+### FM1: Chain halt due to L1 block deployment upgrade transaction failing
+
+- **Description:** The chain could halt if the [L1 block deployment](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/isthmus/derivation.md#l1block-deployment) upgrade transaction can't be included in the upgrade block.
+- **Risk Assessment:** High severity, low likelihood
+- **Mitigations:**
+  1. end-to-end tests created for the L1 block deployment upgrade transaction
+      - [Action test created here](https://github.com/ethereum-optimism/optimism/blob/01ddb2e6a09edf55a7cb2130e0a5b6acd0c2d2fa/op-e2e/actions/upgrades/isthmus_fork_test.go#L290)
+  2. same upgrade path as previous predeploys
+      - example: [EIP-4788 deployment transaction](https://github.com/ethereum-optimism/optimism/blob/b6b74290b5502f45daae57946f969327cdb2d383/op-node/rollup/derive/ecotone_upgrade_transactions.go#L130)
+
```
sebastianst, line 94: I believe that since we decided not to include **upgrade gas** (https://github.com/ethereum-optimism/specs/pull/616) in Isthmus, there's still the failure mode that the upgrade block runs out of gas if a (malicious) chain operator (that owns the `SystemConfig` owner key) sets the chain's block `gasLimit` to <24M. I now believe that this cannot be fully mitigated without upgrade gas.

In that case, the chain could not derive the Isthmus upgrade block, not even sequencer-less auto-derivation. We discussed this with @ajsutton in a Slack thread, the conclusion back then was that this could be mitigated by the network as a whole agreeing by consensus to just use a higher `gasLimit` parameter. However, I now believe this doesn't work when users want to exit the system. There are two cases.
1) chain still on permissioned output roots: the same malicious chain op probably also posts the output roots, so they can just stop posting/refuse posting of a chain that uses higher gasLimit . However, they can always do that, this is not special to the upgrade block running out of gas. This is a fundamental problem of permissioned output roots.
2) chain is on permissionless output roots: in this case, the on-chain FPP that would execute as part of a dispute game would not be aware of the off-chain agreed higher gasLimit but instead use the on-chain gasLimit. So executing the upgrade block in a dispute would fail too and no progress could be made.

Maybe @ajsutton can help me see where this reasoning is wrong.

Note that this problem exists since the first hardfork, so it's nothing new with Isthmus.

ajsutton: The easiest "way out" would be to disable the upgrade again and the chain would then continue without the upgrade transactions so auto-derivation would work again. For permissionless proofs to work correctly, this would require updating the fault proof prestate which the ProxyAdminOwner controls. Where the ProxyAdminOwner includes the chain operator in the multisig, then the chain operator could prevent recovery. However, it completely bricks the chain so there's no incentive for a chain operator to do this.

Ultimately we will need upgrade gas or an increase in the `systemTxMaxGas` in SystemConfig to prevent chain operators from causing this problem, but it's exceptionally low risk if we don't have it.

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,275 @@
...
+- [Action Items](#action-items)
+- [Audit Requirements](#audit-requirements)
+- [Appendix](#appendix)
+  - [Appendix A: Required Code Changes by EIP](#appendix-a-required-code-changes-by-eip)
+  - [Appendix B: Block Header Changes](#appendix-b-block-header-changes)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<!-- _Italics are used to indicate things that need to be replaced._ -->
+
```
Ethnical, line 27:
```suggestion
|  Related References       |     Links                                   |
| ----------------------- | ------------------------------------------- |
|   Operator Fee FMA        | https://github.com/succinctlabs/optimism-design-docs/blob/main/security/fma-operator-fee.md |
```

security/fma-isthmus-l2-pectra.md
sebastianst: Thanks for confirming, so ultimately, recovering from this may need cooperation from the chain operator. This is a fine assumption for permissioned chains anyway, where the chain operator is also relied on to post output roots. For permissionless chains, this depends on whether the chain op is required to update the prestate.
But I also see this as low risk as there aren't any incentives for this.

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,281 @@
...
+| EIP-7002 | Execution layer triggerable withdrawals                    | Affects the L1 block header via the `requests_hash` field (see EIP 7685).                                                                     | none                                          |
+| EIP-7251 | Increase the MAX_EFFECTIVE_BALANCE                         | Affects the L1 block header via the `requests_hash` field (see EIP 7685).                                                                     | none                                          |
+| EIP-7549 | Move committee index outside Attestation                   | None, since it only affects beacon chain.                                                                                                     | none                                          |
+| EIP-7623 | Increase calldata cost                                     | Support change in op-geth; no extra changes are required.                                                                                     | none                                          |
```
geoknee, line 235:
```suggestion
| EIP-7623 | Increase calldata cost                                     | Support change in EL clients (e.g. op-geth, op-reth); no extra changes are required.                                                                                     | none                                          |
```

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,281 @@
...
+- [ ] (NON-BLOCKING)  Make a way to enforce the gas limit of a block can't be lower of 30M (cf this [discord thread](https://discord.com/channels/1244729134312198194/1334603521454247946/1351651773755949117) with @tynes & @sebastianst) 
+- [x] (NON-BLOCKING):  Add the description of the new block header with the new field `request_hash` in this FMA (@meyer9)
+- [ ] (BLOCKING): Add a test that Validate a block with BLS accelerated precompile (@meyer9 ref to this [comment](https://github.com/ethereum-optimism/design-docs/pull/201/files#r1985623601))
+- [ ] (BLOCKING): Add a E2E test that Validate that minting to EOA that contains code (with Type4) is not an issue (@meyer9 assignee for now but could be someone else).
+- [ ]  (NON-BLOCKING): Identify path from the sequencer during the deposit transaction on the L2 that can cause unexpected behavior when an EOA has some code.
+- [ ]  (NON-BLOCKING): Indicate in this FMA reference the kurtosis devnet that allow the perform testing on Isthmus (L2 Pectra).
+- [ ]  (NON-BLOCKING): Tests to make sure that `ressources_limit` and `gas_limit` for each Superchain are not failling by overflowing from the config on L1 (assignee: @geoknee)
```
geoknee, line 206: I have performed this check manually [code here](https://github.com/ethereum-optimism/optimism/blob/371f5e08a9ff708b91af5bc7088e0238e7bf6ebf/op-node/rollup/superchain_test.go#L53-L83) and confirmed that all chains currently in the registry currently have a `gasLimit` of at least 30M and a `maxResourceLimit` of at least 20M. 

It wasn't worth committing this as a CI check, but I think we can add the link above and consider this item done.

security/fma-isthmus-l2-pectra.md
Ethnical: This seems to be too much work, and not easy to use with op-geth.optimism.io since this is showing all the diff against geth but we don't have an obvious way to get isthmus changes only...

security/fma-isthmus-l2-pectra.md
Ethnical: Thanks for both of us! 
I see that there is no incentive to do that in a malicious way. 
I was more scared that this is unintentional that would brick the chain if we are running a chain that is under 30M `GasLimit`.
@geoknee made a nice script in the CI to identify this:
```
dialing client at https://ethereum-rpc.publicnode.commaking/ batch call
finished batch call
    superchain_test.go:76: 60000000
    superchain_test.go:78: Chain: cyber-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: cyber-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: mint-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: mint-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: creator-chain-testnet-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: creator-chain-testnet-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: cyber-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: cyber-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: mode-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: mode-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: unichain-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: unichain-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: automata-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: automata-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: ethernity-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: ethernity-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: ink-mainnet GasLimit 60.0M
    superchain_test.go:80: Chain: ink-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: lisk-mainnet GasLimit 60.0M
    superchain_test.go:80: Chain: lisk-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: ethernity-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: ethernity-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: funki-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: funki-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: metal-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: metal-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: mode-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: mode-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: polynomial-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: polynomial-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: race-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: race-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: sseed-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: sseed-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: xterio-eth-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: xterio-eth-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: pivotal-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: pivotal-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: soneium-minato-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: soneium-minato-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: worldchain-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: worldchain-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: hashkeychain-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: hashkeychain-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: orderly-mainnet GasLimit 180.0M
    superchain_test.go:80: Chain: orderly-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: snax-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: snax-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: soneium-mainnet GasLimit 40.0M
    superchain_test.go:80: Chain: soneium-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: swan-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: swan-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: op-sepolia GasLimit 60.0M
    superchain_test.go:80: Chain: op-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: op-mainnet GasLimit 60.0M
    superchain_test.go:80: Chain: op-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: redstone-mainnet GasLimit 100.0M
    superchain_test.go:80: Chain: redstone-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: unichain-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: unichain-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: worldchain-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: worldchain-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: zora-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: zora-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: funki-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: funki-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: race-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: race-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: shape-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: shape-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: bob-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: bob-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: zora-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: zora-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: lyra-mainnet GasLimit 120.0M
    superchain_test.go:80: Chain: lyra-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: ink-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: ink-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: arena-z-mainnet GasLimit 60.0M
    superchain_test.go:80: Chain: arena-z-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: shape-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: shape-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: swell-mainnet GasLimit 60.0M
    superchain_test.go:80: Chain: swell-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: tbn-mainnet GasLimit 30.0M
    superchain_test.go:80: Chain: tbn-mainnet ResourceConfig 20.0M
    superchain_test.go:78: Chain: arena-z-testnet-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: arena-z-testnet-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: base-sepolia GasLimit 60.0M
    superchain_test.go:80: Chain: base-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: lisk-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: lisk-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: metal-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: metal-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: tbn-sepolia GasLimit 30.0M
    superchain_test.go:80: Chain: tbn-sepolia ResourceConfig 20.0M
    superchain_test.go:78: Chain: base-mainnet GasLimit 116.0M
    superchain_test.go:80: Chain: base-mainnet ResourceConfig 20.0M
--- PASS: TestSuperchainGasLimit (0.19s)
```

security/fma-isthmus-l2-pectra.md
Ethnical: Thanks!
How to make sure we are running this just before the isthmus upgrade to make sure no systemconfig has been updated by mistake after we just ran the script?

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,281 @@
...
+- **Recovery Path(s)**: This would require an emergency client update of the implementation with the bug (e.g. `op-reth`, `op-geth`, ...).
+
+### FM5: BLS Precompiles could cause increased FP Program execution time
+
+- **Description:** If implemented in the FP program, BLS precompiles could greatly increase computation time.
+- **Risk Assessment:** High impact, low likelihood
+- **Mitigations:**
+	1. Fully mitigated by using an accelerated precompile that calls out to L1 instead of calculating inside the program.
+        - [Specs](https://github.com/ethereum-optimism/specs/blob/b5e0fa98881171f658f782597a46b641e8f3dfd0/specs/protocol/isthmus/exec-engine.md#evm-changes)
+- **Detection:** op-program trace runner computation time would increase significantly (should trigger an alert)
```
Ethnical, line 149:
```suggestion
- **Detection:** The `vm-runner` computation time would **increase** significantly, or **panic** due to OOM and trigger an alert through our current set of monitoring. 
```

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,281 @@
...
+- [ ] (NON-BLOCKING)  Make a way to enforce the gas limit of a block can't be lower of 30M (cf this [discord thread](https://discord.com/channels/1244729134312198194/1334603521454247946/1351651773755949117) with @tynes & @sebastianst) 
+- [x] (NON-BLOCKING):  Add the description of the new block header with the new field `request_hash` in this FMA (@meyer9)
+- [ ] (BLOCKING): Add a test that Validate a block with BLS accelerated precompile (@meyer9 ref to this [comment](https://github.com/ethereum-optimism/design-docs/pull/201/files#r1985623601))
+- [ ] (BLOCKING): Add a E2E test that Validate that minting to EOA that contains code (with Type4) is not an issue (@meyer9 assignee for now but could be someone else).
+- [ ]  (NON-BLOCKING): Identify path from the sequencer during the deposit transaction on the L2 that can cause unexpected behavior when an EOA has some code.
+- [ ]  (NON-BLOCKING): Indicate in this FMA reference the kurtosis devnet that allow the perform testing on Isthmus (L2 Pectra).
+- [ ]  (NON-BLOCKING): Tests to make sure that `ressources_limit` and `gas_limit` for each Superchain are not failling by overflowing from the config on L1 (assignee: @geoknee)
```
Ethnical, line 206:
```suggestion
- [x]  (NON-BLOCKING): Tests to make sure that `ressources_limit` and `gas_limit` for each Superchain are not failling by overflowing from the config on L1 (assignee: @geoknee)
        - [ ]  (NON-BLOCKING): Execute this test in the warroom of the upgrade(assignee: anyEvm-safety engineers). 
```

security/fma-isthmus-l2-pectra.md
Ethnical: @meyer9 I don't see this test here?

security/fma-isthmus-l2-pectra.md
Ethnical: Since this would be similar to the previous Failure mode, here we move forward.

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,281 @@
...
+- [x] Check this box to confirm that these items have been considered and updated if necessary.
+
+## Action Items
+
+Below is what needs to be done before launch to reduce the chances of the above failure modes occurring, and to ensure they can be detected and recovered from:
+
+- [ ] Resolve all comments on this document and incorporate them into the document itself (Assignee: document author)
+- [ ] (NON-BLOCKING)  Make a way to enforce the gas limit of a block can't be lower of 30M (cf this [discord thread](https://discord.com/channels/1244729134312198194/1334603521454247946/1351651773755949117) with @tynes & @sebastianst) 
+- [x] (NON-BLOCKING):  Add the description of the new block header with the new field `request_hash` in this FMA (@meyer9)
+- [ ] (BLOCKING): Add a test that Validate a block with BLS accelerated precompile (@meyer9 ref to this [comment](https://github.com/ethereum-optimism/design-docs/pull/201/files#r1985623601))
```
tynes, line 202: cc @meyer9

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,281 @@
...
+  1. Spec disallows type 4 transactions before Isthmus
+	1. End-to-end test
+        - [Action test created here](https://github.com/ethereum-optimism/optimism/blob/01ddb2e6a09edf55a7cb2130e0a5b6acd0c2d2fa/op-e2e/actions/upgrades/isthmus_fork_test.go#L290)
+- **Detection:** Upgraded clients would fork causing an alert on L2 unsafe liveness.
+- **Recovery Path(s)**: Emergency patch to rollup node clients (like `op-node`).
+
+
+### FM8: Smart contracts relying on sender check no longer ensures sender does not have code
+
+- **Description:** As part of EIP-7702, `msg.sender == tx.origin` no longer implies that the caller has no code. If unaddressed, this could cause
```
tynes, line 178: This has been handled, there is a new library for handling this check that has been introduced into our L1 contracts

Ethnical: Correct the changes has been made on L1. 
Also this is fixed on the L2 the modifier `onlyEOA` is used now: 
https://github.com/ethereum-optimism/optimism/blob/9249efc6343208f69283290fc9c5c8f6e7b243f8/packages/contracts-bedrock/src/L2/L2StandardBridge.sol#L104

I would like still to know if there is any operation on the `op-node` that can potentially revert or cause unexpected behavior if interacting with an EIP7702 address such as minting, refund etc. 
But the smart contractsa are correct.

Ethnical: > we should be safe from eip-3607 from just rebasing on top of latest geth, but good call on checking that deposits work as expected, https://eips.ethereum.org/EIPS/eip-3607

From Mark, this is correct but just in case can someone point me the tests that can cover this if there is?

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,282 @@
...
+
+- [x] Check this box to confirm that these items have been considered and updated if necessary.
+
+## Action Items
+
+Below is what needs to be done before launch to reduce the chances of the above failure modes occurring, and to ensure they can be detected and recovered from:
+
+- [ ] Resolve all comments on this document and incorporate them into the document itself (Assignee: document author)
+- [x] (BLOCKING): Add a test that Validate a block with BLS accelerated precompile (@meyer9 ref to this [comment](https://github.com/ethereum-optimism/design-docs/pull/201/files#r1985623601))
+- [ ] (BLOCKING): Add a E2E test that Validate minting to EOA that contains code (through EIP7702) is not an issue (assignee: @meyer9).
```
Ethnical, line 200:
```suggestion
- [x] (BLOCKING): Add a E2E test that Validate minting to EOA that contains code (through EIP7702) is not an issue (assignee: @meyer9). [Implemented here](https://github.com/ethereum-optimism/optimism/pull/15090) 
```

Ethnical (approved):
Thanks @meyer9! and @geoknee for the big help there!
we can merge here, since all the `BLOCKING` items are finished.

security/fma-isthmus-l2-pectra.md
```diff
@@ -0,0 +1,283 @@
...
+- **Recovery Path(s)**: If this occurred, which it can't with the mitigations, we'd have to make op-program efficient enough to execute BLS precompiles in sufficient time.
+
+### FM7: Early fork if batches containing EIP-7702 transactions could be posted before Pectra
+
+- **Description:** If batches containing EIP-7702 transactions are posted and accepted early, the batcher could fork any nodes that upgrade early. A malicious batcher could cause this to happen if we don't have a check to ensure type 4 (SetCode) txs can only appear in a batch after Isthmus is active.
+- **Risk Assessment:** High impact, low likelihood
+- **Mitigations:**
+  1. Spec disallows type 4 transactions before Isthmus
+  1. End-to-end test
+     - [Action test created here](https://github.com/ethereum-optimism/optimism/blob/01ddb2e6a09edf55a7cb2130e0a5b6acd0c2d2fa/op-e2e/actions/upgrades/isthmus_fork_test.go#L290)
```
geoknee, line 171: I think we should be linking to this test here? https://github.com/ethereum-optimism/optimism/blob/develop/op-e2e/actions/proofs/isthmus_setcode_tx_test.go 

Also, I'm not sure we have coverage for the unsafe chain, i.e. a check that the txpool will filter out set code txs if Isthmus is not active.

PR #202, opened 2025-02-11
feat: simplified guardian system

smartcontracts:
Proposes a simplified Guardian role that maintains Stage 1 compatibility with the latest L2Beat Stage 1 definitions.

### DIFF
--- a/protocol/simplified-guardian.md
+++ b/protocol/simplified-guardian.md
@@ -1,0 +1,127 @@
+# Simplified Guardian System
+
+## Purpose
+
+Provides a design for a simplified Guardian role that aligns the OP Stack with the latest update to
+[L2Beat's Stage 1 definition][1].
+
+## Context
+
+L2Beat recently published [an update][1] to the Stages Framework that is scheduled to take effect
+in the middle of 2025. The OP Stack is impacted by this update. Modifications will need to be made
+to guarantee that the OP Stack continues to be compatible with the Stage 1 definition.
+
+## Problem Statement
+
+L2Beat's updated definition specifically impacts ability to use the Deputy Guardian alongside the
+default Guardian role. The Deputy Guardian conflicts with the updated Stage 1 definition because it
+allows a role *other* than the Guardian (assumed to be held by a Security Council) to execute
+unilateral actions that cannot always be easily reversed by the Guardian.
+
+Specifically, L2Beat highlights the following problematic case for OP Mainnet:
+
+1. Deputy Guardian (held by Optimism Foundation) triggers the fallback to the
+  `PermissionedDisputeGame`.
+2. Deputy Guardian convinces a blocking minority of the Security Council by some means to *not*
+  switch back to the `PermissionlessDisputeGame`, to *not* challenge any of the dispute game
+  proposals created by the permissioned Proposer role, and to not trigger the Superchain-wide pause
+  mechanism.
+3. Permissioned Proposer role publishes an invalid proposal and the Security Council fails to
+  challenge the dispute game through any of its available mechanisms.
+4. Malicious party is able to execute an invalid withdrawal based on the invalid proposal.
+
+L2Beat's updated Stage 1 definition, once it goes into effect in the middle of 2025, would define a
+system that *allows* this situation as a Stage 0 system. We must therefore make adjustments to the
+incident response process so that OP Mainnet and other similarly configured systems remain Stage 1
+when this definition goes into effect.
+
+## Proposed Solution
+
+### Remove the Deputy Guardian
+
+We begin by removing the Deputy Guardian role entirely. The Guardian role would be held exclusively
+by the 1/1 Security Council Operations Safe (held by the Security Council). All Guardian actions
+would therefore only be executable by a majority of the Security Council. This eliminates the
+intial problem posed by the updated Stage 1 definition.
+
+### Deputy Pause Module v2
+
+We still want the Optimism Foundation to retain the power to quickly respond to potential
+incidents. We would therefore install a modified version of the Deputy Pause Module into the
+Security Council Operations Safe and permit the Optimism Foundation Safe (or whatever party
+previously held the Deputy Guardian role for that system prior to this change) to manage the Pause
+Deputy account. This permits the Pause Deputy to trigger the pause mechanism freely.
+
+### Pause Expiry
+
+Because the Pause Deputy could trigger the pause mechanism at any time and the unpause function
+would require participation from a majority of the Security Council, we must either (1) allow a
+minority of the Security Council to unpause the system or (2) have the pause automatically expire
+after some period of time. Given the potential security concerns with allowing a minority of the
+Security Council to unpause, we prefer option (2).
+
+We therefore propose that the pause would expire after some period of time to be defined in the
+specification for these changes (on the order of several months). Once expired, the pause would
+have a cooldown period (to be specified later) to prevent anyone with access to the pause from
+repeatedly pausing as soon as the previous pause expires. A majority of the Security Council would
+be able to reset the expiry clock or explicitly unpause the system. An explicit unpause action
+(rather than an expiration) would  similarly reset the cooldown such that the pause can again be
+executed immediately.
+
+### Cluster-wide Pause
+
+Because the pause mechanism would become the main mechanism available to the Optimism Foundation,
+the pause mechanism would also become the basis for an updated incident response process. However,
+the Superchain-wide Pause is far too impactful to be the only available option during an incident.
+Given that we suspect many incidents will take the form of *single* impacted chains or chain
+clusters, it may be valuable to implement a Cluster-wide Pause on top of the Superchain-wide Pause.
+A chain would be paused if *either* the Superchain-wide Pause is active or its Cluster-wide Pause
+is active.
+
+## Considerations
+
+### Cluser-wide Pause Complexity
+
+It should be noted that the Cluster-wide Pause is likely the most significant source of complexity
+in this proposal. We have a number of different options for how to implement this. Here's a
+non-exhaustive list:
+
+- Cluster ID inside of `SystemConfig` with Cluster-wide Pause held inside of `SuperchainConfig`.
+- New `ClusterConfig` contract.
+- Pause within `SystemConfig` and bundle pauses for an entire cluster into single transaction.
+- Only use Superchain-wide pause.
+
+Given that for the foreseeable future we expect to either (1) pause individual chains or (2) pause
+the primary interop cluster and (3) that an issue impacting the entire primary interop cluster
+probably requires a Superchain-wide Pause anyway, it may just be simpler to implement an additional
+pause in the `SystemConfig` contract instead.
+
+### Pause Expiry Security Implications
+
+Pause Expiry has non-trivial security implications in that a bug in the proof system alongside a 
+malicious blocking minority of the Security Council could be catastrophic. L2Beat explciitly argues
+that this security model is better than the model in which a malicious Foundation and a blocking
+minority of the Security Model is catastrophic. Since this is fundamental to the updated Stage 1
+definition, we are required to accept this conclusion. We may want to take this as motivation to
+implement some other proof system as a backup to the primary proof system.
+
+### Permissioned Proof System
+
+A chain will be considered Stage 0 if it falls back to the Permissioned Proof System. We should
+accept this potential for a downgrade and generally rely on the pause mechanism as the primary
+incident response tool.
+
+We propose the following incident response protocol:
+
+- Always attempt to resolve an issue offchain if possible, before a game resolves incorrectly.
+- First reaction when a dispute game resolves incorrectly is to execute the Superchain-wide Pause.
+- We immediately determine if we believe that the bug can be resolved within 2 weeks.
+- If the bug can be resolved within 2 weeks (via Security Council emergency upgrade) then we
+  keep the pause active until the bug is resolved, upgrade the game, blacklist all invalid games or
+  retire all existing games, and unpause.
+- If the bug cannot be resolved within 2 weeks or if 2 weeks have actually passed since the pause
+  was executed then we switch to the Permissioned Proof System, blacklist all invalid games or
+  retire all existing games, and unpause.
+
+<!-- References -->
+[1]: https://forum.l2beat.com/t/stages-update-a-high-level-guiding-principle-for-stage-1/338


protocol/simplified-guardian.md
```diff
@@ -0,0 +1,127 @@
...
+would therefore only be executable by a majority of the Security Council. This eliminates the
+intial problem posed by the updated Stage 1 definition.
+
+### Deputy Pause Module v2
+
+We still want the Optimism Foundation to retain the power to quickly respond to potential
+incidents. We would therefore install a modified version of the Deputy Pause Module into the
+Security Council Operations Safe and permit the Optimism Foundation Safe (or whatever party
+previously held the Deputy Guardian role for that system prior to this change) to manage the Pause
+Deputy account. This permits the Pause Deputy to trigger the pause mechanism freely.
```
maurelian, line 53: The path to this should be as simple as replacing the DGM with the DPM.

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Pause Expiry
+
+Because the Pause Deputy could trigger the pause mechanism at any time and the unpause function
+would require participation from a majority of the Security Council, we must either (1) allow a
```
maurelian, line 58: What is the source of the requirement to 'allow a minority of the SC to unpause'?

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Cluster-wide Pause
+
+Because the pause mechanism would become the main mechanism available to the Optimism Foundation,
+the pause mechanism would also become the basis for an updated incident response process. However,
+the Superchain-wide Pause is far too impactful to be the only available option during an incident.
+Given that we suspect many incidents will take the form of *single* impacted chains or chain
+clusters, it may be valuable to implement a Cluster-wide Pause on top of the Superchain-wide Pause.
+A chain would be paused if *either* the Superchain-wide Pause is active or its Cluster-wide Pause
+is active.
```
maurelian, line 79: Does this require a new ClusterConfig or ClusterPauseManager contract?

maurelian: nvm, I see it below

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,127 @@
...
+## Considerations
+
+### Cluser-wide Pause Complexity
+
+It should be noted that the Cluster-wide Pause is likely the most significant source of complexity
+in this proposal. We have a number of different options for how to implement this. Here's a
+non-exhaustive list:
+
+- Cluster ID inside of `SystemConfig` with Cluster-wide Pause held inside of `SuperchainConfig`.
+- New `ClusterConfig` contract.
```
maurelian, line 90: I strongly dislike this option, as it creates a whole new class of contract.

protocol/simplified-guardian.md
maurelian: Wrong, the DPM assumes it's calling the DGM, so it needs changes.

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,127 @@
...
+that this security model is better than the model in which a malicious Foundation and a blocking
+minority of the Security Model is catastrophic. Since this is fundamental to the updated Stage 1
+definition, we are required to accept this conclusion. We may want to take this as motivation to
+implement some other proof system as a backup to the primary proof system.
+
+### Permissioned Proof System
+
+A chain will be considered Stage 0 if it falls back to the Permissioned Proof System. We should
+accept this potential for a downgrade and generally rely on the pause mechanism as the primary
+incident response tool.
```
maurelian, line 112: agree with this. I'd frame the overall proposal as "making the pause more granular so that using it is more palatable".

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,127 @@
...
+### Pause Expiry Security Implications
+
+Pause Expiry has non-trivial security implications in that a bug in the proof system alongside a 
+malicious blocking minority of the Security Council could be catastrophic. L2Beat explciitly argues
+that this security model is better than the model in which a malicious Foundation and a blocking
+minority of the Security Model is catastrophic. Since this is fundamental to the updated Stage 1
+definition, we are required to accept this conclusion. We may want to take this as motivation to
+implement some other proof system as a backup to the primary proof system.
+
+### Permissioned Proof System
```
smartcontracts, line 108: Need to do product research on the pause mechanism itself

mds1: Reference from @ControlCplusControlV: https://docs.lido.fi/contracts/gate-seal/

protocol/simplified-guardian.md
smartcontracts: Either DPM needs changes or we can simply change the DGM and leave the rest of the system untouched. I'm ok with either approach.

protocol/simplified-guardian.md
smartcontracts: L2Beat Stage 1 requirement that FND + blocking minority cannot cause a permanent liveness failure.

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,127 @@
...
+
+Because the Pause Deputy could trigger the pause mechanism at any time and the unpause function
+would require participation from a majority of the Security Council, we must either (1) allow a
+minority of the Security Council to unpause the system or (2) have the pause automatically expire
+after some period of time. Given the potential security concerns with allowing a minority of the
+Security Council to unpause, we prefer option (2).
+
+We therefore propose that the pause would expire after some period of time to be defined in the
+specification for these changes (on the order of several months). Once expired, the pause would
+have a cooldown period (to be specified later) to prevent anyone with access to the pause from
```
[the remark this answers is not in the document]
smartcontracts: Oh yep, correct. Will fix.

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,127 @@
...
+### Permissioned Proof System
+
+A chain will be considered Stage 0 if it falls back to the Permissioned Proof System. We should
+accept this potential for a downgrade and generally rely on the pause mechanism as the primary
+incident response tool.
+
+We propose the following incident response protocol:
+
+- Always attempt to resolve an issue offchain if possible, before a game resolves incorrectly.
+- First reaction when a dispute game resolves incorrectly is to execute the Superchain-wide Pause.
```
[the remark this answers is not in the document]
smartcontracts: Ah good catch, will update

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,118 @@
...
+incident response process so that OP Mainnet and other similarly configured systems remain Stage 1
+when this definition goes into effect.
+
+## Proposed Solution
+
+### Remove the Deputy Guardian
+
+We begin by removing the Deputy Guardian role entirely. The Guardian role would be held exclusively
+by the 1/1 Security Council Operations Safe (held by the Security Council). All Guardian actions
+would therefore only be executable by a majority of the Security Council. This eliminates the
```
maurelian, line 44: This statement is ambiguous about its implications:

> All Guardian actions would therefore only be executable by a majority of the Security Council. 

ie. given that the `pause()` will still be executable by the Foundation this implies that the pause is no longer a guardian action. I don't think that's the intent so: 

 ```suggestion
by the 1/1 Security Council Operations Safe (held by the Security Council). All Guardian actions other than `pause()`
would therefore only be executable by a majority of the Security Council. This eliminates the
```

File: protocol/simplified-guardian.md [hunk 10]
```diff
@@ -0,0 +1,118 @@
...
+
+A chain would be paused if *either* the Superchain-wide Pause is active or its Cluster-wide Pause
+is active. We would implement the Cluster-wide Pause by allowing the Guardian to pause a specific
+chain/cluster as identified by a `SystemConfig` contract address. That is, the Guardian would call
+`SuperchainConfig.pause(someSystemConfigAddress)`. Other contracts would then check the status of
+the pause by calling `SuperchainConfig.paused(mySystemConfigAddress)` which would return true if
+either that specific system is paused or the entire Superchain is paused.
+
+The interop cluster can use a "fake" shared system config address in the call to `paused(..)` so
+that the entire cluster pauses at the same time.
```
maurelian, line 83: from this snippet, I gather that each OP Chain will need to store `mySystemConfigAddress` somewhere, and that having the same value defines a cluster? 

How will this be kept in sync if a chain moves into a cluster from having been on its own?

Would it be better to use the lockbox address for this, since it is already intended to eventually become a singleton per cluster?

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Permissioned Proof System
+
+A chain will be considered Stage 0 if it falls back to the Permissioned Proof System. We should
+accept this potential for a downgrade and generally rely on the pause mechanism as the primary
+incident response tool.
+
+We propose the following incident response protocol:
+
+- Always attempt to resolve an issue offchain if possible, before a game resolves incorrectly.
```
maurelian, line 104: what does that mean?

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,118 @@
...
+
+We begin by removing the Deputy Guardian role entirely. The Guardian role would be held exclusively
+by the 1/1 Security Council Operations Safe (held by the Security Council). All Guardian actions
+would therefore only be executable by a majority of the Security Council. This eliminates the
+intial problem posed by the updated Stage 1 definition.
+
+### Deputy Pause Module v2
+
+We still want the Optimism Foundation to retain the power to quickly respond to potential
+incidents. We would therefore install a modified version of the Deputy Pause Module into the
```
maurelian, line 50: what are the mods?

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,118 @@
...
+A chain will be considered Stage 0 if it falls back to the Permissioned Proof System. We should
+accept this potential for a downgrade and generally rely on the pause mechanism as the primary
+incident response tool.
+
+We propose the following incident response protocol:
+
+- Always attempt to resolve an issue offchain if possible, before a game resolves incorrectly.
+- Determine if the issue is chain-specific (e.g., missing honest actor) or applies to all chains
+  (e.g., contract bug in dispute game).
+  - If chain-specific, execute the chain-specific pause.
```
maurelian, line 107: Chain and/or cluster?

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,118 @@
...
+  - If applies to all chains, execute the Superchain-wide pause.
+- We immediately determine if we believe that the bug can be resolved within 2 weeks.
+- If the bug can be resolved within 2 weeks (via Security Council emergency upgrade) then we
+  keep the pause active until the bug is resolved, upgrade the game, blacklist all invalid games or
+  retire all existing games, and unpause.
```
maurelian, line 112: generally I think retiring all should be the preferred move to mitigate the risk of missing an invalid game.

protocol/simplified-guardian.md
smartcontracts: Many times when you get an alert for an incident, the issue is with some offchain infrastructure and not a bug with a smart contract. If we can resolve the issue offchain before a game resolves incorrectly, do so. Somewhat self-evident to us internally but useful to make explicit about when we would and would not use a pause.

protocol/simplified-guardian.md
smartcontracts: Specifically, we would remove the reference to the `DeputyGuardianModule` and simply install this directly into the Security Council Guardian Safe. We can alternatively just modify the `DeputyGuardianModule` to remove all functions except for `pause()`. This second option is smaller diff, but leaves extra contracts in place. I'm OK with either option.

protocol/simplified-guardian.md
smartcontracts: Yes, correct. Will update.

protocol/simplified-guardian.md
maurelian: I prefer to remove the DGM and have fewer contracts to keep track of.

protocol/simplified-guardian.md
maurelian: I'm on board with that general direction. 
Since there is already a reference to the SuperchainConfig in the OptimismPortal, it would probably be easier and less diff to do: 

2. Contracts check `OptimismPortal.paused()` which passes the lockbox address as the pause identifier to `SuperchainConfig.paused(identifier)`.

protocol/simplified-guardian.md
maurelian:
```suggestion
- Always attempt to resolve an issue offchain if possible, before a game resolves incorrectly. 
  This would typically occur as a result of an issue in offchain infrastructure, rather than a bug in the contracts.
```

protocol/simplified-guardian.md
smartcontracts: I think it's cleaner if the SystemConfig is the source of truth here instead of the OptimismPortal. I think the diff will look the same either way.

maurelian (approved):

protocol/simplified-guardian.md
mds1: Did we ever do this research?

protocol/simplified-guardian.md
```diff
@@ -0,0 +1,121 @@
...
+  (e.g., contract bug in dispute game).
+  - If chain-specific or cluster-specific, execute the chain-specific or cluster-specific pause.
+  - If applies to all chains, execute the Superchain-wide pause.
+- We immediately determine if we believe that the bug can be resolved within 2 weeks.
+- If the bug can be resolved within 2 weeks (via Security Council emergency upgrade) then we
+  keep the pause active until the bug is resolved, upgrade the game, or retire all existing games,
+  and unpause.
+- If the bug cannot be resolved within 2 weeks or if 2 weeks have actually passed since the pause
+  was executed then we switch to the Permissioned Proof System, or retire all existing games, and
+  unpause.
```
mds1, line 118: Do we actually need to keep the permissioned game? Given that it's existence was the motivation for a lot of the design, can we simplify the design at all by removing permissioned games entirely? The result would just be the last bullet here is gone, and we maintain the pause for longer until the issue can be resolved

ajsutton: Yes I would say we need to keep the permissioned game. If there were an issue found in the way the dispute game itself worked (vs just a bug in the implementation of it), it could take us a significant amount of time to design, implement and test a fix for it.  Having withdrawals completely disabled during that time would incur massive reputational damage and the time pressure to develop the fix would increase the risk of the fix being insufficient or introducing other problems.
I'd quite like to have time to simplify the permissioned game so it isn't a full dispute game implementation though.

PR #203, opened 2025-02-19
feat: crosschain ERC20 (ERC-7802 and ERC-7281)

hexshire:
**Description**

This design document aims to provide a clear implementation path for token issuers who want to make their tokens interoperable across both the Superchain and other ecosystems, while maintaining compatibility with existing bridge infrastructure.

Contracts naming is open to discussion since the idea is to open these implementations to be compatible with any bridge that uses ERC-7802 or ERC-7281.

### DIFF
--- a/protocol/crosschain-erc20.md
+++ b/protocol/crosschain-erc20.md
@@ -4,7 +4,7 @@ This design document aims to provide a clear implementation path for token issue
 
 # **Summary**
 
-xSuperchainERC20 is a token implementation that combines xERC20 ([ERC-7281](https://ethereum-magicians.org/t/erc-7281-sovereign-bridged-tokens/14979)) and SuperchainERC20 ([ERC-7802](https://ethereum-magicians.org/t/erc-7802-crosschain-token-interface/21508)) functionality. This allows tokens to be immediately usable with existing bridge infrastructure while being compatible with the Superchain interop cluster.
+CrosschainERC20 is a token implementation that combines xERC20 ([ERC-7281](https://ethereum-magicians.org/t/erc-7281-sovereign-bridged-tokens/14979)) and SuperchainERC20 ([ERC-7802](https://ethereum-magicians.org/t/erc-7802-crosschain-token-interface/21508)) functionality. This allows tokens to be immediately usable with existing bridge infrastructure while being compatible with the Superchain interop cluster.
 
 # **Problem Statement + Context**
 
@@ -19,7 +19,7 @@ If the asset issuer wants to make their token cross-chain interoperable today or
 
 ## **A. Non-deployed or upgradable tokens**
 
-For tokens that have not yet been deployed, or can be upgraded, we recommend implementing xSuperchainERC20 - a token that combines both ERC-7281 (xERC20) and ERC-7802 standards. This solution is ideal when:
+For tokens that have not yet been deployed, or can be upgraded, we recommend implementing CrosschainERC20 - a token that combines both ERC-7281 (xERC20) and ERC-7802 standards. This solution is ideal when:
 
 - You want your token to work both within the Superchain (via ERC-7802)
 - AND outside the Superchain using third-party bridges (via ERC-7281)
@@ -33,10 +33,10 @@ This approach provides maximum flexibility as:
 
 For tokens that are already deployed and cannot be upgraded, we propose using a Lockbox mechanism:
 
-1. Deploy a xSuperchainERC20 contract that will be the wrapped version of the original token
+1. Deploy a CrosschainERC20 contract that will be the wrapped version of the original token
 2. Deploy a Lockbox contract that:
    - Locks the original ERC20 tokens
-   - Mints xSuperchainERC20 tokens at a 1:1 ratio
+   - Mints CrosschainERC20 tokens at a 1:1 ratio
    - Can be deployed deterministically to have the same address across all chains
 
 Key benefits:
@@ -128,11 +128,11 @@ sequenceDiagram
   P7 -->> P7: RelayERC20()
 ```
 
-## xSuperchainERC20Factory
+## CrosschainERC20Factory
 
 The Factory contract serves as a central deployment mechanism for all the components in our system. It provides methods to:
 
-1. Deploy new xSuperchainERC20 tokens and grant mint/burn permissions to the SuperchainTokenBridge
+1. Deploy new CrosschainERC20 tokens and grant mint/burn permissions to the SuperchainTokenBridge
 2. Deploy Lockboxes for existing tokens that support ERC20 interfaces
 3. Deploy Adapters for existing xERC20 tokens
 
@@ -147,11 +147,11 @@ The Factory contract serves as a central deployment mechanism for all the compon
 
 ## Alternatives Considered for Scenario C
 
-_Worth mentioning that the Lockbox solution described in section B is also compatible with this scenario, providing an alternative option if the token issuer prefers to wrap their xERC20 into a new xSuperchainERC20 token_
+_Worth mentioning that the Lockbox solution described in section B is also compatible with this scenario, providing an alternative option if the token issuer prefers to wrap their xERC20 into a new CrosschainERC20 token_
 
-### **xSuperchainERC20Converter**
+### **CrosschainERC20Converter**
 
-It is a xSuperchainERC20 contract that can mint/burn xERC20 tokens. Similar to the Lockbox mechanism, but instead of locking tokens, it performs a direct conversion.
+It is a CrosschainERC20 contract that can mint/burn xERC20 tokens. Similar to the Lockbox mechanism, but instead of locking tokens, it performs a direct conversion.
 
 _It would be good to add a `pause` modifier on the convert functions to force the liquidity to be moved to one direction._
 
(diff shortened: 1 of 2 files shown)


protocol/crosschain-erc20.md
```diff
@@ -0,0 +1,284 @@
...
+
+## **B. Deployed, non-upgradable tokens (Non-xERC20)**
+
+For tokens that are already deployed and cannot be upgraded, we propose using a Lockbox mechanism:
+
+1. Deploy a CrosschainERC20 contract that will be the wrapped version of the original token
+2. Deploy a Lockbox contract that:
+   - Locks the original ERC20 tokens
+   - Mints CrosschainERC20 tokens at a 1:1 ratio
+   - Can be deployed deterministically to have the same address across all chains
```
zainbacchus, line 33: To clarify, the minted CrosschainERC20 would have a deterministic address across all chains right? Not the Lockbox. The Lockbox would only exist on the hub chain where the token was originally deployed

zainbacchus:
Thanks! Dropped some questions on Option C but otherwise LGTM. 

Strong agree this should not be a predeploy but available similar to the work done for opUSDC: https://github.com/defi-wonderland/opUSDC

protocol/crosschain-erc20.md
hexshire: Zain, you are right on this one. Fixed on 9673db84b660afd5abdaca93fd4fa73b6cb98bad

protocol/crosschain-erc20.md
```diff
@@ -0,0 +1,284 @@
...
+function crosschainMint(address _to, uint256 _amount) onlyBridge external {
+    XERC20.mint(_to, _amount);
+
+    emit CrosschainMint(_to, _amount, msg.sender);
+}
+```
+
+Key benefits:
+
+- The most simple/straightforward solution, token do not need to have the same address.
```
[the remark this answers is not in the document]
hexshire: In this case, the SuperchainTokenBridge is going to call the Adapter (then the Adapter calls the xERC20 mint/burn) and not the token directly. So, we need to be sure the adapter has the same address on every chain. I added more details on this section description. 9673db84b660afd5abdaca93fd4fa73b6cb98bad

protocol/crosschain-erc20.md
dianakocsis: What if the non xERC20 token was already bridged to another chain by a different bridging method? Would you then need the Lockbox on both chains, not just the hub chain?

protocol/crosschain-erc20.md
```diff
@@ -0,0 +1,277 @@
...
+- Maintains the original token's supply while enabling Superchain functionality
+- The deterministic deployment ensures the wrapped token has the same address on all chains
+- Enables usage of the `SuperchainTokenBridge` since the wrapped token's address is consistent across the network
+- Users can always unwrap their tokens to return to the original `ERC20`
+
+## **C. Deployed, non-upgradable `xERC20`**
+
+For tokens already deployed as non-upgradable `xERC20`, upgrading to implement ERC-7802 is not an option. To integrate with the Superchain while maintaining existing bridge infrastructure, an `ERC7802Adapter` can be used. This contract implements `ERC-7802`’s crosschainBurn and crosschainMint functions, translating them into the corresponding `xERC20` mint/burn operations.
+
+Since bridges will interact with the adapter instead of the token contract, the adapter must be deployed deterministically to ensure the same address across all chains.
```
dianakocsis, line 45: So tokens do not need to be deterministically deployed with the adaptor? How is the total supply of a token across all the chains calculated? Does this make it harder?

protocol/crosschain-erc20.md
```diff
@@ -0,0 +1,277 @@
...
+
+1. Within the Superchain: Uses the SuperchainTokenBridge for fast (1-block) and secure transfers
+2. Outside the Superchain: Allows configuration of any EVM-compatible third-party bridge through xERC20's mint/burn permissions
+
+## **B. Deployed, non-upgradable tokens (Non-xERC20)**
+
+For tokens that are already deployed and cannot be upgraded, use a Lockbox mechanism:
+
+1. Deploy a `CrosschainERC20` contract that will be the wrapped version of the original token
+2. Deploy a `Lockbox` contract that:
```
dianakocsis, line 30: is this one Lockbox per token?

protocol/crosschain-erc20.md
dianakocsis: if the xerc20 contract is passed into the constructor of the adaptor, wouldn't the xerc20 address still need to be the same across all chains in order for the adaptor address to be the same across all chains?

dianakocsis: oh I see the factory uses CREATE3

protocol/crosschain-erc20.md
dianakocsis: same with the adaptor? one adaptor per xerc20 per chain?

protocol/crosschain-erc20.md
```diff
@@ -0,0 +1,277 @@
...
+  P6 ->> P5: mint()
+```
+
+### **SuperchainXERC20Bridge**
+
+This approach introduces a new predeploy contract that acts as a bridge between `xERC20` tokens across the Superchain. The implementation requires:
+
+Pre-requisite:
+
+- `xERC20` addresses MUST be the same in both chains
```
dianakocsis, line 217: how common is it for xERC20 addresses to be the same?

protocol/crosschain-erc20.md
```diff
@@ -0,0 +1,277 @@
...
+- Permission management for deployments
+- Single entry point for all deployments
+
+---
+
+## Alternatives Considered for Scenario C
+
+_Worth mentioning that the Lockbox solution described in section B is also compatible with this scenario, providing an alternative option if the token issuer prefers to wrap their `xERC20` into a new `CrosschainERC20` token_
+
+### **CrosschainERC20Converter**
```
dianakocsis, line 135: Not sure I understand this one: so the CrosschainERC20Converter contract is a CrosschainERC20 contract? What does the `undo` call do?

protocol/crosschain-erc20.md
```diff
@@ -0,0 +1,277 @@
...
+
+- The token MUST work both within the Superchain or any bridge that implements `ERC-7802`.
+- The token MUST work outside the Superchain using third-party bridges via `ERC-7281`.
+
+This approach provides maximum flexibility as:
+
+1. Within the Superchain: Uses the SuperchainTokenBridge for fast (1-block) and secure transfers
+2. Outside the Superchain: Allows configuration of any EVM-compatible third-party bridge through xERC20's mint/burn permissions
+
+## **B. Deployed, non-upgradable tokens (Non-xERC20)**
```
dianakocsis, line 25: would this path also be taken for ERC7802 tokens that are non-upgradeable and non-xerc20?

protocol/crosschain-erc20.md
hexshire: Ideally, the Lockbox should be on a single chain or hub chain to avoid liquidity issues with the Lockbox. If there are different methods to bridge the tokens, this could cause a situation where, on a given chain, there are not enough funds in the Lockbox to unlock the original ERC20. In this case, the funds should be returned using the bridge initially used and locked in the hub chain.

protocol/crosschain-erc20.md
hexshire: > So tokens do not need to be deterministically deployed with the adaptor? How is the total supply of a token across all the chains calculated? Does this make it harder?

In this case, the bridge (or SuperchainTokenBridge) calls the Adapter and it calls the Token contract, the Adapter does not hold any token. For this reason, it is not necessary for the Tokens to have the same address, but the Adapters must.
To calculate liquidity: If the tokens have the same address across all chains, the usual methods would be used. However, if the tokens do not have the same address, the variable in the adapter that stores the adapted token's address could be used to redirect the calculation.

protocol/crosschain-erc20.md
hexshire: It is one lockbox per token, but on the adapter side, one adapter is deployed per bridge per token. The purpose of having one adapter per bridge is that limits are unique for that bridge. Having an adapter with both A and B bridges means that if bridge A is hacked, it can use the limits on both A and B.

protocol/crosschain-erc20.md
hexshire: In this case, we are requiring as a prerequisite that the token be deployed at the same address. If this is not the case, the token issuer would need to resolve this situation, as it is outside the scope of this solution.

protocol/crosschain-erc20.md
hexshire: Using the lockbox to move from SuperchainERC20 to CrosschainERC20 is valid.

protocol/crosschain-erc20.md
hexshire: Exactly. Here is an example of the implementation, being `BRIDGE` immutable. 

```solidity
  function crosschainMint(address _to, uint256 _amount) external {
      if (msg.sender != BRIDGE) revert Unauthorized();

      XERC20.mint(_to, _amount);

      emit CrosschainMint(_to, _amount, msg.sender);
  }
 ```

hexshire: It is worth noting that multiple adapters can be deployed per xERC20. A different limit can then be set for each adapter, effectively assigning a distinct limit to each bridge.

hexshire:
The relevant information for this development can now be found in the following README.

https://github.com/defi-wonderland/crosschainERC20/blob/dev/README.md

PR #205, opened 2025-02-19
feat: super root withdrawal proofs

smartcontracts:
Introduces a proposal to include super root validation logic inside of the base optimism portal contract.

### DIFF
--- a/protocol/proofs/generic-withdrawal-proofs.md
+++ b/protocol/proofs/generic-withdrawal-proofs.md
@@ -29,11 +29,11 @@ introducing the code to the `OptimismPortalInterop` portal. Doing so means that 
 extra code on the `OptimismPortal` that is technically unnecessary until interop ships but also
 means that we do not need to ship any additional changes to the `OptimismPortal` before interop.
 
-We could expose a new public variable `usingSuperRoots` that would toggle this function on and off
+We could expose a new public variable `superRootsActive` that would toggle this function on and off
 and would also act as a signal for client-side tooling to automatically switch between the two
-available functions. When `usingSuperRoots` is `false`, the version of `proveWithdrawalTransaction`
-for Super Roots would revert. When `usingSuperRoots` is `true`, the legacy version of the function
-would revert.
+available functions. When `superRootsActive` is `false`, the version of
+`proveWithdrawalTransaction` for Super Roots would revert. When `superRootsActive` is `true`, the
+legacy version of the function would revert.
 
 ## Alternatives Considered
 
@@ -70,13 +70,23 @@ We have decided that we will NOT allow the `OptimismPortal` for one chain to pro
 from another chain when interop launches because of security implications with the correctness of
 the cross-domain message sender. The Super Root withdrawal proof must therefore guarantee that a
 given withdrawal was actually created on the specific chain that corresponds to the given
-`OptimismPortal`. This means that we need to include a chain ID in the `OptimismPortal`.
+`OptimismPortal`.
+
+This means that we need to include a chain ID in the `OptimismPortal`. Instead of adding this into
+the `OptimismPortal` itself, we will add this into the `SystemConfig` and have the portal read the
+chain ID from the portal.
 
 ### Portal Code Differences
 
 We have a strong preference that all chains use the Super Root proof system within a relatively
 short time of the interop launch. We should have a clear plan to shipping this proof system to all
 chains after interop goes live.
+
+### Early Upgrade
+
+If any chain were to use this code *before* switching to Super Roots, user proofs would not
+function properly. We would not expect any safety concerns, but this woudl cause issues for the
+impacted chain and require an urgent upgrade.
 
 ## Failure Modes
 
--- a/protocol/proofs/generic-withdrawal-proofs.md
+++ b/protocol/proofs/generic-withdrawal-proofs.md
@@ -64,6 +64,14 @@ we can blackbox the Fault Dispute Game. We also need to audit that the Super Roo
 used unless it has been activated and that the inclusion of both functions is not dangerous for any
 reason.
 
+### Cross-Chain Withdrawals
+
+We have decided that we will NOT allow the `OptimismPortal` for one chain to process withdrawals
+from another chain when interop launches because of security implications with the correctness of
+the cross-domain message sender. The Super Root withdrawal proof must therefore guarantee that a
+given withdrawal was actually created on the specific chain that corresponds to the given
+`OptimismPortal`. This means that we need to include a chain ID in the `OptimismPortal`.
+
 ## Failure Modes
 
 ### Bug in Proof Validation Logic
--- a/protocol/proofs/generic-withdrawal-proofs.md
+++ b/protocol/proofs/generic-withdrawal-proofs.md
@@ -72,6 +72,12 @@ the cross-domain message sender. The Super Root withdrawal proof must therefore 
 given withdrawal was actually created on the specific chain that corresponds to the given
 `OptimismPortal`. This means that we need to include a chain ID in the `OptimismPortal`.
 
+### Portal Code Differences
+
+We have a strong preference that all chains use the Super Root proof system within a relatively
+short time of the interop launch. We should have a clear plan to shipping this proof system to all
+chains after interop goes live.
+
 ## Failure Modes
 
 ### Bug in Proof Validation Logic
--- a/protocol/proofs/generic-withdrawal-proofs.md
+++ b/protocol/proofs/generic-withdrawal-proofs.md
@@ -35,6 +35,12 @@ available functions. When `superRootsActive` is `false`, the version of
 `proveWithdrawalTransaction` for Super Roots would revert. When `superRootsActive` is `true`, the
 legacy version of the function would revert.
 
+At the time of interop launch, chains in the interop set would switch to the Super Root versions of
+the dispute game and set `superRootsActive` to `true`. Chains not in the interop set would still
+have `superRootsActive` set to `false`. We would then switch these chains to a single-chain Super
+Root proof and either set `superRootsActive` to `true` or remove the old code path at the same
+time.
+
 ## Alternatives Considered
 
 ### Using an Inherited Contract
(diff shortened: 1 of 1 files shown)


protocol/proofs/generic-withdrawal-proofs.md
```diff
@@ -0,0 +1,74 @@
...
+We propose to allow users to prove against Super Roots by adding a new overload of
+`proveWithdrawalTransaction` that allows a user to prove the existence of an Output Root inside of
+the Super Root before performing the standard withdrawal validation inside of the Output Root.
+
+We also specifically propose that this be implemented on the *base* `OptimismPortal` instead of
+introducing the code to the `OptimismPortalInterop` portal. Doing so means that we are including
+extra code on the `OptimismPortal` that is technically unnecessary until interop ships but also
+means that we do not need to ship any additional changes to the `OptimismPortal` before interop.
+
+We could expose a new public variable `usingSuperRoots` that would toggle this function on and off
```
smartcontracts, line 32: I strongly suggest that we do this because (1) it gates the correct function, which is a safety feature and (2) acts as a clear signal for offchain tooling for which proof version to use.

smartcontracts:
Review notes
- Include sequencing of events in the design doc
- Maybe consider switching methods based on game type

clabby (approved):
I agree with this proposal, and it should still be able to use all of the logic from [my PR](https://github.com/ethereum-optimism/optimism/pull/14349).

Good catch on including the chain ID in the portal.

I'd still like for us to explore using the game type as the activation method for the proof verification functionality. It feels much nicer for this to be implicitly activated, rather than having to switch the game type + explicitly activate super root verification. We never ran into this need because we've never switched the format of the output root before, but being able to have games declare their withdrawal inclusion proof verification function is beneficial.

If we don't want to do that, this should still be fine, and we can always upgrade to that later.

protocol/proofs/generic-withdrawal-proofs.md
```diff
@@ -0,0 +1,88 @@
...
+and would also act as a signal for client-side tooling to automatically switch between the two
+available functions. When `usingSuperRoots` is `false`, the version of `proveWithdrawalTransaction`
+for Super Roots would revert. When `usingSuperRoots` is `true`, the legacy version of the function
+would revert.
+
```
maurelian, line 37: We talked about this in the call, I'm not sure if it a decision was reached, but I do think it would reduce the amount of config air that is possible if we were to use the respected game type to set the value of usingSuperRoots. 

Other OK alternatives would be: 
1. a check in both `setUsingSuperRoots()` and `setRespectedGameType()` to ensure the values don't disagree.
2. adding a check to the StandardValidator contract

protocol/proofs/generic-withdrawal-proofs.md
```diff
@@ -0,0 +1,88 @@
...
+used unless it has been activated and that the inclusion of both functions is not dangerous for any
+reason.
+
+### Cross-Chain Withdrawals
+
+We have decided that we will NOT allow the `OptimismPortal` for one chain to process withdrawals
+from another chain when interop launches because of security implications with the correctness of
+the cross-domain message sender. The Super Root withdrawal proof must therefore guarantee that a
+given withdrawal was actually created on the specific chain that corresponds to the given
+`OptimismPortal`. This means that we need to include a chain ID in the `OptimismPortal`.
```
maurelian, line 73: I'd prefer to put it into the SystemConfig and have it read from there.

protocol/proofs/generic-withdrawal-proofs.md
```diff
@@ -0,0 +1,88 @@
...
+
+We could implement this function inside of an inherited contract instead of putting it in the base
+contract. However, this would mean performing two sets of upgrades to the `OptimismPortal` before
+interop ships on an already tight timeline.
+
+## Considerations
+
+### Unused Code in Production
+
+This proposal means we are shipping code that cannot yet be used into production. We believe this
```
ControlCplusControlV, line 50: Definitely worth stating considerations around this code potentially being used *before* we want

protocol/proofs/generic-withdrawal-proofs.md
```diff
@@ -0,0 +1,88 @@
...
+We propose to allow users to prove against Super Roots by adding a new overload of
+`proveWithdrawalTransaction` that allows a user to prove the existence of an Output Root inside of
+the Super Root before performing the standard withdrawal validation inside of the Output Root.
+
+We also specifically propose that this be implemented on the *base* `OptimismPortal` instead of
+introducing the code to the `OptimismPortalInterop` portal. Doing so means that we are including
+extra code on the `OptimismPortal` that is technically unnecessary until interop ships but also
+means that we do not need to ship any additional changes to the `OptimismPortal` before interop.
+
+We could expose a new public variable `usingSuperRoots` that would toggle this function on and off
```
ControlCplusControlV, line 32: Would prefer the name `SuperRootsActive` or similar

ControlCplusControlV (approved):

protocol/proofs/generic-withdrawal-proofs.md
```diff
@@ -0,0 +1,98 @@
...
+`proveWithdrawalTransaction` for Super Roots would revert. When `superRootsActive` is `true`, the
+legacy version of the function would revert.
+
+## Alternatives Considered
+
+### Using an Inherited Contract
+
+We could implement this function inside of an inherited contract instead of putting it in the base
+contract. However, this would mean performing two sets of upgrades to the `OptimismPortal` before
+interop ships on an already tight timeline.
```
mds1, line 44: Why does inheritance mean an extra upgrade?

protocol/proofs/generic-withdrawal-proofs.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Considerations
+
+### Unused Code in Production
+
+This proposal means we are shipping code that cannot yet be used into production. We believe this
+trade-off is worthwhile because it means that we would not have to ship the code later on.
```
mds1, line 51: This is a bit precedent-setting for us, as we typically do not ship dead code to production. It's likely we start doing things like this more often to help manage merging of in-flight features to develop, so it would be nice to have clear guidelines for implementation and testing of dead/feature-flagged code

protocol/proofs/generic-withdrawal-proofs.md
```diff
@@ -0,0 +1,98 @@
...
+
+We have decided that we will NOT allow the `OptimismPortal` for one chain to process withdrawals
+from another chain when interop launches because of security implications with the correctness of
+the cross-domain message sender. The Super Root withdrawal proof must therefore guarantee that a
+given withdrawal was actually created on the specific chain that corresponds to the given
+`OptimismPortal`.
+
+This means that we need to include a chain ID in the `OptimismPortal`. Instead of adding this into
+the `OptimismPortal` itself, we will add this into the `SystemConfig` and have the portal read the
+chain ID from the portal.
```
mds1, line 77: I thought we said we can just key of system config addresses and not need to add a chain ID, or am I mixing things up?

protocol/proofs/generic-withdrawal-proofs.md
```diff
@@ -0,0 +1,98 @@
...
+We also specifically propose that this be implemented on the *base* `OptimismPortal` instead of
+introducing the code to the `OptimismPortalInterop` portal. Doing so means that we are including
+extra code on the `OptimismPortal` that is technically unnecessary until interop ships but also
+means that we do not need to ship any additional changes to the `OptimismPortal` before interop.
+
+We could expose a new public variable `superRootsActive` that would toggle this function on and off
+and would also act as a signal for client-side tooling to automatically switch between the two
+available functions. When `superRootsActive` is `false`, the version of
+`proveWithdrawalTransaction` for Super Roots would revert. When `superRootsActive` is `true`, the
+legacy version of the function would revert.
```
mds1, line 36: We should note that chains not in the initial interop set will keep `superRootsActive` set to `false` even after interop is live, assuming that's the current plan

mds1 (approved):

protocol/proofs/generic-withdrawal-proofs.md
smartcontracts: We'd be doing one upgrade for lockbox already, then we'd need to do another portal upgrade at interop time. So technically still 2 upgrades total but means we'd be doing a portal upgrade at the same time as interop which I really don't love tbh.

protocol/proofs/generic-withdrawal-proofs.md
smartcontracts: Good callout! Where would be the right place for this? Notion? Github?

protocol/proofs/generic-withdrawal-proofs.md
smartcontracts: Ah the system config addresses thing refers to the pause mechanism. Here we need chain IDs.

protocol/proofs/generic-withdrawal-proofs.md
mds1: I would suggest a new design doc to supersede https://github.com/ethereum-optimism/design-docs/blob/main/smart-contract-feature-development.md, with the result being an update to our style guide. So perhaps the design review is really just a review of a proposed PR to the style guide

protocol/proofs/generic-withdrawal-proofs.md
mds1: I see this was added in edc4417, so resolving

protocol/proofs/generic-withdrawal-proofs.md
smartcontracts: Ok I'll make a note on my end that this should get updated but won't block this design doc on it

protocol/proofs/generic-withdrawal-proofs.md
```diff
@@ -0,0 +1,110 @@
...
+
+We have decided that we will NOT allow the `OptimismPortal` for one chain to process withdrawals
+from another chain when interop launches because of security implications with the correctness of
+the cross-domain message sender. The Super Root withdrawal proof must therefore guarantee that a
+given withdrawal was actually created on the specific chain that corresponds to the given
+`OptimismPortal`.
+
+This means that we need to include a chain ID in the `OptimismPortal`. Instead of adding this into
+the `OptimismPortal` itself, we will add this into the `SystemConfig` and have the portal read the
+chain ID from the portal.
```
mdehoog, line 83: `portal` -> `config`?

PR #206, opened 2025-02-20
feat: ETH lockbox redesign

agusduha:
**Description**

Making changes to the current ETH lockbox design doc to apply the new design changes. The dependency set is removed from L1 and the `SuperchainConfig` is no more related with the lockbox. Every portal will migrate to have its own lockbox and later two lockboxes can be merged. We use the "upgrader controller" as the authorized role instead of the system admin to be coordinated with the new OPCM design.

### DIFF
--- a/protocol/eth-shared-lockbox.md
+++ b/protocol/eth-shared-lockbox.md
@@ -6,7 +6,7 @@ This document discusses possible solutions to address the constraints on ETH wit
 
 # Summary
 
-With interoperable ETH, withdrawals may fail if the referenced `OptimismPortal` lacks sufficient ETH—especially for large amounts or on OP Chains with low liquidity—since ETH liquidity isn't shared at the L1 level. To prevent users from being stuck, several solutions have been explored. Currently, the Superchain design favors the `SharedLockbox` and extends the `SuperchainConfig` as a *dependency set manager*, which is considered the most effective solution.
+With interoperable ETH, withdrawals may fail if the referenced `OptimismPortal` lacks sufficient ETH—especially for large amounts or on OP Chains with low liquidity—since ETH liquidity isn't shared at the L1 level. To prevent users from being stuck, several solutions have been explored. Currently, the Superchain design favors the `SharedLockbox`, which is considered the most effective solution.
 
 # Problem Statement + Context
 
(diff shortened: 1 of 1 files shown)


protocol/eth-shared-lockbox.md
```diff
@@ -6,7 +6,7 @@ This document discusses possible solutions to address the constraints on ETH wit
 
 # Summary
 
-With interoperable ETH, withdrawals may fail if the referenced `OptimismPortal` lacks sufficient ETH—especially for large amounts or on OP Chains with low liquidity—since ETH liquidity isn't shared at the L1 level. To prevent users from being stuck, several solutions have been explored. Currently, the Superchain design favors the `SharedLockbox` and extends the `SuperchainConfig` as a *dependency set manager*, which is considered the most effective solution.
+With interoperable ETH, withdrawals may fail if the referenced `OptimismPortal` lacks sufficient ETH—especially for large amounts or on OP Chains with low liquidity—since ETH liquidity isn't shared at the L1 level. To prevent users from being stuck, several solutions have been explored. Currently, the Superchain design favors the `SharedLockbox`, which is considered the most effective solution.
```
maurelian, line 9: It seems to me that the SharedLockbox name is no longer accurate. I can see arguments for why we might want to keep it anyways, if we think it will make it easier to transition to the single SharedLockbox in the future. 

Do we expect that the implementation of the single lockbox will be different?

agusduha: We can use the `ETHLockbox` name directly and assume that it can have one or more portals

protocol/eth-shared-lockbox.md
```diff
@@ -35,125 +35,75 @@ OP Chains will be governed within a common Chain Cluster governance entity (the
...
+The existing problem and considerations motivate us to propose an L1 shared liquidity design through the introduction of a new `SharedLockbox` on L1, which serves as a singleton contract for ETH, given a defined set of interoperable chains. New ETH deposits will be directed to the lockbox, with the same process applied to ETH withdrawals.
 
 ### Spec changes
 
 The core proposed changes are as follows:
 
-- **Modify the `SuperchainConfig` contract**: Extend this contract to manage the dependency set of each chain, interacting with each `SystemConfig`.
-- **Introduce the `SharedLockbox` contract**: This contract acts as an escrow for ETH, receiving deposits and allowing withdrawals from approved `OptimismPortal` contracts. The `SuperchainConfig` contract serves as the source of truth of the lockbox.
-- **Modify the `OptimismPortal`**: To forward ETH into the `SharedLockbox` when `depositTransaction` is called, with the inverse process applying when `finalizeWithdrawal` is called.
+- **Introduce the `SharedLockbox` contract**: This contract acts as an escrow for ETH, receiving deposits and allowing withdrawals from approved `OptimismPortal` contracts.
```
maurelian, line 48: is it proxied?

protocol/eth-shared-lockbox.md
```diff
@@ -35,125 +35,75 @@ OP Chains will be governed within a common Chain Cluster governance entity (the
...
 
-  // Authorize the portal on the shared lockbox
-  SHARED_LOCKBOX.authorizePortal(portal);
+The migration process would:
 
-  emit ChainAdded(_chainId, _systemConfig, portal);
-}
-```
+- Allow any caller to trigger the migration
+- Transfer the entire ETH balance of the `OptimismPortal` to the pre-configured `SharedLockbox`
```
maurelian, line 69: Rather than add a new function, it might be simpler to just do this during the upgrade process, via an `upgrade()` function.

maurelian: That's an implementation detail tho

protocol/eth-shared-lockbox.md
```diff
@@ -35,125 +35,75 @@ OP Chains will be governed within a common Chain Cluster governance entity (the
...
 
-A minimal set of functions should include:
+The merge process has two parts:
 
-- `lockETH`: Accepts ETH from the `depositTransaction` originating from a valid `OptimismPortal`.
-- `unlockETH`: Releases ETH from the `finalizeWithdrawalTransaction` originating from a valid `OptimismPortal`.
+1. Destination lockbox preparation:
 
-Access control for `lockETH` and `unlockETH` is validated against the mapping of authorized `OptimismPortal` addresses.
+- Only the upgrade controller role can approve a specific source lockbox to send ETH
```
maurelian, line 79: How will this auth be managed? Is it done by an upgrade, or is there a function to set the value?
If the latter, I have a request to not use `Ownable` and instead read the upgrade controller from the SuperchainConfig.

protocol/eth-shared-lockbox.md
```diff
@@ -35,125 +35,75 @@ OP Chains will be governed within a common Chain Cluster governance entity (the
...
+- This prevents receiving ETH from unauthorized or incorrect lockboxes
 
-### `OptimismPortal` upgrade process
+2. Source lockbox migration:
 
-A one-time L1 liquidity migration is required for each approved chain. By using an intermediate contract denominated `LiquidityMigrator` during the upgrade, all ETH held by the `OptimismPortal` can be transferred to the `SharedLockbox`. This intermediate contract functions similarly to `StorageSetter` and is used for the sole purpose of transferring the ETH balance. After this migration, the `OptimismPortal` is upgraded to the new version with the desired functionality.
+- Only the upgrade controller role can trigger the migration
```
maurelian, line 84: I think above this said anyone could trigger it.

protocol/eth-shared-lockbox.md
```diff
@@ -35,125 +35,75 @@ OP Chains will be governed within a common Chain Cluster governance entity (the
...
-    Note over L1PAO: End batch
+flowchart TD
+  SuperchainConfig --> OptimismPortal
+  SuperchainConfig --> OptimismPortal'
+  SuperchainConfig --> OptimismPortal''
+
+  OptimismPortal --> ETHLockbox
+  OptimismPortal' --> ETHLockbox'
+  OptimismPortal'' --> ETHLockbox'
 ```
```
maurelian, line 102: what do the arrows in this diagram indicate?

protocol/eth-shared-lockbox.md
mds1, line 1: Design review comments:
- Add a function to lockbox/portal implementations so they can introspect their proxy admin and proxy admin owner for auth purposes
- Second design needed to spec out the interop rollout. This should detail various contract upgrades, what specifically happens for chain in AND not in the initial interop set, at what point interop is officially live (i.e. what is the set of criteria that make it live), and the sequencing for adding more chains to the interop set

protocol/eth-shared-lockbox.md
agusduha: Yes, it should be

protocol/eth-shared-lockbox.md
agusduha: Like we talked, we are going to use the proxy admin owner

protocol/eth-shared-lockbox.md
agusduha: The permissionless trigger was for the `OptimismPortal`, though after talking it will be permissioned now

protocol/eth-shared-lockbox.md
agusduha: Oh it seems that some arrows are the other way, it should indicate the dependency between contracts

protocol/eth-shared-lockbox.md
agusduha: Just pushed some fixes for the first point, the second point is being done here https://github.com/ethereum-optimism/design-docs/pull/207

protocol/eth-shared-lockbox.md
smartcontracts: Let's use a function

protocol/eth-shared-lockbox.md
```diff
@@ -35,125 +35,80 @@ OP Chains will be governed within a common Chain Cluster governance entity (the
...
 
-A minimal set of functions should include:
+2. Source lockbox migration:
 
-- `lockETH`: Accepts ETH from the `depositTransaction` originating from a valid `OptimismPortal`.
-- `unlockETH`: Releases ETH from the `finalizeWithdrawalTransaction` originating from a valid `OptimismPortal`.
+- Only the `ProxyAdmin` owner can trigger the migration
+- The destination lockbox is provided as a parameter during migration
+- Transfer the entire ETH balance to the specified destination lockbox
+- Permanently disable the current lockbox after migration
```
ControlCplusControlV, line 87: I would specify *how* this should be done, its not immediately clear

protocol/eth-shared-lockbox.md
```diff
@@ -35,125 +35,80 @@ OP Chains will be governed within a common Chain Cluster governance entity (the
...
-    if (_chainId == currentId) continue;
+A ETH migration function can be added to the `OptimismPortal` to allow transferring all ETH to the `ETHLockbox` at any time. The `ETHLockbox` address is set during initialization of the `OptimismPortal` and cannot be changed afterwards.
 
-    systemConfigs[currentId].addDependency(_chainId);
+The migration process would:
 
-    systemConfigs[_chainId].addDependency(currentId);
-  }
+- Only the `ProxyAdmin` owner can trigger the migration
+- Transfer the entire ETH balance of the `OptimismPortal` to the pre-configured `ETHLockbox`
```
ControlCplusControlV, line 69: should likely disable the lockbox in some way after as well correct?

ControlCplusControlV (approved):
Looks good to me, but would appreciate some more specificity on how the lockboxes should should be treated after migration

protocol/eth-shared-lockbox.md
agusduha: Yeah, maybe is better just to clarify that is no longer being used, but i wouldn't add any disabling process in the code

protocol/eth-shared-lockbox.md
smartcontracts, line 69: Let's make sure that the `OptimismPortal` is approved to deposit into the `ETHLockbox` when doing this xfer

protocol/eth-shared-lockbox.md
smartcontracts: Agreed we shouldn't disable it in code

smartcontracts (approved):
LGTM with some very minor comments

protocol/eth-shared-lockbox.md
agusduha: Fixed https://github.com/ethereum-optimism/design-docs/pull/206/commits/89c4d62e04eb32fd9f49abb7236f02fcf3ced82d

smartcontracts:
Ok just waiting on approval from @mds1 now

smartcontracts:
Oh and @clabby

protocol/eth-shared-lockbox.md
```diff
@@ -35,125 +35,81 @@ OP Chains will be governed within a common Chain Cluster governance entity (the
...
 
-    uint256 currentId = _dependencySet.at(i);
+### `OptimismPortal` upgrade process
 
-    // Skip recently added chain
-    if (_chainId == currentId) continue;
+A ETH migration function can be added to the `OptimismPortal` to allow transferring all ETH to the `ETHLockbox` at any time. The `ETHLockbox` address is set during initialization of the `OptimismPortal` and cannot be changed afterwards.
 
-    systemConfigs[currentId].addDependency(_chainId);
+The migration process would:
```
mds1, line 66: nit
```suggestion
The migration process would work as follows:
```

protocol/eth-shared-lockbox.md
```diff
@@ -35,125 +35,81 @@ OP Chains will be governed within a common Chain Cluster governance entity (the
...
-  emit ChainAdded(_chainId, _systemConfig, portal);
-}
-```
+The merge process has two parts:
 
-Note that, under the specified flow, the dependency set consistently maintains the form of a [complete graph](https://en.wikipedia.org/wiki/Complete_graph) at all times.
+1. Destination lockbox preparation:
 
-### `SharedLockbox` implementation
+- Only the `ProxyAdmin` owner can approve a specific source lockbox to send ETH
```
mds1, line 80: There should be requirement that migration also requires the ProxyAdmin owner of the source lockbox to match the ProxyAdmin owner of the destination lockbox

mds1 (approved):

mds1 (approved):

PR #211, opened 2025-02-26
Add design doc for interop game config

ajsutton:
**Description**

Design doc to review options to ensure the split depth is suitable for interop games.

### DIFF
--- a/protocol/proofs/interop-config.md
+++ b/protocol/proofs/interop-config.md
@@ -1,0 +1,67 @@
+# Problem Statement + Context
+
+The fault proof system for interop, continues to use the split game where the bottom half of the game bisects the cannon
+execution trace. However the top half of the game no longer progresses from one block to the next, but breaks down the
+transition between one super root to the next into 1024 individual steps. As a result, the trace for the top half of the
+game requires 2048 steps per 2 seconds compared to the pre-interop game when using 2 second block times, or 1024 times
+more when using 1 second block times.
+
+To operate correctly, the game's split depth must always be large enough for there to be enough leaf nodes in the top 
+half of the game to progress from the anchor state to the current chain safe head. Otherwise the valid trace for the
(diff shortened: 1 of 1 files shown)


Inphi (approved):
I agree with the proposed solution.

I believe there's a path to breaking up consolidation thereby requiring much more steps per timestamp to handle a terrible amount of chains. So even 128 steps can quickly get exhausted in a few years. But we'll deal with that when we have to.

As a side note, I think it's better to hardcode the number of steps in the program rather than pull in from the contract. In the event the steps per timestamp needs to be tweaked, it'll be because we're making fundamental changes to the program anyways. So there's little point making it configurable via the dispute game.

ajsutton:
> I believe there's a path to breaking up consolidation thereby requiring much more steps per timestamp to handle a terrible amount of chains. So even 128 steps can quickly get exhausted in a few years. But we'll deal with that when we have to.

Yeah I agree we will be able to break up the consolidation steps. I strongly suspect that will mean moving to a new game type which would then be an opportunity to increase the number of steps if we need to, and we can deal with the impact on bonds/capital efficiency then.

> As a side note, I think it's better to hardcode the number of steps in the program rather than pull in from the contract. In the event the steps per timestamp needs to be tweaked, it'll be because we're making fundamental changes to the program anyways. So there's little point making it configurable via the dispute game.

Definitely if there are other changes like breaking the consolidation into multiple steps. I'd be open to it anyway but could go either way if it was just increasing the steps to support more chains. And we definitely should keep hard coding it now then decide what to do if we need to change it. Fundamentally though, there's no chance we can support 127 chains without changing other game rules.

protolambda (approved):
This sounds reasonable to me

clabby (approved):

clabby:
Change isn't horrible to make in `kona` + rationale seems sane, :shipit:

PR #214, opened 2025-03-12
feat: add interop access list

agusduha:
**Description**

This design introduces a new validation mechanism for cross-chain message execution using transaction access-lists. The `CrossL2Inbox` contract will implement gas introspection to validate `ExecutingMessage` transactions. Each cross-chain message must be pre-declared in the transaction's access list.

Co-authored-by: @skeletor-spaceman @0xng @tynes @protolambda

### DIFF
--- a/protocol/interop-access-list.md
+++ b/protocol/interop-access-list.md
@@ -1,0 +1,103 @@
+# Interop Access List - Design doc
+
+## Purpose
+
+This document outlines the changes required for the `CrossL2Inbox` contract and client implementations to enhance security and prevent denial of service (DoS) attacks in cross-chain message processing.
+
+## Summary
+
+This design introduces a new validation mechanism for cross-chain message execution using transaction access-lists. The `CrossL2Inbox` contract will implement gas introspection to validate `ExecutingMessage` transactions. Each cross-chain message must be pre-declared in the transaction's access list.
+
+## Problem Statement + Context
+
+Interop introduces a property where transactions can't be validated without execution, conflicting with Ethereum's goal of cheap pre-validation to prevent spam. This creates a risk where invalid interop transactions are dropped and therefore never included in the chain, causing sequencers to lose fees. MEV searchers can exploit this by attempting extraction and, if unsuccessful, rendering transactions unprocessable, paradoxically increasing spam. We need a way to cheaply validate cross-chain messages while ensuring spam incurs a cost.
+
+## Proposed Solution
+
+### Overview
+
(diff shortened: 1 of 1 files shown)


protocol/interop-access-list.md
```diff
@@ -0,0 +1,103 @@
...
+        isWarm := iszero(gt(sub(startGas, endGas), WARM_READ_COST))
+    }
+ }
+
+function validateMessage(Identifier calldata _id, bytes32 _msgHash) external {
+    bytes32 checksum = calculateChecksum(_id, _msgHash);
+
+    (bool _isSlotWarm,) = _isWarm(checksum);
+
+    if (!_isSlotWarm) revert NotWarm();
```
protolambda, line 56: Important detail from previous discussion to document here: the storage-warmup is rolled back when the `revert` happens. That means that prior invalid message lookups will not cause later lookups of the same checksum to appear valid.

protocol/interop-access-list.md
```diff
@@ -0,0 +1,103 @@
...
+    CrossL2Inbox-->>L2toL2CDM: revert NotWarm()
+```
+
+### Alternatives considered
+
+Alternative solutions that kept the current dynamic declaration approach were explored. The most prominent one involved adding locking and unlocking functionality to the `CrossL2Inbox`. If the transaction referenced an invalid message, then that transaction would be included at the top of the next block when the `CrossL2Inbox` was locked. Because it was locked, the invalid transaction would revert. If it referenced an valid transaction, then it would be included when the `CrossL2Inbox` was unlocked. A special transaction took care of locking and unlocking the `CrossL2Inbox`.
+
+### Security considerations
+
+- Future changes in gas cost can affect this solution
```
protolambda, line 95: We may want to exand on this note that gas-introspection is fragile, and will need expert review with every EVM storage/call gas behavior change.
A custom opcode or precompile could do the access-list check in the future instead, to not rely on gas-introspection, if the gas-introspection is considered too problematic.

protocol/interop-access-list.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Every message requires at most 3 entries in the access list and at least 2, therefore it will warms at most 3 and at least 2 storage slots, of which we only care about one. It's important that the checksum of the `Identifier` and `messageHash` data parameters provided to the `validateMessage` function doesn't result in one of these slots. We should ensure this scenario is statistically infeasible.
```
protolambda, line 97:
```suggestion
- Every message requires at most 3 entries in the access list and at least 2, therefore it will warms at most 3 and at least 2 storage slots, of which we only care about one. It's important that the checksum of the `Identifier` and `messageHash` data parameters provided to the `validateMessage` function doesn't result in a slot warmed for any other reason than a prior non-reverting `validateMessage`. We should ensure there is no accidental storage collision that warms up an unverified checksum.
```

protolambda (commented):
LGTM, just a few comments

tynes (approved):

protolambda (approved):

PR #216, opened 2025-03-18
Promise Callback Library

hamdiallam:
**Description**

By adding the return data to the `RelayedMessage` event, a callback library can be introduced allowing users to attach continuations dependent on the return. This makes any remote view function callable with low latency via superchain interop.


tynes:
I don't think any amount of design review will be able to give a thumbs up on this design better than working with teams to build interop into their applications. I recommend that we focus on getting the entrypoint primitive into the `L2ToL2CrossDomainMessenger` and then build this on top of that to see what design pattern works the best. With the entrypoint, you can iterate on this design without needing to make any additional changes to the core contracts

ecosystem/promise.md
```diff
@@ -0,0 +1,141 @@
...
+## Notes
+
+- Requires no fundamental contract API changes to the CDM besides the event.
+- Different kinds of callbacks can be introduced on the source, all without change to the destination. `onFailure`, `onTimeout`, etc. The log identifier for the event has timestamp information. Currently the `L2ToL2CrossDomainMessenger` does not allow for failed message delivery but if it did, failure callbacks could be created as well.
+- Multiple callbacks can be registered to the same message. `then()` and `onTimeout()`.
+- Relayers must listen to `CallbackRegistered` events to know when which messages have pending continuations. The tradeoff being additional touch-points for relayers to hook into. In the alternative, if a message was to be sent back from the destination, it would be almost be a no-op in relayer infra since the message outbox would look the same.
+
+# Alternatives Considered
+
+## Callback Entrypoint
```
fainashalts, line 124: The entrypoint solution and this proposal are not mutually exclusive though, right? You are simply proposing that the CDM emit return data that can be used by other contracts if I understand correctly. Entrypoints would still have a place but wouldn't be the preferred way to address this use case?

ecosystem/promise.md
```diff
@@ -0,0 +1,141 @@
...
+A new entrypoint contract is created, `CallbackEntrypoint` that expects callback information to be propogated with the message to the destination. When relaying, the entrypoint contract parses the callback information and sends a message back to the source invoking the callback with the return data of the relay.
+
+![Callback Entrypoint](./promise/callback-entrypoint.png)
+
+## Notes
+
+- Tight coupling between the source & destination chains. Callback information must be propogated between networks
+- Multiple callbacks cant clearly be registered. Must be enshrined in the propogated callback data.
+- Single interface for relayers `relayMessage`.
+- The proposed solution could be implemented via a generic entrypoint to avoid adding the return data to the event. However this feels like more complexity from a code & call path perspective. Propogate callback -> parse callback -> relay -> capture return and sendMessage(). Versus -- RelayedMessage() -> continue().
```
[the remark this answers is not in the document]
hamdiallam: +1

If this involved a lot of change to the messenger, avoiding this and instead leveraging entrypoints would make a lot of sense.

However the change is very minimal and at the same time very high leverage (consuming the return data via the event itself) in the event promise based contract development becomes more interesting.

ecosystem/promise.md
```diff
@@ -0,0 +1,141 @@
...
+        emit RelayedMessage(source, nonce, messageHash, returnData_);
+    }
+}
+```
+
+The returned `messageHash` which identifies a sent messages can be behave like a Promise that the caller can attach continuations to. All state
+related to the callback is managed on the sending chain and works with any remote target.
+
+```solidity
+library Promise {
```
karlfloersch, line 59: I think we should do some thinking around how to make the API for incentivized relay / callbacks as minimal (and flexible) as possible. This makes me wary about using a Library because then it's harder to hide the underlying callback logic and harder to update.

I think it might be easier to use a dedicated contract for registering callbacks that you call vs using a library

hamdiallam: yea totally agree, the functions here are marked as external meaning this library gets deployed as a dedicated contract

no point in declaring this as a lib though, i'll update this to be a contract

tynes:
This document is coupling too many different things. We are talking about a modification to the cross domain messenger, a promise library and an incentivized relay. I recommend creating a design doc that specifically is focused on the changes to the cross domain messenger that you want to get in. I also am not confident in making a bunch of small changes to the messenger without understanding how a specific user's app needs the change. Historically we added features when they sounded cool, which has resulted in a lot of tech debt. I want to see changes driven by an app attempting to integrate and then unable to do so because of the lack of a feature

hamdiallam:
> This document is coupling too many different things. We are talking about a modification to the cross domain messenger, a promise library and an incentivized relay. I recommend creating a design doc that specifically is focused on the changes to the cross domain messenger that you want to get in. I also am not confident in making a bunch of small changes to the messenger without understanding how a specific user's app needs the change. Historically we added features when they sounded cool, which has resulted in a lot of tech debt. I want to see changes driven by an app attempting to integrate and then unable to do so because of the lack of a feature

Nowhere in this design doc is incentivized relays mentioned. I've explicitly made them separate since they are not dependent on each other. In the doc you'll see this is only about chaining side effects to the return value of a remote call, generally useful when writing cross chain contracts, removing the need to write handlers that ping pong back the return value for every possible remote view call

I'll go ahead and pull out the changes to the messenger but you're unnecessarily commingling the two

PR #218, opened 2025-03-19
interop: Topology and Tx Flow for Interop Chain Operation

tynes:
**Description**

Design doc that is meant to define clearly the topology for a cloud
deployment of an OP Stack cluster. This is not meant to define the
only way to deploy a cluster, but is the architecture that we will
do our security review against to ensure that it is not possible to
attack a cluster.

### DIFF
--- a/protocol/interop-topology.md
+++ b/protocol/interop-topology.md
@@ -209,3 +209,4 @@ Ultimately, this design represents a hypothesis which needs real testing before 
 - Put an interval check in place in the mempool
 - Remove build-time checks
 - Test RPC performance (data collectable via Grafana)
+- Consider and add Supervisor-health as a trigger for Conductor Leadership Transfer
--- a/protocol/interop-topology.md
+++ b/protocol/interop-topology.md
@@ -1,0 +1,145 @@
+# Purpose
+
+<!-- This section is also sometimes called “Motivations” or “Goals”. -->
+
+<!-- It is fine to remove this section from the final document,
+but understanding the purpose of the doc when writing is very helpful. -->
+
+This document exists to drive consensus and act on a reference of the preferred topology
+of a cloud deployment of an interop enabled OP Stack cluster.
+
+# Summary
+
+<!-- Most (if not all) documents should have a summary.
+While the length will likely be proportional to the length of the full document,
+the summary should be as succinct as possible. -->
+
+The production topology of an interop cluster checks the validity of cross chain transactions at:
+- cloud ingress (`proxyd`)
+- sentry node mempool ingress
+- sentry node mempool on interval
+- sequencer node mempool on interval
+
+The validity of cross chain transactions are not checked at:
+- sequencer node mempool ingress
+- block builder
+
+It is safe to not check at block building time because if the check passes at ingress
+time and passes again at block building time, it does not exhaustively cover all cases.
+It is possible that the remote reorg happens **after** the local block is sealed.
+In practice, it is far more likely to have an unsafe reorg on the remote chain that is not
+caught **before** the local block is sealed because the time between checking at the mempool
+and checking at block building is so small. Therefore we should not add a remote RPC request
+as part of the block building happy path, but we should still implement and benchmark it.
+
+TODO: include information on recommendation around standard mode/multi node/running multiple supervisors.
+Should include information on the value props of each so we can easily align on roadmap.
+
+# Problem Statement + Context
+
+<!-- Describe the specific problem that the document is seeking to address as well
+as information needed to understand the problem and design space.
+If more information is needed on the costs of the problem,
+this is a good place to that information. -->
+
+It is a goal to remove as many async, blocking operations from the hot path of
+the block builder as possible. Validating an interop cross chain transaction
+requires a remote RPC request to the supervisor. Having this as part of the hot
+path introduces a denial of service risk. Specifically, we do not want to have so
+many interop transactions in a block that it takes longer than the blocktime
+to build a block.
+
+To prevent this sort of issue, we want to move the validation on interop
+transactions to as early as possible in the process, so the hot path of the block builder
+only needs to focus on including transactions.
+
+For context, a cross chain transaction is defined in the [specs](https://github.com/ethereum-optimism/specs/blob/85966e9b809e195d9c22002478222be9c1d3f562/specs/interop/overview.md#interop). Any reference
+to the supervisor means [op-supervisor](https://github.com/ethereum-optimism/design-docs/blob/d732352c2b3e86e0c2110d345ce11a20a49d5966/protocol/supervisor-dataflow.md).
+
+# Proposed Solution
+
+<!-- A high level overview of the proposed solution.
+When there are multiple alternatives there should be an explanation
+of why one solution was picked over other solutions.
+As a rule of thumb, including code snippets (except for defining an external API)
+is likely too low level. -->
+
+## Solution
+
+### `op-supervisor` alternative backend
+
+We add a backend mode to `op-supervisor` that operates specifically by using dynamic calls to `eth_getLogs`
+to validate cross chain messages rather than its local index. This could be accomplished by adding
+new RPC endpoints that do this or could be done with runtime config. When `op-supervisor` runs in this
+mode, it is a "light mode" that only supports `supervisor_validateMessagesV2` and `supervisor_validateAccessList`
+(potentially a subset of their behavior). This would give us a form of "client diversity" with respect
+to validating cross chain messages. This is a low lift way to reduce the likelihood of a forged initiating
+message. A forged initiating message would be tricking the caller into believing that an initiating
+message exists when it actually doesn't, meaning that it could be possible for an invalid executing
+message to finalize.
+
+TODO: feedback to understand what capabilities are possible
+
+### `proxyd`
+
+We update `proxyd` to validate interop messages on cloud ingress. It should check against both the indexed
+backend of `op-supervisor` as well as the alternative backend.
+
+### Sentry Node Mempool Ingress
+
+We update the EL clients to validate interop transactions on ingress to the mempool. This should be a different
+instance of `op-supervisor` than the one that is used by `proxyd` to reduce the likelihood of a nondeterministic
+bug within `op-supervisor`.
+
+### Sentry Node + Sequencer Mempool on Interval
+
+We update the EL clients to validate interop transactions on an interval in the mempool. Generally the mempool
+will revalidate all transactions on each new block, but for an L2 that has 1-2s blocktime, that is quite often.
+It may be the case that we could get away with a batch RPC request every 1-2s, but generally we should not do
+`n` RPC requests on each block where `n` is the number of transactions that include statically declared executing
+messages.
+
+### Block Building
+
+Lets say that it takes 100ms for the transaction to be checked at `proxyd`, checked at the mempool of the sentry node,
+forwarded to the sequencer and pulled into the block builder. The chances of the status of an initiating message
+going from existing to not existing during that timeframe is extremely small. Even if we did check at the block builder,
+it doesn't capture the case of a future unsafe chain reorg happening that causes the message to become invalid.
+Because it is most likely that the remote unsafe reorg comes after the local block is sealed, there is no real
+reason to block the hot path of the chain with the remote lookups.
+
+## Resource Usage
+
+<!-- What is the resource usage of the proposed solution?
+Does it consume a large amount of computational resources or time? -->
+
+Doing a remote RPC request is always going to be an order of magnitude slower than doing a local lookup.
+Therefore we want to ensure that we can parallelize our remote lookups as much as possible. Block building
+is inherently a single threaded process given that the ordering of the transactions is very important.
+
+## Single Point of Failure and Multi Client Considerations
+
+There is a single point of failure with the `op-supervisor`. Adding an alternative backend that dynamically
+fetches data using `eth_getLogs` rather than looking up its local database helps to approximate a second implementation.
+
+<!-- Details on how this change will impact multiple clients. Do we need to plan for changes to both op-geth and op-reth? -->
+
+# Alternatives Considered
+
+<!-- List out a short summary of each possible solution that was considered.
+Comparing the effort of each solution -->
+
+## Block Building
+
+The main alternative to not validating transactions at the block builder is validating transactions
+at the block builder. We would like to have this feature implemented because it can work for simple networks,
+as well as act as an ultimate fallback to keep interop messaging live, but we do not want to run it as
+part of the happy path.
+
+# Risks & Uncertainties
+
+<!-- An overview of what could go wrong.
+Also any open questions that need more work to resolve. -->
+
+We really need to measure everything to validate our hypothesis on the ideal architecture.
+To validate the ideal architecture, we need to measure it and then try to break it.
--- a/protocol/interop-topology.md
+++ b/protocol/interop-topology.md
@@ -133,7 +133,7 @@ it may be prudent to switch the active Sequencer to one with a functional Superv
 ## Solution Side-Ideas
 
 Although they aren't strictly related to TX Flow or Redundancy, here are additional ideas to increase the stability
-of a network. These ideas won't be brought forward into the Solution Summary.
+of a network. These ideas won't be brought forward into the Solution Summary or Action Items.
 
 ### `op-supervisor` alternative backend
 
@@ -157,42 +157,27 @@ We should establish `op-supervisor` checks of transactions at the following poin
 - On ingress to all mempools
 - On regular interval on all mempools
 
-Additionally, the Sequencer should batch the calls 
+Additionally, regular interval checks should use batch calls which validate at least a block's worth of the mempool
+at a time.
 
-The production topology of an interop cluster checks the validity of cross chain transactions at:
-- cloud ingress (`proxyd`)
-- sentry node mempool ingress
-- sentry node mempool on interval
-- sequencer node mempool on interval
-
-The validity of cross chain transactions are not checked at:
+`op-supervisor` checks of transactions should *not* happen at the following points:
 - sequencer node mempool ingress
-- block builder
+- block building (as a synchronous activity)
 
-It is safe to not check at block building time because if the check passes at ingress
-time and passes again at block building time, it does not exhaustively cover all cases.
-It is possible that the remote reorg happens **after** the local block is sealed.
-In practice, it is far more likely to have an unsafe reorg on the remote chain that is not
-caught **before** the local block is sealed because the time between checking at the mempool
-and checking at block building is so small. Therefore we should not add a remote RPC request
-as part of the block building happy path, but we should still implement and benchmark it.
-
-TODO: include information on recommendation around standard mode/multi node/running multiple supervisors.
-Should include information on the value props of each so we can easily align on roadmap.
+When we deploy hosts, we should currently use a "Full Validation Set" of one Supervisor plus N Managed nodes,
+to maximize redundancy and independent operation of validators. When Sequencers are deployed, Conductors should manage
+individual Managed Nodes *across* Supervisors.
 
 # Alternatives Considered
 
-<!-- List out a short summary of each possible solution that was considered.
-Comparing the effort of each solution -->
-
-## Block Building
+## Checking at Block Building Time (Tx Flow Solution)
 
 The main alternative to not validating transactions at the block builder is validating transactions
 at the block builder. We would like to have this feature implemented because it can work for simple networks,
 as well as act as an ultimate fallback to keep interop messaging live, but we do not want to run it as
 part of the happy path.
 
-## Multi-Node (redundancy solution)
+## Multi-Node (Host Redundancy Solution)
 
 One request that has been made previously is to have "Multi-Node" support. In this model,
 multiple Nodes for a single chain are connected to the same Supervisor. To be clear, the Supervisor software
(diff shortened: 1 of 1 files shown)


protocol/interop-topology.md
```diff
@@ -0,0 +1,145 @@
...
+message. A forged initiating message would be tricking the caller into believing that an initiating
+message exists when it actually doesn't, meaning that it could be possible for an invalid executing
+message to finalize.
+
+TODO: feedback to understand what capabilities are possible
+
+### `proxyd`
+
+We update `proxyd` to validate interop messages on cloud ingress. It should check against both the indexed
+backend of `op-supervisor` as well as the alternative backend.
```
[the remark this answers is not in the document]
axelKingsley: Another thing to consider -- the way `CheckAccessList` currently works is:
- All Access Items are checked for *correct inclusion in the events database*. Inclusion in the database only implies Local Unsafety.
- All Access Items are checked to match the provided `MinSafety`, which can be any safety level. If the any check is not sufficiently safe, the check is considered as failed.

We can use this MinSafety to control the filter strictness:
- On Ingress, if we check only that it is local-unsafe, we risk including messages that won't become cross-unsafe later (like due to being after an invalid message)
- So instead, we would want to enforce that it is cross-unsafe
  - If it is cross-unsafe, then cross-validation works to this point and it won't change unless a reorg affects the remote chain

IF the ingress filter acted on `cross-safe`, then it would be *highly effective* at the cost of having very high block latency -- interop transactions would be dropped unless they were pointing at data confirmed on the L1, which could take minutes to be batch published.

IF the ingress filter acted on `local-unsafe`, then it would as fast low-latency as possible, and as wrong as possible Any time a reorg happens on the remote chain, all these previously valid messages would be (potentially) invalidated.

protocol/interop-topology.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Sentry Node Mempool Ingress
+
+We update the EL clients to validate interop transactions on ingress to the mempool. This should be a different
+instance of `op-supervisor` than the one that is used by `proxyd` to reduce the likelihood of a nondeterministic
+bug within `op-supervisor`.
```
axelKingsley, line 92: This is exactly what the mempool `IngressFilter` does today, it checks the supervisor for validity on ingestion, and rejects it otherwise.

protocol/interop-topology.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Sentry Node + Sequencer Mempool on Interval
```
axelKingsley, line 94: +1, regular interval re-checks would be good, and indeed could be batched to the supervisor.

protocol/interop-topology.md
```diff
@@ -0,0 +1,145 @@
...
+
+We update the EL clients to validate interop transactions on an interval in the mempool. Generally the mempool
+will revalidate all transactions on each new block, but for an L2 that has 1-2s blocktime, that is quite often.
+It may be the case that we could get away with a batch RPC request every 1-2s, but generally we should not do
+`n` RPC requests on each block where `n` is the number of transactions that include statically declared executing
+messages.
+
+### Block Building
```
axelKingsley, line 102: I agree with this *assuming* that prior mentioned interval checks are running -- the time between ingress and building could be arbitrarily long with PGA.

(Axel's soapbox: everything gets easier with FIFO!)

protocol/interop-topology.md
```diff
@@ -0,0 +1,145 @@
...
+reason to block the hot path of the chain with the remote lookups.
+
+## Resource Usage
+
+<!-- What is the resource usage of the proposed solution?
+Does it consume a large amount of computational resources or time? -->
+
+Doing a remote RPC request is always going to be an order of magnitude slower than doing a local lookup.
+Therefore we want to ensure that we can parallelize our remote lookups as much as possible. Block building
+is inherently a single threaded process given that the ordering of the transactions is very important.
```
axelKingsley, line 118: Because every transaction has a statically defined Access List, *unless* the transactions being considered for building are interdependent (which is an unlikely use-case), interop transactions *can* indeed be batch submitted.

Say the builder started building, and before it did it looked over the first "100" Tx in its prioritized list. It could locally ilter out all the access list items across all the transactions, and then submit it to the Supervisor as one giant `CheckAccessList` call.

This would allow us to have very time-relevant checks while still not individually checking every Tx while building, which I agree could be prohibitive.

protocol/interop-topology.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Single Point of Failure and Multi Client Considerations
+
+There is a single point of failure with the `op-supervisor`. Adding an alternative backend that dynamically
+fetches data using `eth_getLogs` rather than looking up its local database helps to approximate a second implementation.
```
axelKingsley, line 123: Heavy agree, this is similar to the "Challenger" I proposed in Discord recently -- I want this at a minimum to challenge Supervisor correctness, but it could also be used as a (potentially slow) alternative validation method.

protocol/interop-topology.md
axelKingsley: One issue with this -- if we batched up X Tx's AccessLists and sent it to one big CheckAccessList call, if it fails we won't immediately know *which* transaction fails. That might be a moment where we can break it down and move slower because we know *something* is going to be invalid, and then in the happy-path we can still just use the big batches.

axelKingsley (commented):
I think in general I agree with the premise here: we need to economical with our calls to the supervisor, or else RPC communication will dominate the block-building time, creating the potential that the supervisor can't build fast enough if enough interop transactions occur.

There are *lots* of places for us to dial things in, and I really agree with:
- Proxyd/Ingress pre-checks, which can use `cross-unsafe` as the minimum safety
- Regular interval re-checking in the mempool, because the older a tx is, the less accurate the previous check may be.
- Batching calls to the supervisor where possible

I think we can likely set up a batch-submission scheme for the block builder which puts interop checks very near to the block-building while avoiding making many calls -- basically just batch up all the AccessLists for the next X transactions that are going to be applied to the block. I talk about that in more detail in a comment below.

In general I am concerned with time delta between interop checks and tx inclusion. Optimism L2s currently reorg *very rarely*, but one new reorg vector we are introducting is invalid interop Tx. While it's true that a message may become invalid *after* unsafe inclusion, that doesn't convince me that checking at build time isn't valuable. If I really wanted to I could make the argument that "L2 chains basically never reorg, so we should just check at proxyd, and then never again".

Hard agree that we need to take measurements to best decide where to set the dial to. We'll need to design those experiments.

protocol/interop-topology.md
```diff
@@ -0,0 +1,145 @@
...
+- sentry node mempool ingress
+- sentry node mempool on interval
+- sequencer node mempool on interval
+
+The validity of cross chain transactions are not checked at:
+- sequencer node mempool ingress
+- block builder
+
+It is safe to not check at block building time because if the check passes at ingress
+time and passes again at block building time, it does not exhaustively cover all cases.
```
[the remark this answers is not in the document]
tynes: There are two time periods to think of - the in which a transaction reaches `proxyd` ingress and the duration in which a transaction is `unsafe` before being promoted to `safe`. I am claiming that the time in which a transaction is `unsafe` before becoming `safe`/`finalized` is much larger than the time in which a transaction goes from `proxyd` ingress to being included in a block. Far more reorgs will happen after the block is sealed and its unsafe than during the time between `proxyd` and the block builder

protocol/interop-topology.md
```diff
@@ -0,0 +1,145 @@
...
+of why one solution was picked over other solutions.
+As a rule of thumb, including code snippets (except for defining an external API)
+is likely too low level. -->
+
+## Solution
+
+### `op-supervisor` alternative backend
+
+We add a backend mode to `op-supervisor` that operates specifically by using dynamic calls to `eth_getLogs`
+to validate cross chain messages rather than its local index. This could be accomplished by adding
```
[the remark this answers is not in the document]
tynes: > It's at least 2x as slow as the local index, since there's now another RPC call to some other chain to fetch logs

I am personally ok with increasing the latency at ingress for cross chain transactions to be more certain that they are correctly being dropped when invalid. The failure mode is that somebody is able to trick the supervisor into saying that an executing message is valid when it is not. This could be used to effectively mint an infinite amount of ether

> It might get messy: if there are cascading dependencies, the supervisor may find itself with a stack of logs to fetch, which could be very slow.

Do we actually need to recurse here or could we just do a single RPC call to see if the particular log exists?

protocol/interop-topology.md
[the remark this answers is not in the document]
tynes: This is great detail, thank you @axelKingsley!

We do need to make decisions around what the min safety level is that we plan on running. Intuitively I believe we should do `local-unsafe` given that the product feature that we want is 2s cross chain transactions.

Regarding category 3, I tried to communicate in https://github.com/ethereum-optimism/design-docs/pull/218/files#r2004374259 why I think they are very unlikely to exist often in practice, in practice an unsafe head reorg is a larger category that may include category 3 but also includes cross chain transactions that have already passed all checks and have been sealed in a block

protocol/interop-topology.md
tynes: If you do a batch rpc call then you should be able to get a batch rpc response and know which responses correspond to which requests by matching up indices

tynes:
@axelKingsley Wondering if you could help articulate the differences between standard/managed mode, their value props and how it compares to running 1 supervisor per set of replicas. I would like to contain that context in this document so that we can present a unified piece of information to the platforms team

protocol/interop-topology.md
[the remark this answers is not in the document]
axelKingsley: You need to do all calls, because even if you know that log exists on the remote block, you don't know if the remote block is valid yet. Maybe there's some recursive logic that resolves this... 🤔

protocol/interop-topology.md
axelKingsley: Fair, it could be a batch of CheckAccessList calls. I was imagining that we could compile them into one *massive* Access List and then just make a single (non-batched) call, which loses the Tx resolution. But just baching the calls make sense.

protocol/interop-topology.md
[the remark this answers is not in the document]
axelKingsley: Right, and if these transactions already passed validation and were sealed *before* the remote reorg happened, then the caller who submitted those transactions almost certainly isn't DOSing, unless they had the ability to predict the remote reorg.

All this is to say, remote reorgs and DOS vectors seem to have little overlap if any.

protocol/interop-topology.md
```diff
@@ -0,0 +1,186 @@
...
+In order to have redundancy, we will need multiple Nodes, and also *multiple Supervisors*.
+We should use Conductor to ensure the Sequencers have redundancy as well.
+Therefore, we should arrange the nodes like so:
+
+|             | Chain A | Chain B | Chain C |
+|------------|---------|---------|---------|
+| Supervisor 1 | A1      | B1      | C1      |
+| Supervisor 2 | A2      | B2      | C2      |
+| Supervisor 3 | A3      | B3      | C3      |
+
```
zhwrd, line 143: This makes sense for sequencers. Are there any recommendations for how to run replica nodes that we would be using for things like RPC, security monitoring, etc? I'm assuming we could run a similar setup where we would want multiple supervisors and nodes per "availability group", or are there other options to consider?

protocol/interop-topology.md
```diff
@@ -0,0 +1,186 @@
...
+In this model, each chain has one Conductor, which joins all the Sequencers for a given network. And each heterogeneous group of Sequencers is joined by a Supervisor.
+This model gives us redundancy for both Sequencers *and* Supervisors. If an entire Supervisor were to go down,
+there are still two full validation stacks processing the chain correctly.
+
+There may need to be additional considerations the Conductor makes in order to determine failover,
+but these are not well defined yet. For example, if the Supervisor of the active Sequencer went down,
+it may be prudent to switch the active Sequencer to one with a functional Supervisor.
```
zhwrd, line 150: If the supervisor goes down, this would not necessarily result in unsafe head stall correct? What are the implications of the active sequencer's supervisor going down & does the impact change over length of downtime?

protocol/interop-topology.md
axelKingsley: There are ~3 arrangements, but among them I think this "Full Stack" arrangement is best for each "availability group", at least internally.

- Full Stack Availability Group (One Supervisor + N Nodes, multiple times): The highest quantity of hosts, but really good redundancy, and easy to reason about.
- Multi-Node (One Supervisor + N*m Nodes): This is what we tried to get working on previous devnets, and while it is a mostly stable feature in the happy path, we are still uncovering situations in the unhappy paths where this needs to be carefully managed.
- Standard Mode: I wrote about the modes [here](https://www.notion.so/oplabs/Supervisor-Modes-Disambiguation-1bcf153ee1628084aeb0e52d5e1b9fc6), which describes Standard Mode. Standard Mode is currently *not implemented*


In short: yes I agree, a similar (identical) setup for RPC and utility nodes would be best.

protocol/interop-topology.md
axelKingsley: Correct, if the supervisor goes down it will not impact the Unsafe Head, as that is totally local and gossip based.

The impacts of a Supervisor going down (on Sequencers) look like:
- Immediately: all cross-safety halts
- Immediately: all L1 traversal and derivation stops (because the supervisor feeds the managed nodes the L1)
- Batcher can likely still collect the unsafe data and publish to L1
- Output roots won't be posted because no one is deriving from L1
- After 12 hours, due to no new Safe Heads being discovered, the nodes may revert to the deposit-only system, but I think what would actually happen is that the node would stall altogether since it can't get L1 data in the first place.
- Any nodes who do their own L1 derivation (like a Standard Mode Node, which is not yet implemented) may be able to follow the batched safe data, and theoretically could post output roots too. But if the Sequencer itself isn't following the batches, this feels like a risky situation to be in.

TLDR: if the Supervisor is down only the Local Unsafe chain will advance, and after the 12h window presumably everything will stop.

axelKingsley: Another thing to note -- if the Supervisor is unavailable, the Sequencer should be denying interop transactions anyway. So, if it checks with the Supervisor at reasonable times, then it should notice that it shouldn't publish an interop-inclusive block, and will drop those transactions.

zhwrd: so in that sense for interop liveness, it is important to ensure supervisor is up and we would likely want to do a leadership transfer if the supervisor goes down on the active sequencer?

protocol/interop-topology.md
```diff
@@ -0,0 +1,217 @@
...
+
+It is safe to not check at block building time because if the check passes at ingress
+time and passes again at block building time, it does not exhaustively cover all cases.
+It is possible that the remote reorg happens **after** the local block is sealed.
+In practice, it is far more likely to have an unsafe reorg on the remote chain that is not
+caught **before** the local block is sealed because the time between checking at the mempool
+and checking at block building is so small. Therefore we should not add a remote RPC request
+as part of the block building happy path, but we should still implement and benchmark it.
+
+TODO: include information on recommendation around standard mode/multi node/running multiple supervisors.
```
tynes, line 180: Just noting this todo in case you were planning on coming back to it as part of this doc or a future doc

protocol/interop-topology.md
```diff
@@ -0,0 +1,217 @@
...
+Should include information on the value props of each so we can easily align on roadmap.
+
+# Alternatives Considered
+
+<!-- List out a short summary of each possible solution that was considered.
+Comparing the effort of each solution -->
+
+## Block Building
+
```
tynes, line 189: The biggest uncertainty around doing the checks in the block builder is for `rbuilder`, when it runs in a TEE. Technically, its possible to block transactions with a firewall before they reach the block builder, so the transactions would need to be encrypted to the TEE itself. I also think that the chain would need to validate a TEE proof alongside batches, because I do not think there is a way to guarantee the private key for the batcher isn't in 2 locations - inside and outside of the TEE. See https://github.com/ethereum-optimism/design-docs/pull/229 for a design that would enable validating a TEE proof onchain

We should descope the concept of using a TEE completely until we are able to end to end understand the security model

tynes:
This looks great to me, thank you @axelKingsley! I cannot approve it given that I opened the PR, but would if i could

protocol/interop-topology.md
axelKingsley: Ah, this should be as a separate doc, as topology *without* Standard or Multi is the current recommendation, and I would like to lay out these options later in another doc. I'll remove this TODO.

protocol/interop-topology.md
axelKingsley: I would say further that the scope of this document applies to what OP Labs plans on doing with our infrastructure, and strictly speaking is a matter of policy. Since we don't run `rbuilder`, I don't see much to consider.

+1: an operator using a TEE will need to decide on Supervisor checks at times that make sense to them

tynes: This does not mean that its not useful to check at the block builder level, we have that feature implemented in `op-geth`. We may ultimately delete that code or keep it around and use it as a fallback or explicitly only use it for local devnets. We need to ship the code and measure its performance to truly know the ideal architecture

protocol/interop-topology.md
axelKingsley: Right, and I think that's something we don't have built out yet. Optimistically, if the Supervisor going down affects the Sequencer in a way that impacts its own liveness, existing transfer mechanisms would kick in. But pragmatically, we should probably add some specialized checks.

I'll add this to the document soon, this is a great idea, thanks!

zhwrd (approved):

protocol/interop-topology.md
```diff
@@ -0,0 +1,211 @@
...
+We can update `proxyd` to validate interop messages on cloud ingress. It should check against both the indexed
+backend of `op-supervisor` as well as the alternative backend.
+Because interop transactions are defined by their Access List,
+`proxyd` does not have to execute any transactions to make this request.
+This filter will eliminate all interop transactions made in bad faith, as they will be obviously invalid.
+
+It may be prudent for `proxyd` to wait and re-test a transaction after a short timeout (`1s` for example)
+to allow through transactions that are valid against the bleeding edge of chain content. `proxyd` can have its own
+`op-supervisor` and `op-node` cluster specifically to provide cross safety queries without putting any load on other
+parts of the network.
```
jelias2, line 69: Currently proxyd is used to typically query EL backends, so in the context of having an "op-supervisor and op-node cluster" would this actually refer to a op-supervisor and "op-EL" cluster? 

Is there any information that must be queried from op-node specifically that is not available from an EL client?

protocol/interop-topology.md
```diff
@@ -0,0 +1,212 @@
...
+Therefore, we should arrange the nodes like so:
+
+|             | Chain A | Chain B | Chain C |
+|------------|---------|---------|---------|
+| Supervisor 1 | A1      | B1      | C1      |
+| Supervisor 2 | A2      | B2      | C2      |
+| Supervisor 3 | A3      | B3      | C3      |
+
+In this model, each chain has one Conductor, which joins all the Sequencers for a given network. And each heterogeneous group of Sequencers is joined by a Supervisor.
+This model gives us redundancy for both Sequencers *and* Supervisors. If an entire Supervisor were to go down,
```
jelias2, line 126: How would the sequencer determine if the active supervisor is lagging and should be switched to a new supervisor?

protocol/interop-topology.md
```diff
@@ -0,0 +1,212 @@
...
+- Managing multiple Nodes sync status from a single Supervisor is tricky -- you have to be able to replay
+all the correct data on whatever node is behind, must be able to resolve conflicts between reported blocks,
+and errors on one Node may or may not end up affecting the other Nodes. While this feature has some testing,
+the wide range of possible interplay means we don't have high confidence in Multi-Node as a redundancy solution.
+- Multi-Node is *only* a Node redundancy solution, and the Supervisor managing multiple Nodes is still a single
+point of failure. If the Supervisor fails, *every* Node under it is unable to sync also, so there must *still*
+be a diversification of Node:Supervisor. At the point where we split them up, it makes no sense to have higher quanitities
+than 1:1:1 Node:Chain:Supervisor.
+
+# Risks & Uncertainties
```
jelias2, line 196: One risk scenario I could see is if a person/machine had direct access to the sequencer RPC, if they were able to spam the direct sequencer endpoint with invalid interop transactions, these would only be checked on the 'low frequency all mempool check', which could result in lots of invalid blocks being produced

protocol/interop-topology.md
```diff
@@ -0,0 +1,212 @@
...
+We really need to measure everything to validate our hypothesis on the ideal architecture.
+To validate the ideal architecture, we need to measure it and then try to break it.
+
+Incorrect assumptions, or unexpected emergent behaviors in the network, could result in validation not happening at the right times,
+causing excessive replacement blocks. Conversely, we could also fail to reduce load on the block builder, still leading to slow
+block building or stalls.
+
+Ultimately, this design represents a hypothesis which needs real testing before it can be challenged and updated.
+
+# Action Items from this Document
```
jelias2, line 207: What metrics would be most important to validate the performance of the system? 
Some that come to mind.

- Amount of tx filtered by each level (ingress, mempool ingress, mempool polling)
- supervisor validation response time
- # Invalid interop blocks produced
- how close supervisor can maintain to chain tip. ( In some monitoring scenarios the minimal resolution is a 1m)
- Tx delays, how much overhead does this add to the current tx flow

protocol/interop-topology.md
axelKingsley: Totally, a malicious privileged actor could create a replacement block by jumping over the Tx filters. But they couldn't do more than one replacement at a time (since all the other invalid tx they submit would get wiped out at the same time).

protocol/interop-topology.md
axelKingsley: These are excellent yes!

- Number of Tx filtered at each stage would be very educational
- Supervisor validation response time we already have a ticket for measuring and adding to dashboards
- When an invalid block is produced, i would expect some amount of `Error` alerting, it would be scary if this were common enough to track as a metric
- Determining supervisor tip tracking, I think we would just run health monitors on the Managed Nodes
- Hard to determine Tx Delay if we don't have a sense of the Tx latency on non-interop transactions.

protocol/interop-topology.md
axelKingsley: Yes, except the supervisor does not need to query the EL/CL in realtime, it has already indexed the data it will use for response.

PR #222, opened 2025-03-19
design doc: interop monitoring

tynes:
Monitoring service is a critical service for detecting and responding to potential invalid messages across the superchain.

### DIFF
--- a/protocol/interop-monitoring.md
+++ b/protocol/interop-monitoring.md
@@ -13,23 +13,23 @@ Monitoring Service for Executing Messages.
 ## Summary + Problem Statement + Context
 
 Given assumptions in the [cloud topology](https://github.com/ethereum-optimism/design-docs/pull/218),
-it is generally not possible to guarantee that invalid `ExecutingMessage`s do not finalize without multiple
+it is generally not possible to guarantee that invalid `Executing Message`s do not finalize without multiple
 implementations of `op-supervisor`. With only a single implementation, a bug becomes consensus.
 In the worst case, this can mint an infinite amount of ether. Given this risk, we need to have monitoring,
-alerting, and a runbook for handling invalid `ExecutingMessage`s being included in the chain.
+alerting, and a runbook for handling invalid `Executing Message`s being included in the chain.
 
-We want to be alterted when there is an invalid `ExecutingMessage`. We are implementing preventative
-measures, but the downside risk is existential if an invalid `ExecutingMessage` finalizes,
+We want to be alterted when there is an invalid `Executing Message`. We are implementing preventative
+measures, but the downside risk is existential if an invalid `Executing Message` finalizes,
 so we need to have ways to detect and prevent that.
 
 ## Proposed Solution
 
-We should implement a monitoring service that validates all of the `ExecutingMessage` logs
-produced by the entire cluster and validates them against transaction access lists
+We should implement a monitoring service that validates all of the `Executing Message` logs
+produced by the entire Superchain and validates them against transaction access lists
 and remote nodes. We use this service to alert oncall engineers as well as potentially automatically
-pausing the batcher/transaction ingress if an invalid `ExecutingMessage` is included.
+pausing the batcher/transaction ingress if an invalid `Executing Message` is included.
 
-This "Cross Message Monitor" should have the following features:
+This "Executing Message Monitor" should have the following features:
 
 ### Monitoring Strategies like `dispute-mon`
 
@@ -40,16 +40,15 @@ can refer to in order to determine network health:
 - How many Incorrect Forecasts or Incorrect Results
 - Warning and Error Logs from the Monitor
 
-Cross Message Monitor can crib directly from these statistics, but focused on Interop:
+Executing Message Monitor can crib directly from these statistics, but focused on Interop:
 - How many `Executing Message`s are emitted by the `CrossL2Inbox` per block per chain
 - How many `Executing Message`s Messages point at each Chain in the dependency set
 - How many `Executing Message`s are known valid, per safety level
 - How many `Executing Message`s are known invalid, per safety level
 - How many `Executing Message`s are not yet known valid/invalid, per safety level
 - How many `Executing Message`s *changed validity* over time (indicating remote reorg)
-- How many `Executing Message`s were resolved via Block Replacement
 
-Almost all `Executing Message` metrics emitted by the Cross Message Monitor should have dimensions:
+Almost all `Executing Message` metrics emitted by the Executing Message Monitor should have dimensions:
 - What chain the `Executing Message` in question is on
 - What chain the `Executing Message` is referring to (the chain of the initiating message)
 - Timestamp of Block
@@ -61,9 +60,9 @@ data is not allways sufficiently available to validate `Executing Message`s, and
 cause cascades of Valid/Invalid messages.
 
 Therefore, it is insufficent to check a message just once. Instead, every Executing Message
-detected by the Cross Message Monitor will be considered an ongoing process, like games are
+detected by the Executing Message Monitor will be considered an ongoing process, like games are
 for the Dispute Monitor. From the time the `Executing Message` is discovered, until the `Executing Message` is included by a
-Cross Safe block height which is now L1 finalized, the `Executing Message` should be repeatedly re-checked.
+Cross-Safe block height which is now L1 finalized, the `Executing Message` should be repeatedly re-checked.
 
 This means that when the status of the `Executing Message` flips, special alerts can be emitted to indicate
 a remote reorg has likely occured. Or, when a single invalid message creates a cascade of
@@ -74,11 +73,11 @@ invalidation, each `Executing Message` can resolve individually.
 The [access list](https://github.com/ethereum-optimism/design-docs/blob/9e919c5b173fe8fc89949b012f6f70a0bc3247f6/protocol/interop-access-list.md)
 design guarantees the fact that all executing messages can be validated without the need to execute the transaction. Any calls to the `CrossL2Inbox`
 that do not include the statically declared executing message in the access list will revert rather than needing to be dropped. This prevents
-a denial of service attack where the MEV searcher can simply produce an invalid `ExecutingMessage` after their MEV attempt fails.
+failing Interop Transactions from putting unpaid load onto the block builder.
 
 Given that the decided upon approach depends strictly on the current EVM resource pricing via storage slot cost introspection, we should have
-monitoring to alert us if someone is able to trick the `CrossL2Inbox` into producing an `ExecutingMessage` when the access list entry is
-not declared. We think this is impossible, but given this is such a critical security property, it is important to monitor.
+monitoring to alert us if someone is able to trick the `CrossL2Inbox` into producing an `Executing Message` when the access list entry is
+not declared or differs from the Executing Message. We think this is impossible, but given this is such a critical security property, it is important to monitor.
 
 Each message can be checked for this once, when it is detected and added to the monitoring set.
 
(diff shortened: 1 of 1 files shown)


protocol/interop-monitoring.md
```diff
@@ -0,0 +1,132 @@
...
+by the sequencer offchain.
+
+We may also want to consider a way to alert partners in the interop set ahead of time that an unsafe head reorg is coming
+if an invalid `ExecutingMessage` is observed in an unsafe block. If they turn off their cross chain message ingress fast enough,
+it could be possible that they can prevent a contingent reorg. The liveness of the chain can continue with no issues until the
+remote chain goes through its unsafe head reorg, then it can open up its cross chain message ingress again.
+
+#### Safe/Finalized Blocks
+
+If an invalid `ExecutingMessage` ends up in a safe block, that means that a bug becomes consensus (without multiclient).
```
axelKingsley, line 94: I'm going to be taking an editorial pass on this doc, but just wanted to point out this is not the case -- invalid messages may appear in Locally Safe blocks, but if they are in *Cross Safe* blocks there is a bugged consensus.

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Alert Behaviors
+
+Though it will need evaluation over time, we already know the sorts of operator responses we want when certain situations are detected
+by the monitor.
+
+#### Unsafe Blocks
+
+We want to be able to detect when an invalid `ExecutingMessage` is included in an unsafe block and trigger an altert to the
+oncall engineering team. Additionally, we should consider pausing the batcher automatically if an invalid `ExecutingMessage`
```
tynes, line 93: The batcher pause existed to prevent consecutive rate-limited mints of ether. Now that we don't have interoperable assets, I do not think that we need the batcher pause

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,134 @@
...
+has been detected in an unsafe block and triggering an unsafe head reorg. It is preferable to not waste blobs and trigger
+an unsafe head reorg by batch submitting the invalid block. The problem with this is that it is indistinguishable from
+a malicious sequencer triggering a reorg to extract MEV. Either way an unsafe head reorg is going to happen, its just whether
+or not its due to the data being posted and then resulting in a replacement deposits only block or if its manually done
+by the sequencer offchain.
+
+We may also want to consider a way to alert partners in the interop set ahead of time that an unsafe head reorg is coming
```
tynes, line 100: We probably want an admin api on `proxyd` that can turn off the interop ingress

protocol/interop-monitoring.md
axelKingsley: :+1: can reduce this, but I'll leave in words that mention this is still an option to the operatior.

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+This document is meant to align on a strategy for monitoring interop and propose a
+Monitoring Service for Executing Messages.
+
+## Summary + Problem Statement + Context
+
+Given assumptions in the [cloud topology](https://github.com/ethereum-optimism/design-docs/pull/218),
+it is generally not possible to guarantee that invalid `ExecutingMessage`s do not finalize without multiple
+implementations of `op-supervisor`. With only a single implementation, a bug becomes consensus.
+In the worst case, this can mint an infinite amount of ether. Given this risk, we need a monitoring,
+alerting and runbook for handling invalid `ExecutingMessage`s being included in the chain.
```
geoknee, line 19:
```suggestion
In the worst case, this can mint an infinite amount of ether. Given this risk, we need to have monitoring,
alerting, and a runbook for handling invalid `ExecutingMessage`s being included in the chain.
```

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+
+We want to be alterted when there is an invalid `ExecutingMessage`. We are implementing preventative
+measures, but the downside risk is existential if an invalid `ExecutingMessage` finalizes,
+so we need to have ways to detect and prevent that.
+
+## Proposed Solution
+
+We should implement a monitoring service that validates all of the `ExecutingMessage` logs
+produce by the entire cluster and validates them against transaction access lists
```
geoknee, line 28:
```suggestion
We should implement a monitoring service that validates all of the `ExecutingMessage` logs
produced by the entire cluster and validates them against transaction access lists
```

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+Dispute Monitor is a service already implemented and deployed for tracking Fault Proof Disputes.
+Rather than just be a simple alert when games are invalid, it serves up various statistics that an operator
+can refer to in order to determine network health:
+- How many games are being monitored, how many of each status
+- How many Incorrect Forecasts or Incorrect Results
+- Warning and Error Logs from the Monitor
+
+Cross Message Monitor can crib directly from these statistics, but focused on Interop:
+- How many Executing Messages emitted by the CrossL2Inbox per block per chain
+- How many `Executing Message`s Messages point at each Chain in the Superchain
```
geoknee, line 45:
```suggestion
- How many `Executing Message`s Messages point at each Chain in the dependency set
```

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+Executing Messages can change validity over the course of the Unsafe Chain,
+data is not allways sufficiently available to validate `Executing Message`s, and transitive `Executing Message`s can
+cause cascades of Valid/Invalid messages.
+
+Therefore, it is insufficent to check a message just once. Instead, every Executing Message
+detected by the Cross Message Monitor will be considered an ongoing process, like games are
+for the Dispute Monitor. From the time the `Executing Message` is discovered, until the `Executing Message` is included by a
+Cross Safe block height which is now L1 finalized, the `Executing Message` should be repeatedly re-checked.
+
+This means that when the status of the `Executing Message` flips, special alerts can be emitted to indicate
```
geoknee, line 68: We presumably do not want to alert for the case where the message begins "unknown" and changes to "valid", as this is the happy path?

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+### Alert Behaviors
+
+Though it will need evaluation over time, we already know the sorts of operator responses we want when certain situations are detected
+by the monitor.
+
+#### Unsafe Blocks
+
+We want to be able to detect when an invalid `ExecutingMessage` is included in an unsafe block and trigger an altert to the
+oncall engineering team. It is preferable to not waste blobs and trigger an unsafe head reorg by batch submitting the invalid block as soon as possible,
+therefore the operator may want to accelerate batch submission when this alert arrives. Unless nodes on the network are
```
geoknee, line 94: This is not currently easy to do with the batcher. But I don't think it is a big lift to add an admin API to submit as soon as possible, or as soon as a block with a certain height is added to a channel.

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+#### Unsafe Blocks
+
+We want to be able to detect when an invalid `ExecutingMessage` is included in an unsafe block and trigger an altert to the
+oncall engineering team. It is preferable to not waste blobs and trigger an unsafe head reorg by batch submitting the invalid block as soon as possible,
+therefore the operator may want to accelerate batch submission when this alert arrives. Unless nodes on the network are 
+able to accept an Unsafe->Unsafe block replacement (and they are not), the Sequencer's only path forward is to see the 
+invalid block commited to L1, at which point it will be replaced. Doing this faster will minimize reorg sizes.
+
+We may also want to consider a way to alert partners in the interop set ahead of time that an unsafe head reorg is coming
+if an invalid `ExecutingMessage` is observed in an unsafe block. If they turn off their cross chain message ingress fast enough,
```
geoknee, line 99: If we want it to be fast, shouldn't we consider automating it and not requiring a human in the loop? They can still be alerted, but the evasive manoeuvre happens regardless

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+it could be possible that they can prevent a contingent reorg. The liveness of the chain can continue with no issues until the
```
geoknee, line 100: Does a "contingent reorg" mean a secondary reorg caused by an executing message referencing an initiating message which was in an orphaned block? It seems like the worst case here is a never ending cycle of contingent reorgs.

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+- Shut off Executing Message Ingress at `proxyd`
+- Force remove Executing Messages from block builder mempools.
+These triggers should occur automatically when an invalid `ExecutingMessage` is discovered at the Unsafe Block stage, in order to reduce cascades.
+
+#### Safe/Finalized Blocks
+
+If an invalid `ExecutingMessage` ends up in a safe block, it is an expectation of the Protocol that the block is Invalid,
+and must be replaced with a Deposit Only Block. This situation should page the operator to monitor the situation, and every
+individual invalid `Executing Message` in a Safe Block should be very easy to see and monitor individually. The operator is monitoring
+to ensure a Block Replacement occurs and the invalid messges are no longer known to the chain.
```
geoknee, line 113:
```suggestion
to ensure a Block Replacement occurs and the invalid messages are no longer part of the canonical chain.
```

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+
+If Cross-Validation should promote the block to Cross-Safe, this is an all-hands-on-deck consensus bug, which would naturally
+have its own alerts associated in addition to the prior expectation of an operator monitoring the situation.
```
geoknee, line 116: Presumably the runbook would say this should be escalated and the bridge paused (unless it can be established that the invalid message is benign)?

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Resource Usage
+
+This new service will need minimal CPU/Disk and can be stateless. It will need a connection to one of each Node
```
geoknee, line 120: If it is to be stateless, doesn't this make it difficult to fully track the status of an x-msg (executing message) over time, if (for example) the service is restarted?

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+for the Superchain it is monitoring.
```
geoknee, line 121: Be good to add more detail about how many `xmsg-mon` instances should be deployed in total, and to use the "full validation stack" terminology if appropriate.

For example, should the nodes the `xmsg-mon` connects to be managed by the same supervisor (I guess yes)?

protocol/interop-monitoring.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Alternatives Considered
+
+No real alternatives considered. Monitoring should happen as a matter of course when deploying new services.
+
+Having additional Cross-Validation software besides Supervisor would lessen the criticality of this software.
+
+## Risks & Uncertainties
+
+- The Monitoring Service may be insufficent, and we may not catch what we need to. Real experience will inform updates to this service.
```
geoknee, line 139:
```suggestion
- The Monitoring Service may be insufficent, and we may not catch what we need to. Real experience will inform updates to this service.
- The Monitoring Service may cause a lot of RPC traffic and generate a lot of data, putting strain on the infrastructure.
```

geoknee (approved):

protocol/interop-monitoring.md
axelKingsley: "cascade" or "recursive" may have been better words, but yes I think you've got it: one reorg causes messages to become invalid. reliance on those now-invalidated messages creates more reorgs, and repeat...

The solution is an AutoStop mechanism for self protection. I think there's some argument to be made that these never ending cycles are actually very difficult to sustain, BUT with AutoStop the expectation is that chains won't experience more than one reorg per invalid message incident.

https://github.com/ethereum-optimism/design-docs/pull/287/files

protocol/interop-monitoring.md
axelKingsley: Yes, that is one weakness as designed, the service should be running to collect new jobs in the first place, and would drop ongoing jobs if restarted.

The question can be expanded to: even if persisted, if the Monitoring Service is restarted, must it also backfill the gap?

I could see this going either way- the complexity of persisting and backfilling jobs seems like it would make this a lot more work to build. But, the risk of having blind or incomplete metrics is scary! Maybe there is a middle-ground where we can run monitors in redundancy and be tolerant of them failing.

PR #223, opened 2025-03-20
FMA for withdrawals root in block header

geoknee:
Adds an FMA for the "withdrawals root in block header" feature, planned for inclusion in the Isthmus hardfork.

### DIFF
--- a/security/fma-withdrawals-root-in-block-header.md
+++ b/security/fma-withdrawals-root-in-block-header.md
@@ -87,6 +87,24 @@ Below are references for this project:
 - **Recovery Path(s)**:
   The changes to the genesis state would need to be reverted and rescheduled into a hardfork if still desired. Chains which were sequenced with the modified genesis sate logic may need to be repaired with a special hardfork.
 
+### FM3: Failure of p2p network due to bug in new topic/message serde logic
+
+- **Description:**
+  Because this feature introduces a new p2p gossip topic and message serialization format, a bug can mean the failure of p2p gossip for any chain with Isthmus active. This would cause an unsafe chain halt.
+
+- **Risk Assessment:**
+
+  **High impact, low likelihood**
+
+  **Mitigations:**
+  We rely on end-to-end testing (including fuzzing) to catch any bugs in this code path. We could run extended fuzzing campaigns.
+
+- **Detection:**
+  Continuous integration, or Kurtosis and/or devnet testing should catch this. Failing that, the bug makes it to production, our alerting infrastructure would notify us.
+
+- **Recovery Path(s)**:
+  The bug would need to be patched and new op-node release cut and rolled out.
+
 ## Generic failure modes:
 
 See the [generic FMA](./fma-generic-hardfork.md):
--- a/security/fma-withdrawals-root-in-block-header.md
+++ b/security/fma-withdrawals-root-in-block-header.md
@@ -32,36 +32,31 @@ Below are references for this project:
 
 ## Failure Modes and Recovery Paths
 
-### FM1: Chain split due to divergent `withdrawalsRoot`
+### FM1: Fault proofs system failure do to inaccurate `withdrawalsRoot`
 
 - **Description:** 
-  If different clients compute a different `withdrawalsRoot` for the same block, this will cause a chain split because it is a consensus parameter.
-- **Risk Assessment:**
-  High impact, low likelihood.
-  **Mitigations:**
- 
-- **Detection:** 
-  Replica healthcheck would fire P1 alerts after a few minutes.
-- **Recovery Path(s)**:
-
-
-### FM2: Fault proofs system failure do to inaccurate `withdrawalsRoot`
-
-- **Description:** 
-  Even if all clients agree and there is no chain split, if the `withdrawalsRoot` is incorrect, critical infra used to propose, challenge and validate fault proof games may fail.
-  Because the information is present both in state and in the block header, there is the opportunity for an "off by one" error to arise. 
+  Even if all clients agree and there is no chain split, if the `withdrawalsRoot` is incorrect, critical infra used to propose, challenge and validate fault proof games may fail when it is switched to use the `withdrawalsRoot` header field instead of querying the information from an archive node in the usual way.  Because the information is present both in state and in the block header, there is the opportunity for an "off by one" error to arise, if there is confusion  about whether the root applies to the state before or after the block is applied. If there are any problems with activation, this would also cause the same issue.
 - **Risk Assessment:**
 
   **Mitigations:**
+  Fault proofs infrastructure (i.e. `op-challenger`) should be run against the newly populated `withdrawalsRoot` field in a testing environment, to confirm it still functions and agrees with a legacy-configured system.
 
 - **Detection:** 
+  Fault proof monitoring systems would detect this failure mode.
 
 - **Recovery Path(s)**:
+  Fault proof infrastructure would need to be patched. The pre-signed pause may need to be invoked. 
 
 
 ### Generic items we need to take into account: 
-## Action Items
+See the [generic FMA](./fma-generic-hardfork.md):
+* Chain halt at activation
+* Activation failure 
+* Invalid setImplementation execution
+* Chain split (across clients)
 
+## Action Items
+- [ ]   Fault proofs infrastructure (i.e. `op-challenger`) should be run against the newly populated `withdrawalsRoot` field in a testing environment, to confirm it still functions and agrees with a legacy-configured system.
 
 ## Audit Requirements
 
--- a/security/fma-withdrawals-root-in-block-header.md
+++ b/security/fma-withdrawals-root-in-block-header.md
@@ -36,7 +36,7 @@ Below are references for this project:
 
 ## Failure Modes and Recovery Paths
 
-### FM1: Withdrawals downtime do to inaccurate `withdrawalsRoot`
+### FM1: Withdrawals downtime do to inaccurate `withdrawalsRoot` in the block header
 
 - **Description:**
   If the `withdrawalsRoot` in the block header is incorrect, critical infra used to enable withdrawals may fail. Namely, output proposals and challenges would be incorrect, affecting chains with permissioned and chains with permissionless proofs. This is because these components will, with the activation of the Isthmus fork, use the `withdrawalsRoot` header field instead of querying the information from an archive node in the usual way. Output roots are returned by the op-node [`optimism_outputAtBlock`](https://docs.optimism.io/operators/node-operators/json-rpc#optimism_outputatblock) RPC method, and this behaves differently under Isthmus -- when handling a request for an output root (it no longer delegates an `eth_getProof` to op-geth and just reads the information from the block header).
--- a/security/fma-withdrawals-root-in-block-header.md
+++ b/security/fma-withdrawals-root-in-block-header.md
@@ -107,11 +107,11 @@ See the [generic FMA](./fma-generic-hardfork.md):
 - [x] (BLOCKING): We have implemented extensive unit and end-to-end testing of the activation flow: https://github.com/ethereum-optimism/optimism/blob/develop/op-e2e/actions/upgrades/isthmus_fork_test.go
 - [ ] (BLOCKING): We have implemented multi-client testing with kurtosis and/or devnets to reduce the chance of bugs. This should be in the form of an acceptance tests which target all client types in the network.
 - [ ] (BLOCKING) We should ensure that our usual suite of alerts applies to devnets and are routed to protocol engineers signing off on the devnet completion.
+- [x] (BLOCKING): Run fuzzing on the v4 gossip p2p more than 10s (assignee: @Ethnical @geoknee) https://github.com/ethereum-optimism/optimism/pull/15068
+- [ ] (BLOCKING): We will be testing the activation on our [devnets](https://devnets.optimism.io/) and testnets.
+- [ ] (non-BLOCKING): Making sure, that component that now fetch the withdrawalHash by this have also a fallback mechanism to previous Merkle Tree method in case there is a bug (assignee: @XXXXX for now).
+- [ ] (non-BLOCKING): Creating a monitoring that differential testing from the merkle tree inclusion computation and the block.header request `withdrawalRoot`(assignee: @Ethnical)](https://github.com/ethereum-optimism/design-docs/pull/223/files)https://github.com/ethereum-optimism/design-docs/pull/223/files
 - [ ] (non-BLOCKING): We have implemented fuzz testing in a kurtosis multi-client devnet to reduce the chance of bugs
-- [ ] (BLOCKING): We will be testing the activation on our devnets and testnets.
-- [ ] (NON-BLOCKING): Making sure, that component that now fetch the withdrawalHash by this have also a fallback mechanism to previous Merkle Tree method in case there is a bug (assignee: @XXXXX for now).
-- [ ] (NON-BLOCKING): Creating a monitoring that differential testing from the merkle tree inclusion computation and the block.header request `withdrawalRoot`(assignee: @Ethnical)](https://github.com/ethereum-optimism/design-docs/pull/223/files)https://github.com/ethereum-optimism/design-docs/pull/223/files
-- [ ] (BLOCKING): Run fuzzing on the v4 gossip p2p more than 10s (assignee: @Ethnical @geoknee)
 
 ## Audit Requirements
 
--- a/security/fma-withdrawals-root-in-block-header.md
+++ b/security/fma-withdrawals-root-in-block-header.md
@@ -98,7 +98,7 @@ Below are references for this project:
   **High impact, low likelihood**
 
   **Mitigations:**
-  We rely on end-to-end testing (including fuzzing) to catch any bugs in this code path. We could run extended fuzzing campaigns.
+  We rely on [end-to-end testing](https://github.com/ethereum-optimism/optimism/blob/9249efc6343208f69283290fc9c5c8f6e7b243f8/op-service/eth/ssz_test.go#L251) (including fuzzing) to catch any bugs in this code path. We could run extended fuzzing campaigns.
 
 - **Detection:**
   Continuous integration, or Kurtosis and/or devnet testing should catch this. Failing that, the bug makes it to production, our alerting infrastructure would notify us.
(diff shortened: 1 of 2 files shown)


security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Generic failure modes: 
+See the [generic FMA](./fma-generic-hardfork.md):
+* Chain halt at activation (there is a change to the engine API, which elevates this risk)
+* Activation failure
+* Invalid setImplementation execution
+* Chain split (across clients)
+
+## Specific Action Items
+- [ ] (BLOCKING) e2e tests must check for consistency between output roots returned from op-node and those constructed manually in the old way
```
geoknee, line 78: We currently check consistency of withdrawal root itself https://github.com/ethereum-optimism/optimism/blob/6a436fe9ac9acb215b0f4b9f87ccd3832f4d6b72/op-e2e/actions/upgrades/isthmus_fork_test.go#L286-L301 which is probably sufficient

security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,90 @@
...
+- [ ] (non-BLOCKING) op-node could be furnished with an override to make it serve output roots in the legacy fashion; this would also aid in testing (see above item). It would even allow us to run the two systems side by side for a time before fully switching over.
+- [ ] (non-BLOCKING) op-geth could be made to log a critical error triggering an alert if ever the `withdrawals` list in the block body is non empty (post Isthmus)
+
+## Generic Action Items
+- [x] (BLOCKING): We have implemented extensive unit and end-to-end testing of the activation flow: https://github.com/ethereum-optimism/optimism/blob/develop/op-e2e/actions/upgrades/isthmus_fork_test.go
+- [ ] (BLOCKING): We have implemented multi-client testing to reduce the chance of bugs (the above test could be migrated to a fault proof test where it can run on kona)
```
geoknee, line 84: Kurtosis / devnet multiclient testing might be a better / lower lift way of performing the multi client testing.

security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,90 @@
...
+|                    |                                                    |
+| ------------------ | -------------------------------------------------- |
+| Author             | George Knee                                        |
+| Created at         | 2025-03-20                                         |
+| Initial Reviewers  | Mark Tyneway                                       |
+| Need Approval From | Tom Assas                                          |
+| Status             | DRAFT                                              |
+
+## Introduction
+
```
tynes, line 25: > making it a part of the information stored by all (including non-archive) nodes

This is generally called history

tynes:
A couple additional cases I can think of:

- A bug with the creation of a genesis block creates a chain split from the genesis. The spec says that the genesis block should have the genesis account root, which may change over time if the predeploy is modified in ways that change its initial storage. This sort of change would happen by a different team, they should not be able to break it through changing the smart contract.
- The p2p gossip is incorrect. A new topic and p2p message serialization is introduced. There could be non deterministic encoding/decoding of the binary format resulting in the p2p network breaking. We do have test coverage of this

geoknee:
Thanks @tynes I added two more failure modes, PTAL.

security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,134 @@
...
+- **Detection:**
+  Fault proof monitoring systems may not detect this failure mode immediately, until an actor running patched software made a proposal or challenge.
+
+- **Recovery Path(s)**:
+  Fault proof infra would nee to be pointed at a patched op-node. The patch would restore the old behaviour for generating output roots.
+
+### FM2: Chain split from genesis due to future changes to genesis state tooling
+
+- **Description:**
+  The genesis state is, with the activation of Isthmus at genesis, now part of the genesis block. Therefore changes to the tooling which generates the genesis block from the genesis state can cause a chain split if there is a bug introduced in the future or even if the genesis state is changed intentionally but in an uncoordinated manner.
```
Ethnical, line 76: I would love clarification on this, for me this case is related to the fact that only "old" nodes are causing a chain split with new node since some features are not included? but somehow seems to be not only related to this feature right? 
any new feature that is included would cause the same behavior. 
I am not sure to follow here, if this possible to have a more detail recap?

geoknee: I agree, I think the chain split problem is:

1. Not specific to this feature, since modifying the genesis state can cause a chain split anyway since the state root would change.
2. Not likely to come up in production since the genesis is usually created once and then never modified. @tynes can you think of a scenario (outside of testing) where the genesis is regenerated?

geoknee: Got it, clarified here [e2e8d16](https://github.com/ethereum-optimism/design-docs/pull/223/commits/e2e8d1646baf430189c8d66965a7fe9ee31024dd)

Ethnical: Oh thank for clarifying here!

security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,133 @@
...
+
+- **Risk Assessment:**
+
+  **High impact, low likelihood**
+
+  **Mitigations:**
+  This should be caught by cross-client testing (kurtosis and devnets)
+
+- **Detection:**
+  If the bug made it to production, replica healthcheck alerts would fire as nodes diverged.
```
Ethnical, line 86: This is clearer, but I would like to know at what moment we will receive this alerts in the process. 
Can we just add at what moment the node would be "tested" and then will fire alerts?

security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,133 @@
...
+
+- **Recovery Path(s)**:
+  The buggy client(s) would need to be patched to bring them back into compliance with the specification.
+
+### FM3: Failure of p2p network due to bug in new topic/message serde logic
+
+- **Description:**
```
Ethnical, line 93: Just to make sure here, this is only **availability** issue correct?

security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,133 @@
...
+  Because this feature introduces a new p2p gossip topic and message serialization format, a bug can mean the failure of p2p gossip for any chain with Isthmus active. This would cause an unsafe chain halt.
+
+- **Risk Assessment:**
+
+  **High impact, low likelihood**
+
+  **Mitigations:**
+  We rely on end-to-end testing (including fuzzing) to catch any bugs in this code path. We could run extended fuzzing campaigns.
```
Ethnical, line 101: Can we show the current fuzzing testing written linked here.

Ethnical:
Just to be sure @geoknee:
This is not possible to have a integrity issue correct?
Since we can still recompute the merkle tree inclusion of this `withdrawalHash` right?
If yes, I think this is worth noting somewhere in this FMA.

security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,133 @@
...
+- [ ] (non-BLOCKING) op-node could be furnished with an override to make it serve output roots in the legacy fashion; this would also aid in testing (see above item). It would even allow us to run the two systems side by side for a time before fully switching over.
+- [ ] (non-BLOCKING) op-geth could be made to log a critical error triggering an alert if ever the `withdrawals` list in the block body is non empty (post Isthmus)
+
+## Generic Action Items
+
+- [x] (BLOCKING): We have implemented extensive unit and end-to-end testing of the activation flow: https://github.com/ethereum-optimism/optimism/blob/develop/op-e2e/actions/upgrades/isthmus_fork_test.go
+- [ ] (BLOCKING): We have implemented multi-client testing to reduce the chance of bugs (the above test could be migrated to a fault proof test where it can run on kona)
+- [ ] (non-BLOCKING): We have implemented fuzz testing in a kurtosis multi-client devnet to reduce the chance of bugs
+- [ ] (BLOCKING): We will be testing the activation on our devnets and testnets.
+
```
Ethnical, line 130:
```suggestion
- [ ] (NON-BLOCKING): Making sure, that component that now fetch the withdrawalHash by this have also a fallback mechanism to previous Merkle Tree method in case there is a bug.
- [ ] (NON-BLOCKING): Creating a monitoring that differential testing from the merkle tree inclusion computation and the block.header request (assignee: @Ethnical)
```

security/fma-withdrawals-root-in-block-header.md
Ethnical: As @geoknee mentioned and I completely agree we should add this one at generic FMA for HF for all the next HF in the future.

security/fma-withdrawals-root-in-block-header.md
Ethnical: This value was already set but not used in the P2P directly since this was the same value hardcoded for any node no reason to use it. 

```
baseFeePerGas        199783
difficulty           0
extraData            0x00000000fa00000006
gasLimit             60000000
gasUsed              7819168
hash                 0x35c7d6036dff00e6f070b5a287dec35936c775bd5fa34130c30ab8e27ef7c1d9
logsBloom            0x0082a8010000001000400000000002000100000800000110000c00002000010000000000002000000014001020000000004404000400220000010140402c02400000000000080028020000080010000020000080104000200000000080000408002140000204040002200100010008000000010020400400000000520008300000000020100210000000c000000000019000040100210000000100102204000002100000100000000018000001000010800041100002001000400000000000000100100700000000000000000002000000000114001010000000000202042880251040080204010040428c000000010080002000001208400400080000002040
miner                0x4200000000000000000000000000000000000011
mixHash              0xd61da8e428ed9083e49da38e0d2a3b75020924651d6c5b676dd33abb834d1ef6
nonce                0x0000000000000000
number               133703028
parentHash           0xbae595baeb8cc927e1b379b9c3d0c5f40cf0d687ab184638bb9cfd8f27a6206a
parentBeaconRoot     0x630622c87f0cf9d5f2f5c5cd53086dc45a2e49622f622de1075be855d161d9cc
transactionsRoot     0x5cca6a9f842561bda798d2d521299c8c4ef8e62ca10535e0bb4e2fc6d61545ce
receiptsRoot         0x1b13433bcc449f3ea476a2be021577cb3774d9260efde0feec29ea2a1bfc3d44
sha3Uncles           0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347
size                 10843
stateRoot            0x7edacdce97732010e1c3e5325664456c9ff287634a0bde900f7b38c9a72ee68a
timestamp            1743004833 (Wed, 26 Mar 2025 16:00:33 +0000)
withdrawalsRoot      0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421
totalDifficulty
blobGasUsed          0
excessBlobGas        0
requestsHash
transactions:        [

```

security/fma-withdrawals-root-in-block-header.md
Ethnical, line 101:
```suggestion
  We rely on [end-to-end testing](https://github.com/ethereum-optimism/optimism/blob/9249efc6343208f69283290fc9c5c8f6e7b243f8/op-service/eth/ssz_test.go#L251) (including fuzzing) to catch any bugs in this code path. We could run extended fuzzing campaigns.
```

security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,133 @@
...
+- **Risk Assessment:**
+
+  **High impact, low likelihood**
+
+  **Mitigations:**
+  We rely on [end-to-end testing](https://github.com/ethereum-optimism/optimism/blob/9249efc6343208f69283290fc9c5c8f6e7b243f8/op-service/eth/ssz_test.go#L251) (including fuzzing) to catch any bugs in this code path. We could run extended fuzzing campaigns.
+
+- **Detection:**
+  Continuous integration, or Kurtosis and/or devnet testing should catch this. Failing that, the bug makes it to production, our alerting infrastructure would notify us.
```
Ethnical, line 104: This would lead to unsafe head stall, how we can make sure here that engineers can detect this issue?
with the error handling in the logs?

security/fma-withdrawals-root-in-block-header.md
geoknee: We have this acceptance test /Users/georgeknee/code/ethereum-optimism/optimism/op-acceptance-tests/tests/isthmus/withdrawal_root_test.go

security/fma-withdrawals-root-in-block-header.md
tynes: we do have alerting in prod for observing chain splits. the chances of a bug that come from a chainsplit after observing a few correct blocks is very small, because it would imply something about the state root calculation is incorrect, which would result in a chain split anyways. so i think just seeing that a few blocks with different withdrawal roots across multiple clients should give confidence that this will not result in a chain split

security/fma-withdrawals-root-in-block-header.md
tynes: a devnet-sdk test should suffice for this, which we already have one

tynes:
This generally looks good to me, understanding https://github.com/ethereum-optimism/design-docs/pull/223#discussion_r2014590043 or being sure that we would be able to detect an unsafe head stall caused by this would be good

security/fma-withdrawals-root-in-block-header.md
geoknee: Do you mean this test https://github.com/ethereum-optimism/optimism/blob/develop/op-acceptance-tests/tests/isthmus/withdrawal_root_test.go ? I think it needs to be strengthened to make it a true multi-client test.

security/fma-withdrawals-root-in-block-header.md
geoknee: As discussed on discord, we can use a combination of:
1. Routing alerts from devnets and doing manual inspections of metrics and
2. Adding an acceptance test which validates no chain split after a few blocks

security/fma-withdrawals-root-in-block-header.md
geoknee: We should route alerts from devnets somewhere that engineers see them.

geoknee: Added an action item for this! Please resolve this thread if you are happy @Ethnical

security/fma-withdrawals-root-in-block-header.md
geoknee: Agreed, we can detect this as described here https://github.com/ethereum-optimism/design-docs/pull/223#discussion_r2015020862 . There is an action item to cover this now. 

@Ethnical please resolve if you are happy

security/fma-withdrawals-root-in-block-header.md
geoknee: It is done [d72d299](https://github.com/ethereum-optimism/design-docs/pull/223/commits/d72d299d1c825984652d894102fa63a4457f8d56). Please resolve if happy @Ethnical

security/fma-withdrawals-root-in-block-header.md
geoknee: Have added action items for these, @Ethnical please resolve the thread if you are happy.

security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,118 @@
...
+- [ ] (BLOCKING): We have implemented multi-client testing with kurtosis and/or devnets to reduce the chance of bugs. This should be in the form of an acceptance tests which target all client types in the network.
+- [ ] (BLOCKING) We should ensure that our usual suite of alerts applies to devnets and are routed to protocol engineers signing off on the devnet completion.
+- [ ] (non-BLOCKING): We have implemented fuzz testing in a kurtosis multi-client devnet to reduce the chance of bugs
+- [ ] (BLOCKING): We will be testing the activation on our devnets and testnets.
+- [ ] (NON-BLOCKING): Making sure, that component that now fetch the withdrawalHash by this have also a fallback mechanism to previous Merkle Tree method in case there is a bug (assignee: @XXXXX for now).
+- [ ] (NON-BLOCKING): Creating a monitoring that differential testing from the merkle tree inclusion computation and the block.header request `withdrawalRoot`(assignee: @Ethnical)](https://github.com/ethereum-optimism/design-docs/pull/223/files)https://github.com/ethereum-optimism/design-docs/pull/223/files
```
geoknee, line 113: @Ethnical why the link back to this PR?

security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,118 @@
...
+- [ ] (non-BLOCKING) op-geth could be made to log a critical error triggering an alert if ever the `withdrawals` list in the block body is non empty (post Isthmus)
+
+## Generic Action Items
+
+- [x] (BLOCKING): We have implemented extensive unit and end-to-end testing of the activation flow: https://github.com/ethereum-optimism/optimism/blob/develop/op-e2e/actions/upgrades/isthmus_fork_test.go
+- [ ] (BLOCKING): We have implemented multi-client testing with kurtosis and/or devnets to reduce the chance of bugs. This should be in the form of an acceptance tests which target all client types in the network.
+- [ ] (BLOCKING) We should ensure that our usual suite of alerts applies to devnets and are routed to protocol engineers signing off on the devnet completion.
+- [x] (BLOCKING): Run fuzzing on the v4 gossip p2p more than 10s (assignee: @Ethnical @geoknee) https://github.com/ethereum-optimism/optimism/pull/15068
+- [ ] (BLOCKING): We will be testing the activation on our [devnets](https://devnets.optimism.io/) and testnets.
```
geoknee, line 111:
```suggestion
- [x] (BLOCKING): We tested the activation on our [devnets](https://devnets.optimism.io/interop-rc-alpha.html).
```

security/fma-withdrawals-root-in-block-header.md
Ethnical: Thanks!

security/fma-withdrawals-root-in-block-header.md
Ethnical: Correct this is an error! Thanks!

security/fma-withdrawals-root-in-block-header.md
```diff
@@ -0,0 +1,118 @@
...
+
+Below are references for this project:
+
+- [Spec](https://specs.optimism.io/protocol/isthmus/exec-engine.html#l2tol1messagepasser-storage-root-in-header)
+- [op-geth PR](https://github.com/ethereum-optimism/op-geth/pull/451)
+- [monorepo PR](https://github.com/ethereum-optimism/optimism/pull/13962)
+
+## Failure Modes and Recovery Paths
+
+### FM1: Withdrawals downtime do to inaccurate `withdrawalsRoot`
```
geoknee, line 39:
```suggestion
### FM1: Withdrawals downtime do to inaccurate `withdrawalsRoot` in the block header
```

Ethnical (approved):

Ethnical:
Thanks for the hardwork from @geoknee! 
He was really clear in his explanation to help us to cover most of the scope!

PR #226, opened 2025-03-21
feat: hashed cross domain submarine sends

hamdiallam:
**Description**

Provide a public utility to conduct hashed cross domain sends, primarily for DeFi applications.

### DIFF
--- a/ecosystem/submarine-sends.md
+++ b/ecosystem/submarine-sends.md
@@ -4,32 +4,29 @@ Superchain interop enables new low-latency DeFi actions across chains. The proto
 
 ## Goals
 
-- Cross chain DeFi actions do not expose the user to additional MEV risk
-- 3rdparty can relay submarine transactions to remove friction for users
+- Cross chain DeFi actions do not expose the user to additional 3rdparty risk
 
 ## Non Goals
 
--
+- Masquerade submarine sends from regular transfer of funds.
 
 # Problem Statement + Context
 
 When using the `L2ToL2CrossDomainMessenger`, the target address and calldata is immediately made public when the message is sent. The outbox on the source chain becomes a visible mempool of transactions for a given destination.
 
-For many use cases, this can be fine but in DeFi can undermine fairness, due to front running. [Submarine Sends](https://libsubmarine.org/) is a way to mitigate these kinds of problems, where the transaction is made private until executed in a commit-reveal scheme.
-
-[Submarine Sends](https://libsubmarine.org/) also obfuscates the the flow of funds by generating a fresh address the funds are escrowed apart of the commitment phase. Making it impossible differentiate between a simple eth send and a user privately interacting with a dapp.
+For many use cases, this can be fine but in DeFi can undermine fairness, due to front running. [Submarine Sends](https://libsubmarine.org/) is a way to mitigate these kinds of problems, where the transaction is made private until executed in a commit / reveal structure.
 
 # Proposed Solution
 
-This is not an important feature to replicate for Superchain interactions as the `L2ToL2CrossDomainMessenger` is the single entrypoint for the native flow of funds across all Superchain L2s. As long as the target contract and calldata is made hidden, a user observing outgoing messages wont be able to differentiate between the the different types of submarine sends.
+We can take inspiration from by replicating a commit / reveal scheme for data sent between chains. By providing a standard interface for submarine sends, cross chain transactions can be made between contracts where the intent remains hidden until the reveal on the destination chain.
 
-- TODO: maybe note that people can detect which messages are submarine sends prior to the reveal
+Unlike [Submarine Sends](https://libsubmarine.org/) which also provides the ability to make the commit transaction indistinguishable from a normal value transfer, the funds and commitment have to be propogated through the `L2ToL2CrossDomainMessenger` which can allow for an external entinty to differentiate between a hashed cross domain submarine send and not. However this is not a problem since DeFi attacks rely on understanding the intent of the transaction which we show remains masked.
 
-By replicating a commit-reveal format for interoperable transactions, user can privately send and relay messages between chains.
+- Maybe there's some additional design where the destination escrow address is fresh on every call, achieving the same property but that is left out of scope for now.
 
 ## Commitment
 
-The commitment is simply the sensitive information in the outbound `SentMessage` which is the `target` and `message`. For submarine sends where the same target & message can be used more than once, salt can be used to ensure unique commitments.
+The commitment for a submarine send is the sensitive information in the outbound `SentMessage` event, `target` and `message`. For submarine sends where the same target & message can be used more than once, a salt can be provided to ensure unique commitments per transaction.
 
 ```solidity
 bytes memory payload = abi.encodePacked(salt, target, message)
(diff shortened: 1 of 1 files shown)


ecosystem/submarine-sends.md
```diff
@@ -0,0 +1,167 @@
...
+
+For many use cases, this can be fine but in DeFi can undermine fairness, due to front running. [Submarine Sends](https://libsubmarine.org/) is a way to mitigate these kinds of problems, where the transaction is made private until executed in a commit / reveal structure.
+
+# Proposed Solution
+
+We can take inspiration by replicating a commit / reveal scheme for data sent between chains. Enabling cross chain transactions between contracts where the intent remains hidden until the reveal on the destination chain.
+
+Unlike [Submarine Sends](https://libsubmarine.org/) which also provides the ability to make the commit transaction indistinguishable from a normal value transfer, the funds and commitment have to be accounted for and propogated through the `L2ToL2CrossDomainMessenger` which can allow for an external entinty to differentiate between a hashed cross domain submarine send and not. However this is not a problem since DeFi attacks rely on understanding the intent of the transaction which we show remains masked.
```
simplyoptimistic, line 23: nit: `entinty` -> `entity`

ecosystem/submarine-sends.md
```diff
@@ -0,0 +1,167 @@
...
+## Commitment
+
+The commitment for a submarine send is the sensitive information in the outbound `SentMessage` event, `target` and `message`. For submarine sends where the same target & message can be used more than once, a salt can be provided to ensure unique commitments per transaction.
+
+```solidity
+bytes memory payload = abi.encodePacked(salt, target, message)
+bytes32 commitment = keccak256(payload)
+```
+
+By posting just the commitment, **only the the submitting user with knowledge of the preimage would be able to reveal the transaction** when relaying. If this user encounters issues between submission and relay, the preimage for that data is also lost.
```
pedrovalido, line 36: nit: `only the the submitting` -> `only the submitting`

ecosystem/submarine-sends.md
```diff
@@ -0,0 +1,167 @@
...
+3. Reveal the payload with the submarine contract. The commitment hash can be recomputed and checked.
+   ```solidity
+   function reveal(bytes32 commitment, uint256 salt, address target, bytes calldata message) external
+   ```
+
+## Cross Domain Submarine
+
+By sending the commitment hash between L2s via the L2ToL2CrossDomainMessenger, we gain the ability to privately send transactions. However, submarine sends are typically accompanied with funds with the transaction. i.e Bridge and Swap.
+
+The SuperchainTokenBridge & ETH bridge are designed as single cross chain messages. Thus the cross domain submarine contract needs to be able to receive and escrow sent funds prior to excuting the reveal transaction. Thus, a cross domain submarine transaction requires 2 relays on the destination
```
pedrovalido, line 79: Nit: `prior to excuting` -> `prior to executing`

ecosystem/submarine-sends.md
```diff
@@ -0,0 +1,167 @@
...
+
+Making the right assertions is already an error-prone process and we're adding some complexity for developers. However this complexity is likely worth the cost since a developer must be intentionally aware how submarine functionality works when integrating their dapp & contracts.
+
+## Commitment Uniqueness
+
+The salt can ensure commitments are always unique. However it is the responsibility of the dapp to ensure this. It's already unlikely in DeFi that calldata might be the exact same as they depend on token inputs and minimum expected outputs. And even if they are the same, they must be pending during the same time window.
+
+However an attacker that realizes that a dapp is using the same salt could try to fontrun some expected (target,message) combos with submarine sends that where the target always fails. All this would achieve in the current sample code snippet is halting delivery of submarine sends with the same commitment (already unlikely).
```
simplyoptimistic, line 165: nit: `fontrun` -> `frontrun`

ecosystem/submarine-sends.md
```diff
@@ -0,0 +1,167 @@
...
+
+By sending the commitment hash between L2s via the L2ToL2CrossDomainMessenger, we gain the ability to privately send transactions. However, submarine sends are typically accompanied with funds with the transaction. i.e Bridge and Swap.
+
+The SuperchainTokenBridge & ETH bridge are designed as single cross chain messages. Thus the cross domain submarine contract needs to be able to receive and escrow sent funds prior to excuting the reveal transaction. Thus, a cross domain submarine transaction requires 2 relays on the destination
+
+1. Send and escrowed funds in the submarine contract
+2. Relay message the propogated commitment and details about funds owed to the target.
+
+Sample snippet on how this can work for ETH. Generalizable to SuperchainERC20s
+
```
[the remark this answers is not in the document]
hamdiallam: good q!

Yep just like how apps integrate with the SuperchainTokenBridge or L2ToL2CrossDomainMessenger, they can integrate with this. Probably would have a handler contracts on both sides that integrates with the DEX

This design retains the same ordering properties of the token bridge and messenger. There is no enforced ordering. Relayers can relay messages in whatever order they would like.

Since the submarine sender specifies the relayer ahead of time (labs likely would provide a public goods one). I would imagine here we would prolly relay in order as much as possible. However this would not be enshrined in the sumbarine contract.

If you wanted to enforce ordering though, you could do that in the handler contract. what are your thoughts?

ecosystem/submarine-sends.md
```diff
@@ -0,0 +1,167 @@
...
+The commitment for a submarine send is the sensitive information in the outbound `SentMessage` event, `target` and `message`. For submarine sends where the same target & message can be used more than once, a salt can be provided to ensure unique commitments per transaction.
+
+```solidity
+bytes memory payload = abi.encodePacked(salt, target, message)
+bytes32 commitment = keccak256(payload)
+```
+
+By posting just the commitment, **only the the submitting user with knowledge of the preimage would be able to reveal the transaction** when relaying. If this user encounters issues between submission and relay, the preimage for that data is also lost.
+
+To solve for this, the payload is encrypted and posted as calldata to the submitting transaction. By using ECDH, Elliptic Curve Diffie-Hellman, the data can be encrypted against any public key, allowing for the user to designate a 3rdparty to conduct the relay on their behalf.
```
marktoda, line 38: I'd suggest leaving this preimage posting unspecified. If a user wants to post it encrypted onchain they can, but many implementations may prefer an out of band dissemination

ecosystem/submarine-sends.md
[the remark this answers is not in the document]
simplyoptimistic: Ideally ordering should be enforced on the protocol level because any sort of cross chain action that involves an initial token transfer followed by some defi action would have this race condition, meaning the protocol would have to implement this logic. It could be as simple as having certain messages be contingent on a transfer (so until your token transfer passes, we cannot progress). I think this is something hyperlane is going to implement. 

Re: the submarine send contract, it would make sense if this could be wrapped together with an interchain account router type feature, as it would mean that crosschain features that require no new contracts could just involve a crosschain transfer followed by multicalls to execute the remainder of the defi logic.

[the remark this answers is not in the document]
hamdiallam: > because any sort of cross chain action that involves an initial token transfer followed by some defi action would have this race condition

Maybe I misunderstood. The ordering for this part is enforced. The hashed defi action cannot occur before the funds are available in the destination. And this can be batched together so that it does not require separate txs.

Ordering for that is enforced here:
```solidity
   // Ensure delivery before marking the commitment as pending
   require(messenger.successfulMessages(sentETHMsg));
```

The ordering that is not enforced is between two unrelated submarine actions.

Can you elaborate on the interchain account feature here? The sender is agnostic of any particular `target`. So it would be compatible with any crosschain action

ecosystem/submarine-sends.md
hamdiallam: yea that's a good point! This all can be encoded as opaque bytes represented with 1 argument. Everything else should be good

ecosystem/submarine-sends.md
airtoonricardo, line 23: nit: `propogated` => `propagated`

PR #227, opened 2025-03-24
feat: l2tol2cdm RelayedMessage return value

hamdiallam:
**Description**

Increase leverage of the L2ToL2CDM by including the return value in the RelayedMessage event

### DIFF
--- a/ecosystem/l2tol2cdm-relayedmessage-return.md
+++ b/ecosystem/l2tol2cdm-relayedmessage-return.md
@@ -1,0 +1,85 @@
+# Purpose
+
+The ability to read events with Superchain interop is incredibly powerful. We should make sure we maximize the potential of all the default events emmitted when interacting with the OP-Stack predeploys.
+
+## Goals
+
+- Faciliate cross chain async programming
+
+# Problem Statement + Context
+
+As we horizontally scale, contract development will look more asynchronous. A lot of cross chain messages are predominalty write-based.
+
+- Bridge Funds
+- Invoke calldata on via a remotely controlled proxy account
+
+Read-based cross chain messages are less common but will increasingly become so in an asynchronous world. Solidity has a `view` modifier specifically targeted for off chain reading -- it would be natural to also want to also query this state onchain very quickly.
+
+- Fetch remote swap quotes
+- Query balance/ownership status
+- Chain on subsequent write actions, only after success (deposit -> borrow -> brige)
+
+In order to support use cases like these, custom handler contracts per-application must be written to capture these side effects and return data backwards or chain on additional side effects. For example
+
+```solidity
+contract ERC20BalanceGateway {
+    L2ToL2CrossDomainMessenger messenger;
+
+    // Gateway contract that can be used to query ERC20s with a specified handler
+    function query(IERC20 token, address account, bytes4 selector) external {
+        uint256 balance = token.balanceOf(account);
+
+        // Message sent back with the value as an encoded argument
+        (address sender, uint256 source) = messenger.crossDomainMessageContext();
+        messenger.sendMessage(source, sender, abi.encode(selector, balance));
+    }
+}
+```
+
+# Proposed Solution
+
+When the `L2ToL2CrossDomainMessenger` invokes a target with calldata, the return data of that call is captured and returned within the `relayMessage` function. However this return data is only actionable for the caller of this function, typically the relayer, or a faciliating smart contract.
+
+`relayMessage` emits the `RelayedMessage` event on successful execution which itself can be consumed remotely via the `CrossL2Inbox` predeploy. By including the return data in this event, it becomes immediately visible to the sending chain (and all others in the interop set) without any additional message or handler contract.
+
+```solidity
+contract L2ToL2CrossDomainMessenger {
+    function relayMessage(Identifier calldata _id, bytes calldata _sentMessage) {
+        // include return data
+        (, returnData_) = target.call{ value: msg.value }(message);
+        emit RelayedMessage(source, nonce, messageHash, returnData_);
+    }
+}
+```
+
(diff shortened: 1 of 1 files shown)


tynes:
This is missing the [resource usage](https://github.com/ethereum-optimism/design-docs/blob/main/assets/design-doc-template.md#resource-usage) section. How will this impact the happy path (no returndata) vs worst case (returndata bomb) in gas costs and database size?

ecosystem/l2tol2cdm-relayedmessage-return.md
```diff
@@ -0,0 +1,85 @@
...
+- Faciliate cross chain async programming
+
+# Problem Statement + Context
+
+As we horizontally scale, contract development will look more asynchronous. A lot of cross chain messages are predominalty write-based.
+
+- Bridge Funds
+- Invoke calldata on via a remotely controlled proxy account
+
+Read-based cross chain messages are less common but will increasingly become so in an asynchronous world. Solidity has a `view` modifier specifically targeted for reading -- it would be natural to also want to also query this state onchain very quickly.
```
tynes, line 16: For some usecases, its not safe to read remote data without knowing the timestamp of the read. You could be tricked into reading stale data for some usecases. Is there an easy way to see the timestamp of the read?

tynes:
I recommend adding the table to the header that was recently added to the template: https://github.com/ethereum-optimism/design-docs/blob/main/assets/design-doc-template.md#project-name-design-doc

ecosystem/l2tol2cdm-relayedmessage-return.md
hamdiallam: `Identifier` of the remote log plays in nicely here. The consumer on the source chain can choose to reject the `RelayedMessage` event if the timestamp of the event is stale

Nothing needs to change on the destination.

hamdiallam:
> This is missing the [resource usage](https://github.com/ethereum-optimism/design-docs/blob/main/assets/design-doc-template.md#resource-usage) section. How will this impact the happy path (no returndata) vs worst case (returndata bomb) in gas costs and database size?

ah good catch, let me add this

ecosystem/l2tol2cdm-relayedmessage-return.md
```diff
@@ -0,0 +1,105 @@
...
+contract L2ToL2CrossDomainMessenger {
+    function relayMessage(Identifier calldata _id, bytes calldata _sentMessage) {
+        // include return data
+        (, returnData_) = target.call{ value: msg.value }(message);
+        emit RelayedMessage(source, nonce, messageHash, returnData_);
+    }
+}
+```
+
+The `messageHash` known ahead of time can be correlated with the corresponding `RelayedMessage`. Thus the return value of any read function or return value of a write call becomes immediately actionable for the sender, without requiring special integration.
```
tremarkley, line 65: In order to read/validate the `RelayedMessage`, the identifier of the `RelayedMessage` is a requirement. 
Wouldn't an integration be required for constructing the identifier  of the `RelayedMessage`  since you cannot calculate the identifier ahead of time?

hamdiallam: by ahead of time here, I meants with the return value of `sendMessage`. You can directly chain the continuation since the caller just needs to assert the msg hash matches with the `RelayedMessage` event.

I'll make this a bit more clear

tremarkley: I was referring to the message identifier not being able to be calculated ahead of time and not the message hash. The identifier is needed in order to trigger the continuation, right?

hamdiallam: yea after relay is when the continuation would be invoked. you need the return value

the identifier isnt computed ahead of time. It can just be consumed right away after emission

tremarkley: Yeah and so that identifier and message will need to be fetched or indexed offchain. Do you see this requiring some additional offchain infra to find and relay these events back to the source chain?

tynes:
We generally have consensus on this approach, but will add the hash of the returndata to the event rather than the full returndata itself. The returndata can be unwrapped on the originating chain. The reason that the hash is used is to make the costs more constant, reducing risk for paying a relayer on the local chain for their execution on the remote chain. There is a concern that extra data needs to be posted to the calldata, which puts pressure on DA. This is a valid concern, if its abi encoded returndata, then it likely compresses pretty well, but ultimately depends on the returndata itself

ecosystem/l2tol2cdm-relayedmessage-return.md
hamdiallam: yep, we'll definitely need indexing of pending promises that relayers will have to handle separately. Especially now that the data is hashed.

This plays in nicely for protocols that have to propagate the return data backwards, which incurs extra gas cost

ecosystem/l2tol2cdm-relayedmessage-return.md
```diff
@@ -0,0 +1,105 @@
...
+
+The extra gas cost associated with including a variable sized input into the `RelayedMessage` log. There's a gas cost of 8 per non-zero byte (4 for zeros) in log data. This is acceptable for a couple of reasons
+
+1. Dynamically sized return values are generally discouraged, and it is good practice in solidity to ensure bounds here when writing contracts.
+2. The caller to `relayMessage` is already gas-sensitive to the return since it is already internally captured and returned.
+3. Just as these gas costs are a concern in a single-chain development experience, the developer looking to query a contract cross chain should also be aware of the gas costs of invoking any dynmically sized return values.
+
+### Single Point of Failure and Multi Client Considerations
+
+N/A as this is a smart contract change.
```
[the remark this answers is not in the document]
hamdiallam: Updated!

tynes (approved):

tynes:
@hamdiallam going to merge this, next steps are a specs + source code change

PR #229, opened 2025-03-25
Add design doc for smart batching feature

jbearer:
**Description**

Design document for a feature we call "smart batching", i.e. allowing the batch inbox to be a smart contract for customizable batching or sequencing logic.

**Tests**

No tests, design doc only.

### DIFF
--- a/protocol/smart-batching.md
+++ b/protocol/smart-batching.md
@@ -1,0 +1,78 @@
+# Smart Batching: Design Doc
+
+|                    |                                                    |
+| ------------------ | -------------------------------------------------- |
+| Author             | @jbearer (Espresso Systems)                        |
+| Created at         | 2025-03-25                                         |
+| Initial Reviewers  | Eli Haims, Sam McIngvale                           |
+| Need Approval From |                                                    |
+| Status             | In Review                                          |
+
+## Purpose
+
+Enabling plugin logic for application-specific sequencing or batching behavior without the need to fork the derivation pipeline.
+
+## Summary
+
+Enables chains to deploy a contract at the batch inbox address with custom logic for filtering
+batches. The derivation pipeline is modified to ignore batches where the contract reverts. This
+expands the design space for OP stack rollups running the canonical derivation pipeline.
+
+Example uses:
+
+- TEE-sequencer: L1 enforces that the rollup sequencer runs in a TEE, which enforces additional sequencing policies:
+    - Unichain wants to use a TEE for Flashblocks and verifiable priority ordering. Currently the derivation pipeline cannot check the TEE attestation, so the current implementation is no better than a trusted sequencer. This proposed change would provide an avenue to fix that. It will also be needed for the UVN.
+    - The Celo sequencer will run a TEE that first posts its transaction batches to a fast BFT consensus (run by Espresso Network) and only posts batches to the L1 that were already confirmed by this BFT consensus. This allows users and protocols (exchanges, bridges, etc) to obtain rollup soft confirmations (pre-L1 finality) from the fast BFT consensus that are more secure than confirmations coming from a regular centralized sequencer. Other OP stack chains are interested in adding this protective layer as well.
+- L1 directly verifies that the sequenced batches have been confirmed by an external layer (e.g. Espresso) instead of relying on a TEE.
+- Enabling various decentralized sequencing policies, by deploying an inbox contract with logic for determining which sequencer is eligible to post any given batch (similar to [RFP](https://github.com/ethereum-optimism/ecosystem-contributions/issues/63) that Espresso completed for OP Labs)
+
+## Problem Statement + Context
+
+It is desirable for many chains to customize the behavior of the OP stack. This is always possible
+by forking and modifying the derivation pipeline, but it is desirable to be able to run the canonical stack, with custom behavior as plugins. OP stack natively supports customization of certain aspects of the rollup (such as alt-DA layers) but lacks sufficient customizability to deploy new batching and sequencing policies.
+
+A simple way for rollups to deploy custom batch posting rules without needing to run a fork of
+validators and RPC nodes is to deploy a *batch inbox* contract on L1, which acts as a pre-filter to
+the derivation pipeline, rejecting any batches that don't satisfy the custom rules while
+allowing the derivation pipeline to ingest the remaining batches and execute them as normal.
+
+Unfortunately, this is not quite possible with the current derivation pipeline. While the inbox
+address can be configured to the address of a contract, the contract cannot stop batches from being ingested by the derivation pipeline, because **the derivation pipeline does not check revert status of transactions**. Thus, even if the contract reverts, the reverting transactions can still end up included in an L1 block, and the derivation pipeline will still parse and execute their calldata or
+blob data.
+
+## Proposed Solution
+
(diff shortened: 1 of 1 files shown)


protocol/smart-batching.md
```diff
@@ -0,0 +1,78 @@
...
+
+Modify the L1 derivation stage of the pipeline (`calldata_source` and `blob_data_source`) to fetch receipts for each L1 block they process, and skip any transaction where the `status` field of the receipt is 0 (reverted transactions).
+
+### Compatibility
+
+The change is backwards compatible. If the inbox address is not a contract, transactions to it will never revert. Thus the only action necessary to opt in is to deploy a contract and upgrade to change the batch inbox address.
+
+### Resource Usage
+
+The derivation pipeline already fetches transaction receipts in several places (e.g. attributes
```
tynes, line 53: I believe this may mean that L1 execution will have to happen in the proof system, to recreate the execution in the L1 inbox to create the receipt. We may be able to get around it by unwrapping the receipts from the receipts root in the header if we have a trusted L1 blockhash, but would want to check with @Inphi @ajsutton @clabby to understand the tradeoffs here

protocol/smart-batching.md
```diff
@@ -0,0 +1,78 @@
...
+derivation) and it caches them when fetched. Thus, adding an additional place where they are fetched does not meaningfully increase the cost of fetching L1 data.
+
+There may be additional costs related to the deployment and operation of the contract, but that is
+rollup-specific and opt-in.
+
+### Single Point of Failure and Multi Client Considerations
+
```
tynes, line 60: `op-node` is a single point of failure. What is the worst case that can happen because of that? Changes to code that are a single source of failure mean that bugs introduced become consensus, unless there is an explicit runbook with social layer buy in around rolling back the chain in case of a bug

protocol/smart-batching.md
```diff
@@ -0,0 +1,78 @@
...
+A simple way for rollups to deploy custom batch posting rules without needing to run a fork of
+validators and RPC nodes is to deploy a *batch inbox* contract on L1, which acts as a pre-filter to
+the derivation pipeline, rejecting any batches that don't satisfy the custom rules while
+allowing the derivation pipeline to ingest the remaining batches and execute them as normal.
+
+Unfortunately, this is not quite possible with the current derivation pipeline. While the inbox
+address can be configured to the address of a contract, the contract cannot stop batches from being ingested by the derivation pipeline, because **the derivation pipeline does not check revert status of transactions**. Thus, even if the contract reverts, the reverting transactions can still end up included in an L1 block, and the derivation pipeline will still parse and execute their calldata or
+blob data.
+
+## Proposed Solution
```
tynes, line 43: Right now its not possible to reconfigure the batch inbox given that existing chains send to an address with no know private key generally. Any thoughts on how to handle this?
- Enable ability to modify batch inbox address dynamically
- Only allow new chains to use this feature?

protocol/smart-batching.md
jbearer: > We may be able to get around it by unwrapping the receipts from the receipts root in the header if we have a trusted L1 blockhash

This was how I imagined it would work. Is there not already something like this in the proof system, since the derivation pipeline already depends on receipts (e.g. for attributes derivation)?

protocol/smart-batching.md
jbearer: Hmm so your saying it's not currently possible to do a chain upgrade to change the inbox address? If not we would definitely want to add this ability

protocol/smart-batching.md
jbearer: Added some brief discussion on this

protocol/smart-batching.md
tynes: It is not currently possible for a chain to change its batch inbox. It is hardcoded as part of the chains genesis config

protocol/smart-batching.md
tynes: Thank you!

protocol/smart-batching.md
jbearer: Gotcha, I'll work on adding this, thanks

jbearer: @tynes it looks like the inbox address is currently read from the system config contract, so we propose simply adding an owner-only function to set this address in the contract. Does that work? [2139117](https://github.com/ethereum-optimism/design-docs/pull/229/commits/2139117c3bff4078ce79bfbe2b450a05393edb06)

protocol/smart-batching.md
```diff
@@ -44,6 +44,8 @@ blob data.
 
 Modify the L1 derivation stage of the pipeline (`calldata_source` and `blob_data_source`) to fetch receipts for each L1 block they process, and skip any transaction where the `status` field of the receipt is 0 (reverted transactions).
 
+To support existing chains opting into this feature, we also make it possible for chains to update their inbox address, by adding an owner-only `setBatchInbox` method to the `SystemConfig.sol` contract. Chains can then use the existing mechanisms for updating the `SystemConfig` to change their batch inbox address to that of a smart contract.
```
tynes, line 47: This will also need to take into account passing the information into the L2 system, so there will likely need to be an [UpdateType](https://github.com/ethereum-optimism/optimism/blob/aec52922e9aa194b5288f2cd466b5799a3d402e5/packages/contracts-bedrock/src/L1/SystemConfig.sol#L28) and the [L1 attributes tx calldata](https://github.com/ethereum-optimism/optimism/blob/aec52922e9aa194b5288f2cd466b5799a3d402e5/packages/contracts-bedrock/src/L2/L1Block.sol#L195) will need to include the batch inbox address. Its not ideal to keep bloating the size of the L1 attributes tx but its a clear way to continue to add config and would result in a larger diff to the proof system if we didn't include it in the calldata

protocol/smart-batching.md
tynes: https://github.com/ethereum-optimism/design-docs/pull/229#discussion_r2042870057

protocol/smart-batching.md
```diff
@@ -0,0 +1,84 @@
...
+expands the design space for OP stack rollups running the canonical derivation pipeline.
+
+Example uses:
+
+- TEE-sequencer: L1 enforces that the rollup sequencer runs in a TEE, which enforces additional sequencing policies:
+    - Unichain wants to use a TEE for Flashblocks and verifiable priority ordering. Currently the derivation pipeline cannot check the TEE attestation, so the current implementation is no better than a trusted sequencer. This proposed change would provide an avenue to fix that. It will also be needed for the UVN.
+    - The Celo sequencer will run a TEE that first posts its transaction batches to a fast BFT consensus (run by Espresso Network) and only posts batches to the L1 that were already confirmed by this BFT consensus. This allows users and protocols (exchanges, bridges, etc) to obtain rollup soft confirmations (pre-L1 finality) from the fast BFT consensus that are more secure than confirmations coming from a regular centralized sequencer. Other OP stack chains are interested in adding this protective layer as well.
+- L1 directly verifies that the sequenced batches have been confirmed by an external layer (e.g. Espresso) instead of relying on a TEE.
```
samlaf, line 26: What does "confirmed" mean here? Are you talking about EspressoDA here or something else? Because one thing we are currently looking at for EigenDA is to verify a certificate of availability in the batcher inbox.

PR #233, opened 2025-03-26
FMA: OP-Supervisor

axelKingsley:
Failure Mode Analysis for `op-supervisor`, a new consensus-critical piece of code for the op-stack.

### DIFF
--- a/security/fma-supervisor.md
+++ b/security/fma-supervisor.md
@@ -114,7 +114,7 @@ And all Failure Modes are some subtype. Incorrect responses are much worse than 
     - The message validity code in the Supervisor is the most core aspect of its implementation. It has reasonable unit testing for database, accessors, apis, and all validity checking code. It has some E2E tests, and some situationally comprehensive Action tests, as well as local Kurtosis and Devnet exposure. However, this component is not battle tested.
     - *Even When* the Supervisor behaves totally correctly, this case may occur if some data cross-unsafe data is used to build a block which later becomes invalid due to a reorg on the referenced chain. In this situation, the same outcome is felt by the network.
 - Mitigations
-    - The Sequencer could detect cross-unsafe head stall and issue a reorg on the unsafe chain in order to avoid invalid L1 inclusions. Depending on the heuristic used, this could create regular unsafe reorgs with low threshold, or larger, less common ones. This also saves operators from wasted L1 fees when a batch would contain unwanted data.
+    - The Sequencer *should* detect cross-unsafe head stall and issue a reorg on the unsafe chain in order to avoid invalid L1 inclusions. Depending on the heuristic used, this could create regular unsafe reorgs with low threshold, or larger, less common ones. This also saves operators from wasted L1 fees when a batch would contain unwanted data.
     - When promoting local-unsafe to cross-unsafe, the Supervisor can additionally detect if the data it is stalled on is already cross-safe or not. If it is, it can proactively notify the Sequencer that the current chain is hopeless to be valid, creating a more eager reorg point.
     - The Batcher can decline to post beyond the current cross-unsafe head. This will avoid the publishing of bad data so the sequencer may reorg it out, saving the replacement based reorg. If it went on long enough, the Batcher would prevent any new data from being posted to L1, effectively creating a safe-head stall until the sequencer resolved the issue. This *could* be a preferred scenario for some chains.
     - We need to develop and use production-realistic networks in large scale testing to exercise failure cases and get confidence that the system behaves and recovers as expected.
@@ -145,6 +145,7 @@ And all Failure Modes are some subtype. Incorrect responses are much worse than 
     - Currently, the code used to issue resets to Managed Nodes is insufficiently tested. This is due to limitations in our ability to construct lifelike-yet-erroneous scenarios for Nodes to sync against. **As far as it has been battle-tested thus**-**far**, Managed Nodes are stable (for example, the “Denver Devnet” has run for 40 days with minimal operations, while supporting real builder traffic).
 - Mitigations
     - If a reset would significantly roll back the Sequencer, a chain with a Conductor Set *should* be able to Identify that the Node is unhealthy and elect a new Active Sequencer. In this case, there would be no interruption to the chain, as the tip is continued by the new Active Sequencer.
+    - We will be cleaning up the number of Node<>Supervisor messages in current operation, which will allow us to hook closer metrics to this. If A Sequencer ever gets a reset signal, it may be worthy of an alert on its own (even if the reset is due to a legitimate reason, they are rare enough to be tracked)
 - Recovery
     - With respect to the Node, an arbitrary amount of re-sync may be required.
     - Recovery to the network depends entirely on the impact of the node outage.
@@ -166,7 +167,7 @@ And all Failure Modes are some subtype. Incorrect responses are much worse than 
 - Description
     - The Supervisor is managing one Managed Node per chain, and using the data reported from their derivation to calculate cross-safety.
     - For some reason, a Managed Node is not able to sync the chain quickly, or at all (perhaps the node is down or is failing)
-    - Without the data from this Node, the Supervisor cannot advance safety for any other chain which depends on this stalled cahin.
+    - Without the data from this Node, the Supervisor cannot advance safety for any other chain which depends on this stalled chain.
     - The Supervisor also won’t be able to answer any validity questions about the un-synced portion of the chain (which is why safety doesn’t advance)
     - Assuming they take some dependency on un-syncing chain, other chains will eventually stall their cross-unsafe and cross-safe chains
 - Risk Assessment
@@ -178,14 +179,26 @@ And all Failure Modes are some subtype. Incorrect responses are much worse than 
         - This feature is mostly ready in the happy path, but there are known gaps in managing secondary Managed Nodes during block replacements and resets.
         - This feature *also* needs much more robust testing in lifelike environments. Previous development cycles were spent in the devnet trying to enable this feature, which was slow and risky. To get this feature working well, we need to leverage our new network-wide testing abilities.
 
+## FM4c: Supervisor has Insufficient Performance
+
+- Description
+    - Like FM4b, updates are not happening on the Supervisor quickly enough and it is falling behind.
+    - In this instance however, it is due to Supervisor performance, not Nodes.
+    - This is a liveness threat to the Supervisor only.
+    - Nodes who rely on the Supervisor may not be able to get all the queries they need answered, leading to protectively dropped interop transactions.
+- Risk Assessment
+    - Low Impact, Low likelihood
+    - The Supervisor does not do strenuous calculations, mostly just DB lookups.
+    - Nodes and their Execution Engines are likely to be the bottleneck because they have to process the Gas of the block.
+
 ## FM5: Increased Operational/Infrastructure Load Results in Fewer Nodes
 
 - Description
-    - In the current design, validating a given chain in the Supechain requires validaiton of *all* chains. Conceptually,
+    - In the current design, validating a given chain in the Supechain requires validation of *all* chains. Conceptually,
     this is unavoidable because inter-chain message validity requires inter-chain validation.
     - The way a user would achieve this today is by running one Node (`op-geth` and `op-node`) for each Chain in the Interopating Set, and would additionally run a Supervisor to manage these Managed Nodes.
     - If operators want redundancy, they are advised to create entire redundant stacks (N nodes and Supervisor)
-    - Operators may not appreciate the increased burdeon and could decline to validate the network at all
+    - Operators may not appreciate the increased burden and could decline to validate the network at all
 - Risk Assessment
     - Sliding Scale Impact and Likelihood
     - Network security is not based on number of validators, but fewer people running the OP Stack means less validation, mindshare, etc.
@@ -194,7 +207,7 @@ And all Failure Modes are some subtype. Incorrect responses are much worse than 
         - In Standard Mode, a Node *is not* managed by a Supervisor, and instead runs its on L1 derivation
         - Periodically, the Node reaches out to *some* trusted Supervisor to fetch the Cross-Heads and any Replacement signals
         - The Node uses this data to confirm that the chain it has derived is also cross-safe, and to handle invalid blocks
-    - With Standard Mode, operators would only *need* to run a single Node, in exchange for the interop aspects of validation being a trusted activity. We expect this Mode to be valuable to offset operator burdeon in cases where fully trustless validation isn't critical.
+    - With Standard Mode, operators would only *need* to run a single Node, in exchange for the interop aspects of validation being a trusted activity. We expect this Mode to be valuable to offset operator burden in cases where fully trustless validation isn't critical.
 
 ## FM6: Standard Mode Nodes Trust a Malicious Supervisor
 
@@ -205,7 +218,7 @@ And all Failure Modes are some subtype. Incorrect responses are much worse than 
     - The result depends on the way in which the data is incorrect:
         - If the data ignores a block Invalidation/Replacement, the Node will be deceived into following a chain where an invalid interop message exists, allowing for invalid interop behaviors (like inappropriate minting) *on that Node* (and presumably all Nodes who also trust this Supervisor)
         - If the data claims a block should be replaced, the Node would similarly perform the replacement, leaving this Node at that state.
-        - If the data simply isn't consistently applicable to what the Node independently derived, the Node has nothing to be decieved about, but also can't make forward progress, and would need to halt at this point.
+        - If the data simply isn't consistently applicable to what the Node independently derived, the Node has nothing to be deceived about, but also can't make forward progress, and would need to halt at this point.
 - Risk Assessment
     - Low Impact, Unknown Likelihood
     - If there were a tactical reason to use a Trusted Sequencer Endpoint to fool a Node as part of a larger exploit, this would be an attractive thing to attempt.
@@ -213,3 +226,24 @@ And all Failure Modes are some subtype. Incorrect responses are much worse than 
     - At most, a dishonest Supervisor can mislead the Nodes connected to it. So long as block producers *are not* using one of these Trusted Endpoints, block production is unaffected.
 - Mitigation
     - Standard Mode could allow for *multiple* Supervisor Endpoints to be specified, they could confirm that all endpoints agree, preventing dishonesty from one party from deceiving the Node.
+
+# Action Item Summary
+
+Across all these Failure Modes, the following are explicitly identified improvements and mitigations we should make soon:
+- Alternative implementations should exist to catch instances where Supervisor has a bug.
+- Higher quality testing (NAT, Test-SDK, DSL) for:
+    - Rigorus syncing behaviors, including node resets and multi-node syncing.
+    - High volume indexing for performance.
+    - Chaos-Monkey style testing for Correctness.
+- Smarter behaviors for existing components:
+    - Batcher to only submit up-to cross-unsafe.
+    - Sequencer to look for cross-unsafe stalls to initiate reorg.
+    - Admin APIs to sideline interop messages during incidents.
+    - Conductor to consider Supervisor liveness for Leadership Transfers.
+    - Standard Mode should be implemented to allow operators to run a single Node to validate.
+        - Standard Mode should check with multiple Supervisors to avoid deception.
+
+When the Supervisor fails, the worst thing it could do is report the wrong information for a given message.
+This incorrect data can mislead a network into building an invalid chain which must be dropped once it is
+known to be invalid. At worst, if no one catches the incorrect data, the network can be mislead into an entirely
+non-canonical state indefinitely, threatening the validity of the network itself.
--- a/security/fma-supervisor.md
+++ b/security/fma-supervisor.md
@@ -91,6 +91,9 @@ And all Failure Modes are some subtype. Incorrect responses are much worse than 
     - Other Node Operators can adopt similar redundancy measures, but may follow their own infrastructure designs . Operators should treat the Supervisor as consensus critical, and manage it similar to how they would manage their L1 source.
     - The impact if a Sequencer's Supervisor goes down and cannot be replaced is an Interop Liveness failure, where the block building
     won't include interop messages. This is higher impact, as it avoidably harms UX by dropping interop messages which were presumably valid.
+- Action Items
+    - Monitor for a lack of Interop Messages in blocks, as an indicator that the Supervisor driving block building is down.
+    Can also be used to monitor for other chains' percieved interop liveness.
 
 ## FM1b: Supervisor Is Corrupted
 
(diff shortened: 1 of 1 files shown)


security/fma-supervisor.md
```diff
@@ -0,0 +1,215 @@
...
+    - *Even When* the Supervisor behaves totally correctly, this case may occur if some data cross-unsafe data is used to build a block which later becomes invalid due to a reorg on the referenced chain. In this situation, the same outcome is felt by the network.
+- Mitigations
+    - The Sequencer could detect cross-unsafe head stall and issue a reorg on the unsafe chain in order to avoid invalid L1 inclusions. Depending on the heuristic used, this could create regular unsafe reorgs with low threshold, or larger, less common ones. This also saves operators from wasted L1 fees when a batch would contain unwanted data.
+    - When promoting local-unsafe to cross-unsafe, the Supervisor can additionally detect if the data it is stalled on is already cross-safe or not. If it is, it can proactively notify the Sequencer that the current chain is hopeless to be valid, creating a more eager reorg point.
```
protolambda, line 118: This is worth prioritizing more

security/fma-supervisor.md
```diff
@@ -0,0 +1,215 @@
...
+    - The Batcher can decline to post beyond the current cross-unsafe head. This will avoid the publishing of bad data so the sequencer may reorg it out, saving the replacement based reorg. If it went on long enough, the Batcher would prevent any new data from being posted to L1, effectively creating a safe-head stall until the sequencer resolved the issue. This *could* be a preferred scenario for some chains.
```
protolambda, line 119: Safety over liveness, I think this is valid to do by default.

That said, we should invest more in the syncing of the unsafe chain, so that low-latency batch-submission is not as important for stability. And the longer we defer batch-submission, the larger the gap in L1 fees and actual batch-costs may be.

security/fma-supervisor.md
```diff
@@ -0,0 +1,215 @@
...
+    - Even Higher Impact, Low Likelihood.
+- Recovery
+    - Within the 12h Safe Head window, if the sequencer is repaired and rewound, it could correctly interpret the batch data to replace a block, and then would have the job of rebuilding the chain from that point. All users would need to upgrade to a version without this derivation bug, and resync the chain from the failed position. Operators who *did not* upgrade and resync would be left on a dead branch that is no longer being updated.
+
+## FM3: The Supervisor Issues a Reset to the Sequencer
+
+- Description
+    - The Supervisor manages Nodes in Managed Mode, meaning they listen to the Supervisor for signals for what derivation activities to take next.
+    - One such activity is to reset the node to specific heads.
+    - Due to some misbehavior, the Supervisor could issue a reset to the Sequencer.
```
protolambda, line 140: I think we should clean up the noisy events between the node and the supervisor.

The event-metrics, and the new metrics capability I implemented in the op-geth library (error codes! method inspection! including RPC subscription metrics) can help identify when something unexpected happens.

security/fma-supervisor.md
```diff
@@ -0,0 +1,215 @@
...
+    - Assuming they take some dependency on un-syncing chain, other chains will eventually stall their cross-unsafe and cross-safe chains
+- Risk Assessment
+    - Low Impact, Low likelihood
+    - The Supervisor knows the boundaries of data it can/can’t report on, and won’t answer *incorrectly* just because it doesn’t have data.
+    - If the Managed Nodes are Sequencers who need interop data to build blocks, they will be unable to validate interop messages, and will therefore not include them in block building
+- Mitigation
+    - Supervisor supports a feature in which *multiple* Nodes of a given chain can be connected as Managed Nodes. If one Node goes down, syncing can continue with the backup.
+        - This feature is mostly ready in the happy path, but there are known gaps in managing secondary Managed Nodes during block replacements and resets.
+        - This feature *also* needs much more robust testing in lifelike environments. Previous development cycles were spent in the devnet trying to enable this feature, which was slow and risky. To get this feature working well, we need to leverage our new network-wide testing abilities.
```
protolambda, line 179: This is definitely an area that lacks testing due to prior prioritizations over this work.

Fortunately, @pcw109550 is starting on some sync testing. And I'm pushing for devnet-sdk to support the sync-testing context (multiple nodes, multi-supervisor, etc.) to support that.

security/fma-supervisor.md
```diff
@@ -0,0 +1,215 @@
...
+
+## FM5: Increased Operational/Infrastructure Load Results in Fewer Nodes
+
```
protolambda, line 182: I think we should add some RPC controls to the sequencer to rate-limit or temporarily entirely cancel inclusion of cross-chain messages, so we can use it if there's ever an issue with load or security.

security/fma-supervisor.md
```diff
@@ -0,0 +1,215 @@
...
+        - If the data claims a block should be replaced, the Node would similarly perform the replacement, leaving this Node at that state.
+        - If the data simply isn't consistently applicable to what the Node independently derived, the Node has nothing to be decieved about, but also can't make forward progress, and would need to halt at this point.
+- Risk Assessment
+    - Low Impact, Unknown Likelihood
+    - If there were a tactical reason to use a Trusted Sequencer Endpoint to fool a Node as part of a larger exploit, this would be an attractive thing to attempt.
+    - Most entities who would provide a Trusted Endpoint have intrinsic incentive to keep their endpoint valid and secure. Much like a provider like Alchemy doesn't want the Node you connect to to lie.
+    - At most, a dishonest Supervisor can mislead the Nodes connected to it. So long as block producers *are not* using one of these Trusted Endpoints, block production is unaffected.
+- Mitigation
+    - Standard Mode could allow for *multiple* Supervisor Endpoints to be specified, they could confirm that all endpoints agree, preventing dishonesty from one party from deceiving the Node.
```
protolambda, line 215: Not viable now, but later with interop, we can potentially include proofs for trustless sync. Since all L2 data is really just some computation over L1 data, which the op-node can verify if they've access to a L1 header chain.

protolambda (commented):
Looks good. Commented on a few of the items, but no blockers.

security/fma-supervisor.md
axelKingsley: I agree, the logs show a lot to clean up. But this FM would exist anyway, if there were a buggy signal, the Supervisor could direct the Sequencer to reset its Engine.

I'll add this to mitigations.

security/fma-supervisor.md
```diff
@@ -0,0 +1,249 @@
...
+To start setting context, here is a basic overview of how Interop works at the protocol level:
+
+- **There are Two Chains, A and B** who participate in interop with one another.
+- Any time a log is emitted on A, it can be referenced on B.
+- And likewise, any logs on B can be referenced on A.
+- The way a chain can reference logs from other chains is by creating an *Executing Message*
+    - Executing Messages are emitted as Logs by the Cross-L2-Inbox contract when called.
+    - Each Executing Message contains indexing information and a hash.
+- By the rules of the protocol, Executing Messages are always valid on the canonical chain because:
+    - If they are valid, then they are valid.
```
teddyknox, line 26: Could you elaborate what you mean here?

security/fma-supervisor.md
```diff
@@ -0,0 +1,249 @@
...
+
+- Local Unsafe: Represents the basic P2P gossiped chain with no interop validity considered.
+- Cross Unsafe: Represents the farthest point within the Local Unsafe chain which optimistically appears to be valid with the Cross Unsafe data of other chains.
+- Local Safe: Represents the basic L1 derived chain with no interop validity considered.
+- Cross Safe: Represents the farthest point within the Local Safe chain which can be validated with Cross Safe data of other chains from this L1 source.
+- Finalization: As always, the L2 data is finalized when the L1 data from which it is derived becomes Finalized. However, Finalization only meaningfully applies to the *Cross Safe* head, because the local safe chain may be discovered invalid post-finalization if data that was awaited turns out to be invalid.
+
+While the protocol rules of interop create a very simple and effective method of keeping interoperation secure, it creates the incentive for chain operators to *not* include these transactions in their blocks. If they do, then the promotion from Local Safe to Cross Safe will fail and their chain will experience a reorg in order to replace the invalid block.
```
teddyknox, line 38: > it creates the incentive for chain operators to *not* include these transactions

it creates the incentive for chain operators to *not* include forged interop transactions?

axelKingsley: "these" -> Invalid Transactions. Will update it.

security/fma-supervisor.md
axelKingsley: I'll do a pass on the text in the FMA too, but what I mean by this is:

An `Executing Message` is a Log Event from this particular Smart Contract. *By the Rules Of the Protocol*, we know **all** `Executing Messages` fit some invariants (ie, they are valid). It is impossible for there to be an EM which *does not* fit the invariant, because if such a thing existed in a block, that entire block itself would be invalid.

Therefore, Executing Messages are either Valid, or if they are not Valid, the Block they exist in is non-canoncial (by way of replacement) anyway, so Invalid Messages don't ever really "exist" in the canonical chain.

"If they are Valid, then they are Valid" + "If they are Invalid, then they don't exist" = "Only Valid EMs Exist"

security/fma-supervisor.md
```diff
@@ -0,0 +1,269 @@
...
+# Action Item Summary
+
+Across all these Failure Modes, the following are explicitly identified improvements and mitigations we should make soon:
+- Alternative implementations should exist to catch instances where Supervisor has a bug.
+- Higher quality testing (NAT, Test-SDK, DSL) for:
+    - Rigorus syncing behaviors, including node resets and multi-node syncing.
+    - High volume indexing for performance.
+    - Chaos-Monkey style testing for Correctness.
+- Smarter behaviors for existing components:
+    - Batcher to only submit up-to cross-unsafe.
```
tynes, line 259: Can you articulate the value proposition of only batch submitting up to cross unsafe? If i understand correctly, cross unsafe means that you have observed the existence of the initiating messages in a block, but they have not been finalized themselves yet. We would reduce risk of batch submitting something that would be reorg'd out

security/fma-supervisor.md
```diff
@@ -0,0 +1,269 @@
...
+
+## FM1a: Supervisor Is Totally Unavailable
+
+- Description
+    - The Supervisor is entirely unavailable, as if it were disconnected.
+    - No Message/Block Validity could be determined for any Node connected to this Supervisor.
+    - Any Node in Managed Mode would be unable to advance their chain state at all besides local-unsafe.
+    - Any Node in Standard Mode would be unable to advance cross-unsafe and cross-safe, but could still advance local-unsafe and local-safe. And because Validity can’t be checked, the Node may prefer not to trust Safe data either.
+- Risk Assessment
+    - Low Impact, High likelihood.
```
K-Ho, line 88: End user impact would be similar to an unsafe head stall: the longer the supervisor is unavailable, the more $ is lost via things like changing on-chain prices making trades no longer profitable, or the delay resulting in users not being able to take time-sensitive actions. Similar to unsafe head stalls, the longer the supervisor is unavailable for, the more loss there is. Responding swiftly here is important and should be treated the same as how we respond to unsafe head stalls currently

security/fma-supervisor.md
```diff
@@ -0,0 +1,269 @@
...
+    - In this scenario, all of the above applies, and the recovery is delayed by our ability to restore a working Supervisor.
+- Risk Assessment
+    - Low Impact, Low likelihood.
+    - The Supervisor has robust database trimming to clear invalid states and partial writes.
+    - This code, like all Supervisor code, could be tested more.
+- Mitigations
+    - There is a feature of the Supervisor that databases can be replicated over HTTP. If a Supervisor needs to be quickly synced it can bootstrap its databases from an existing Supervisor.
+    - Supervisors themselves should be made redundant, so if one goes down, another may serve in its place. Supervisor *replacement* is not something that is well tested, and more likely you’d switch to use that backup Supervisor *and* its Managed Nodes.
+
+## FM2a: Supervisor Incorrectly Determines Message Validity in Response to Sequencer - Unsafe Block Only
```
tynes, line 106: Action item: communicate clearly to applications that depend on unsafe blocks the new risks that come with interop. Across is an example application

security/fma-supervisor.md
```diff
@@ -0,0 +1,269 @@
...
+
+- Description
+    - If the Supervisor were to claim an Executing Message is valid when it actually isn’t, Sequencers may use this determination to include an invalid message in their local-unsafe chain.
+    - Assuming the Supervisor correctly evaluates the cross-promotion (a later step), it will experience a cross-unsafe head stall at this error point, because of the cross-invalid message.
+    - The Sequencer does not care about cross-unsafe chain stalling, because unsafe data is subject to change, and as far as it knows, all the messages it has included are valid, so the *expectation* is that other dependent data will unstick it.
+    - Eventually, the block is published to the L1 and becomes local-safe data.
+    - When the Supervisor attempts to promote the local-safe data to cross-safe, it discovers the invalid message and issues an invalidation and replacement.
+    - The Replacement block is applied to the block which contained the invalid message, and the chain has now reorg’d out all blocks from the invalid message to the safe head (effectively resetting the chain back to the stalled cross-unsafe head).
+- Risk Assessment
+    - Medium Impact, Medium Likelihood.
```
K-Ho, line 116: If a third party bridge or centralized exchange trusts unsafe interop messages, this is a risk. Dangerous scenario:
1. Chain A includes initiating message for "move 1000 ETH to Chain B"
2. Chain B sees the unsafe initiating message
3. Executing message to mint 1000 ETH on Chain B is executed
4. Attacker moves 1000 ETH out to L1 via a third party bridge or CEX which just trusts the unsafe head
5. An unsafe chain reorg happens on Chain A, and the initiating message is reorged out (this could happen maliciously, or even via just an L1 reorg)
6. Attacker has stolen 1000 ETH


Mitigation: Ensure that any applications which enabling moving substantial value *out* of the Superchain understand this risk, and ideally require making sure the chain is safe

security/fma-supervisor.md
```diff
@@ -0,0 +1,269 @@
...
+    - When promoting local-unsafe to cross-unsafe, the Supervisor can additionally detect if the data it is stalled on is already cross-safe or not. If it is, it can proactively notify the Sequencer that the current chain is hopeless to be valid, creating a more eager reorg point.
+    - The Batcher can decline to post beyond the current cross-unsafe head. This will avoid the publishing of bad data so the sequencer may reorg it out, saving the replacement based reorg. If it went on long enough, the Batcher would prevent any new data from being posted to L1, effectively creating a safe-head stall until the sequencer resolved the issue. This *could* be a preferred scenario for some chains.
+    - We need to develop and use production-realistic networks in large scale testing to exercise failure cases and get confidence that the system behaves and recovers as expected.
+
+## FM2b: Supervisor Doesn’t Catch Block Invalidation of Safe Data
```
tynes, line 126: We could come up with a plan to roll back the chain, but this would break exchange integrations. This is critical severity and the most risky part of interop

security/fma-supervisor.md
```diff
@@ -0,0 +1,269 @@
...
+
+- Description
+    - FM2a has occurred, but additionally, the Supervisor doesn’t catch the invalid message when promoting from local-safe to cross-safe.
+    - An output root that builds on this incorrect cross-safe head is published to L1.
+    - At this point, any validators who rely on the Supervisor are following an incorrectly derived chain — blocks between the invalid message and the end of the batch should be replaced by deposit only blocks.
+    - Any validators who *do not* rely on the failing Supervisor will see the correct chain, but there are currently no alternative implementations to use.
+    - An output root posted from this incorrect state would be open to be fault-proven.
+- Risk Assessment
+    - Even Higher Impact, Low Likelihood.
```
K-Ho, line 135: Clarifying from the FMA review, the reason this is low likelihood, while 2a is medium likelihood is because 2a could happen even if there are no bugs (e.g. L1 has a major reorg) whereas 2b should never happen unless there is a bug in the supervisor.

This is the most critical impact of any of these. This can allow a user to mint uncapped amounts of ETH, and quickly exit that ETH from the superchain via CEXes, third party bridges, etc. 
These bridges/cexes look at the safe head, and in this scenario, their checks would pass and enable these withdrawals.
To figure out:
1. How much value exited out of the superchain would this be over how much time?
2. How would we even respond if this happened? somehow halt all chains? pause L2->L1 withdrawals? How would we notify all third party bridges/cexes?

security/fma-supervisor.md
```diff
@@ -0,0 +1,269 @@
...
+- Recovery
+    - Within the 12h Safe Head window, if the sequencer is repaired and rewound, it could correctly interpret the batch data to replace a block, and then would have the job of rebuilding the chain from that point. All users would need to upgrade to a version without this derivation bug, and resync the chain from the failed position. Operators who *did not* upgrade and resync would be left on a dead branch that is no longer being updated.
```
tynes, line 137: Things that can help with this issue:
- multiple clients
- 2 step cross chain messages plus pause button
- monitoring and runbook for triggering a reorg
- only build blocks based on safe cross chain messages
- use sequencer policy to only include ether transfers that are safe

security/fma-supervisor.md
axelKingsley: I did add this in the AIs at the bottom

security/fma-supervisor.md
axelKingsley: Exactly, you got it right. Even publishing `cross unsafe` leaves you open to invalidation depending on the remote chain, but if you haven't yet seen the initiating message you're waiting on, and especially if the remote unsafe chain appears to conflict, the Batcher reduces the risk of batch submitting invalid (to be replaced) data.

I call this out in more detail in the FM2 section where it's first introduced. This is just the summary of AIs at the end.

>  The Batcher can decline to post beyond the current cross-unsafe head. This will avoid the publishing of bad data so the sequencer may reorg it out, saving the replacement based reorg. If it went on long enough, the Batcher would prevent any new data from being posted to L1, effectively creating a safe-head stall until the sequencer resolved the issue. This *could* be a preferred scenario for some chains.

security/fma-supervisor.md
axelKingsley: Just to clarify - the impact here is indeed a `cross unsafe` and beyond stall, but this would *only* be the case for for the individual nodes connected to the affected supervisor. This is not like an unsafe head stall, and rather like your `op-node` lost its L1 connection.

If the Supervisor that goes down happens to be the one a Sequencer is reliant on, the outcome would be that the block builder would stop including interop messages (because it can't validate them), but blocks would still be produced. We could call this something like an "Interop Stall", and it would indeed impact time sensitive interop activities.

I will increase the Impact here and call out the difference in affect when it's the Sequencer.

security/fma-supervisor.md
axelKingsley: Minor clarification - an L1 reorg is actually not a vector for this, because the unsafe chain isn't based in any L1. And from the other perspective, any L1 reorg which changes initiating messages *must also* change the executing messages downstream from it. So L1 reorgs apply pretty transparently to the interop protocol.

security/fma-supervisor.md
axelKingsley: Thanks, I added these as additional considerations.

security/fma-supervisor.md
```diff
@@ -0,0 +1,283 @@
...
+And all Failure Modes are some subtype. Incorrect responses are much worse than no responses, as they may mislead the network. At maximum, a network which is mislead may have fraudulent interop messages played on it, resulting in arbitrary damage to that network. I say “mislead” because all invalid interop messages are invalid by the protocol, and this fact can be checked using L1 data. But because there is limited client diversity, bugs may have greater impact.
+
+## FM1a: Supervisor Is Totally Unavailable
+
+- Description
+    - The Supervisor is entirely unavailable, as if it were disconnected.
+    - No Message/Block Validity could be determined for any Node connected to this Supervisor.
+    - Any Node in Managed Mode would be unable to advance their chain state at all besides local-unsafe.
+    - Any Node in Standard Mode would be unable to advance cross-unsafe and cross-safe, but could still advance local-unsafe and local-safe. And because Validity can’t be checked, the Node may prefer not to trust Safe data either.
+- Risk Assessment
```
smartcontracts, line 87: Recommended action items:
- Monitor for a lack of interop messages, could be done relatively easily with Hexagate

security/fma-supervisor.md
```diff
@@ -0,0 +1,283 @@
...
+
+- Description
+    - FM2a has occurred, but additionally, the Supervisor doesn’t catch the invalid message when promoting from local-safe to cross-safe.
+    - An output root that builds on this incorrect cross-safe head is published to L1.
+    - At this point, any validators who rely on the Supervisor are following an incorrectly derived chain — blocks between the invalid message and the end of the batch should be replaced by deposit only blocks.
+    - Any validators who *do not* rely on the failing Supervisor will see the correct chain, but there are currently no alternative implementations to use.
+    - An output root posted from this incorrect state would be open to be fault-proven.
+- Risk Assessment
+    - Even Higher Impact, Low Likelihood.
+- Recovery
```
smartcontracts, line 138: Action items:
- Improved testing of this codepath across the board
- Implement ETH rate limit
- Implement pause mechanism on Sequencer ingress
- Implement batch submission pause, potentially
- Automated pause based on monitoring

security/fma-supervisor.md
```diff
@@ -0,0 +1,283 @@
...
+    - When promoting local-unsafe to cross-unsafe, the Supervisor can additionally detect if the data it is stalled on is already cross-safe or not. If it is, it can proactively notify the Sequencer that the current chain is hopeless to be valid, creating a more eager reorg point.
+    - The Batcher can decline to post beyond the current cross-unsafe head. This will avoid the publishing of bad data so the sequencer may reorg it out, saving the replacement based reorg. If it went on long enough, the Batcher would prevent any new data from being posted to L1, effectively creating a safe-head stall until the sequencer resolved the issue. This *could* be a preferred scenario for some chains.
+    - We need to develop and use production-realistic networks in large scale testing to exercise failure cases and get confidence that the system behaves and recovers as expected.
```
smartcontracts, line 126: Recommended actions:
- Monitoring for invalid interop messages
- Monitoring for key invariants (ETH supply?)

security/fma-supervisor.md
axelKingsley: For this FM: Interop Messages may still be arriving on the ingress/block-building side, but couldn't be processed by the block builder (so they'd get dropped), and cross-heads would halt.

But yes we could monitor for a suspicious lack of interop messages in blocks and use that as a signal that some part of the system is down, including the Sequencer's Supervisor.

security/fma-supervisor.md
```diff
@@ -0,0 +1,314 @@
...
+- Action Items
+    - Improved testing is required across the board for cross-head promotion, as this is the area most likely to
+    break consensus.
+    - Implement an ETH rate limit which prevents more than X Eth from being transferred in a single transaction.
+    This should limit the blast radius of any undiscovered bugs.
+    - Implement a pause mechanism which declines interop transactions from being included in block building.
+        - This pause mechanism should be triggerable via admin API
+        - And also should be responsive to automated Alert/Monitoring based triggers.
+        - Once paused, unpause should be manual to start with.
+    - Optionally: Implement a matching Batcher pause to avoid publishing invalid blocks in batches where avoidable.
```
smartcontracts, line 158: We need to make a determination on this, possible it will be required, but need to analyze further. Approving for now, it gets the gist across.

smartcontracts (approved):

PR #249, opened 2025-04-01
FMA: Interop Transaction Handling

axelKingsley:


security/fma-interop-tx-handling.md
```diff
@@ -0,0 +1,137 @@
...
+- If an Executing Message is ever Invalid, it is discarded and not retried.
+- *No* Checks are done at Block Building time.
+
+This FMA describes the potential negative consequences of this design. We have selected this design because it maximizes
+the opportunities for Invalid Messages to be caught and discarded, while also leaving the block building hot-path from
+having to take on new compute.
+
+# FMs:
+
+## FM1: Checks Fail to Catch an Invalid Executing Message
```
tynes, line 68: Is this specific to ingress checks?

tynes: I think we should have a section specific for ingress and a section specific for finalization. Both of these problems can have different solutions

tynes:
The framing of two extremes is a great introduction to this topic

security/fma-interop-tx-handling.md
```diff
@@ -0,0 +1,137 @@
...
+    Supervisor.
+    - If this happens, the Supervisor may become slow to respond, slow to perform its other sync duties, or may crash outright.
+    - When a Supervisor crashes, any connected Nodes can't keep their Cross-Heads up to date, and Managed Nodes won't get L1 updates either.
+- Risk Assessment
+    - Medium Impact ; Low Likelihood
+    - The Supervisor is designed to respond to Check requests. Even though it hasn't been load tested in realistic settings, there is very little computational overhead when responding to an RPC request.
+    - Supervisors can be scaled and replicated to serve high-need sections of the infrastructure. Supervisors
+    sync identically (assuming a matching L1), so two of them should be able to share traffic.
+- Mitigations
+    - When the Supervisor is down, any block builder or mempool filter *should* treat unavailability as
```
tynes, line 122: The liveness of cross chain messages is decoupled from the liveness of the chain. Its possible to keep the chain running without the ability to process cross chain messages when the supervisor goes down

tynes:
Is there any consideration with the batcher? Did we add some new interop specific config?

tynes:
There could be a failure mode of disk growth given we have to index a lot of data. The solution seems to be pruning, which we will know the depth at which we can prune once we get the final expiry window

SozinM:
Another more esoteric option for checks:
When building block N we validate all cross transaction for to block N+1 with the timeout 2 seconds.
Then, while building block N+1 we would have very fresh transactions that we could include.
This assumes that it's possible to validate batch under 2 seconds, so they would be on time for the next block building round.
This will load supervisor as it would make a lot of checks every 2 seconds.
To improve this a bit we could validate only number of transaction that would most likely be included into the block (if we have 100k txs in mempool it's obvious that we won't include them all into the next block)

security/fma-interop-tx-handling.md
```diff
@@ -0,0 +1,137 @@
...
+    service.
+    - When `proxyd` goes down, it takes outstanding requests with it, meaning some requests fail to be handled,
+    but excess load is shed.
+- Mitigations
+    - `proxyd` could feature a pressure-relief setting, where if too much time is spent waiting on the Supervisor,
+    no additional Interop Messages will be accepted through this gateway.
+    - We should deploy at least one "Utility Supervisor" to respond to Checks from `proxyd` instances.
+    The size and quantitiy of the Supervisor(s) could be scaled if needed. (Note: A Supervisor also requires Nodes
+    of each network to function)
+## FM3b: Transaction Volume causes DOS Failures of Supervisor
```
protolambda, line 110: Something to call out here, that's different in ingress than tx-pool usage (*) is that transactions can declare a very very long access-list that they cannot actually afford to include. Since the gas-limit vs gas-used and sender balance are not checked yet.

Two possible solutions:
1) rate-limit the access-list checks. E.g. configure a burst (= per tx access-list size limit) and a rate (= total load on supervisor) that may be checked.
2) do balance/nonce check before access-list check, in ingress.

Option 1 may be sufficient, if the rate is reasonable, even if txs do not pay. This largely depends on supervisor performance. On the bright side, these checks should be relatively quick (binary search a flat DB) and can run in parallel (different proxyd RPC handler routines, different supervisor RPC handler routines).

Option 2 here is going to add load to EL nodes, and thus less favorable from a DoS perspective. And it's easy to get out of sync (since you would need to look at the pending nonce, not the latest state, and either may lag behind on a verifier node).

security/fma-interop-tx-handling.md
axelKingsley: Right, that's what I'm trying to say here - when block builders can't assess Interop Validity (by way of not having a Supervisor), they can choose to drop interop messages. In this way, the liveness of interop is managed separately to preserve the chain liveness.

I'll do a pass on this content to make sure its closer to your description.

security/fma-interop-tx-handling.md
axelKingsley: This FM is overlap with the Supervisor FM2a (not 2b). I am not sure if it needs to be kept in this document, but I wanted it around to demonstrate that if these layers fail, an invalid message gets into a block.

To your question about if it's specific to some check -- this would have to apply to *every* check layer, not just one. If ingress is failing but mempool checks are working, it still gets filtered. If mempool is failing but ingress is working, it gets filtered.

I'll rework this FM to be more descriptive of how all the layers could fail together.

security/fma-interop-tx-handling.md
axelKingsley: Regarding Option 2 getting out of sync: the EL Nodes could do all this checking through an API, right? We could make `checkInterop` an API of Geth, which does the balance checks and then calls the Supervisor if needed. Then we can manage just one checking implementation, in geth.

axelKingsley: And, I will add this callout and Action Item to this document, thanks.

axelKingsley:
> Another more esoteric option for checks: When building block N we validate all cross transaction for to block N+1 with the timeout 2 seconds. Then, while building block N+1 we would have very fresh transactions that we could include. This assumes that it's possible to validate batch under 2 seconds, so they would be on time for the next block building round. This will load supervisor as it would make a lot of checks every 2 seconds. To improve this a bit we could validate only number of transaction that would most likely be included into the block (if we have 100k txs in mempool it's obvious that we won't include them all into the next block)

@SozinM thanks for this idea! I think this is functionally equivalent to our decision to batch-evaluate these messages on a recurring timer. If the timer happened to match the block building time, you'd achieve the same effect, where transactions get evaluated in anticipation of the next block. Hooking it directly to block number sounds neat, but I wouldn't want to put validation in the way of our block building timing.

And I totally agree regarding only validating on block's worth in advance. The nice thing is that we can statically determine if transactions are interop or not, so if there are 100k tx in the pool, we will only need to look over them once to identify the interop ones, and batch-check *those* (which would be a small subset of all possible tx)

axelKingsley:
> There could be a failure mode of disk growth given we have to index a lot of data. The solution seems to be pruning, which we will know the depth at which we can prune once we get the final expiry window

Even with fairly aggressive estimates, a cloud host would not be strained by the Supervisor's disk usage: https://github.com/ethereum-optimism/optimism/issues/14919

security/fma-interop-tx-handling.md
```diff
@@ -0,0 +1,168 @@
...
+- Description
+    - This overlaps Supervisor FMA's FM2a, where an invalid message is included in a built block.
+    - An interop transaction is checked mulitiple times (Proxyd, Sentry Node, Sequencer), and each check must fail
+    in order for the transaction to make it all the way to block building.
+    - The most likely cause for all filter layers to fail is a bug in the Supervisor.
+    - Individual layers may fail to filter a transaction if it decides the transaction doesn't need to be
+    checked for Interop Validity. For example, if it appears to have no Access List, it does not need to be checked.
+    - Recurring filters (ones which regularly re-validate Transactions in a Mempool) may fail to filter
+    a Transaction if the frequency of the check is too low.
+- Risk Assessment
```
smartcontracts, line 78: Recommended action items:
- Monitor for invalid messages
- Monitor for critical messages

security/fma-interop-tx-handling.md
```diff
@@ -0,0 +1,168 @@
...
+    the Supervisor FMA.
+    - Negative UX and Customer Perception from building invalid block content.
+
+## FM2: Checks Discard Valid Message
+- Description
+    - Due to a bug in either the Supervisor, or the Callers, some or all Executing Messages
+    aren't being included in blocks.
+    - When this happens, there is nothing invalid being produced by block builders, but no Interop
+    Messages are being included.
+- Risk Assessment
```
smartcontracts, line 91: Recommended action items:
- Monitor for a lack of interop messages

security/fma-interop-tx-handling.md
```diff
@@ -0,0 +1,168 @@
...
+- All Executing Message are checked once at Node Mempool ingress (not counting Sequencer).
+- All Executing Message in Node Mempools are Batched at checked on a regular interval.
+- If an Executing Message is ever Invalid, it is discarded and not retried.
+- *No* Checks are done at Block Building time.
+
+This FMA describes the potential negative consequences of this design. We have selected this design because it maximizes
+the opportunities for Invalid Messages to be caught and discarded, while also leaving the block building hot-path from
+having to take on new compute.
+
+# FMs:
```
smartcontracts, line 66: From async threat modeling:
- Monitor for reorgs

smartcontracts (approved):

security/fma-interop-tx-handling.md
axelKingsley: what is a critical message?

smartcontracts (approved):

Issue #238, opened 2025-03-27
FMA: Access List DoS Protection


tynes:
This ticket represents the completion of the FMA for the access list based denial of service protection. The design doc can be seen here: https://github.com/ethereum-optimism/design-docs/blob/f3aa2db64b1676b5e58ae602acf2ebdba34b617c/protocol/interop-access-list.md

agusduha:
Wonderland is going to work on this going through the `CrossL2Inbox` changes to support access list. Also review the Msg passing FMA to sync with this one (https://github.com/ethereum-optimism/design-docs/blob/main/security/fma-message-passing-contracts.md)

agusduha:
PR: https://github.com/ethereum-optimism/design-docs/pull/252

agusduha:
@mds1 Fixed and ready to merge

mds1:
Merged! As mentioned above we'll leave this issue open until we have the follow up PR changing the status to Final

mds1:
I have also reassigned to @agusduha since there are now actions to implement

agusduha:
Moving forward with some action implementation, there is a documentation PR ready to review https://github.com/ethereum-optimism/specs/pull/664

agusduha:
Finalize PR: https://github.com/ethereum-optimism/design-docs/pull/267

PR #252, opened 2025-04-02
feat: add interop access list FMA

agusduha:
**Description**

Adds Failure Modes and Recovery Path Analysis (FMA) for the Interop Access List mechanism. 

FMs:
- Gas Introspection Breaks Due to EVM Changes or Compiler Optimizations
- Storage slot collisions
- Unexpected Storage Slot Warming Behavior

**Additional context**

- Design PR: https://github.com/ethereum-optimism/design-docs/pull/214
- Specs PR: https://github.com/ethereum-optimism/specs/pull/612

**Metadata**

Issue: https://github.com/ethereum-optimism/design-docs/issues/238

### DIFF
--- a/security/fma-interop-access-list.md
+++ b/security/fma-interop-access-list.md
@@ -53,7 +53,7 @@ Below are references for this project:
 - **Description:** The checksum calculation for message validation uses storage slots that could potentially collide with other slots from the access list, leading to false validation of messages.
 - **Risk Assessment:** High
   - Potential Impact: Critical. Storage collisions could allow unauthorized messages to be validated. Additionally, this would break the ability to filter invalid executing messages at ingress, creating a spam vector where MEV searchers could attempt arbitrage by submitting invalid messages with no inclusion cost.
-  - Likelihood: Extremely Low. Storage slots are calculated using a 248-bit (31 byte) hash space, providing cryptographically secure collision resistance. A collision would require finding a hash collision, which is computationally infeasible with 248 bits of security.
+  - Likelihood: Extremely Low. Storage slots are calculated using a 248-bit (31 byte) hash space, providing cryptographically secure collision resistance. A collision would require finding a hash collision, which makes it infeasible to have a collision as long as the implementation is correct.
 - **Mitigations**
   - Implement robust checksum calculation that minimizes collision risk
   - Add tests verifying no collisions occur with expected storage patterns
--- a/security/fma-interop-access-list.md
+++ b/security/fma-interop-access-list.md
@@ -3,7 +3,7 @@
 | Author              | Agusduha, Skeletor-spaceman  |
 | ------------------- | ---------------------------- |
 | Created at          | 2025-02-05                   |
-| Initial Reviewers   | [Initial Reviewer Names]     |
+| Initial Reviewers   | Mark Tyneway                 |
 | Needs Approval From | Matt Solomon, Kelvin Fichter |
 | Status              | Draft                        |
 
@@ -90,8 +90,8 @@ See [fma-generic-contracts.md](https://github.com/ethereum-optimism/design-docs/
 
 ## Action Items
 
-- [ ] FM1: Implement comprehensive gas schedule testing
 - [ ] FM1: Document gas schedule dependencies
+- [ ] FM1: Implement comprehensive gas schedule testing
 - [ ] FM2: Create storage layout documentation
 - [ ] FM2: Implement slots collision tests
 - [ ] FM3: Document EVM warming behavior dependencies and transaction-scoped assumptions
--- a/security/fma-interop-access-list.md
+++ b/security/fma-interop-access-list.md
@@ -5,7 +5,7 @@
 | Created at          | 2025-02-05                   |
 | Initial Reviewers   | Mark Tyneway                 |
 | Needs Approval From | Matt Solomon, Kelvin Fichter |
-| Status              | Draft                        |
+| Status              | Implementing Actions |
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


security/fma-interop-access-list.md
```diff
@@ -0,0 +1,121 @@
...
+  - Document gas schedule dependencies clearly for future upgrades
+- **Detection:** Monitor for network upgrades that modify storage access gas costs. Run validation tests before accepting upgrades.
+- **Recovery Path(s):** Deploy contract upgrade with adjusted thresholds if gas schedule changes significantly.
+
+### FM2: Storage Slot Collision
+
+- **Description:** The checksum calculation for message validation uses storage slots that could potentially collide with other slots from the access list, leading to false validation of messages.
+- **Risk Assessment:** High
+  - Potential Impact: Critical. Storage collisions could allow unauthorized messages to be validated.
+  - Likelihood: Low. Storage slots are carefully calculated using Identifier and message hash parameter. The way both are encoded in the contract’s logic needs to be flawed for this case to happen.
```
tynes, line 46: How many bits of security do we have?

security/fma-interop-access-list.md
```diff
@@ -0,0 +1,121 @@
...
+  - Add tests that verify validation works across different gas cost scenarios
+  - Document gas schedule dependencies clearly for future upgrades
+- **Detection:** Monitor for network upgrades that modify storage access gas costs. Run validation tests before accepting upgrades.
+- **Recovery Path(s):** Deploy contract upgrade with adjusted thresholds if gas schedule changes significantly.
+
+### FM2: Storage Slot Collision
+
+- **Description:** The checksum calculation for message validation uses storage slots that could potentially collide with other slots from the access list, leading to false validation of messages.
+- **Risk Assessment:** High
+  - Potential Impact: Critical. Storage collisions could allow unauthorized messages to be validated.
```
tynes, line 45: The impact of this is that we would not be able to filter messages at ingress that created invalid executing messages. This creates a spam incentive for MEV searchers that can attempt an arb and create an invalid executing message at no cost (guarantee of no inclusion)

security/fma-interop-access-list.md
```diff
@@ -0,0 +1,121 @@
...
+- **Mitigations**
+  - Implement robust checksum calculation that minimizes collision risk
+  - Add tests verifying no collisions occur with expected storage patterns
+  - Document storage layout and slot calculation methodology
+- **Detection:** Monitor for unexpected message validations that could indicate storage collisions.
+- **Recovery Path(s):** Deploy contract upgrade with revised storage slot calculation if collisions are detected and update ALL off-chain tools and SDKs that generate these hashes on the access list.
+
+### FM3: Access List Manipulation
+
+- **Description:** Malicious actors could attempt to manipulate transaction access lists to force validation of unauthorized messages.
```
tynes, line 56: This seems to be similar to having the user sign a malicious payload? ie the app has the user send a tx that sweeps all their ETH

security/fma-interop-access-list.md
```diff
@@ -0,0 +1,121 @@
...
+- **Risk Assessment:** High
+  - Potential Impact: High. Could allow unauthorized message execution.
+  - Likelihood: Medium. Access lists are user-controlled transaction parameters.
+- **Mitigations:**
+  - Ensure checksum calculation is cryptographically sound
+  - Implement proper validation of access list entries
+  - Add checks for maximum access list size
```
tynes, line 63: The action item here is to have a paramaterizable max access list size as a policy on ingress/mempool. This is similar to the max tx size for p2p gossip being smaller than the max tx size based on using all gas for calldata in a block

security/fma-interop-access-list.md
```diff
@@ -0,0 +1,121 @@
...
+  - Potential Impact: Medium. Could cause temporary disruption of cross-chain messaging.
+  - Likelihood: Medium. Relayers need to adopt new patterns for access list construction.
+- **Mitigations:**
+  - Provide clear documentation and examples for relayer implementations
+  - Create reference implementations and test suites
+  - Add monitoring for failed validation attempts
+- **Detection:** Track validation failure rates and patterns across relayers.
+- **Recovery Path(s):** Work with relayer implementations to fix access list construction.
+
+### FM5: Compiler Optimizations Break Gas Introspection
```
tynes, line 80: This is a very good failure mode. I think that e2e testing should be able to catch this, if we have an e2e test that asserts the revert. Doing it as an e2e test ensures nothing about the foundry environment impacts the gas consumption

tynes:
Another failure mode would be that the slot is somehow warmed even tho the item was not in the access list. We have validated that if we call the function and warms the slot by doing a read and it returns `isWarm=false` then it reverts, cooling the slot again. This seems like it shouldn't work this way from the pov of Ethereum protocol, given warming is meant to represent the value is cached. If its read and then the call reverts, the value can be kept in the cache? Perhaps it rolls back due to a possible dos attack for the cache?

security/fma-interop-access-list.md
tynes: There are 248 bits (31 bytes) of security. This is plenty, no risk of collision

security/fma-interop-access-list.md
```diff
@@ -0,0 +1,107 @@
+# Interop Access List: Failure Modes and Recovery Path Analysis
+
+| Author              | Agusduha, Skeletor-spaceman  |
+| ------------------- | ---------------------------- |
+| Created at          | 2025-02-05                   |
+| Initial Reviewers   | Mark Tyneway                 |
+| Needs Approval From | Matt Solomon, Kelvin Fichter |
+| Status              | Draft                        |
```
mds1, line 8:
```suggestion
| Status              | Implementing Actions |
```

mds1 (approved):

PR #264, opened 2025-04-14
feat: Kona Rollup Node Architecture Doc

refcell:
### Description

Writes up an architecture document for the `kona-node` (aka the Kona Rollup Node).
Considers how this architecture compares to the `op-node`.
Uses diagrams to visualize how the various actors tie together.

### DIFF
--- a/protocol/kona-node-arch.md
+++ b/protocol/kona-node-arch.md
@@ -171,7 +171,6 @@ make progress in parallel.
 If the orchestration layer is called "Rollup Node", a full modular architecture
 for the Kona Rollup Node looks like the following.
 
-
 ```
       ┌────────────────────────────────────────────────────────────────────┐
       │                                                                    │
@@ -192,6 +191,52 @@ for the Kona Rollup Node looks like the following.
       └────────────┘
 ```
 
+Notice, the "Rpc Actor" doesn't have communication lines drawn to other actors.
+This is because the JSON RPC server has rpc modules that it registers which
+handle various rpc requests. In effect, each actor registers its rpc module
+with the Rpc Actor at construction. The rpc modules in turn grab or perform
+actions on other actors via message passing. Effectively, the Rpc Actor
+messages with all other actors.
+
+### Sequencer vs Validator Mode
+
+Up to this point, Kona's rollup node architecture has been considered
+from the perspective of a _validator_ node. That is, a rollup node
+that only derives the l2 chain, but doesn't _build_ (aka "sequence")
+the L2 chain. In order to extend support for sequencing the L2 chain,
+the rollup node needs to accept transactions, build l2 blocks, and
+broadcast these blocks as gossip over the p2p network.
+
+Part of the architecture to consider is being able to plug in
+"block building" through a simple API. Similar to the `op-node`
+allowing sequencing to be toggled on and off, the `kona-node` should
+support sequencing via a toggle cli argument.
+
+While sequencing is optional, the most kona-native route would be to
+create a sequencer actor that responds to various events. This extends
+the architecture as follows.
+
+```
+      ┌────────────────────────────────────────────────────────────────────────────────────┐
+      │                                                                                    │
+  ┌───┤                                 Rollup Node Service                                │
+  │   │                                                                                    │
+  │   └──────────────────────────┬────────────────┬────────────────┬────────────────┬──────┘
+  │                              │                │                │                │
+  │   ┌────────────┐             │                │                │                │
+  │   │L2 Sequencer│             │                │                │                │
+  ├─► │            ├───┐         ▼                ▼                ▼                ▼
+  │   │   Gossip   │   │   ┌────────────┐   ┌────────────┐   ┌───────────┐    ┌───────────┐
+  │   └────────────┘   │   │            │   │            │   │           │    │           │
+  │                    ├──►│ Derivation │──►│ Engine API │   │ Sequencer │    │ Rpc Actor │
+  │   ┌────────────┐   │   │            │   │            │   │           │    │           │
+  │   │     DA     │   │   └────────────┘   └──────┬─────┘   └───────────┘    └───────────┘
+  └─► │            ├───┘          ▲                │
+      │   Watcher  │              └────────────────┘
+      └────────────┘
+
+
+```
 
 # Alternatives Considered
 
--- a/protocol/kona-node-arch.md
+++ b/protocol/kona-node-arch.md
@@ -71,6 +71,8 @@ in performance, safety, and concurrency. The implementation will:
 
 ## Architecture
 
+### Building Blocks from First Principles
+
 The Rollup Node is a core component of the OP Stack.
 It is responsible for constructing the canonical safe L2 chain.
 In order to produce the chain, the Rollup Node listens to two sources
@@ -90,11 +92,11 @@ somehow producing the L2 Chain.
 ┌────────────┐
 │L2 Sequencer│
 │            ├────────────────────┐
-│   Gossip   │                    │    ┌────────────┐    ┌────────────┐
-└────────────┘                    │    │            │    │            │
-                                  ├──► │     ???    │──► │  L2 Chain  │
-┌────────────┐    ┌────────────┐  │    │            │    │            │
-│     DA     │    │            │  │    └────────────┘    └────────────┘
+│   Gossip   │                    │    ┌────────────┐
+└────────────┘                    │    │            │
+                                  ├──► │     ???    │──►  < L2 Chain >
+┌────────────┐    ┌────────────┐  │    │            │
+│     DA     │    │            │  │    └────────────┘
 │            ├──► │ Derivation ├──┘
 │   Watcher  │    │            │
 └────────────┘    └────────────┘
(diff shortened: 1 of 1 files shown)


protocol/kona-node-arch.md
```diff
@@ -0,0 +1,272 @@
...
+  │   └──────────────────────────┬────────────────┬────────────────┬────────────────┬──────┘
+  │                              │                │                │                │
+  │   ┌────────────┐             │                │                │                │
+  │   │L2 Sequencer│             │                │                │                │
+  ├─► │            ├───┐         ▼                ▼                ▼                ▼
+  │   │   Gossip   │   │   ┌────────────┐   ┌────────────┐   ┌───────────┐    ┌───────────┐
+  │   └────────────┘   │   │            │   │            │   │           │    │           │
+  │                    ├──►│ Derivation │──►│ Engine API │   │ Sequencer │    │ Rpc Actor │
+  │   ┌────────────┐   │   │            │   │            │   │           │    │           │
+  │   │     DA     │   │   └────────────┘   └──────┬─────┘   └───────────┘    └───────────┘
```
refcell, line 233: Probably worth connecting the sequencer to other components of the rollup node.

protocol/kona-node-arch.md
```diff
@@ -0,0 +1,291 @@
...
+What this document addresses is how the components communicate. Namely, do
+components share memory or use an actor=based architecture with message-passing.
+The remainder of this section will provide reasoning to choose the latter.
+
+In an actor-based system, each component is constructed as an "Actor". There's
+a Derivation Actor, DA Watcher Actor, P2P Actor (L2 sequencer gossip),
+and Engine Actor.
+
+Instead of having some top-level object "own" components, actors are spawned
+as threads, and communication between actors happens through channels with
```
Inphi, line 162: are these channels synchronous? If not, what's the queueing policy and how do events/ message receipts handled?

Inphi (commented):
Is interop being considered for this architecture? It would be nice if we can easily attach a component that communicates with op-supervisor.

refcell:
> Is interop being considered for this architecture? It would be nice if we can easily attach a component that communicates with op-supervisor.

Thanks for reviewing Mofi!

My understanding is that the majority of the interop changes occur at the tx pool / block building stage. Since the `kona-node` will be designed to be modular, block building should be easily plugged into the node. The goal is for this modularity to allow interop and other features / changes to be easily supported going forward.

Is that a sufficient answer? Are there other parts to interop that we need to consider besides block building?

protocol/kona-node-arch.md
refcell: Good question. The answer is it really depends on the actors and the relationships between those actors. Since the channels don't have to be uniform, I haven't specified anything here. Most likely, it'll change as we go along.

clabby:
> Are there other parts to interop that we need to consider besides block building?

@axelKingsley recently gave me a pretty good view of the "managed node" concept they've added with the `op-supervisor`. With the "managed" backend, the node and supervisor communicate, and the supervisor instructs the node to perform certain actions - i.e., the supervisor instructs when to replace a block due to bad messages, when to reset the derivation pipeline due to a reorg, etc. This is an implementation detail, but unfortunately an implementation detail we'll have to follow in order to be compatible with `op-supervisor`.

Reached out to Axel and got the following docs on the subject:
* https://www.notion.so/oplabs/Supervisor-Modes-Disambiguation-1bcf153ee1628084aeb0e52d5e1b9fc6?pvs=4
* https://github.com/ethereum-optimism/design-docs/pull/171/files?short_path=21411c4#diff-21411c434a0cd42cf633c6e72b4bf02de2e3c267dd0ce7a2ada516a22fa0049f
* https://github.com/ethereum-optimism/optimism/blob/develop/op-supervisor/README.md#managed-mode

protocol/kona-node-arch.md
```diff
@@ -0,0 +1,291 @@
...
+│   Watcher  │              └────────────────┘
+└────────────┘
+```
+
+### A Modular Architecture
+
+The components described above are all critical to the rollup node.
+Whether these components are part of the architecture is not up for discussion.
+What this document addresses is how the components communicate. Namely, do
+components share memory or use an actor=based architecture with message-passing.
```
theochap, line 154: Micro nit:
```suggestion
components share memory or use an actor-based architecture with message-passing.
```

protocol/kona-node-arch.md
```diff
@@ -0,0 +1,291 @@
...
+broadcast these blocks as gossip over the p2p network.
+
+Part of the architecture to consider is being able to plug in
+"block building" through a simple API. The `op-node` separates
+the sequencer logic from the rest of the node driver, allowing
+it to toggle sequencing on and off via a simple CLI flag. Kona
+can take this one step further. Using a minimal API, the
+`kona-node` should allow sequencing to be toggled on and off,
+but also let users easily slot in their own block building and
+sequencing logic using the given API.
```
theochap, line 217: That's a really interesting concept. Sounds like this would be pretty interesting for sequencers that would want to have custom MEV algorithms or those that want to do some advanced profit estimation based on transaction inclusion. Are these the use-cases you thought of? Or did you have something different in mind

protocol/kona-node-arch.md
```diff
@@ -0,0 +1,291 @@
...
+"Rollup Node Service" orchestrator. This more closely resembles the `op-node`
+architecture where "derivers" are registered via a "registry". Various
+event types are emitted and broadcasted to rollup node components.
+
+While this works effectively for the `op-node` it introduces significant
+overhead and risk for Kona's Rollup Node. Since the `kona-node` is
+parallized, mishandling or even spontaneous flakes where messages are dropped,
+can result in an unrecoverable deadlock. By establishing messaging channels
+directly between actors, there's less "surface area" for message passing
+to be improperly configured.
```
theochap, line 276: I think insisting on the parallelization is a great idea. I was wondering if the optimism spec was built to handle internal component parallelism or if this is something we should be careful of in the implementation

axelKingsley:
> > Are there other parts to interop that we need to consider besides block building?
> 
> @axelKingsley recently gave me a pretty good view of the "managed node" concept they've added with the `op-supervisor`. With the "managed" backend, the node and supervisor communicate, and the supervisor instructs the node to perform certain actions and vice versa - i.e., the supervisor instructs when to update certain heads, and the node instructs the supervisor to inform database updates etc. This is an implementation detail, but unfortunately an implementation detail we'll have to follow in order to be compatible with `op-supervisor`.
> 
> Reached out to Axel and got the following docs on the subject:
> 
> * https://www.notion.so/oplabs/Supervisor-Modes-Disambiguation-1bcf153ee1628084aeb0e52d5e1b9fc6?pvs=4
> * https://github.com/ethereum-optimism/design-docs/pull/171/files?short_path=21411c4#diff-21411c434a0cd42cf633c6e72b4bf02de2e3c267dd0ce7a2ada516a22fa0049f
> * https://github.com/ethereum-optimism/optimism/blob/develop/op-supervisor/README.md#managed-mode

Managed Mode is an important thing to understand for the Supervisor's operation, but I don't know that it will be required to replicate in other clients. Managed Mode's purpose is to provide the Supervisor with a stable collection of Nodes from which to index log events. The Golang Node implementation already satisfies this requirement, and while it would be cool for the Supervisor to support other clients, the only benefit we would get is heterogeneous indexing sources.

There is **another** mode which we have named Standard Mode (questionable name at this point) which I think would be a better integration between the Supervisor and an Alternative Client. In Standard Mode, Kona would do its local derivation and L1 consolidation, and then would do a *further* consolidation with a Supervisor over an API call. This is ill-defined, but you could imagine Kona supplies the last L1 block it derived from, and the Supervisor responds with the appropriate L2 cross-unsafe, cross-safe, and any required block replacements.

Standard Mode was first imagined to support chain operators who didn't want to run a full Supervisor, but would also work as a basic interface into interop for alt clients. You could then separately focus on the longer term goal of including Interop Validation into Kona, to remove reliance on golang implemenations altogether.

The Golang implemenation does not yet support Standard Mode. It is just outside our priority scope, either @teddyknox or I will be working on it. However, we could do with some outside spec input, so if Kona wants to be the first Standard Mode implementor that'd be nifty.

clabby:
I'm a fan of @axelKingsley's proposal to _not_ support the `op-supervisor`'s "managed mode" at first (or ever.) From what I can tell, managed mode in the `op-node` + `op-supervisor` is a tight coupling between the two implementations that we likely do not want to subscribe to arbitrarily.

This gives us the flexibility to wait until we have our own supervisor implementation for a tighter integration with a "managed mode"-esq setup. While an important feature to support, we can run interop without it, and will likely land on a better result with a supervisor that's designed around the `kona-node` rather than the `op-node` specifically.

protocol/kona-node-arch.md
clabby: At some point, we should definitely seek to remove the `rollup-boost` shim in favor of a native layer that allows for custom block building functionality, as an example. Things like direct `rbuilder` integration would be incredible.

axelKingsley:
> This sort-of surfaces the question - does the op-supervisor team intend for the supervisor to be like an extension of a rollup node implementation, or like an independent service that we should be able to plug-and-play? For the EL, we've standardized the endpoint for txpool pre-validation, but on the CL, the integration looks much tighter.

The Supervisor's identity has changed over time, and I don't know that we currently maintain a clear answer to this question. But I will stick my neck out to say: The Supervisor is an Extension of the Rollup Node Implementation... and as a second-order effect, it can serve that information out for plug-and-play support on Alt Clients.

The reality of Superchain Validation is that it really does require a *meta-derivation*, which the Supervisor provides. Once the parallel work of local derivation is done, cross validation determines replacement blocks, which *is* part of Derivation. IF a client is responsible for implementing derivation (and it is), then it is responsible for implementing cross-chain validation.

The second-order effect of being plug-and-play is, I think, opportunistic. We designed the Supervisor with RPC connections because the op-node is already a mature piece of software and can do derivation work. The API interface to check messages was a natural need for supporting filtering in the execution layer, and is naturally callable by anyone.

protocol/kona-node-arch.md
refcell: It'll be something we need to be very careful with. The good thing is we have so many methods of stress testing this architecture through kurtosis, local syncing, action tests, etc etc against a matrix of chains. Testing early is how I hope we can find any architecture-related bugs quicker.

protocol/kona-node-arch.md
```diff
@@ -0,0 +1,291 @@
...
+│L2 Sequencer│
+│            ├───┐
+│   Gossip   │   │   ┌────────────┐   ┌────────────┐
+└────────────┘   │   │            │   │            │
+                 ├──►│ Derivation │──►│ Engine API │──►  < L2 Chain >
+┌────────────┐   │   │            │   │            │
+│     DA     │   │   └────────────┘   └──────┬─────┘
+│            ├───┘          ▲                │
+│   Watcher  │              └────────────────┘
+└────────────┘
```
axelKingsley, line 146: In previous illustrations, the Unsafe Gossip was being routed directly to the Engine API. I *think* that is still the case here, even when Holocene introduces block replacement, ya?

And actually, I am not sure: do block replacements occur on invalid *unsafe* blocks? I would assume not since they are not part of derivation (and therefore not part of *steady* derivation).

protocol/kona-node-arch.md
```diff
@@ -0,0 +1,291 @@
...
+      │                                                                    │
+  ┌───┤                        Rollup Node Service                         │
+  │   │                                                                    │
+  │   └──────────────────────────┬────────────────┬─────────────────┬──────┘
+  │                              │                │                 │
+  │   ┌────────────┐             │                │                 │
+  │   │L2 Sequencer│             │                │                 │
+  ├─► │            ├───┐         ▼                ▼                 ▼
+  │   │   Gossip   │   │   ┌────────────┐   ┌────────────┐     ┌───────────┐
+  │   └────────────┘   │   │            │   │            │     │           │
```
axelKingsley, line 185: Same comment re: Gossip going to the Engine instead of to Derivation

protocol/kona-node-arch.md
tynes: Parallelism for `op-node` is in a draft PR here: https://github.com/ethereum-optimism/optimism/pull/11100

protocol/kona-node-arch.md
refcell: Oops good catch - as discussed on the call, will update this. Bad diagram on my end 😅

refcell:
[here](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/supervisor-dataflow.md) looks to be the core supervisor document that considers the role that the supervisor plays in the OP Stack. Specifically, how the supervisor works with the `op-node`. There may be other documents to reference, but a few of them seem to be in notion.

protocol/kona-node-arch.md
refcell: Updated!

protocol/kona-node-arch.md
```diff
@@ -0,0 +1,291 @@
...
+
+Concretely, this would look like a large-variant enum that holds all
+message types. Actors would then only send and receive messages from the
+"Rollup Node Service" orchestrator. This more closely resembles the `op-node`
+architecture where "derivers" are registered via a "registry". Various
+event types are emitted and broadcasted to rollup node components.
+
+While this works effectively for the `op-node` it introduces significant
+overhead and risk for Kona's Rollup Node. Since the `kona-node` is
+parallized, mishandling or even spontaneous flakes where messages are dropped,
```
samlaf, line 273:
```suggestion
parallelized, mishandling or even spontaneous flakes where messages are dropped,
```
I think..?

PR #266, opened 2025-04-16
feat: L2ToL2CDM SentMessage ctx

hamdiallam:
**Additional context**

Additional of internal context that can be propogated across N cross domain messages

Example High level approach to incentivized relays: https://www.notion.so/oplabs/Incentivized-Message-Relays-1b2f153ee162800e9f36ceab34efa7d9?pvs=4


hamdiallam:
Notes from Design Review: Review also covered incentivized delivery in #272 

General:
* all in favor of adding the context opaque bytes. We can do this ahead of time since it is an api breaking change with confidence due to it's general usefulness
* GasTanks are abstracted enough that can support payments for senders, msgHash, and more. (they can also refund the relayers on origin, destination or any other superchain).
* Not entirely sure about call depth in the propagated ctx, but we need a way to being able to:
  * reference the parent msg-hash on child sentMessages
  * We might be able to infer ordering of relayed messages on the GasTank via a chain of hashes with the parent message
* all in favor of adding the RelayedMessageGasReceipt event with minor modifications to be future-proof
* __important__ we need to go through integration and dapp examples to see where this works and breaks.

Next Steps:
 1. Update this document scoped to just propagating opaque context bytes with the SentMessage event. This will be empty to start
 2. Update or create a new document for `RelayedMessageGasReceipt` which can be tied with #272. When some prototypes  of the gas tank is created, we can then make use of the context bytes with a first version (tx.origin, rootMsgHash) and emit the the gas receipt event.

clabby (approved):
I'm for this change for the flexibility it gives us going forwards. Better to add it now than later, and the method of population is easy to change without breaking the event topic / schema in the future.

ecosystem/l2tol2cdm-sentmsg-ctx.md
```diff
@@ -0,0 +1,93 @@
...
+<!-- A high level overview of the proposed solution.
+When there are multiple alternatives there should be an explanation
+of why one solution was picked over other solutions.
+As a rule of thumb, including code snippets (except for defining an external API)
+is likely too low level. -->
+
+Introudce opaque context bytes field in `SentMessage`. This field MUST be versioned such that it can evolve with new or removed fields.
+
+```solidity
+event SentMessage(uint256 indexed destination, address indexed target, uint256 indexed messageNonce, address sender, bytes message, bytes context);
```
agusduha, line 50:
```suggestion
event SentMessage(uint256 indexed destination, address indexed target, uint256 indexed messageNonce, address sender, bytes message, bytes originContext);
```

PR #272, opened 2025-04-22
feat: incentivized cross domain message delivery

hamdiallam:
**Description**

Incentivized relays settlement framework leveraging the data provided in #266


ecosystem/incentivized-relays-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+        require(messenger.sentMessages[rootMsgHash]);
+
+        // compute total cost
+        uint256 claimCost = CLAIM_OVERHEAD * block.basefee;
+        uint256 cost = relayCost + claimCost;
+        require(balanceOf[txOrigin] >= cost);
+
+        // compensate the relayer
+        balanceOf[txOrigin] -= cost;
+        new SafeSend{ value: cost }(payable(relayer));
```
tremarkley, line 96: should we we emit a message after this? might be helpful for tracking which messages have been claimed.

ecosystem/incentivized-relays-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+
+        uint256 newBalance = balanceOf[msg.sender] + amount;
+        require(newBalance < MAX_DEPOSIT);
+
+        balanceOf[msg.sender] = newBalance;
+    }
+}
+```
+
+We cap the deposits in order to cut scope in supporting withdrawals. This makes our first interation super simple since as withdrawal support introduces a race between deposited funds and a relayer compensating themselves for message delivery. With a relatively low max deposit, we eliminate the risk of a large amount of stuck funds for a given account whilst the amount being sufficient enough to cover hundreds of sub-cent transactions.
```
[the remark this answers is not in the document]
fainashalts: yes I think that's right Harry!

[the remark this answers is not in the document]
hamdiallam: I think all the way till mainnet is an open question. This lets us get to a first version and maybe we have time to iterate on a version with withdrawal support for mainnet

ecosystem/incentivized-relays-mvp.md
hamdiallam: yea agreed, an event here is good

ecosystem/incentivized-relays-mvp.md
```diff
@@ -0,0 +1,152 @@
...
+
+The changes proposed in [#266](https://github.com/ethereum-optimism/design-docs/pull/266), provides the foundation creating the first iteration of this settlement system -- without enshrinment in the core protocol contracts. The `RelayedMessageGasReceipt` includes contextual information on the gas consumed, and the propogated (`tx.origin`, `rootMessageHash`,`callDepth`) upon nested cross domain messages that can be used to appropriately charge `tx.origin`.
+
+This settlement system is a permissionless CREATE2 deployment, `L2ToL2CrossDomainGasTank`, where tx senders can hold an ETH deposit, used to asynchronously pay relayers for delivered messages.
+
+```solidity
+contract L2ToL2CrossDomainGasTank {
+    uint256 constant MAX_DEPOSIT = 0.01 ether;
+
+    mapping(address => uint256) public balanceOf;
```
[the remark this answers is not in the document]
hamdiallam: yea this is a great point! I was actually thinking of having support for this not in replacement but for 3rdparty integrator that wants a different fee payer than`tx.origin`.

The reason it can't replace is that this requires native integration with someone using the L2ToL2CrossDomainMessenger. You'd have to wrap every `sendMessage` call with a call to `deposit`. Instead of a replacement, this would be an additive API.

One easy usecase. Lets say Layerzero wants to integrate this feature. Someone sends a message from Solana -> OPM that then spawns a bunch of operations. You dont want the LZ relayer from Sol->OPM to be the fee payer. However, the LZ DVNthat integrates the L2ToL2CDM can use this api to specifically just push whatever gas was provided from the original tx as the deposit for the given root msg hash. Then everything just works. And if the provided gas wasn't enough, the person can top up the given root msg hash with more funds

Does this make sense?

ecosystem/incentivized-relays-mvp.md
tremarkley: another thing I thought of, wouldnt someone be able to claim the same gas receipt multiple times with this? do we need a mapping of claimed msgs to prevent this?

ecosystem/incentivized-relays-mvp.md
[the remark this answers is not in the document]
tremarkley: > The reason it can't replace is that this requires native integration with someone using the L2ToL2CrossDomainMessenger. You'd have to wrap every sendMessage call with a call to deposit. Instead of a replacement, this would be an additive API.

when you say additive API, are you saying that the gas tank will support balance tracking and claiming by tx.origin and by msg hash?

ecosystem/incentivized-relays-mvp.md
hamdiallam: lmfao yes good call 🤣

ecosystem/incentivized-relays-mvp.md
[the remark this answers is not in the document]
hamdiallam: by default claiming by tx.origin unless the `rootMsgHash` has an explicit balance set.

If the rootMsgHash has a balance set, i think it should not claim by tx.origin. Is there a scenario where you think it should fallback to tx.origin after depleting the funds associated with the msg hash?

ecosystem/incentivized-relays-mvp.md
hamdiallam: updated with the claim id

ecosystem/incentivized-relays-mvp.md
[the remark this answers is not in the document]
tremarkley: > If the rootMsgHash has a balance set, i think it should not claim by tx.origin. Is there a scenario where you think it should fallback to tx.origin after depleting the funds associated with the msg hash?

Hmm, if we are going to support balances by tx.origin and by msgHash then we should fallback to tx.origin. Otherwise, a bad actor could potentially try to make it more difficult for a msg to relay by setting a very small balance to any msg hash they want to prevent from being relayed, and then the user would have to intervene and increase the balance on that msg hash in order to get it to relay.

The other option here is to just simplify things and only support msgHash balances. Even though it will always require an extra deposit, can't that be simplified by something like 7702?

ecosystem/incentivized-relays-mvp.md
```diff
@@ -0,0 +1,162 @@
...
+        // ensure unclaimed, and mark the claim
+        bytes32 claimId = keccak256(abi.encode(rootMsgHash, callDepth));
+        require(!claimed[claimId]);
+
+        claimed[claimId] = true
+
+        // compute total cost
+        uint256 claimCost = CLAIM_OVERHEAD * block.basefee;
+        uint256 cost = relayCost + claimCost;
+        require(balanceOf[txOrigin] >= cost);
```
tremarkley, line 100: even if there is not enough balance to cover should we still let the relayer claim? I could see a scenario where the relayer has already relayed a message and even though the gas tank doesnt have the full balance to cover the cost, the relayer still wants to claim the remaining funds in the gas tank. wdyt?

fainashalts: wouldn't the tx not go through if there's not enough balance prior? I guess its possible that the simulated gas price and the actual are different, is that the scenario you had in mind?

ecosystem/incentivized-relays-mvp.md
fainashalts: @tremarkley interesting! curious how we could account for this. permissioned set of relayers doesn't feel like the right approach. or at least its not a long-term solution.

ecosystem/incentivized-relays-mvp.md
```diff
@@ -0,0 +1,162 @@
...
+
+        // parse the receipt
+        require(payload[:32] == RelayedMessageGasReceipt.selector);
+        (bytes32 rootMsgHash, address txOrigin, uint256 callDepth, address relayer, uint256 relayCost) = _decode(payload);
+
+        // ensure message was sent from this chain
+        require(messenger.sentMessages[rootMsgHash]);
+
+        // ensure unclaimed, and mark the claim
+        bytes32 claimId = keccak256(abi.encode(rootMsgHash, callDepth));
```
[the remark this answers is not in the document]
hamdiallam: yep i really should just use the msg hash here. Idk why i'm creating a new unique identifier when one exists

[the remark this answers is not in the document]
fainashalts: What if we use full msgHash + chainId + receipt blockNumber; this guarantees uniqueness even for identical sub-calls (A→B then A→C), right? Just thinking of ways to further protect against double spend.

ecosystem/incentivized-relays-mvp.md
fainashalts: Gave this some more thought - If the relayer spent gas it should be able to drain whatever funds are present if the cost > what's in the vault I feel. Should we add PartialClaim ability? do a `min(cost, balance)` and emit a `PartialClaim` event for accounting maybe? curious what you think! this should really only be an edge case since we estimate that there is enough to fund the relay.

ecosystem/incentivized-relays-mvp.md
```diff
@@ -0,0 +1,162 @@
...
+}
+```
+
+A claim is uniquely identified by the `(rootMsgHash, callDepth)` tuple.
+
+As relayers deliver messages, they can claim the cost against the original tx sender deposit with the gas tank. This design introduces some off-chain complexity for the relayer.
+
+1. The relayer must simulate the call and ensure the emitted tx origin has a deposit that will cover the cost on the originating chain.
+2. The relayer should also track the pending claimable `RelayedMessageGasReceipt` of the `rootMsgHash` callstack to maximize the likelihood that the held deposit is sufficient.
+3. The relayer should claim the receipt within a reasonable time frame to ensure they are compensated. An unrelated relayer not checking (2) might claim newer transitive message that depletes the funds.
```
fainashalts, line 115: should we add `blockTimestamp` to `RelayedMessageGasReceipt` and let the gas tank owner set `maxReceiptAge`?

ecosystem/incentivized-relays-mvp.md
```diff
@@ -0,0 +1,162 @@
...
+    mapping(address => uint256) public balanceOf;
+
+    function deposit() nonReentrant external payable {
+        uint256 amount = msg.value;
+        require(amount > 0);
+
+        uint256 newBalance = balanceOf[msg.sender] + amount;
+        require(newBalance < MAX_DEPOSIT);
+
+        balanceOf[msg.sender] = newBalance;
```
tremarkley, line 66: I think it would be good to emit a deposit event

karlfloersch:
Idea -- Add a intermediate version of this where there is a flat fee which is paid to a single relayer. Even though it doesn't work in all cases, it would get us to an 80% solution without having to implement the more complex logic in the onchain bounty

ecosystem/incentivized-relays-mvp.md
hamdiallam: updated with the events & using the right claim id.

Agreed permissioned relayers here doesn't feel right but I added a section on the Permissioned version of the gas tank itself based on @karlfloersch's feedback

ecosystem/incentivized-relays-mvp.md
hamdiallam: Updated so that partial claims are supported!! So at least it'll go through even if it's not eough

hamdiallam:
> Idea -- Add a intermediate version of this where there is a flat fee which is paid to a single relayer. Even though it doesn't work in all cases, it would get us to an 80% solution without having to implement the more complex logic in the onchain bounty

Check out the alternatives considered sections. Added a bit about an intermediate permissioned gas tank iteration that we can do

PR #274, opened 2025-04-25
feat(interop): L2toL2CrossDomainMessenger Autorelayer

tremarkley:
Closes https://github.com/ethereum-optimism/ecosystem/issues/750

### DIFF
--- a/ecosystem/auto-relayer/auto-relayer.md
+++ b/ecosystem/auto-relayer/auto-relayer.md
@@ -179,6 +179,7 @@ The following metrics will be used in order to track the overall health of the r
 - memory available on tx relayer processor redis
 - memory available on gas refund processor redis
 - time between message sent and message relayed
+- balance of Drippie contracts
 
 ### Gas Price
 **_Open question: do we need to allow a max gas price to be specified per message?_**


ecosystem/auto-relayer/auto-relayer.md
```diff
@@ -0,0 +1,209 @@
...
+4. Updates whether the claim succeeded or failed in the `Refunds` table in  `event-log-indexer-postgres`.
+
+### Indexing
+Ponder will be used for indexing. The following events will be indexed by ponder:
+- `SentMessage`: Emitted by `L2toL2CrossDomainMessenger` whenever a message is sent to a destination.
+- `RelayedMessageGasReceipt`: Emitted by `L2toL2CrossDomainMessenger` whenever a message is relayed.
+- `Claim`: Emitted by `L2ToL2CrossDomainGasTank` when a refund is claimed.
+- `Deposit`: Emitted by `L2ToL2CrossDomainGasTank` when funds are deposited.
+
+### Nonce management
```
pharger, line 143: @tremarkley does this mean that we can only scale with the number of EOAs that we have funded? So theoretically only one transaction per block?

ecosystem/auto-relayer/auto-relayer.md
```diff
@@ -0,0 +1,209 @@
...
+- balance of each wallet
+- storage available on indexer db
+- memory available on tx relayer processor redis
+- memory available on gas refund processor redis
+- time between message sent and message relayed
+
+### Gas Price
+**_Open question: do we need to allow a max gas price to be specified per message?_**
+
+**_Open question: are we okay with the relayer paying the priority fee and not being refunded for it?_**
```
pharger, line 186: Can we include this in the API design for ad hoc requests? Feels like we would want to set to high such that transactions always get filled immediately

ecosystem/auto-relayer/auto-relayer.md
tremarkley: there will be multiple EOAs per chain, operating in parallel. Each EOA can at most do one transaction per block, since each EOA waits on tx receipt before processing the next message, therefore, the number of txs per block is equal to the number of EOAs we have on that chain. For example, if we have 10 EOAs on OP mainnet, that means we could have up to 10 relayer txs per block.

ecosystem/auto-relayer/auto-relayer.md
```diff
@@ -0,0 +1,210 @@
...
+
+The gas fee charged by the relayer is based on `block.basefee` and not the `tx.gasprice` (see [#266](https://github.com/ethereum-optimism/design-docs/pull/266/files)), which means that the priority fee set by the relayer will not be charged to the user. In later iterations of this service, we will consider exploring a priority fee mechanism, where users can leverage a priority fee in order to increase the speed at which their messages are relayed.
+
+### Testing
+Since the relayer is a production critical service, it is important that we have a high degree of confidence that the changes  pushed to it do not break the service. As a part of this effort we will need to run integration tests that spin up the relayer against [kurtosis interop devnets](https://github.com/ethereum-optimism/optimism/tree/develop/kurtosis-devnet).
+
+## Failure Mode Analysis
```
fainashalts, line 194: Some of this is covered above, like how stuck transactions are handled. But just wanted to share my list of what I think should be covered in FMA in case its scheduled while I'm out:

- redis outage
- postgres lag
- stuck transactions
- funded wallet is empty
- chain reorg
- chain halt
- making sure keys are stored properly and we're rotating them

ecosystem/auto-relayer/auto-relayer.md
```diff
@@ -0,0 +1,210 @@
...
+- [Ponder](https://github.com/ponder-sh/ponder) for indexing of onchain events.
+- [Bullmq](https://github.com/taskforcesh/bullmq) for message queue.
+- Postgres database. 
+
+### Resource Usage
+
+#### Database Resource Usage
+The most significant source of resource usage for the database will be the indexing of all `SentMessage` and `RelayMessage` events from the `L2toL2CrossDomainMessenger`. Initially we will store all of these events without a retention period, however, as resource usage increases we can make optimizations to remove messages that have expired or are no longer needed because they have been relayed.
+
+### Transaction submission
```
fainashalts, line 121: Could we have Ponder stream events directly into a Redis/Bullmq queue via its event hooks instead of persisting everything first to Postgres and then polling? I think that might cut relay latency and reduce write load on the DB.

ecosystem/auto-relayer/auto-relayer.md
fainashalts: I'm wondering if we could go for dynamic concurrency - use a Prometheus-driven autoscaler that spawns additional workers (I assume one EOA per worker) when the queue age exceeds some metric, maybe 2x block time. Then we don't have to have a fixed high number of workers per chain but can still avoid the EOA bottleneck. 

If its not that big a deal to just fund enough wallets on each chain then this may be overkill. What do you think?

ecosystem/auto-relayer/auto-relayer.md
tremarkley: I like that. We should add event hooks to ponder to directly put new events on the queue.  If a balance in the fee vault, can cover that message, then the message can immediately be relayed. Persisting is still useful in the case where the message cant be relayed immediately at emission (because fee vault balance is too low) and then fee is added to the fee vault post-message being sent.

ecosystem/auto-relayer/auto-relayer.md
tremarkley: oh nice, this could help reduce complexity of keeping wallets funded. I'll explore this a bit, but I do think for now the drippie approach is pretty simple and if we are having to scale to an amount of EOAs that is difficult to manage even with drippie, then we should explore a more dynamic based approach.

PR #275, opened 2025-04-28
feat: update design doc for the dispute game creator pattern

stevennevins:
**Description**

Updating the design doc for the update to the DisputeGames to remove the immutable arguements from their constructor and instead have the values be passed as part of the CWIA payload for each proxy.  This allows the chain specific data to be abstracted from the deployed implementation and for different chains to share the same implementation

**Tests**

**Additional context**

**Metadata**


protocol/proofs/dispute-game-creators.md
```diff
@@ -0,0 +1,100 @@
+# Purpose
+
+The current fault-proof system requires a redeployment of the `DisputeGame.sol`, `DisputeGameFactory.sol`, and `AnchorStateRegistry.sol`. Due to chain-specific immutable variables (like `_weth`, `_anchorStateRegistry`, `_l2ChainId`) in its constructor. This necessitates redeploying associated `DisputeGameFactory.sol` and `AnchorStateRegistry.sol` contracts as well. This document proposes changes to simplify the Fault Dispute Game contract setup, enabling reuse of a single implementation across chains by externalizing chain-specific configuration.
```
ControlCplusControlV, line 3: Should specify when it requires a redeployment (ie when deploying a new chain, and when upgrading). It's also worth calling out the different type of immutables (CWIA vs Immutables) the dispute game contracts have right now. It's also not so much as externalizing as it is relocating, they should still be contained somewhere vaguely attached to the dispute game

ControlCplusControlV (commented):
Looks mostly good to me, we should have more direct language imo, I also don't think there are that many additional risks with regards to deployment issues across a lot of different chains

protocol/proofs/dispute-game-creators.md
```diff
@@ -0,0 +1,100 @@
...
+    IBigStepper vm,                 // address
+    IDelayedWETH weth,               // address
+    IAnchorStateRegistry registry,  // address
+    uint256 l2ChainId               // uint256 (e.g., uint256(10))
+);
+```
+
+# Alternatives Considered
+
+One alternative considered was to pass all chain-specific configuration purely within the `_extraData` field during game creation. This is similar to the proposed approach but would rely solely on `_extraData` instead of utilizing the structured `gameArgs` stored in the factory alongside the CWIA mechanism. While feasible, encoding all configuration into `_extraData` was deemed less explicit and potentially harder to manage compared to storing base configuration per `GameType` in the factory's `gameArgs` and combining it with instance-specific `_extraData` during cloning.
```
ControlCplusControlV, line 89: Another big issue is we'd have to expand `_extraData`'s size massively

stevennevins: curious what the issue would be with this vs what we have now?

protocol/proofs/dispute-game-creators.md
```diff
@@ -0,0 +1,100 @@
...
+
+# Risks & Uncertainties
+
+The primary risk with this proposed approach lies in the configuration of the `DisputeGameFactory.sol`. Since the factory would be responsible for providing the correct chain-specific parameters (`gameArgs`) for each `GameType`, any misconfiguration (e.g., incorrect addresses, IDs, or improperly encoded `bytes`) would result in incorrectly configured `FaultDisputeGame` proxies being created. Careful deployment and verification procedures for setting `gameArgs` would be crucial.
+
+Additionally, while using a single canonical implementation would reduce deployment overhead and potential for divergent implementations, it also concentrates risk. A bug in the single `FaultDisputeGame.sol` implementation would affect all chains using it, compared to the current model where a bug might be isolated to a specific chain's deployment. Thorough auditing and testing of the canonical implementation would be paramount.
```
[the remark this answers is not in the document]
stevennevins: Yeah agree here, will adjust

protocol/proofs/dispute-game-creators.md
```diff
@@ -0,0 +1,99 @@
...
+    IBigStepper vm,                 // address
+    IDelayedWETH weth,               // address
+    IAnchorStateRegistry registry,  // address
+    uint256 l2ChainId               // uint256 (e.g., uint256(10))
+);
+```
+
+# Alternatives Considered
+
+One alternative considered was to pass all chain-specific configuration purely within the `_extraData` field during game creation. This is similar to the proposed approach but would rely solely on `_extraData` instead of utilizing the structured `gameArgs` stored in the factory alongside the CWIA mechanism. While feasible, encoding all configuration into `_extraData` was deemed less explicit and potentially harder to manage compared to storing base configuration per `GameType` in the factory's `gameArgs` and combining it with instance-specific `_extraData` during cloning.
```
ajsutton, line 89: I think the big problem with this is that it would be completely insecure. Since game creation is permissionless and the extraData is specified when creating the game it would allow permissionless updating of the prestate and other critical inputs.

protocol/proofs/dispute-game-creators.md
```diff
@@ -0,0 +1,99 @@
...
+
+The goal is to allow a single, canonical `FaultDisputeGame.sol` implementation to be reused, with chain-specific parameters injected during the proxy creation process, simplifying deployments and upgrades.
+
+An important nuance here is that each `FaultDisputeGame.sol` deployment has both a minimal proxy with immutable args deployed that points at an implementation specific to the `FaultDisputeGameFactory.sol`. The two sets of immutable args are first specific to the fault dispute game, then the second set on the implementation are specific to the rollup.
+
+# Proposed Solution
+
+The `FaultDisputeGame.sol` contract be modified to remove chain-specific immutable variables (like `WETH`, `ANCHOR_STATE_REGISTRY`, `L2_CHAIN_ID`, etc.). Instead, these configuration parameters should be read dynamically from the `clones-with-immutable-args` (CWIA) data payload.
```
maurelian, line 19: So, are keeping CWIA in this solution?

protocol/proofs/dispute-game-creators.md
```diff
@@ -0,0 +1,99 @@
...
+
+The `DisputeGameFactory.sol` should be updated to facilitate this. It would include a `gameArgs` mapping (`GameType => bytes`) to store the necessary chain-specific configuration data for each game type. When `create()` is called, the factory would retrieve the appropriate `gameArgs` for the requested `GameType`, concatenate it with the game-specific data (`rootClaim`, `extraData`, etc.), and pass this combined `bytes` payload to the `clone` function.
+
+This approach would allow a single canonical `FaultDisputeGame` implementation to be deployed and reused across multiple chains or configurations. The factory would handle injecting the specific parameters needed for each instance via the CWIA payload. The implementation contract would still retain immutable arguments related to the core bisection game logic (e.g., `MAX_GAME_DEPTH`, `SPLIT_DEPTH`), but all chain-specific configuration would be externalized to the factory and the cloning process. This would simplify deployment for new chains, requiring only a factory configuration update and proxy deployment rather than a full implementation redeployment.
```
maurelian, line 23: is the clone just a simple proxy then?

protocol/proofs/dispute-game-creators.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+*Example `_args` payload structure (illustrative):*
+
+The `_args` bytes payload passed to `setImplementation` and stored in `gameArgs` would contain the ABI-encoded values for the base configuration parameters read by the `FaultDisputeGame` getters (starting from offset 0x74 in the full CWIA payload). This typically includes:
+
+```solidity
+// Example _args construction (conceptual, order based on implementation tests)
+bytes memory exampleArgs = abi.encodePacked(
+    GameType gameType,              // uint8 (e.g., uint8(0))
```
ajsutton, line 74: I wouldn't expect game type to be passed to the dispute game. Would be better to include this in the original deployment - deploying one implementation contract per game type seems reasonable and doesn't make it chain specific.

protocol/proofs/dispute-game-creators.md
```diff
@@ -0,0 +1,99 @@
...
+    Claim absolutePrestate,         // bytes32
+    uint256 maxGameDepth,           // uint256 (e.g., uint256(2 ** 3))
+    uint256 splitDepth,             // uint256 (e.g., uint256(2 ** 2))
+    Duration clockExtension,        // uint64/Duration (e.g., Duration.wrap(3 hours))
+    Duration maxClockDuration,      // uint64/Duration (e.g., Duration.wrap(3.5 days))
```
ajsutton, line 79: These are constants in the standard config. Can we just include them in the implementation contract since they will be the same for every L2 we support?

protocol/proofs/dispute-game-creators.md
```diff
@@ -0,0 +1,99 @@
...
+    IBigStepper vm,                 // address
+    IDelayedWETH weth,               // address
+    IAnchorStateRegistry registry,  // address
+    uint256 l2ChainId               // uint256 (e.g., uint256(10))
```
ajsutton, line 83: This is not needed for super games.  Given this will ship after interop we probably should only update the super contracts and leave FaultDisputeGame and PermissionedDisputeGame as is since they'll be deleted.  We can then remove this chain ID arg.

protocol/proofs/dispute-game-creators.md
stevennevins: yeah makes sense!


[a4f020a](https://github.com/ethereum-optimism/design-docs/pull/275/commits/a4f020a9325b47b073f789d5cfdcb3d1cc8bd88e)

protocol/proofs/dispute-game-creators.md
stevennevins: [d00e17f](https://github.com/ethereum-optimism/design-docs/pull/275/commits/d00e17f24d5b9569fcf4378da97ecbdf6988243f)

protocol/proofs/dispute-game-creators.md
stevennevins: Yeah agree one game type to one implementation seems reasonable [4184ccb](https://github.com/ethereum-optimism/design-docs/pull/275/commits/4184ccb26361e7cc922b94f44394601a44627c88)

protocol/proofs/dispute-game-creators.md
stevennevins: Yeah, definitely a good call out. I might have left out a security consideration in this, but it's definitely a pretty flimsy option

protocol/proofs/dispute-game-creators.md
```diff
@@ -0,0 +1,91 @@
...
+
+# Summary
+
+This proposal recommends modifying the `FaultDisputeGame.sol` contract to remove chain-specific immutable arguments. These parameters would instead be provided during proxy creation via a `bytes` payload constructed by the `DisputeGameFactory.sol`. The factory should be updated to store base configuration per `GameType` in a `gameArgs` mapping and combine it with game-specific data (`rootClaim`, `extraData`) at creation time of the proxy contract. This would allow new chain deployments to use proxies pointing to a canonical implementation, configured by the factory, eliminating the need for full contract redeployments.
```
ajsutton, line 7:
```suggestion
This proposal recommends modifying the `SupertDisputeGame.sol` contract to remove chain-specific immutable arguments. These parameters would instead be provided during proxy creation via a `bytes` payload constructed by the `DisputeGameFactory.sol`. The factory should be updated to store base configuration per `GameType` in a `gameArgs` mapping and combine it with game-specific data (`rootClaim`, `extraData`) at creation time of the proxy contract. This would allow new chain deployments to use proxies pointing to a canonical implementation, configured by the factory, eliminating the need for full contract redeployments.
```

Given we're shipping after interop we want to update the interop dispute games. No need to change the pre-interop games as they won't be relevant by the time this ships.

I guess we may have to update `FaultDisputeGame` as well if its still in the code base when this needs to merge but the focus should be on SuperDisputeGame and we should only update `FaultDisputeGame` if it makes this development easier than to just leave it as-is.

ajsutton: I'd just rename `FaultDisputeGame.sol` everywhere to `SuperDisputeGame.sol`.

ajsutton (approved):
LGTM. I'd just replace all the `FaultDisputeGame.sol` with `SuperDisputeGame.sol` since that's the actual dispute game implementation that will matter by the time this ships.

PR #293, opened 2025-06-24
feat: sequencer-trusting Following Mode

clabby:
## Overview

Introduces a new mode for the rollup node post-interop, where `cross-safe` promotion is delegated to the sequencer. In this mode, no supervisor needs to be consulted, and the operator trusts an attestation from the sequencer that message validity has been checked.


protocol/cross-safe-gossip.md
```diff
@@ -0,0 +1,93 @@
...
+[cross-safety][cross-safe].
+
+## Proposed Solution
+
+"Following Mode," at a high level, consists of [rollup nodes][rollup-node] subscribing to a new gossip topic that
+receives _signed [`cross-safe`][cross-safe] blocks_.
+
+1. When the sequencing node consults with its own [supervisor][supervisor] and determines that a block can be promoted
+to [`cross-safe`][cross-safe], it will sign the payload hash of the new [cross-safe][cross-safe] block as specified in
+["Rollup Node P2P"](https://specs.optimism.io/protocol/rollup-node-p2p.html#block-signatures) and gossip it to peers
```
clabby, line 42: gossip "it" - unclear, should gossip the whole signed payload envelope.

axelKingsley:
Sequencers do not have any special authority to deem a block "Cross Valid" because they do not have any special knowledge of other chains. An attestation network *could* be a good way to promote blocks which are Cross Valid, but there is no particular value to trusting a Sequencer here.

Even once a Sequencer *could* deem the block they created as Cross-Valid, that attestation may be invalidated by any other Sequencer in the set, simply by equivocating the Unsafe Blocks that are being depended upon. 

Counter-thought: why shouldn't we allow the Sequencer to publish "Follow Safe" blocks, so that nodes don't need to do L1 derivation in the first place? (I do not actually like this idea, just posing it as comparable-art)

clabby:
> Sequencers do not have any special authority to deem a block "Cross Valid" because they do not have any special knowledge of other chains. An attestation network _could_ be a good way to promote blocks which are Cross Valid, but there is no particular value to trusting a Sequencer here.
> 
> Even once a Sequencer _could_ deem the block they created as Cross-Valid, that attestation may be invalidated by any other Sequencer in the set, simply by equivocating the Unsafe Blocks that are being depended upon.
> 
> Counter-thought: why shouldn't we allow the Sequencer to publish "Follow Safe" blocks, so that nodes don't need to do L1 derivation in the first place? (I do not actually like this idea, just posing it as comparable-art)

Thanks for review! 

The sequencer doesn't have any special authority to deem a block cross-valid, but it does know of one, and can attest to the fact that it's listened to <insert some black-box authority that isn't the sequencer> to promote its own cross safe head. It can then use its signing key to attest to the fact that it observed said <insert some black-box authority that isn't the sequencer>'s verdict.

There's definitely a conversation to be had about whether or not this should actually be within the realm of the sequencer's role; We could always introduce a completely separate role to do this job, not overloading what we consider to be the sequencer today.

The design doc reuses this role as is primarily to maintain simplicity. Already, this key is available within the rollup node implementations where we need it to be, and as-is, we'd be able to implement this fairly easily without much plumbing at all.

ajsutton:
My thinking on this is more that people like RPC providers could run a small number of "complete" nodes that have op-supervisor and full op-mainnet and unichain-mainnet op-node/op-geth instances and then configure as many other rpc serving nodes as they need to run in following mode and trust the gossip from their own op-supervisor nodes.

That minimises the changes to their architecture, allows them to still decide what level of reliability they want for safe head signals (number of complete nodes) and continue scaling the rpc serving nodes for specific chains as needed without worrying about how to actually hook them to a supervisor at all. One important detail though is how to handle gossip that's only intended for nodes in your own fleet and not spam the whole gossip network with it. Maybe that complexity makes it not worth it....

It's possibly useful for the sequencer to publish this gossip as well for people like home users that want valid blocks but trust the cross-chain checks but that is a somewhat weird trust model.

PR #294, opened 2025-06-30
Add configurable calldata gas design document

niran:
This PR introduces a design document for configurable calldata gas pricing.


protocol/configurable-calldata-gas.md
```diff
@@ -0,0 +1,77 @@
...
+Pricing calldata correctly should make batcher sequencer throttling unnecessary and eliminate the priority fee auctions we see today when calldata is congested.
+
+---
+
+# Alternatives Considered
+
+* **Continue hard-coding gas costs**. Updating gas costs once or twice per year on a similar cadence to L1 hard forks is likely frequent enough for most chains, but it doesn't address differences in gas targets between OP Stack chains without each chain maintaining a persistent fork of the client.
+* **Full fee-abstraction** (see [`specs#73`](https://github.com/ethereum-optimism/specs/issues/73)). Only addresses fees, but we need to manipulate gas consumption for the fee market to function properly.
+* **Expose a single configuration value**. We don't foresee `eip7623StandardTokenCost` ever being modified from its default value of `4`, but it's included in the proposal for completeness.
+* **Change the actual calldata gas cost, not just the floor**. This diverges from how the L1 community has been addressing calldata costs and impacts more transactions than are necessary to solve the problem.
```
BrianBland, line 66: This approach seems simpler and also more directly addresses the issue of L2 calldata. EIP-7623 was implemented with a floor function in order to discourage calldata-focused transactions, steering them towards blobs instead. If all L2 data must be batched to L1, then we may want to meter all calldata equally.

protocol/configurable-calldata-gas.md
```diff
@@ -0,0 +1,77 @@
...
+| Unichain | 15M | ≈ 2.9× |
+| Mode | 15M | ≈ 2.9× | 
+
+Because so many OP Stack rollups have underpriced calldata, these chains rely on [batcher sequencer throttling](https://docs.optimism.io/operators/chain-operators/configuration/batcher#batcher-sequencer-throttling) to prevent large backlogs of batches from accumulating. The downside of this approach is that it "breaks the fee market." Users must outbid each other's priority fees in a first-price auction instead of expecting all transactions willing to pay the current base fee to be included in blocks. Many applications do not expect to actively participate in priority fee auctions because EIP 1559 has been so successful at eliminating them, so they experience these periods as a denial of service. (The base fee also plummets to zero during these periods, so even when the throttle is no longer binding, it takes time for the priority fee auction to end.)
+
+Pricing calldata correctly should make batcher sequencer throttling unnecessary and eliminate the priority fee auctions we see today when calldata is congested.
+
+---
+
+# Alternatives Considered
```
BrianBland, line 61: Another option may be to lean more heavily into the FastLZ function used for L1 data fee computation. We could choose to count all compressed data in excess of the floor (currently 100 bytes) as additional calldata tokens and then charge some configurable `TOTAL_COST_PER_COMPRESSED_TOKEN` gas price for each of these.
This approach more directly incentivizes compressible transactions, instead of simply treating zero bytes as a special case.

protocol/configurable-calldata-gas.md
```diff
@@ -0,0 +1,77 @@
...
+# Summary
+
+[EIP 7623](https://eips.ethereum.org/EIPS/eip-7623) uses `STANDARD_TOKEN_COST` and `TOTAL_COST_FLOOR_PER_TOKEN` parameters to calculate the floor cost of each transaction. This proposal makes these parameters configurable via `SystemConfig` instead of hard-coded in the client.
+
+We introduce **two new configuration values** that can be modified via the rollup's `SystemConfig` contract:
+
+| Name | Type | Default | Meaning |
+|------|------|-------------------|---------|
+| `eip7623StandardTokenCost` | `uint32` | `4` | The ratio of the cost of a non-zero byte in `tx.data` to the cost of a zero byte in `tx.data` |
+| `eip7623TotalCostFloorPerToken` | `uint32` | `10` | The cost of a zero byte in `tx.data` when calculating the total cost floor |
```
BrianBland, line 25: Can we use a uint16 or other smaller type? uint32 definitely seems like overkill for the first parameter, and we can almost certainly pack both of these into 32 or fewer bytes

niran: 2f36340876b4371a9b283098ac2659e0f56994fc

protocol/configurable-calldata-gas.md
niran: I think the floor is actually better. It caps the amount of calldata that can be used in a block without reducing the amount of execution that can happen in a block.

protocol/configurable-calldata-gas.md
niran: Oh, interesting! In that case, I agree that it makes more sense to double down on that rather that zero vs non-zero bytes. Updated in 9cdb74a.

protocol/configurable-calldata-gas.md
```diff
@@ -0,0 +1,85 @@
...
+| Unichain | 15M | 1s | ≈ 5.9× | 235 |
+| OP Mainnet | 20M | 2s | ≈ 3.9× | 158 |
+| Ink  | 15M | 2s | ≈ 2.9× | 120 |
+| Mode | 15M | 2s | ≈ 2.9× | 120 |
+
+Because so many OP Stack rollups have underpriced calldata, these chains rely on [batcher sequencer throttling](https://docs.optimism.io/operators/chain-operators/configuration/batcher#batcher-sequencer-throttling) to prevent large backlogs of batches from accumulating. The downside of this approach is that it "breaks the fee market." Users must outbid each other's priority fees in a first-price auction instead of expecting all transactions willing to pay the current base fee to be included in blocks. Many applications do not expect to actively participate in priority fee auctions because EIP 1559 has been so successful at eliminating them, so they experience these periods as a denial of service. (The base fee also plummets to zero during these periods, so even when the throttle is no longer binding, it takes time for the priority fee auction to end.)
+
+Pricing calldata correctly should make batcher sequencer throttling unnecessary and eliminate the priority fee auctions we see today when calldata is congested.
```
sebastianst, line 65: Sequencer throttling is still necessary because L1 congestion can occur due to spam on other rollups, other disruptions on L1 etc. It's an important backstop solution.

PR #295, opened 2025-07-02
Add minimum base fee design doc

niran:
**Description**

**Tests**

**Additional context**

**Metadata**

### DIFF
--- a/protocol/minimum-base-fee.md
+++ b/protocol/minimum-base-fee.md
@@ -1,0 +1,96 @@
+# Minimum Base Fee
+
+#### Metadata
+
+Authors: @niran
+Created: 2025-07-01
+
+---
+
+# Purpose
+ 
+This document proposes establishing a configurable minimum base fee for OP Stack chains to shorten the length of priority fee auctions.
+
+# Summary
+
+We introduce a new configuration value that can be modified via `SystemConfig`:
+
+| Name             | Type    | Default | Meaning |
+|------------------|---------|---------|---------|
+| `minBaseFeeLog2` | `uint8` | `0`     | The log2 of the minimum base fee that can be charged for a transaction |
+
+These values are updated via a new function in `SystemConfig`:
+
+```solidity
+function setMinBaseFeeLog2(uint8 minBaseFeeLog2) external onlyOwner;
+```
+
+This function will emit a `ConfigUpdate` log-event, with a new `UpdateType`: `UpdateType.MINIMUM_BASE_FEE`.
+
+Like [Holocene's dynamic EIP 1559 parameters](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/holocene/exec-engine.md#eip-1559-parameters-in-block-header), the minimum base fee is included in the block header's `extraData` field.
+
+## Minimum Base Fee in Block Header
+
+The `extraData` header field of each block must have the following format:
+
+| Name             | Type               | Byte Offset |
+| ---------------- | ------------------ | ----------- |
+| `version`        | `u8`               | `[0, 1)`    |
+| `denominator`    | `u32 (big-endian)` | `[1, 5)`    |
+| `elasticity`     | `u32 (big-endian)` | `[5, 9)`    |
+| `minBaseFeeLog2` | `u8`               | `[9, 10)`   |
+
+Additionally,
+
+- `version` must be 1
+- `denominator` must be non-zero
+- there is no additional data beyond these 10 bytes
+
+Note that `extraData` has a maximum capacity of 32 bytes (to fit in the L1 beacon-chain `extraData` data-type) and its
+format may be modified/extended by future upgrades.
+
+## Minimum Base Fee in `PayloadAttributesV3`
+
+The [`PayloadAttributesV3`](https://github.com/ethereum-optimism/specs/blob/a773587fca6756f8468164613daa79fcee7bbbe4/specs/protocol/exec-engine.md#extended-payloadattributesv3)
+type is modified to add an extra byte for `minBaseFeeLog2` to the `eip1559Params` field:
+
+```rs
+PayloadAttributesV3: {
+    timestamp: QUANTITY
+    prevRandao: DATA (32 bytes)
+    suggestedFeeRecipient: DATA (20 bytes)
+    withdrawals: array of WithdrawalV1
+    parentBeaconBlockRoot: DATA (32 bytes)
+    transactions: array of DATA
+    noTxPool: bool
+    gasLimit: QUANTITY or null
+    eip1559Params: DATA (9 bytes) or null
+}
+```
+
+# Problem Statement + Context
+
+Ethereum's EIP 1559 has successfully made first-price auctions for transaction fees rare. Users used to always have to set a fee that was high enough to outbid all other transactions to be included in a block. Now they can set a maximum fee that is high enough to cover the base fee with a margin for base fee increases caused by congestion and expect to be included in a block without ever having to outbid other transactions. In some conditions, Ethereum-based chains still experience first-price auctions via priority fees, and this happens more often on OP Stack chains. During these periods, many users experience a denial of service. They submit transactions that they expect to be included in a block but they are not included for long periods of time (and may be dropped from the mempool entirely).
+
+Priority fee auctions occur when blocks are near the gas limit or when the base fee is significantly below the market-clearing fee. Other than full blocks, the most common cause of priority fee auctions today is [batcher sequencer throttling](https://docs.optimism.io/operators/chain-operators/configuration/batcher#batcher-sequencer-throttling), which intentionally processes fewer transactions to avoid a backlog of batches that need to be posted to L1. Similarly, if the sequencer runs out of time to execute transactions for a block, the transactions that get included will be the ones with the highest priority fees.
+
+It can also be argued that **scaled** rollups will run well below their gas target in typical conditions. Ethereum's fees are _congestion_ fees, and scaled rollups should rarely be congested: their maximum capacity should be much higher than typical usage. A minimum base fee would be charged in the vast majority of cases, and higher fees would only be charged when the rollup needs to discourage transactions near its capacity. In this scenario, we would need to think of congestion often starting from base fees that are the lowest the chain will allow. Today, that's 1 wei, which takes far too long to adjust to the market prices that are needed to prevent first-price auctions using priority fees.
+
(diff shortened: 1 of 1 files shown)


protocol/minimum-base-fee.md
```diff
@@ -0,0 +1,96 @@
...
+
+It can also be argued that **scaled** rollups will run well below their gas target in typical conditions. Ethereum's fees are _congestion_ fees, and scaled rollups should rarely be congested: their maximum capacity should be much higher than typical usage. A minimum base fee would be charged in the vast majority of cases, and higher fees would only be charged when the rollup needs to discourage transactions near its capacity. In this scenario, we would need to think of congestion often starting from base fees that are the lowest the chain will allow. Today, that's 1 wei, which takes far too long to adjust to the market prices that are needed to prevent first-price auctions using priority fees.
```
BrianBland, line 77: I would reframe `gas target` as `congestion threshold`, as it cannot truly be an effective target when either minimum fees or exogenous costs (L1 DA, operator fees) influence the market

protocol/minimum-base-fee.md
```diff
@@ -0,0 +1,96 @@
...
+ 
+This document proposes establishing a configurable minimum base fee for OP Stack chains to shorten the length of priority fee auctions.
+
+# Summary
+
+We introduce a new configuration value that can be modified via `SystemConfig`:
+
+| Name             | Type    | Default | Meaning |
+|------------------|---------|---------|---------|
+| `minBaseFeeLog2` | `uint8` | `0`     | The log2 of the minimum base fee that can be charged for a transaction |
```
BrianBland, line 20: This technically changes the minimum base fee from 0 to 1 wei, although maybe that is an impossible condition in the OP stack as all blocks must contain the attributes transaction.

protocol/minimum-base-fee.md
```diff
@@ -0,0 +1,96 @@
...
+
+| Minimum Base Fee (gwei) | USDC Transfer Cost (ETH) | Cost (USD, ~$2,500/ETH) |
+| ----------------------- | ------------------------ | ----------------------- |
+| 0.00001                 | 4.05e-10                 | $0.0000010              |
+| **0.0001**              | **4.05e-9**              | **$0.0000101**          |
+| **0.001**               | **4.05e-8**              | **$0.000101**           |
+| **0.01**                | **4.05e-7**              | **$0.00101**            |
+| 0.1                     | 4.05e-6                  | $0.0101                 |
+| 1                       | 4.05e-5                  | $0.101                  |
```
BrianBland, line 90: While only minimally different, this should show example values which can be attained with the log2 config

protocol/minimum-base-fee.md
```diff
@@ -0,0 +1,99 @@
...
+We introduce a new configuration value that can be modified via `SystemConfig`:
+
+| Name             | Type    | Default | Meaning |
+|------------------|---------|---------|---------|
+| `minBaseFeeLog2` | `uint8` | `0`     | The log2 of the minimum base fee that can be charged for a transaction |
+
+These values are updated via a new function in `SystemConfig`:
+
+```solidity
+function setMinBaseFeeLog2(uint8 minBaseFeeLog2) external onlyOwner;
```
ajsutton, line 25: It may be worth adding some info around the trust/security properties of allowing the chain operator to configure this. In particular do we need a maximum value (u8 is already fairly limiting but can still set a pretty high base fee because its log2). I _think_ it probably works out o not changing trust assumptions because a malicious chain operator could set a very large L1 fee scalar and all these fees are circumvented via deposit transactions so it can't introduce a censorship risk. Seems like we should call out those security impacts (or non-impacts) though to prompt reviewers to think about ways it might be abused.

protocol/minimum-base-fee.md
```diff
@@ -0,0 +1,99 @@
...
+
+The current prevailing base fee on Base today is slightly below 0.001 gwei. It takes 411 blocks (13.7 minutes) for the base fee to increase from 1 wei to 0.001 gwei if every block is completely full at the gas limit. (In practice, we see 70% full blocks in these scenarios, so the fees take even longer to increase.) If we could configure a minimum base fee of 0.0001 gwei, we could speed up the fee adjustment (and the corresponding end of the priority fee auction) to 59 blocks, or a little under two minutes. We could even eliminate the priority fee auctions that occur while the chain is adjusting from the minimum to the typical market fee price by setting the minimum to that price to begin with. [Arbitrum One sets a minimum of 0.01 gwei](https://docs.arbitrum.io/how-arbitrum-works/gas-fees#child-chain-gas-pricing). At recent ETH prices around $2500, any of these three minimum base fees (which correspond to `2**20` through `2**24`) would be much less than one cent for a USDC transfer (about 40,500 gas). As ETH prices change, rollup operators will want to adjust the minimum base fee accordingly.
```
joshklop, line 81: Thank you for the well-written design doc!

As I understand it, the doc highlights or hints at two related UX problems, namely:

1. DoS during sudden demand spikes due to sluggish base fee updates.
2. The cognitive burden of participating in priority fee auctions.

The configurable minimum base fee seems like the simplest way to address the first problem without increasing base fee volatility in general. If you don't mind including reasons against adjusting the `BASE_FEE_MAX_CHANGE_DENOMINATOR`, I think it would strengthen the motivation for the specific proposed design.

As mentioned, it even partially addresses the second problem.

> We could even eliminate the priority fee auctions that occur while the chain is adjusting from the minimum to the typical market fee price by setting the minimum to that price to begin with.

We could go a step further and completely eliminate priority fee auctions by ignoring tips (or by letting the operator hardcode the tip, which may be a superior way to implement the operator fee; related: https://github.com/ethereum-optimism/design-docs/pull/299).

This so-called "Tipless" 1559 variant is analyzed in more detail [here](https://timroughgarden.org/papers/eip1559.pdf#page=43.61). It has some unrelated side-benefits and negligible drawbacks.

Have you considered the Tipless fee market in the context of the minimum base fee or otherwise? I’m curious if you have any thoughts or opinions. If it sounds promising, I may draft a separate design doc to avoid overloading this one.

Thanks again for taking the time to draft this.

niran: The best case against a higher EIP 1559 denominator is illustrated in Figure 3 of [this research paper on AIMD](https://arxiv.org/pdf/2110.04753#page=7), a proposal for an alternative base fee adjustment formula. It shows how the faster adjustment rates produced by lower denominators increase the volatility of the base fee. Some users will see longer wait times for their transactions to be confirmed when the base fee increases quickly, even if the increase in transaction volume was very temporary. A minimum base fee allows rollup operators to calibrate their denominator for the base fee volatility they'd like to see near the equilibrium base fee without worrying about the extreme cases.

I'll add this to the document!

> Have you considered the Tipless fee market in the context of the minimum base fee or otherwise?

Nope, it's new to me. Thanks for sharing! My first impression is that it would eliminate the first price auctions, but it would do so by removing the market mechanism for selecting which transactions should be included in full blocks. Instead, arbitrary transactions would be included and the others would see degraded performance. That is, the bad experiences don't go away, they just get allocated in a different way.

protocol/minimum-base-fee.md
```diff
@@ -0,0 +1,99 @@
...
+| 20 | 0.001048576 | 4.25e-8 | $0.000106 |
+| 21 | 0.002097152 | 8.49e-8 | $0.000212 |
+| 22 | 0.004194304 | 1.70e-7 | $0.000425 |
+| 23 | 0.008388608 | 3.40e-7 | $0.000849 |
+| 24 | 0.016777216 | 6.79e-7 | $0.00170 |
+| 25 | 0.033554432 | 1.36e-6 | $0.00340 |
+| 26 | 0.067108864 | 2.72e-6 | $0.00679 |
+| 27 | 0.134217728 | 5.44e-6 | $0.0136 |
+
+# Alternatives Considered
```
geoknee, line 95: Could we add a "backwards compatibility" section, to anticipate the effects on L2 users and applications?

May also be worth a note somewhere that this feature could be effectively disabled by setting `minBaseFee` to `0` (significantly reducing the risk of the feature).

protocol/minimum-base-fee.md
```diff
@@ -0,0 +1,99 @@
...
+
+* **Hard-code a minimum base fee** (like [EIP 7762](https://eips.ethereum.org/EIPS/eip-7762)). This is the simplest solution, but it doesn't give the chain operator any flexibility to adjust the minimum base fee. Establishing a Superchain-wide consensus on a minimum base fee seems difficult, especially if the value will have to wait for the next fork to be changed.
+* **Link the minimum base fee to an outside price signal** (like [EIP 7918](https://eips.ethereum.org/EIPS/eip-7918)). This approach avoids choosing a magic number for the minimum base fee and has more elegant economics: the fee doesn't drop when we have a good signal that reducing the base fee wouldn't increase gas usage. The downside is that it's harder to reason about the right values to use in the formula, which makes it hard to build a consensus around. A configurable minimum base fee gives us an option that L1 doesn't have: it allows rollup operators to change the minimum base fee as needed to ensure the fee market is working properly.
```
geoknee, line 98: IIUC these two EIPs apply to `blobBaseFee`, and not regular `baseFee`. Are there EIPs for adding a minimum base fee?

In general I'm interested in how we might justify adopting a different approach from L1 in addressing what seems like a common problem. I think you hint at some of the reasons above, but would be great to concentrate that argument in it's own section. In general we try to maintain as small as possible diff to L1 (of course certain things demand a diff and that's fine).

protocol/minimum-base-fee.md
joshklop: > A minimum base fee allows rollup operators to calibrate their denominator for the base fee volatility they'd like to see near the equilibrium base fee without worrying about the extreme cases.

Makes sense, thanks! The graphs are convincing.

> That is, the bad experiences don't go away, they just get allocated in a different way.

Good point. Other than not needing to worry about setting a tip, users likely won't see much benefit. Probably not worth the L1 diff.

protocol/minimum-base-fee.md
niran: 2e1c7200a4651bb31f3bd00446653d985e2a6ba1

protocol/minimum-base-fee.md
```diff
@@ -0,0 +1,99 @@
...
+
+| log2(minBaseFee) | Minimum Base Fee (gwei) | USDC Transfer Cost (ETH) | Cost (USD, ~$2,500/ETH) |
```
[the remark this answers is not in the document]
niran: 048b4f77ce96846291b78def46301492ef243878

protocol/minimum-base-fee.md
niran: d0f5bf845c38c46efece544751467eb9b23f4b0f

PR #297, opened 2025-07-15
feat(protocol): Feature Decoupling

refcell:
### Description

This document proposes decoupling features from hardforks in the OP Stack.

The primary change proposed is to add a feature map to `RollupConfig`s in
the superchain-registry that map a feature to the hardfork it is active in.

By mapping features to hardforks this way, features can be granularly toggled
on/off allowing for individual feature testing, custom chain configurations, and
phased rollouts while maintaining coordinated across the superchain via the
superchain-registry (also called "SCR").

Shoutout to @geoknee and @sebastianst for the early reviews on this document.

### DIFF
--- a/protocol/decoupled-hardfork-features.md
+++ b/protocol/decoupled-hardfork-features.md
@@ -90,11 +90,13 @@ If a feature is absent from the map, it remains inactive.
     - Look up the hardfork for the feature.
     - Retrieves the hardfork's timestamp (e.g., via existing fields like `IsthmusTime`).
     - Returns true if `timestamp >= hardforkTimestamp`.
-- **Client-Level Overrides:** Extend CLI flags for direct per-feature timestamp overrides
-  (e.g., `--override.feature.eip4844=isthmus` or `--override.feature.eip4844.timestamp=123456`),
-  similar to current hardfork overrides.
 - **Superchain Compatibility:** For chains with `superchain_time`, inherit mappings from
   a central registry, but allow per-chain overrides.
+- **(OPTIONAL) Client-Level Overrides:** Extend CLI flags for direct per-feature timestamp
+  overrides (e.g., `--override.feature.eip4844=isthmus` or
+  `--override.feature.eip4844.timestamp=123456`), similar to current hardfork overrides.
+  Notice that CLI overrides are completely optional and are left to the discretion of
+  individual implementations.
 
 These changes allow *both* chain operators and client/protocol developers to customize
 which features are active in which hardfork.
@@ -245,6 +247,49 @@ max_sequencer_drift = 600
 ```
 
 # Alternatives Considered
+
+### **Feature Fallbacks**
+
+This is an added alternative first suggested on github by @ajsutton. Most
+of the ideas here are adapted from @ajsutton.
+
+"Feature Fallbacks" is an alternative approach to feature decoupling
+that avoids making any changes at all to the superchain registry. Instead,
+feature fallbacks happen at the client implementation level where
+individual features within a hardfork have their own activation selectors
+on the `RollupConfig` that fallback to their hardforks.
+
+For example, an activation check for the operator fee that is part of the
+Isthmus hardfork could look like the following.
+
+```
+func IsOperatorFeeEnabled() {
+  return IsIsthmusEnabled()
+}
+```
+
+> [!NOTE]
+>
+> Notice that with this approach, the method could just be modified to
+> return false if the feature needs to be disabled for some reason.
+
+It's worth noting that this is less flexible for testing since the features
+are not configurable through the `RollupConfig` programmatically, but there's
+no risk of misconfiguration. This approach is typically used on L1 though
+because the features that are actually in a hardfork doesn't change often.
+
+Feature fallbacks like this make sense if this decoupling only exists during
+development of the hard fork. Once the hardfork scope is finalized, individual
+feature toggles are removed from the implementation, locking in what's part
+of the hardfork. The benefit of this approach is that no changes to the
+superchain registry are required. On the flipside, it adds work on the backend
+to remove individual feature toggles. One consideration is development may
+revert to previous bad habits, where the feature toggle is never added to begin
+with because it's considered "well scoped enough" or "fairly committed to".
+
+On the flipside, adding features to the superchain registry would make sense
+if the intent is to keep features decoupled from hard forks permanently.
+
 
 ### **Individual Feature Timestamps**
 


protocol/decoupled-hardfork-features.md
```diff
@@ -0,0 +1,308 @@
...
+# Proposed Solution
+
+Decouple features from hardforks by modifying `RollupConfig`s to include a
+mapping from feature names to hardfork names (e.g., a `map[Feature]Fork`
+where the key is feature like "operatorFee" and value is a hardfork like "isthmus").
+If a feature is absent from the map, it remains inactive.
+
+### Key Changes
+
+- **Superchain Registry Updates:** Add a mapping from features to their hardfork
```
mslipper, line 84: Do we really need to update the Superchain registry here? I think it would be much simpler to hardcode the hardfork activations in the clients themselves once we agree on a scope of an upgrade. The code deduplication we get from this doesn't seem worth it, and different clients may implement features differently (e.g. a Reth client may have different feature flag names from a Nethermind client), or have a completely different set of feature flags.

protocol/decoupled-hardfork-features.md
```diff
@@ -0,0 +1,308 @@
+# Purpose
+
+The purpose of this design document is to propose a mechanism for decoupling
+individual features from their hardforks in the OP Stack. This will enhance
+flexibility in feature activation, enabling granular control for testing,
+custom chain configurations, and phased rollouts while maintaining
+coordination across the superchain. The design aims to improve modularity
```
joshklop, line 7: > custom chain configurations

Is there a concern that some chains may activate features in different forks, leading to pathological configurations that are untested and unaudited? i.e., even if a set of features is sound and reasonable, a subset may not be.

I'm not sure how this interacts with the definition of "standard config", but at first glance it seems prudent to mandate that the same features be deployed at the same forks.

protocol/decoupled-hardfork-features.md
```diff
@@ -0,0 +1,308 @@
...
+implementations of OP Stack components went ahead and implemented the
+feature. For months, the feature sat idle. After it was eventually merged,
+there were issues in the execution client changes for the feature. Removing
+implementations of the feature was considered, but the effort to remove it
+was not seen as worthwhile. After a while, the feature was finally
+complete. It effectively alone delayed the hardfork by a month or so.
+
+This could have been avoided by being able to implement the feature using
+a feature-specific toggle in the `RollupConfig`. The feature toggle would
+be toggled on in the [`superchain-registry`][scr] once the feature is ready.
```
ajsutton, line 39: Putting the feature toggle in the rollup config is one way, but we could have equally feature toggled it by using `if IsOperatorFeeEnabled()` in place of `if IsIsthmus()` and while we expect operator fee to ship with Isthmus it's implementation could just delegate to `IsIsthmus`.  Though in the case of operator fees there were also smart contract changes involved which would also have needed to be feature toggled (and notably those don't have access to the rollup config).

protocol/decoupled-hardfork-features.md
```diff
@@ -0,0 +1,308 @@
...
+In this way, features could be implemented and left in client software
+without being connected to hardfork. The key part: *no changes would
+need to be made to client software to enable the feature*, only enabling
```
ajsutton, line 42: Both runtime and compile time feature toggles can be useful in different cases, but it is definitely easier to test things when using runtime feature toggles (e.g. like the suggested option in the rollup config or a CLI arg). The `IsOperatorFee()` method delegating to `IsIsthmus()` would be an example of a compile time feature and particularly when done like that would be impossible to test in the non-default configuration.

Generally, the ease of testing makes runtime feature toggles preferable, but they do have a greater risk of misconfiguration.

protocol/decoupled-hardfork-features.md
```diff
@@ -0,0 +1,308 @@
...
+
+
+# Problem Statement
+
+Bundling features within hardforks limits flexibility:
+
+- All features in a hardfork activate at once, preventing selective enablement
+  for testing or debugging.
+- Custom chains or testnets cannot easily remap or delay individual features
+  without forking entire client implementations.
```
ajsutton, line 53: Is this something we actually want to support? While a hard fork is in development it can be useful to be able to run devnets with a subset of features that are ready for that testing, but having custom chains pick and choose features from within hard forks creates an explosion of feature combinations that we would need to test and support forever.

I think we should be very, very careful about opening up that Pandora's box.

protocol/decoupled-hardfork-features.md
```diff
@@ -0,0 +1,308 @@
...
+- CLI overrides are limited to hardfork timestamps, not per-feature adjustments.
```
ajsutton, line 54: Note: This is true of the overrides we've currently added, but doesn't have to be true. We could also support CLI overrides for specific in-development features.  Though for consensus affecting things the rollup config is likely a better place since it can be more easily shared between op-geth and op-reth.

protocol/decoupled-hardfork-features.md
```diff
@@ -0,0 +1,308 @@
...
+    - Overrides in the `op-node` exist [here][op-overrides]. In the `kona-node`,
+      [here][kona-overrides].
+
+
+# Proposed Solution
+
+Decouple features from hardforks by modifying `RollupConfig`s to include a
+mapping from feature names to hardfork names (e.g., a `map[Feature]Fork`
+where the key is feature like "operatorFee" and value is a hardfork like "isthmus").
+If a feature is absent from the map, it remains inactive.
```
ajsutton, line 80: Is this done during development and then the set of features in a hard fork and hard coded when they are shipped (either in the release that sets the activation timestamp for testnet or mainnet or post-main net activation)? Or is the intent to keep the individual feature toggles forever and continue to have to list them in the rollup config?

protocol/decoupled-hardfork-features.md
```diff
@@ -0,0 +1,308 @@
...
+  configs if an override is needed.
+- **RollupConfig Update:** Add a field to map feature to its hardfork. For example,
+  a `features` field in the `RollupConfig` is a map from feature to hardfork name.
+- **Activation Check:** Add methods like `IsFeatureActive(timestamp uint64) bool`, which:
+    - Look up the hardfork for the feature.
+    - Retrieves the hardfork's timestamp (e.g., via existing fields like `IsthmusTime`).
+    - Returns true if `timestamp >= hardforkTimestamp`.
+- **Client-Level Overrides:** Extend CLI flags for direct per-feature timestamp overrides
+  (e.g., `--override.feature.eip4844=isthmus` or `--override.feature.eip4844.timestamp=123456`),
+  similar to current hardfork overrides.
```
ajsutton, line 95: Do we actually need to do both of these? Seems like having it _either_ in the rollup config _or_ as a CLI flag would be sufficient and reduce the amount of code needing to be written for each feature.

Even if we create a framework that does make it easy to do both (and we should make it easy regardless), it seems like having two places to specify this would add confusion without really giving any extra flexibility. Would be interested if there's real use cases where you couldn't just edit the rollup config to change what's enabled.

protocol/decoupled-hardfork-features.md
ajsutton: It seems like adding this to the superchain registry makes sense _if_ the intent is to keep features decoupled from hard forks permanently - ie one chain could have operator fee as part of Isthmus, another activates it as part of Jovian and another never activates it.

Whereas if the decoupling only exists during development of the hard fork and we then when the hard fork scope is finalized and everything is confirmed as ready to ship, we remove the individual feature toggles, locking in what is or isn't part of Isthmus, then I'd say we don't need it in the superchain registry since most devnets aren't in the superchain registry and we can keep things simpler.

protocol/decoupled-hardfork-features.md
```diff
@@ -0,0 +1,308 @@
...
+  AnchorStateRegistryProxy = "0x1c68ECfbf9C8B1E6C0677965b3B9Ecf9A104305b"
+  DelayedWETHProxy = "0x21429aF66058BC3e4aE4a8f2EC4531AaC433ecbC"
+  DisputeGameFactoryProxy = "0xe5965Ab5962eDc7477C8520243A95517CD252fA9"
+  FaultDisputeGame = "0x5738a876359b48A65d35482C93B43e2c1147B32B"
+  MIPS = "0xF027F4A985560fb13324e943edf55ad6F1d15Dc1"
+  PermissionedDisputeGame = "0x1Ae178eBFEECd51709432EA5f37845Da0414EdFe"
+  PreimageOracle = "0x1fb8cdFc6831fc866Ed9C51aF8817Da5c287aDD3"
+```
+
+# Alternatives Considered
```
ajsutton, line 247: It's probably also worth listing the hard coded feature toggle I mentioned above as an option here. e.g.:

```
func IsOperatorFeeEnabled() {
  return IsIsthmusEnabled()
}
```
which could be replaced with `return false` if we need to disable the feature for some reason. Less flexible for testing, but no risk of misconfiguration. This approach is typically used on L1 though because which features are actually in a hard fork actually don't change very often.

mslipper:
I will follow up with product to get their input on whether or not we should enable toggles outside of development.

protocol/decoupled-hardfork-features.md
```diff
@@ -0,0 +1,308 @@
...
+One could argue that the problem statement is a result of the hardfork upgrade process
+- not the actual implementation method. Instead of handling this at the implementation
+level, fix the upgrade process so that a hardfork’s feature set is properly established
+prior to implementation.
+
+# Risks
+
+- **Increased Complexity:** Adding maps/methods complicates config management and client
+  logic, risking bugs in activation checks.
+- **Coordination Challenges:** Decoupling might lead to inconsistent feature enablement
```
mslipper, line 291: I feel strongly that once we decide what goes into a hardfork, we disable the ability to modify the feature flags that go within it.

protocol/decoupled-hardfork-features.md
refcell: Do I understand correctly that the main change you're suggesting here is not to feature toggle in the SCR, but instead just feature toggle in the `RollupConfig` implementations in clients?

I actually think this is a great idea. My one concern though is if we think a few years down the road, without centralizing on a set of features, we could see clients meaningfully diverge as they implement features in different ways. Maybe this is a feature and not a bug (no pun intended), but it's something to keep in mind. Being explicit + centralized about feature names and what hardfork they target could prevent a lot of tech debt build up and potential chain split surface area.

protocol/decoupled-hardfork-features.md
refcell: Two thoughts here:
1. If features are defined in SCR, it makes it difficult or at least a very explicit change in client software to enable or disable features out-of-band (for the intended hardfork)
2. In contrast, if features are only handled in development, changes to these features in forked clients are somewhat more expected.

protocol/decoupled-hardfork-features.md
refcell: Good point. I will note that this is an optional change for posterity. I only included to be cohesive about how features could be configured.

protocol/decoupled-hardfork-features.md
pauldowman: I don't think feature flags should be reused as configuration options. I think it's tempting because without looking closely they're so similar. But it ends up messy because they serve different purposes.

The reason is that feature flags should be lightweight and easy to add and remove (and should be aggressively removed or they become a mess), but configuration options need versioning and deprecation once customers are using them.

ajsutton: Open source code doesn't mean you have to support every crazy thing anyone does. We very much get to choose what we support and if other people fork to add things we chose not to support that's fine, but that's their problem, not ours. If it's an important thing that we think we should support, then we make it a feature - either built-in or by providing explicit extension points to allow others to build it in a supported way.

But we just can't afford to support every fork and every feature anyone ever builds.

PR #299, opened 2025-07-16
feat: operator fee scalar fix

yuwen01:
This is the current operator fee formula.

```
operatorFee = operatorFeeConstant + operatorFeeScalar * gasUsed / 1e6
```

The `1e-6` multiplier on the `operatorFeeScalar` is not useful for operators as it results in a fee that is too low to be useful. This design doc proposes the following change to the operator fee formula.

```
operatorFee = operatorFeeConstant + operatorFeeScalar * gasUsed * 10
```

With the new fee formula, here are some examples of some possible fees being charged. 
For the dollar cost column, we assume an ETH price of 2.5k.

| Gas Used | `operatorFeeScalar` | `operatorFeeScalar * gasUsed * 100` | Dollar Cost |
|----------|-------------------|----------------------------------|-------------|
| 30,000,000 | 1 | 3 gwei | $0.0000075 |
| 30,000,000 | 2^32 | 1.29e10 gwei | $32,250.00 |
| 1,000,000 | 1 | 0.1 gwei | $0.00000025 |
| 1,000,000 | 2^32 | 4.29e8 gwei | $1,072.50 |

Fixes https://github.com/ethereum-optimism/optimism/issues/16623

### DIFF
--- a/protocol/operator-fee-scalar-fix.md
+++ b/protocol/operator-fee-scalar-fix.md
@@ -30,7 +30,7 @@ operatorFee = operatorFeeConstant + operatorFeeScalar * gasUsed / 1e6
 The `1e-6` multiplier on the `operatorFeeScalar` is not useful for operators as it results in a fee that is too low to be useful. This design doc proposes the following change to the operator fee formula.
 
 ```
-operatorFee = operatorFeeConstant + operatorFeeScalar * gasUsed * 10
+operatorFee = operatorFeeConstant + operatorFeeScalar * gasUsed * 100
 ```
 
 <!-- Most (if not all) documents should have a summary.
@@ -63,19 +63,19 @@ is likely too low level. -->
 We propose the following change to the operator fee formula.
 
 ```
-operatorFee = operatorFeeConstant + operatorFeeScalar * gasUsed * 10
+operatorFee = operatorFeeConstant + operatorFeeScalar * gasUsed * 100
 ```
 
 Where `operatorFeeScalar` and `operatorFeeConstant` remain `u32` and `u64` respectively.
 
 Examine the following to get a sense of the range of fees this formula supports. For the dollar cost column, we assume an ETH price of 2.5k.
 
-| Gas Used | `operatorFeeScalar` | `operatorFeeScalar * gasUsed * 10` | Dollar Cost |
+| Gas Used | `operatorFeeScalar` | `operatorFeeScalar * gasUsed * 100` | Dollar Cost |
 |----------|-------------------|----------------------------------|-------------|
-| 30,000,000 | 1 | 0.3 gwei | $0.00000075 |
-| 30,000,000 | 2^32 | 1.29e9 gwei | $3,225.00 |
-| 1,000,000 | 1 | 0.01 gwei | $0.000000025 |
-| 1,000,000 | 2^32 | 4.29e7 gwei | $107.25 |
+| 30,000,000 | 1 | 3 gwei | $0.0000075 |
+| 30,000,000 | 2^32 | 1.29e10 gwei | $32,250.00 |
+| 1,000,000 | 1 | 0.1 gwei | $0.00000025 |
+| 1,000,000 | 2^32 | 4.29e8 gwei | $1,072.50 |
 
 Given that the current scalar component fees are negligible, all `operatorFeeScalar` values are set to 0 upon upgrade. Then, chain operators will be responsible for adjusting them according to the new formula. The `operatorFeeConstant` will not change upon transition.
 
--- a/protocol/operator-fee-scalar-fix.md
+++ b/protocol/operator-fee-scalar-fix.md
@@ -68,16 +68,20 @@ operatorFee = operatorFeeConstant + operatorFeeScalar * gasUsed * 10
 
 Where `operatorFeeScalar` and `operatorFeeConstant` remain `u32` and `u64` respectively.
 
-Examine the following to get a sense of the range of fees this formula supports.
+Examine the following to get a sense of the range of fees this formula supports. For the dollar cost column, we assume an ETH price of 2.5k.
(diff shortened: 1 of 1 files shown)


tynes:
Could you add some additional fields to the table that make it easy to see how much USD is charged? ie an ETH/USD assumption and then a USD fee charged. This will help to make sure that the thing that impacts the end user is accounted for

tynes:
Do you mind calling out the parts of the codebase that will need to change? I believe no change to `SystemConfig` or `L1Block` given that the type for the scalar isn't changing, there is a change to the `GasPriceOracle` and the fee charging logic in the state transition function

protocol/operator-fee-scalar-fix.md
```diff
@@ -0,0 +1,95 @@
...
+## Alternatives considered
+
+Instead of changing the scalar value, it is also possible to address this problem by changing the `operatorFeeScalar` to a `u64`. With this solution, we would need to add a new field to the `L1Attributes` transaction, deprecate the `setOperatorFeeScalars` function in the `SystemConfig` contract, and add a new function `setJovianOperatorFeeScalars`. Due to the increased complexity, this solution is not preferred.
+
+# Risks & Uncertainties
+
+<!-- An overview of what could go wrong.
+Also any open questions that need more work to resolve. -->
+
+If anyone is using the `operatorFeeScalar` in production, automatically setting it to 0 upon upgrade may cause issues. Since the fee is so low right now, it is likely that any fees lost will be negligible.
```
geoknee, line 95: This seems right and far less of a risk than an nonzero value suddenly being scaled by 8 orders of magnitude. But that does underline the importance of the "automatically setting it 0" part (see other comment).

geoknee, line 95: If we are certain no one is using operator fee in production, it might be possible to make the change as a soft fork and discard the old consensus code. Do you have thoughts on that?

sebastianst: I agree, this is an interesting question also for Product. If noone is using this feature right now, because it's essentially useless in its current form, we may as well just fix it outside of a fork.

protocol/operator-fee-scalar-fix.md
```diff
@@ -0,0 +1,95 @@
...
+Where `operatorFeeScalar` and `operatorFeeConstant` remain `u32` and `u64` respectively.
+
+Examine the following to get a sense of the range of fees this formula supports. For the dollar cost column, we assume an ETH price of 2.5k.
+
+| Gas Used | `operatorFeeScalar` | `operatorFeeScalar * gasUsed * 100` | Dollar Cost |
+|----------|-------------------|----------------------------------|-------------|
+| 30,000,000 | 1 | 3 gwei | $0.0000075 |
+| 30,000,000 | 2^32 | 1.29e10 gwei | $32,250.00 |
+| 1,000,000 | 1 | 0.1 gwei | $0.00000025 |
+| 1,000,000 | 2^32 | 4.29e8 gwei | $1,072.50 |
```
sebastianst, line 78: I think this table supports that `100` is a good factor, it seems to cover all imaginable ranges.

protocol/operator-fee-scalar-fix.md
```diff
@@ -0,0 +1,95 @@
...
+
+Given that the current scalar component fees are negligible, all `operatorFeeScalar` values are set to 0 upon upgrade. Then, chain operators will be responsible for adjusting them according to the new formula. The `operatorFeeConstant` will not change upon transition.
```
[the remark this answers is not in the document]
geoknee: Right -- does this mean we have to modify the L1BlockContract / derive a transaction to do that? It needs to "stick"  until the next update of the SystemConfig contract, right?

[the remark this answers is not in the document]
sebastianst: We would probably implement it as an update deposit transaction.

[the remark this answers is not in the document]
geoknee: With this migration strategy, if a chain had a nonzero scalar set on L1, it would be out of sync with the actual value on the chain until it was updated again.

protocol/operator-fee-scalar-fix.md
sebastianst: It would also be interesting to add two more entries for a tx that only consumes, say, 21,000 gas.

protocol/operator-fee-scalar-fix.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Scope of changes
+
+This fix will require changes to the `GasPriceOracle` contract as well as execution clients like `op-geth`, in order to reflect the new operator fee formula.
+
+## Alternatives considered
```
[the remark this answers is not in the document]
geoknee: In the review meeting, we agreed that this doesn't make as much sense here where we would like better resolution to tune profit margins etc.

protocol/operator-fee-scalar-fix.md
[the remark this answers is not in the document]
geoknee: We are used to that situation and can address it with docs / comms.

protocol/operator-fee-scalar-fix.md
geoknee: Design review meeting, we aligned on the idea that we should ship this as a regular fork. If any chain wants to test early, they will be able to do that even with an incomplete (Jovian) feature set. Not recommended for production networks.

protocol/operator-fee-scalar-fix.md
geoknee: Let's add another row to this table with the minimum gas to get full confidence on this table.

geoknee (approved):
We got good alignment on the review call. Just need to update the table and then we can merge the PR (ratify it).

PR #302, opened 2025-07-17
Add design doc on OPD bundled artifacts

mslipper:


protocol/op-deployer-embedded-artifacts.md
```diff
@@ -0,0 +1,67 @@
...
+This change would increase the size of the OP Deployer binary by ~60MB after removing test code from the artifacts.
+It is our belief that this is an acceptable tradeoff, particularly since users would have to download the artifacts
+anyway.
+
+For local development, we can make the embedding optional. This will avoid increasing Go compile times when we're not
+cutting an official OP Deployer release.
+
+As part of this solution, we will start developing OP Deployer on the same proposal branches as the contracts 
+themselves. This way, OP Deployer is released side-by-side with the contracts, and we can ensure that they always 
+stay in sync.
```
ajsutton, line 55: How viable would it be to just merge op-deployer and contract releases? ie tagging a contacts release triggers a build of op-deployer containing those contracts and publishes the binary and/or docker image as "the release". It may also publish the contracts in some other format if that's useful.

Just seems like if we want people to use op-deployer to deploy and upgrade contracts, then having releases just be the tool you run to deploy or upgrade seems like it would make a lot of sense. Like shipping an installer.

janjakubnanista: Yeah second that - if we are embedding, i would ensure that we have at least one op-deployer release per contracts release.

The thing to figure out is the versioning. Since op-deployer can also receive features & bugfixes, its version is not synonymous with the contracts version. If though these versions don't match, we'll end up in a similar situation as we are now - we would need to carry around some mapping of which op-deployer version has which contracts.

protocol/op-deployer-embedded-artifacts.md
```diff
@@ -0,0 +1,67 @@
...
+  Deployer, for example, only supports smart contracts from `op-contracts/v2.0.0`. However, the version of the smart
+  contracts can sometimes change by minor or patch versions. This makes it difficult to determine which version of
+  the contracts each version of OP Deployer supports, and can lead to unexpected behavior or broken deployments.
+
+In addition, maintaining the tagged locators themselves is non-trivial amount of work:
+
+- New releases require contract artifacts to be built, pushed to GCP, and added to `standard.go`. This means that
+  every new contract release requires a corresponding code change to OP Deployer.
+- Because the artifacts each locator points to are on GCP, they need to be validated client-side before being used.
+  This adds complexity, overhead, and potential for security vulnerabilities.
```
maurelian, line 33: For my own understanding this validation is done using the values in standard.go right?

protocol/op-deployer-embedded-artifacts.md
maurelian: +1. 
My sense is that this would just be a matter of tooling, and that removing the need to add hashes to standard.go will make op-deployer and op-contracts releasable from the same commit.

mslipper: I'm onboard with this. It should be simple to do. My strawman would be to simply version OP Deployer using the contract versions - e.g., a patch fix to op-deployer would bump op-contracts/vX rather than op-deployer/vX.

protocol/op-deployer-embedded-artifacts.md
```diff
@@ -0,0 +1,67 @@
...
+## Proposed Solution
+
+Rather than continuing to support tagged locators, we propose embedding the contract artifacts for a given release
+directly inside the OP Deployer binary itself using `go:embed`. HTTPs and file locators will continue to work as
+before. The CLI will select the tagged locator by default if it exists, so users will not generally have to specify
+a locator at all when invoking OP Deployer.
+
+This change would increase the size of the OP Deployer binary by ~60MB after removing test code from the artifacts.
+It is our belief that this is an acceptable tradeoff, particularly since users would have to download the artifacts
+anyway.
```
bitwiseguy, line 48: Does the ~60MB include both contract source code AND forge scripts? I'm assuming we'd want to embed both

protocol/op-deployer-embedded-artifacts.md
```diff
@@ -0,0 +1,67 @@
...
+
+As part of this solution, we will start developing OP Deployer on the same proposal branches as the contracts 
+themselves. This way, OP Deployer is released side-by-side with the contracts, and we can ensure that they always 
+stay in sync.
+
+## Impact on Developer Experience
+
+Tools like Kurtosis will continue to be able to use HTTPs and file locators. If users want to use a tagged locator,
+they will need to change the version of OP Deployer they use rather than change the locator itself. This will likely
+impact Netchef the most.
```
bitwiseguy, line 61: Which kinds of contract changes in https/file locators compared to the tagged contract releases do we expect to be supported by the same op-deployer version? If an https/file locator is used, is it expected that a non-tagged version of `op-deployer` will often be used?

protocol/op-deployer-embedded-artifacts.md
bitwiseguy: Definitely seems more reliable to trigger op-deployer and contract releases from the same tag/commit, instead of relying on someone to do both steps manually each time.

protocol/op-deployer-embedded-artifacts.md
```diff
@@ -0,0 +1,67 @@
...
+- New releases require contract artifacts to be built, pushed to GCP, and added to `standard.go`. This means that
+  every new contract release requires a corresponding code change to OP Deployer.
+- Because the artifacts each locator points to are on GCP, they need to be validated client-side before being used.
+  This adds complexity, overhead, and potential for security vulnerabilities.
+- It's much less clear which commit a given set of artifacts belong to. Artifacts are built using a
+  content-addressable hash of their source code. A commit that doesn't change the source code won't change the hash,
+  so the commit for a given contract release may not match the commit in the uploaded artifacts. While not a huge
+  issue, this makes determining the provenance of a given set of artifacts quite difficult.
+
+## Proposed Solution
```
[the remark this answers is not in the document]
mslipper: Would prefer not to backport if we don't have to.

protocol/op-deployer-embedded-artifacts.md
mslipper: Yep

PR #305, opened 2025-07-24
feat: add custom gas token

Joxess:
**Description**

This document introduces a new design for enabling Custom Gas Token ("CGT") support on OP Stack. It replaces the previous experimental approach with a minimal, flexible, and future-proof architecture. The goal is to support a wide range of use cases while laying the groundwork for standardization across the ecosystem.


atenalt:
excellent job for this!
BTW, I want to check for existing CGT rollup, do they can upgrade to this in future? We hope the existing CGT can also migrate to this.

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,318 @@
...
+
+| Item | Old Design | New Design |
+| --- | --- | --- |
+| Approach | L1 token is used as gas asset. | The native asset exists in its own and may be convertible into an ERC20. |
+| Native asset is bridgeable? | Yes, through `DepositERC20Transaction` and native withdrawals. | Yes, but not enshrined in core contracts. The chain governor must implement custom logic to enable it, e.g., coupling bridging with the `LiquidityController`. |
+| Token Implementation Flexibility | Restricted to 18 decimals and transfer properties. | There are potentially no limitations, as long as it is coupled adequately with `NativeAssetLiquidity`. |
+| WETH predeploy | Reserved for the wrapped version of the custom gas token, taking metadata from `L1Block` (`SystemConfig`). | Reserved for the wrapped version of the custom gas token. Metadata is taken from `L1Block` (`LiquidityController`). |
+| Native Asset Supply | Held in `OptimismPortal`, originating from the original L1 token contract. | Minted at genesis via `NativeAssetLiquidity`. Manageable under any rules, through the controller, which may depend on an existing ERC20. |
```
pahor167, line 315: For chains that use existing CGT will it be expected to move existing L1 token supply from OptimismPortal to L1CGTStandardBridge or will L1CGTStandardBridge rather manage the supply in OptimismPortal ?

Joxess:
> excellent job for this! BTW, I want to check for existing CGT rollup, can they upgrade to this in future? We hope the existing CGT can also migrate to this.

Yes, one of the goals is to allow migrations!

protocol/custom-gas-token.md
Joxess: Yes, chains need to migrate the liquidity held in `OptimismPortal` into their implemented custom bridge.

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,318 @@
...
+
+As an alternative to the proposed architecture, both predeploys might be merged into a single one. 
+
+## Risks & Uncertainties
+
+- Any minters granteed to interact with the `LiquidityController` require proper audits to ensure native asset logic can’t be broken in production, as the recovery would be costly.
+    - There are two potential paths to minimize the impact: adding rate limits, or minters to withdraw the needed supply and virtually “close” or don’t need to use the `CustomAssetLiquidity` anymore.
+- There is an open discussion on how to support tokens that have other than 18 decimals. This is a concern for the minter that is placed on top of the `LiquidityController`.
+    - One possible solution might be adding `decimals()` in the `LiquidityController`, or fully handling it through the minters.
+- Existing OP Stack chains using old designs would need to upgrade to the new version, the solution of which is actively being architected.
```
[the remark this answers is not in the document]
Joxess: We want to offer some out-of-the-box features, and at the same time, allow enough flexibility for the liquidity management criteria that the chain wants to implement.

tynes:
A poc implementation can be found here: https://github.com/defi-wonderland/optimism/pull/478

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,330 @@
...
+
+The `WETH` (now `WNA`) predeploy would stay the same as their old version, to preserve backward compatibility with existing chains.
+
+To align with the new design, the metadata, specifically the `name()` and `symbol()` functions, will now be sourced from the `LiquidityController` instead of the `SystemConfig` contract used in the previous design. These functions will return the name and symbol of the wrapped native asset, prefixed with "W" and "Wrapped".
+
+### Gas-related Components
+
+The CGT sets `Types.WithdrawalNetwork.L2` as chain type in each `FeeVaults` contracts on the constructor, without requiring contract changes. Funds are going to be sent to a desired L2 address.
+
+To accurately charge the fees for execution and data availability, chain governors may now adjust a `minBaseFee` and `operatorFee`. In the future, other parameters or mechanisms might be included to properly account for the cost for each activity, such as adding a price oracle that converts data cost expressed in ETH into the ASSET.
```
mds1, line 178: Can you expand more on how these two parameters should be used to charge for fees? When would you adjust each and how do you know what to set it to?

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,330 @@
...
+## Impact on Developer Experience
+
+The developers will present several changes with respect to an OP Stack chain that uses ETH as native asset:
+
+- Since native asset management is decoupled from core components, obtaining native assets might vary chain-by-chain. That means that bridging, conversions, and features utilized would be different. It’s suggested to create some library or templates that cover basic uses cases (e.g., L2-native, bridging from L1) to standardize most implementations.
+    - Chain Governors are encouraged to audit their implementations.
+    - Developers would need to look at the chain’s docs to understand how to obtain native assets.
+- `OptimismPortal` and `L2ToL1MessagePasser` don’t contain the API anymore to bridge native assets.
+- Native ETH bridging is disabled, now enabled through L1-WETH, minting an `OptimismMintableERC20` representation.
+- Supersim needs to be aware when CGT is being used.
```
mds1, line 282: I haven't seen this aspect of the design mentioned anywhere else, but it sounds like the L1CGT bridge is now the entry point for all ETH/native asset transfers? This is a breaking change we'll need to make sure we communicate, possibly requiring viem changes also, cc @fainashalts @ZakAyesh

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,330 @@
...
+EVM-equivalence means the native asset doesn’t have any custom feature other than the described in the EVM specs. It could be possible for the native asset to be modified to allow features such as blacklisting, whitelisting, staking, and voting at the cost of losing EVM equivalence. This would likely affect the current proof systems and fundamental APIs such as `depositTransaction`. This alternative would case similar issues on software maintenance and existing toolings, as stated above.
+
+**Merge `NativeAssetLiquidity` and `LiquidityController`.**
+
+As an alternative to the proposed architecture, both predeploys might be merged into a single one. 
+
+## Risks & Uncertainties
+
+- Any minters granted to interact with the `LiquidityController` require proper audits to ensure native asset logic can’t be broken in production, as the recovery would be costly.
+    - There are two potential paths to minimize the impact: adding rate limits, and burn the unneeded supply during chain's deployment.
```
mds1, line 305: I don't understand the burn use case, and why the `burn` code comment in this doc says "ideally to be called only once". If the idea is to call burn at genesis only, couldn't you just modify the genesis state to not need a burn?

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,330 @@
...
+- There is an open discussion on how to support tokens that have other than 18 decimals. This is a concern for the minter that is placed on top of the `LiquidityController`.
+    - One possible solution might be adding `decimals()` in the `LiquidityController`, or fully handling it through the minters.
```
mds1, line 307: Is solving this a blocker to finalizing/merging this design doc? I think it should be as it feels like a fundamental thing and we want to be certain it doesn't require nontrivial design changes

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,330 @@
...
+
+## Purpose
+
+Enable OP Stack chains to use an asset other than ETH as their native fee currency by introducing a minimal, flexible, and standardizable Custom Gas Token (CGT) architecture. 
+
+## Summary
+
+The proposed Custom Gas Token upgrade lets any OP Stack chain introduce its native asset as the gas currency with almost no core-code intrusion: a single `isCustomGasToken()` flag turns off ETH transfer flows in all bridging methods, while three new pre-deploys, `NativeAssetLiquidity`, a contract with pre-minted assets, `LiquidityController`, an owner-governed mint/burn router, and `L2CGTBridge`, an reserved address that contains a contract with the API calls, hand supply control to chain governors or authorized “minter” contracts that can plug in anything from ERC-20 converters to third-party bridgind models or emission schedules. Wrapped-asset compatibility is preserved, and the entire system launches at genesis by funding the vault and letting the governor release an initial working balance. Overall, the design keeps the OP Stack lean, token-agnostic, and future-proof while unlocking custom economics and flexibility to make the native asset expressive in its manner.
```
mds1, line 16: What do you think of merging LiquidityController and NativeAssetLiquidity? We have a lot of contracts, and these two are fairly small and logically related, so combining them might be simpler

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,330 @@
...
+
+## Problem Statement + Context
+
+The prior CGT design anchored the gas token to an L1 ERC‑20, hard‑coding bridge paths, and cluttering core contracts with token metadata. These restrictions make it hard to handle the following scenarios properly:
+
+- Integrates with a token that doesn’t live on L1.
+- Launch chains whose native asset does not yet exist or have an ERC20 representation somewhere.
```
mds1, line 23: What restrictions are there on the ERC20? Lots of edge cases to consider https://github.com/d-xo/weird-erc20

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,330 @@
...
+
+### Gas-related Components
+
+The CGT sets `Types.WithdrawalNetwork.L2` as chain type in each `FeeVaults` contracts on the constructor, without requiring contract changes. Funds are going to be sent to a desired L2 address.
+
+To accurately charge the fees for execution and data availability, chain governors may now adjust a `minBaseFee` and `operatorFee`. In the future, other parameters or mechanisms might be included to properly account for the cost for each activity, such as adding a price oracle that converts data cost expressed in ETH into the ASSET.
+
+### Chain Deployment
+
+During the deployment in genesis, the CGT version comes up with the `isCustomGasToken()` flag activated, adding the new `NativeAssetLiquidity`, `LiquidityController` and `L2CGTBridge` predeploy contracts.
```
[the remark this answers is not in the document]
Joxess: `L1CGTBridge` and `L2CGTBridge` are the entry contracts for depositing and withdrawing native assets. They are provided as reference points, meaning we do not define any mandatory implementation for CGT Chains. These are *external* bridge contracts, coupled with `NativeAssetLiquidity` and `LiquidityController`, to be used as the chain operator sees fit. We included them to demonstrate how the L1-L2 bridging flow could work in alignment with the old implementation, though chain operators may choose a completely different approach.

That said, we introduced `L2CGTBridge` as a reserved address for a proxy predeploy, intended to act as a minter for `LiquidityController` by default. However, it could have any custom implementation. We are still discussing whether this predeploy is sufficiently useful or if another approach would be preferable.

protocol/custom-gas-token.md
Joxess: Until now, when the chain uses ETH as native currency, it uses the typical `Standardbridge` / `CrossDomainMessenger` / `OptimismPortal` / `L2ToL1MessagePasser` flows, this is untouched.

When `isCustomGasToken` is marked as true, it disables all the bridging flow for ETH (`msg.value`) above, and then it could be replaced by a `CGTBridge` as we described as a way to bridge the native asset.

protocol/custom-gas-token.md
Joxess: We talked with @tynes about this, the reason is that it will simplify the consensus config for all the chains

tynes:
Something that we need to be explicit about is how the deployment of these contracts are handled. Per the current implementation (https://github.com/defi-wonderland/optimism/pull/478) we will only deploy the contracts if it is a custom gas token chain and we need to initialize them with some chain specific config. This means that the op-node rollup config needs to have the following information: `isCGT: bool`, `tokenName: string`, `tokenSymbol: string`

protocol/custom-gas-token.md
Joxess: Burn function removed 2a7a99c, we agreed on just using the `L2ToL1MessagePasser` `burn` function if desired, so this will not be a feature of this design.

protocol/custom-gas-token.md
tynes: There are no restrictions on the type of ERC20 because the token bridge is not coupled to the design. A bridge can be built that supports any sort of token, even non standard ones. When we build a token bridge, we will likely make it work with tokens that have 18 decimals and are a standard ERC20 token

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,459 @@
...
+As a consequence, native asset mints and burns are decoupled from system transactions (deposits) in normal operations, and those are moved into the application layer with the new introduced predeploys.
+
+**`NativeAssetLiquidity` and `LiquidityController` Predeploys**
+
+A very large amount of native assets is pre-minted and stored in `NativeAssetLiquidity`, which is managed through the `LiquidityController`, which has the rights to withdraw and deposit to this contract.
+
+A code example of both contracts would look like this:
+
+```solidity
+contract NativeAssetLiquidity {
```
drinkcoffee, line 58: If the only contract that can call this contract is the LiquidityController, does it make sense for the contracts to be separate?

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,459 @@
...
+    Note over LiquidityController,recipient: [Flow B] Convert ERC20 → native asset
+    Address B->>ERC20Converter: getNativeAsset(ERC20, amount, recipient)
+    ERC20Converter->>ERC20Converter: lock / burn ERC20
+    ERC20Converter->>LiquidityController: mint(amount, recipient)
+    LiquidityController->>Liquidity: deposit(amount, recipient)
+    Liquidity-->>recipient: native asset
+
+    %% --- Flow C: burn native asset and unlock ERC20 ---
+    Note over LiquidityController,recipient: [Flow C] Convert native asset → ERC20
+    Address B->>ERC20Converter: getERC20(recipient)  %% msg.value = native asset
```
drinkcoffee, line 148: Maybe write this as
`Address B->>ERC20Converter: getERC20{msg.value = amount}(recipient)  `

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,459 @@
...
+- Introduce an ERC20 in L2 that contains utility functions (governance, DeFi); meanwhile, their native asset version serves as the medium to pay for gas.
+- Do not depend on an existing asset, emulating how some L1 currently works.
+
+The design allows for the inclusion of any desired features, such as rate limits, emission schedules, and a max. cap the supply, coupled with any bridge mechanism, etc. The architecture also doesn’t restrict the use of any token framework (e.g., OFT, xERC20, SuperchainERC20, etc.) to be coupled with the native asset eventually.
+
+Existing CGT chains using the old design can perform a hard fork to set such contracts and seed the native asset liquidity.
+
+### Wrapped Native Asset
+
+The `WETH` (now `WNA`) predeploy would stay the same as its old version, to preserve backward compatibility with existing chains.
```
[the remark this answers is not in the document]
tynes: The legacy version can be found here: https://github.com/ethereum-optimism/specs/blob/5887eb2f74821b4125c2f47523d6d54379fd46d8/specs/experimental/custom-gas-token.md

It would be good to include this hyperlink in the design doc itself cc @Joxess

protocol/custom-gas-token.md
tynes: The general idea of separating them was to reduce the amount of logic in the liquidity contract to the most bare minimum given if there is a bug in it, its game over because an extreme amount of native asset could be minted

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,459 @@
...
+
+## Proposed Solution
+
+### Design Principles
+
+The presented design has been thought out based on the following principles:
+
+- Stick with a minimal set of changes.
+- Remain token-agnostic and completely flexible for any use cases.
+- Preserve OP Stack [foundational properties](https://specs.optimism.io/background.html#foundations), which also make Stage 2 requirements achievable.
```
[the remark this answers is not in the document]
tynes: Here is an article describing the stages: https://medium.com/l2beat/introducing-stages-a-framework-to-evaluate-rollups-maturity-d290bb22befe

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,459 @@
...
+
+**`isCustomGasToken()` in L1 and L2**
+
+The `isCustomGasToken()` view function becomes the unique one that brings awareness to the system about whether or not the CGT mode is being used. It is a boolean flag that other functions in other contracts use to check if they can receive ETH values or not.
+
+In L1 placed in `OptimismPortal`, it will block `depositTransaction` calls that contain `msg.value`. The same will occur in their pair functions `sendMessage` in `L1CrossDomainMessenger` and ETH bridging in `L1StandardBridge`. The `SystemConfig` reads this setting from the `OptimismPortal`, since it is a relevant configuration parameter for the chain.
+
+In L2 placed in `L1Block`, it will block `initiateWithdrawal` call, which contains `msg.value` in `L2ToL1MessagePasser`. The same will happen in `sendMessage` in `L2CrossDomainMessenger` and ETH bridging methods in `L2StandardBridge` and `FeeVaults`.
+
+As a consequence, native asset mints and burns are decoupled from system transactions (deposits) in normal operations, and those are moved into the application layer with the new introduced predeploys.
```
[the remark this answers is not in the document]
Joxess: thank you!

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,459 @@
...
+    LiquidityController->>LiquidityController: isAuthorizedMinter()
+    LiquidityController->>Liquidity: withdraw(amount, recipient)
+    Liquidity-->>recipient: native asset
+
+    %% --- Flow B: lock ERC20 and mint native asset ---
+    Note over LiquidityController,recipient: [Flow B] Convert ERC20 → native asset
+    Address B->>ERC20Converter: getNativeAsset(ERC20, amount, recipient)
+    ERC20Converter->>ERC20Converter: lock / burn ERC20
+    ERC20Converter->>LiquidityController: mint(amount, recipient)
+    LiquidityController->>Liquidity: deposit(amount, recipient)
```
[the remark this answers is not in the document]
Joxess: Fixed! The issue was in the diagram

protocol/custom-gas-token.md
Joxess: Thanks!

protocol/custom-gas-token.md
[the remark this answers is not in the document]
Joxess: added! Thank you!

protocol/custom-gas-token.md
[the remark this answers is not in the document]
drinkcoffee: Thank you - very helpful.

protocol/custom-gas-token.md
drinkcoffee: I agree that the fewer lines of code, the lower the probability of error. However, having separate contracts means every mint and burn takes more gas. Also, given the LiquidityController completely controls the NativeAssetLiquidity contract, it will come down to the security of the LiquidityController contract.

protocol/custom-gas-token.md
drinkcoffee: There are examples of chains using tokens that have decimals other than 18 as their gas token: for instance Circle's newly announced chain that will USDC for gas (6 decimals):
https://www.circle.com/blog/introducing-arc-an-open-layer-1-blockchain-purpose-built-for-stablecoin-finance

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,459 @@
...
+    actor U as User
+    participant SB as L2StandardBridge
+    participant XDM as L2CrossDomainMessenger
+    participant OP as L2ToL1MessagePasser
+    participant SC as L1Block
+    
+    U->>SB: withdraw / bridgeETH / receive(...)
+    SB->>XDM: sendMessage(...)
+    XDM->>OP: initiateWithdrawal(...)
+    OP->>SC: isCustomGasToken()
```
drinkcoffee, line 375: Add "returns true" at the end of isCustomGasToken()

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,459 @@
...
+    OP->>OP: revert
+
+    U->>XDM: sendMessage(...), value > 0
+    XDM->>OP: initiateWithdrawal(...)
+    OP->>SC: isCustomGasToken()
```
drinkcoffee, line 380: Add "returns true" at the end of isCustomGasToken()

protocol/custom-gas-token.md
Joxess: Fixed, thank you!

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,459 @@
...
+    OP->>OP: revert
+
+    U->>OP: initiateWithdrawal(...), value > 0
+    OP->>SC: isCustomGasToken()
```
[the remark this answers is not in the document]
Joxess: Fixed, thank you!

protocol/custom-gas-token.md
Joxess: Fixed, thank you!

protocol/custom-gas-token.md
```diff
@@ -0,0 +1,462 @@
...
+    - One possible solution might be adding `decimals()` in the `LiquidityController`, or fully handling it through the minters.
+- Existing OP Stack chains using old designs would need to upgrade to the new version, the solution of which is actively being architected.
+- The design is proposed to facilitate its standardization, but the discussion regarding enabling interoperability is still open.
+    - For example, `ETHLiquidity` and `NativeAssetLiquidity` might share the same address, and the ERC20 version of the native asset might be created with a deterministic method to enable interoperability through a `SuperchainERC20` version.
+    - A new `SuperchainWETH` version might be introduced to enable interoperability for the ETH transfer.
+- `L1FeeVault` might not receive the amount in value to pay for data availability, since the mismatch is corrected via `minBaseFee` and `operatorFees`. The introduction of parameters or oracles is actively being discussed.
+- Native asset custom features are in early research phase, but the design is minimal enough not to block a future upgrade of this kind.
+
```
alcueca, line 312: These are the most significant risks from the CGT Threat Model:

```suggestion
### NativeAssetLiquidity Drained
A bug or a compromise on `NativeAssetLiquidity` or the `LiquidityController` could lead to a massive asset loss either in the canonical bridge or third party bridges. Given the low complexity of `LiquidityController`, the most likely vector would be an access control bug or attack.

This is already mitigated by the airgap in each bridge, and by the existing measures to avoid bugs and L2 predeploy compromises. No new mitigations are required for this risk.

### CGTs unduly minted or burned
Errors or compromises affecting chain operators could cause CGTs to be minted or burned not following the protocol. The most likely would be:
- L1 or L2 misconfigurations of `isCustomGAsToken` or `LiquidityController`.
- AccessControl, Replayability, ERC20 or CGT bridge smart contract bugs.
- `NativeAssetLiquidity` balance overflows
- Misconfiguration of `L1ToL2MessagePasser`

These risks would be mostly borne by chain operators, and the developer documentation should reflect these concerns so that they are not overlooked. Additionally, we will need automated checks to verify that the `isCustomGAsToken` always matches in the L1 and L2 parts of a chain configuration.

### Very Inaccurate Fees
L1 misconfigurations could lead to very inaccurate fees being charged. This will need to be highlighted in the developer documentation for chain operators to be aware.

### Chain operators see wrong amounts of wrapped asset
The current version of CGT only accepts 18-decimal CGTs, or the wrapper contract will malfunction. This will need to be highlighted in the documentation.
```

tynes (approved):

protocol/custom-gas-token.md
tynes: You are definitely right that it takes up more gas, its a marginal amount of gas, but this sort of conversation can only be resolved with a shared optimization target

tynes:
Merging this given its had sufficient review

PR #308, opened 2025-07-28
update minimum base fee design doc

wlawt:
**Description**

Updates the minimum base fee design document to include a note about decoding `extraData` and its best-effort for all blocks in the chain's history. For more context see: https://github.com/wlawt/optimism/pull/10#discussion_r2232020344

It also updates the # of bytes for `eip1559Params` back to 8 bytes to maintain cleaner backwards compatibility. For more context see: https://github.com/wlawt/optimism/pull/10#discussion_r2229144428

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A

### DIFF
--- a/protocol/minimum-base-fee.md
+++ b/protocol/minimum-base-fee.md
@@ -47,7 +47,7 @@ Additionally,
 - there is no additional data beyond these 10 bytes
 
 Note that `extraData` has a maximum capacity of 32 bytes (to fit in the L1 beacon-chain `extraData` data-type) and its
-format may be modified/extended by future upgrades.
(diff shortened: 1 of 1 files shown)


protocol/minimum-base-fee.md
```diff
@@ -64,7 +64,7 @@ PayloadAttributesV3: {
     transactions: array of DATA
     noTxPool: bool
     gasLimit: QUANTITY or null
-    eip1559Params: DATA (9 bytes) or null
+    eip1559Params: DATA (8 bytes) or null
```
niran, line 67: We're missing `minBaseFeeLog2` in the payload attributes, which also means we're probably not handling it properly in `ForkChoiceUpdated`. That means op-node can't instruct op-geth to build a payload with the current value of the setting.

PR #310, opened 2025-07-29
design doc for the ocpm superchain upgrade fix

AmadiMichael:
design doc for the ocpm superchain upgrade fix

### DIFF
--- a/protocol/opcm-superchain-upgrade-fix.md
+++ b/protocol/opcm-superchain-upgrade-fix.md
@@ -1,0 +1,103 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [[Project Name]: Design Doc](#project-name-design-doc)
+  - [Summary](#summary)
+  - [Problem Statement + Context](#problem-statement--context)
+  - [Proposed Solution](#proposed-solution)
+  - [Failure Mode Analysis](#failure-mode-analysis)
+  - [Alternatives Considered](#alternatives-considered)
+  - [Risks & Uncertainties](#risks--uncertainties)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# [Project Name]: Design Doc
+
+|                    |                                                    |
+| ------------------ | -------------------------------------------------- |
+| Author             | Michael Amadi                                      |
+| Created at         | _2025-07-29_                                       |
+| Initial Reviewers  | Matt Solomon, John Mardlin                 |
+| Need Approval From | _To be assigned_                                    |
+| Status             | _Draft_ |
+
+## Summary
+
+Currently, the OPCM's `upgrade()` function optionally upgrades the superchainConfigProxy based on the following checks:
+
+For [op-contracts/v2.0.0](https://github.com/ethereum-optimism/optimism/blob/8d0dd96e494b2ba154587877351e87788336a4ec/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L477)
+```solidity
+        // If the SuperchainConfig is not already upgraded, upgrade it.
+        if (superchainProxyAdmin.getProxyImplementation(address(superchainConfig)) != impls.superchainConfigImpl) {
+            // Attempt to upgrade. If the ProxyAdmin is not the SuperchainConfig's admin, this will revert.
+            upgradeTo(superchainProxyAdmin, address(superchainConfig), impls.superchainConfigImpl);
+        }
+```
+
+and [op-contracts/v4.0.0](https://github.com/ethereum-optimism/optimism/blob/54c19f6acb7a6d3505f884bae601733d3d54a3a6/packages/contracts-bedrock/src/L1/OPContractsManager.sol#L615)
+```solidity
+        // If the SuperchainConfig is not already upgraded, upgrade it. NOTE that this type of
+        // upgrade means that chains can ONLY be upgraded via this OPCM contract if they use the
+        // same SuperchainConfig contract. We will assert this later.
+        if (_superchainProxyAdmin.getProxyImplementation(address(_superchainConfig)) != impls.superchainConfigImpl) {
+            // Attempt to upgrade. If the ProxyAdmin is not the SuperchainConfig's admin, this will revert.
+            upgradeToAndCall(
+                _superchainProxyAdmin,
+                address(_superchainConfig),
+                impls.superchainConfigImpl,
+                abi.encodeCall(ISuperchainConfig.upgrade, ())
+            );
+```
+
+The check basically says that, if the implementation of the superchainConfig is not the same as the implementation the OPCM has stored, then upgrade it.
+
+To upgrade the superchainConfig, the superchainProxyAdmin will need to call the OPCM's `upgrade()` function and optionally pass in values to upgrade any other chain's contracts as long as they're that chain's proxyAdmin.
+
+## Problem Statement + Context
+
+This works for the most part but has a flaw. Assume the following:
+- There are 2 chains in the Superchain, ChainA and ChainB
+- ChainA's proxyAdmin is also the SuperchainConfig's ProxyAdmin
+- The SuperchainConfig's implementation is Impl0
+- The OPCM has stored Impl1 as the implementation to upgrade to
+
+If ChainA's proxyAdmin (also the SuperchainConfig's ProxyAdmin) calls the OPCM's `upgrade()` function, the check above (if (superchainProxyAdmin.getProxyImplementation(address(superchainConfig)) != impls.superchainConfigImpl)) will be true and it will upgrade the SuperchainConfig to Impl1 and also upgrade ChainA's L1 contracts.
+
+If ChainB's proxyAdmin calls the OPCM's `upgrade()` function, the check above (if (superchainProxyAdmin.getProxyImplementation(address(superchainConfig)) != impls.superchainConfigImpl)) will be false and so it will not enter the if block (if it did it will revert since ChainB's proxyAdmin is not the SuperchainConfig's ProxyAdmin). So it will skip this and go ahead to upgrade ChainB's L1 contracts.
+
+Now the a problem arises here:
+- The SuperchainConfig's implementation is Impl1
+- A new OPCM is created with an expected implementation of Impl2
+- The OPCM's `upgrade()` function is called and the SuperchainConfig's implementation is upgraded to Impl2
+- A new chain, ChainC comes in or is behind and has to use the old OPCM first
+
+When it calls the OPCM's `upgrade()` function, the check above (if (superchainProxyAdmin.getProxyImplementation(address(superchainConfig)) != impls.superchainConfigImpl)) will be true and it will attempt to upgrade the SuperchainConfig which will fail since it's ProxyAdmin will not be the SuperchainConfig's ProxyAdmin.
+
+This means all upgrades from that OPCM is not possible any longer.
+
+## Proposed Solution
+
+A proposed solution for this is to change the check:
+- Create a variable `bool isUpgraded` in the OPCMUpgrader contract.
+- When `OPCM.upgrade()` is called, it delegates call to `OPCMUpgrader.upgrade()` as usual.
+- `OPCMUpgrader.upgrade()` calls back into itself (to be able to access it's storage) and checks if the superchainConfig is already upgraded by checking the `isUpgraded` variable.
+- If it is not upgraded, it calls back into itself once more to set the `isUpgraded` variable to true and then upgrade the superchainConfig.
+- If it is already upgraded, it will skip the upgrade and continue execution.
+
+While at it, it has also been decided to add support for different superchainConfigs i.e different Superchains. We can easily do this by replacing the `isUpgraded` variable with a mapping `mapping(ISuperchainConfig superchainConfig => bool isUpgraded)`. This way we can check which superchainConfig is being upgraded and set the `isUpgraded` variable to true for that superchainConfig while also preventing the same superchainConfig from being upgraded again.
+
+## Failure Mode Analysis
+
+- **Protecting the `setSuperchainUpgraded(ISuperchainConfig superchainConfig)` function from being called directly by a malicious actor:**
+    The way the OPCM sets the `isUpgraded` variable is by calling  `OPCMUpgrader.setSuperchainUpgraded(ISuperchainConfig superchainConfig)` function. This is safe and does not need checks because if a malicious actor tries doing this, it will fail when trying to actually upgrade the superchainConfig. However, A malicious actor could call the `setSuperchainUpgraded(ISuperchainConfig superchainConfig)` function directly (not via an OPCM upgrade call) to set the `isUpgraded` variable to true for any superchainConfig while not being that Superchain's ProxyAdmin. The obvious way to do this is to check that the msg.sender is the Superchain's ProxyAdmin. This is possible for future OPCM versions because the proxyAdmin variable is accessible via the SuperchainConfig contract. For v2.0.0 and v4.0.0, this is not possible.
+
+## Alternatives Considered
+
+- Define the mapping `mapping(ISuperchainConfig superchainConfig => bool isUpgraded) public isSuperchainUpgraded` and `setSuperchainUpgraded(ISuperchainConfig superchainConfig)` in the OPCM contract:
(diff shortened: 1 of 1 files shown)


protocol/opcm-superchain-upgrade-fix.md
```diff
@@ -0,0 +1,103 @@
...
+                _superchainProxyAdmin,
+                address(_superchainConfig),
+                impls.superchainConfigImpl,
+                abi.encodeCall(ISuperchainConfig.upgrade, ())
+            );
+```
+
+The check basically says that, if the implementation of the superchainConfig is not the same as the implementation the OPCM has stored, then upgrade it.
+
+To upgrade the superchainConfig, the superchainProxyAdmin will need to call the OPCM's `upgrade()` function and optionally pass in values to upgrade any other chain's contracts as long as they're that chain's proxyAdmin.
```
maurelian, line 55:
```suggestion
To upgrade the superchainConfig, the superchainProxyAdminOwner will need to call the OPCM's `upgrade()` function and optionally pass in values to upgrade any other chain's contracts as long as they're that chain's proxyAdmin.
```

protocol/opcm-superchain-upgrade-fix.md
```diff
@@ -0,0 +1,103 @@
...
+- The SuperchainConfig's implementation is Impl1
+- A new OPCM is created with an expected implementation of Impl2
+- The OPCM's `upgrade()` function is called and the SuperchainConfig's implementation is upgraded to Impl2
+- A new chain, ChainC comes in or is behind and has to use the old OPCM first
+
+When it calls the OPCM's `upgrade()` function, the check above (if (superchainProxyAdmin.getProxyImplementation(address(superchainConfig)) != impls.superchainConfigImpl)) will be true and it will attempt to upgrade the SuperchainConfig which will fail since it's ProxyAdmin will not be the SuperchainConfig's ProxyAdmin.
+
+This means all upgrades from that OPCM is not possible any longer.
+
+## Proposed Solution
```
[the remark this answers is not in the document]
AmadiMichael: can you re-review?

protocol/opcm-superchain-upgrade-fix.md
```diff
@@ -0,0 +1,131 @@
...
+| Multiple Chains in Single Upgrade | Multiple chains are being upgraded in a single call to upgrade() | All chains being upgraded must belong to the same superchain. The upgrade function should verify this by checking that all chains have the same SuperchainConfig, otherwise it should revert. |
+
+## Failure Mode Analysis
+
+- **Protecting the `setSuperchainUpgraded(ISuperchainConfig superchainConfig)` function from being called directly by a malicious actor:**
+    - Explainer: The way the OPCM sets the `isSuperchainUpgraded` variable is by calling  `OPCMUpgrader.setSuperchainUpgraded(ISuperchainConfig superchainConfig)` function. This is safe and does not need checks because if a malicious actor tries doing this, it will fail when trying to actually upgrade the superchainConfig. However, A malicious actor could call the `setSuperchainUpgraded(ISuperchainConfig superchainConfig)` function directly (not via an OPCM upgrade call) to set the `isSuperchainUpgraded` variable to true for any superchainConfig while not being that Superchain's ProxyAdmin.
```
maurelian, line 117: > This is safe and does not need checks because if a malicious actor tries doing this, it will fail when trying to actually upgrade the superchainConfig

This seems to imply that the upgrade itself happens within `setSuperchainUpgraded()` is that correct?

AmadiMichael: oh nope, i was talking about when calling the `setSuperchainUpgraded` from the upgrade function. it is safe since the upgrade will happen next and if the caller is not the superchainProxyAdmin, the tx will revert

protocol/opcm-superchain-upgrade-fix.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Problem Statement + Context
+
+This works for the most part but has a flaw. Assume the following:
+- There are 2 chains in the Superchain, ChainA and ChainB
+- ChainA's proxyAdmin is also the SuperchainConfig's ProxyAdmin
+- The SuperchainConfig's implementation is Impl0
+- The OPCM has stored Impl1 as the implementation to upgrade to
+
+If ChainA's proxyAdmin (also the SuperchainConfig's ProxyAdmin) calls the OPCM's `upgrade()` function, the check above (if (superchainProxyAdmin.getProxyImplementation(address(superchainConfig)) != impls.superchainConfigImpl)) will be true and it will upgrade the SuperchainConfig to Impl1 and also upgrade ChainA's L1 contracts.
```
mds1, line 66: We say ProxyAdmin a lot in this section, but is it correct that really it's the ProxyAdmin Owner that matter?

protocol/opcm-superchain-upgrade-fix.md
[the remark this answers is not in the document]
mds1: Towards @maurelian's point, can we move the "Requirements and Expected behaviour" above, so it comes before "Proposed Solution"?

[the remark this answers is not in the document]
mds1: Actually, I see now that the "Requirements and Expected behaviour" is really talking about the Requirements and Expected behaviour needed to successfully implement the proposed solution. I believe what @maurelian is asking for (which I agree with) is a section that comes before "Proposed Solution" to explain what customer and developer requirements are, i.e. for various situations what is the expected behavior when calling `upgrade`? In slack @maurelian gave some scenarios, here are a few to start with:

| Scenario Number | PAO == superchainPAO? | Target version is latest? | SuperchainConfig already on latest? | Expected Output |
|---|-----------------------|---------------------------|------------------------------------|-----------------|
| 1 | Yes | Yes | Yes | TODO |
| 2 | Yes | Yes | No  | TODO |
| 3 | No  | Yes | Yes | TODO |
| 4 | No  | Yes | No  | TODO |
| 5 | Yes | No  | Yes | TODO |
| 6 | Yes | No  | No  | TODO |
| 7 | No  | No  | Yes | TODO |
| 8 | No  | No  | No  | TODO |


What should happen in each of those scenarios? It would be great to include a table like this will all the TODOs populated. 

Can you think through any other missing scenarios (i.e. headers), and add the completed version to a new "Customer Requirements and Expected Behavior" section before "Proposed Solution"? I think we can also delete the "Requirements and Expected behaviour" header and make that part of the "Proposed Solution" header.

I just really want to make sure we are clear on the "Customer Requirements and Expected Behavior" before agreeing on a solution, and we'll want to get approval from Product on this design doc as well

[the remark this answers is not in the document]
mds1: Alternatively, makes for a nice flowchart
```mermaid
graph TD
    A{PAO == superchainPAO?}
    A -- Yes --> B{Target version is latest?}
    A -- No  --> E{Target version is latest?}

    B -- Yes --> C{SuperchainConfig already upgraded?}
    B -- No  --> D{SuperchainConfig already upgraded?}

    E -- Yes --> F{SuperchainConfig already upgraded?}
    E -- No  --> G{SuperchainConfig already upgraded?}

    C -- Yes --> S1["TODO"]
    C -- No  --> S2["TODO"]

    D -- Yes --> S5["TODO"]
    D -- No  --> S6["TODO"]

    F -- Yes --> S3["TODO"]
    F -- No  --> S4["TODO"]

    G -- Yes --> S7["TODO"]
    G -- No  --> S8["TODO"]
```

protocol/opcm-superchain-upgrade-fix.md
```diff
@@ -0,0 +1,131 @@
...
+## Proposed Solution
+
+A proposed solution for this is to change the check:
+- Create a variable `bool isSuperchainUpgraded` in the OPCMUpgrader contract.
+- When `OPCM.upgrade()` is called, it delegates call to `OPCMUpgrader.upgrade()` as usual.
+- `OPCMUpgrader.upgrade()` calls back into itself (to be able to access it's storage) and checks if the superchainConfig is already upgraded by checking the `isSuperchainUpgraded` variable.
+- If it is not upgraded, it calls back into itself once more to set the `isSuperchainUpgraded` variable to true and then upgrade the superchainConfig.
+- If it is already upgraded, it will skip the upgrade and continue execution.
+
+While at it, it is proposed to also add support for different superchainConfigs i.e different Superchains. We can easily do this by replacing the `isSuperchainUpgraded` variable with a mapping `mapping(ISuperchainConfig superchainConfig => bool isSuperchainUpgraded)`. This way we can check which superchainConfig is being upgraded and set the `isSuperchainUpgraded` variable to true for that superchainConfig while also preventing the same superchainConfig from being upgraded again.
```
mds1, line 90: Hmm the control flow here feels a bit confusing and error prone. Is there a solution where we add inputs so the caller can simply specify whether they want to try upgrading the superchainConfig, with the following logic:
- Add `bool shouldUpgradeSuperchainConfig` as an input to the `upgrade` method`
- If `isRc == false`, this input value MUST be true. Revert if this condition is violated
- Otherwise, proceed to do the upgrade as normal and try upgrading superchainConfig when requested

Alternatively, can we just make the `(superchainProxyAdmin.getProxyImplementation(address(superchainConfig)) != impls.superchainConfigImpl))` check smarter to do a greater than/less than semver comparison instead of strict inequality?

protocol/opcm-superchain-upgrade-fix.md
AmadiMichael: fixed

protocol/opcm-superchain-upgrade-fix.md
mds1: Got it, so would it be fair to say there's two solutions here:
- Short term patches for the existing OPCM where we don't want to change the interface
- Longer term solution (i.e. for all future OPCMs) where we are ok with breaking the interface to simplify things

protocol/opcm-superchain-upgrade-fix.md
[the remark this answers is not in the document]
maurelian: The second column here doesn't make sense to me. 
In any situation where the `ChainAPAO != superchainPAO` you will always get a revert if you are trying to upgrade contracts with a different PAO.

The more relevant consideration is `ChainAProxyAdmin == superchainProxyAdmin?`, because that determines whether or not the current interface enables us to access the superchainProxyAdmin.

protocol/opcm-superchain-upgrade-fix.md
```diff
@@ -0,0 +1,148 @@
...
+```solidity
+ISuperchainConfig _superchainConfig = _opChainInputs[0].systemConfig.optimismPortal().superchainConfig();
+for (uint256 i = 0; i < _opChainInputs.length; i++) {
+    if (_opChainInputs[i].systemConfig.optimismPortal().superchainConfig() != _superchainConfig) {
+        revert SuperchainConfigMismatch();
+    }
+}
+```
+
+## Requirements and Expected behaviour
```
maurelian, line 121: We should remove this section since we added a similar one above.

protocol/opcm-superchain-upgrade-fix.md
```diff
@@ -0,0 +1,131 @@
...
+
+| Scenario | Description | Expected Behavior |
+|----------|-------------|------------------|
+| Dedicated SuperchainConfig ProxyAdmin | A SuperchainConfig has its own ProxyAdmin that is not shared with any of it's OP chains | When OPCM.upgrade() is called by this SuperchainProxyAdminOwner, it should check if the SuperchainConfig is already upgraded via a isSuperchainUpgraded mapping. If not upgraded, it should upgrade the SuperchainConfig and set isSuperchainUpgraded for that SuperchainConfig to true. No other contract upgrades would occur since this ProxyAdmin only controls the SuperchainConfig. |
```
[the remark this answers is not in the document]
AmadiMichael: if we call upgrade with the superchainPAO and pass in an empty opChainConfig array

protocol/opcm-superchain-upgrade-fix.md
AmadiMichael: yes exactly

protocol/opcm-superchain-upgrade-fix.md
AmadiMichael: done

protocol/opcm-superchain-upgrade-fix.md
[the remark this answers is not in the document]
AmadiMichael: yeah that makes sense, changed ✅

mds1 (approved):

protocol/opcm-superchain-upgrade-fix.md
```diff
@@ -65,66 +68,45 @@ This works for the most part but has a flaw. Assume the following:
...
-
-## Failure Mode Analysis
+### Proposal 2: SuperchainConfig Upgrade Check Fix
+A proposed solution for this is to change the check to version comparisons. We can hardcode an expected previous and target versions for the SuperchainConfig and compare it to the actual version when performing upgrades. E.g
+- For `The SuperchainConfig has not been upgraded by the OPCM being called` in `OPCM.upgradeSuperchainConfig()`, we check if the SuperchainConfig's actual version is less than the target version. If it is, we can upgrade the SuperchainConfig. Otherwise we revert. We do this check to ensure that SuperchainConfigs are not accidentally downgraded.
+- For `ChainA's SuperchainConfig has been upgraded by the OPCM being called` in `OPCM.upgrade()`, we check if the SuperchainConfig's actual version is greater than or equal to the expected target version. If it is, we can proceed to upgrade the OP Chains L1 contracts. Otherwise we revert. We do not make a strict check here so that OP Chains that are more than 2 SuperchainConfig versions behind can still be upgraded.
```
maurelian, line 102: Comment only, I thought about if just making it a strict check might be better (because simpler), ie.

```solidity
require(keccak256(superchainConfig.version()) == keccak256("x.y.z"))
```

but I think the complexity is worthwhile to allow inserting an intermediate version into the upgrade graph.

protocol/opcm-superchain-upgrade-fix.md
```diff
@@ -65,66 +68,45 @@ This works for the most part but has a flaw. Assume the following:
...
-- Define the mapping `mapping(ISuperchainConfig superchainConfig => bool isSuperchainUpgraded) public isSuperchainUpgraded` and `setSuperchainUpgraded(ISuperchainConfig superchainConfig)` in the OPCM contract:
-    This was considered but was not implementable because OPCM components (since they are created before the OPCM) do not have any knowledge of what the OPCM address is in order to call the `isSuperchainUpgraded(ISuperchainConfig superchainConfig)` and `setSuperchainUpgraded(ISuperchainConfig superchainConfig)` functions.
 
-## Risks & Uncertainties
+## Failure Mode Analysis
 
-<!-- An overview of what could go wrong.
-Also any open questions that need more work to resolve. -->
+- **A chain might be upgraded when it's superchainConfig is not upgraded:**
+    - Mitigation: We can implement a simple loop to assert that all chains in the array are of the same superchain and have tests for this.
```
maurelian, line 112: Above it says:

> Note this means that it is possible for OPChains of different superchains to be upgraded in one call to `OPCM.upgrade()` as long as they share the same proxyAdminOwner. This means that we must check that each OPChain's superchainConfig is at the correct version before proceeding to upgrade it, otherwise we revert the transaction.

Which way should we go, allow this situation or not?

AmadiMichael: we want to allow.
> Note this means that it is possible for OPChains of different superchains to be upgraded in one call to OPCM.upgrade() as long as they share the same proxyAdminOwner. This means that we must check that each OPChain's superchainConfig is at the correct version before proceeding to upgrade it, otherwise we revert the transaction.

I have fixed the FMA

protocol/opcm-superchain-upgrade-fix.md
```diff
@@ -65,66 +68,45 @@ This works for the most part but has a flaw. Assume the following:
...
-    - Mitigation: We can implement a simple loop to assert that all chains in the array are of the same superchain and have tests for this.
 
-## Alternatives Considered
+### Proposal 3: Support for different superchainConfigs i.e different Superchains
+While at it, it is proposed to also add support for different superchainConfigs i.e different Superchains. We can easily achieve this by getting the superchainConfig of each OPChain from its optimismPortal (gotten from the systemConfig in the OPChainConfig struct) rather than using the hardcoded SuperchainConfig address in the OPCM. Note this means that it is possible for OPChains of different superchains to be upgraded in one call to `OPCM.upgrade()` as long as they share the same proxyAdminOwner. This means that we must check that each OPChain's superchainConfig is at the correct version before proceeding to upgrade it, otherwise we revert the transaction.
```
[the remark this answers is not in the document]
AmadiMichael: Yes this is in the plans

protocol/opcm-superchain-upgrade-fix.md
AmadiMichael: yes exactly

maurelian (approved):

PR #317, opened 2025-08-12
Jovian: Calldata footprint block limit

sebastianst:
A calldata footprint block limit is introduced to mitigate DA spam and prevent priority fee auctions. By tracking
calldata footprint alongside gas, this approach adjusts the block gas limit to account for high calldata usage without
altering individual transaction gas mechanics. Preliminary analysis shows minimal impact on most blocks on production
networks like Base or OP Mainnet.


protocol/da-footprint-block-limit.md
```diff
@@ -0,0 +1,299 @@
...
+this is a good place to that information. -->
+
+The current L1 cost model doesn't account for *limited L1 data availability (DA) throughput*. 
+This can lead to network congestion for batchers if the throughput on L2s surpasses that of L1 blob space, or if blob
+space is congested for other reasons.
+
+Current L1 throughput is 6 blobs per block at target with a max of 9 blobs per block.
+This is approx. `128 kB * 6 / 12 s = 64 kB/s` or `96 kB/s`, resp. The L1 cost is proportional to the L1 origin’s base
+fee and blob base fee. But Most OP Stack chains have gas targets high enough to accept far more than 128kb of calldata
+per (2s) block. So the calldata floor cost doesn't nearly limit calldata throughput enough to prevent.
```
geoknee, line 38: Trails off here. I think you mean priority fee auctions. Recommend introducing PFA acronym.

protocol/da-footprint-block-limit.md
```diff
@@ -0,0 +1,299 @@
...
+
+The current measure to counter high DA throughput that's being deployed on productio networks is batcher throttling.
```
geoknee, line 40:
```suggestion
The current countermeasure for high DA throughput that's being deployed on production networks is batcher throttling.
```

protocol/da-footprint-block-limit.md
```diff
@@ -0,0 +1,299 @@
...
+  would have been limited to include fewer transactions.
+* *Exc. Gas Usage*: Number and percent of blocks for wich the total DA footprint exceeds the total gas usage. For those blocks, the `block.gas_used`
+  field would equal the total block DA footprint instead of the total transaction gas used.
+* *Avg. Utilization*: Average utilization of DA footprint resource.
+* *Max Utilization*: Maximum utilization of DA footprint resource.
+
+#### OP Mainnet
+
+**[Random Sampling](https://htmlpreview.github.io/?https://github.com/ethereum-optimism/op-analytics/blob/main/notebooks/adhoc/jovian_analysis/notebooks/saved_output_html/op_random_2025-04-29_2025-08-26.html):**
+_Sample Size:_ `51,840`
```
geoknee, line 136:
```suggestion
_Sample Size:_ `51,840` blocks
```
?

protocol/da-footprint-block-limit.md
```diff
@@ -0,0 +1,299 @@
...
+faster), but the block still wouldn't have been limited. Only if the DA footprint also went over the gas limit, would it
+have limited that block's transactions to a smaller DA footprint. For those blocks, we also looked at the
+distribution of the excess DA footprint as an indicator of how much those blocks would have been limited.
+
+The following tables show, for each analyzed chain, the resulting statistics.
+* *Scalar*: DA footprint gas scalar value.
+* *Effective Limit*: The DA usage limit that the given gas scalar would imply (`block_gas_limit / da_footprint_gas_scalar`), in estimated compressed bytes.
+* *Exceeding Limit*: Number and percent of blocks for wich the total DA footprint exceeds the gas limit. Those blocks
+  would have been limited to include fewer transactions.
+* *Exc. Gas Usage*: Number and percent of blocks for wich the total DA footprint exceeds the total gas usage. For those blocks, the `block.gas_used`
```
geoknee, line 128:
```suggestion
* *Exceeding Limit*: Number and percent of blocks for which the total DA footprint exceeds the gas limit. Those blocks
  would have been limited to include fewer transactions.
* *Exc. Gas Usage*: Number and percent of blocks for which the total DA footprint exceeds the total gas usage. For those blocks, the `block.gas_used`
```

protocol/da-footprint-block-limit.md
```diff
@@ -0,0 +1,299 @@
...
+total block DA footprint.
+
+The DA footprint cost is comparable to the calldata floor cost. If it's set to a higher value, say 400, you
+can expect _roughly_ a 1/10 of incompressible calldata to fit into a block compared to only limiting it by
+transaction gas usage alone. If the block is mostly low-calldata transactions that use much more regular gas than DA
+footprint, this new mechanism wouldn’t have any effect. This is the case for the vast majority of blocks.
+
+### DA footprint gas scalar
+
+I propose to make the DA footprint gas scalar parameter configurable via the `SystemConfig`. The next Ethereum hardfork
```
geoknee, line 94: Perhaps clarify _who_ can change the configuration, i.e. `SystemConfigOwner`.

protocol/da-footprint-block-limit.md
```diff
@@ -0,0 +1,299 @@
...
+
+| Scalar | Effective Limit | Exceeding Limit | Exc. Gas Usage | Avg. Utilization | Max Utilization |
+|--------|-----------------|-----------------|----------------|------------------|-----------------|
+| 160    | 250,000         | 0               | 0              | 1.1%             | 40.0%           |
+| 400    | 100,000         | 0               | 1              | 2.7%             | 99.9%           |
+| 600    |  66,666         | 4 (0.01%)       | 8 (0.02%)      | 4.1%             | 149.9%          |
+| 800    |  50,000         | 17 (0.03%)      | 23 (0.04%)     | 5.5%             | 199.8%          |
+
+
+#### Base
```
geoknee, line 158: We should double check these figures against estimates from the base team cc @nira.

geoknee (changes requested):
From today's review session:
* We would like to double check the data analysis to ensure it matches napkin math e.g. around throttling
* We would like to consider if we can modify the design to allow for a higher target compared to the gas limit
* We are aligned on the overall design as the best we have yet considered for addressing the core problem
* We are also aligned on having the scalar be configurable

protocol/da-footprint-block-limit.md
```diff
@@ -0,0 +1,299 @@
...
+da_footprint = da_usage_estimate * da_footprint_gas_scalar
+```
+
+Here, `fastlzSize` is the length of the FastLZ-compressed RLP-encoding of a transaction.
+
+Now when evaluating whether a transactions still fits into a block, we take the maximum over the two resources, regular
+total gas use of all transactions, and new total DA footprint, and see if this max is still below the block gas limit.
+The Multidimensional Gas Metering design introduces a new block field `gas_metered` to differentiate it from `gas_used`
+as the sum total of all transaction's gas used. However, it is proposed to just repurpose a block's `gas_used` field to
+hold the maximum over both resources' totals:
```
niran, line 74: I would expect `gas_used` header fields that don't match the sum of the gas used by each transaction to break something somewhere. I don't know what would break, but I'd be surprised if nothing is relying on that invariant.

protocol/da-footprint-block-limit.md
```diff
@@ -0,0 +1,299 @@
...
+usage, then there wouldn't have been any impact, since the new block gas usage would be defined as the max over the
+total transaction gas usages and DA footprints. If it went over the gas usage, but not gas limit, then the impact would
+only have been that the new block gas usage would have been set higher (so the base fee may have potentially increased
+faster), but the block still wouldn't have been limited. Only if the DA footprint also went over the gas limit, would it
+have limited that block's transactions to a smaller DA footprint. For those blocks, we also looked at the
+distribution of the excess DA footprint as an indicator of how much those blocks would have been limited.
+
+The following tables show, for each analyzed chain, the resulting statistics.
+* *Scalar*: DA footprint gas scalar value.
+* *Effective Limit*: The DA usage limit that the given gas scalar would imply (`block_gas_limit / da_footprint_gas_scalar`), in estimated compressed bytes.
```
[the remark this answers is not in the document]
niran: I think we can proceed with just a single scalar. Here's how I think we'd approach setting the scalar.

No chain can sustain more than 64 kb/s of compressed data because that's the target throughput for the entire blobspace, so we all have `l1_target_throughput = 64000`.

```
da_footprint_gas_scalar = block_gas_limit / (block_time * l1_target_throughput * estimation_ratio * elasticity_multiplier)
```

For Base, our estimated DA sizes in production seem to be about 50-100% higher than the actual sizes, giving us an `estimation_ratio = 1.5`. With an elasticity of 3, a block time of two seconds, and a block gas limit of 150M, we get:

```
da_footprint_gas_scalar = 150,000,000 / (2 * 64000 * 1.5 * 3) = 260.4
```

The base fee should only begin to rise when the DA footprint is above the block gas target. To double-check that, we calculate `da_usage_target = block_gas_limit / elasticity_multiplier / da_footprint_gas_scalar = 150,000,000 / 3 / 260 = 192,307`, which is a good point for Base to start pricing out DA usage. Unfortunately, this produces a `da_usage_limit = block_gas_limit / da_footprint_gas_scalar = 150,000,000 / 260 = 576,923`, which is significantly higher than the `always_throttle` value we already use. But I think that's okay! We would get base fees that increase when estimated DA usage is above 192kb per block, and would rely on sequencer throttling to enforce the maximum in practice.

Assuming an `estimation_ratio = 1.5` for all OP Stack chains, here are the values that price out DA throughput above 64 kb/s.

OP Mainnet = 40,000,000 / (2 * 64000 * 1.5 * 2) = 104.2
Ink = 30,000,000 / (1 * 64000 * 1.5 * 2) = 156.25
Unichain = 30,000,000 / (1 * 64000 * 1.5 * 2) = 156.25
Soneium = 40,000,000 / (2 * 64000 * 1.5 * 2) = 104.2
Mode = 30,000,000 / (2 * 64000 * 1.5 * 2) = 78.125
World Chain = 60,000,000 / (2 * 64000 * 1.5 * 2) = 156.25

Since most chains don't need to be concerned about a DA target that prices out usage, they can focus on using the scalar for the DA **limit** rather than the target. Multiplying each of those values by `blob_limit / blob_target = 9 / 6` will produce scalars that prevent a single chain from ever exceeding the throughput of the blob limit on its own. Scalar values higher than that can be used to target a particular share of blob throughput.

Setting all chains to `da_footprint_gas_scalar = 260` shouldn't cause problems for any chain (though World Chain would want to change this if they approach 60% of blob throughput). The original proposal of 800 is probably fine for chains that expect to use less than 10% of blob throughput.

protocol/da-footprint-block-limit.md
sebastianst: done

protocol/da-footprint-block-limit.md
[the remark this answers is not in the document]
sebastianst: Thanks @niran for your thorough analysis of the scalar value! After fixing several flaws in our analysis, we have updated the numbers for all chains and it seems that a scalar value of 400 would have minimal to no impact: back testing with this scalar showed that for almost no blocks do their DA footprint exceed the gas used. This value would be slightly higher than your proposed default of 260, but given that the observed actual impact would be minimal, I'd favor to go with this higher value as a default to achieve stronger ecosystem-wide protection against DA spam by default. Individual chains may still choose to increase their DA usage targets and limits by lowering the scalar.

Starting with a scalar of 600 did we observe meaningful impact for some chains, where the DA footprint went over the total gas used for a significant percentage. Interestingly, for the Base random sample, even with 600 it did go over the gas usage only in 5.4% of blocks.

PR #320, opened 2025-08-12
feat: liveness module v2

smartcontracts:
This PR includes a design document that proposes an updated LivenessModule to be installed into all safes managed by OP Labs and the Optimism Foundation, and potentially the Security Council.

### DIFF
--- a/safe-modules/liveness-module-v2.md
+++ b/safe-modules/liveness-module-v2.md
@@ -129,8 +129,7 @@ gain control of the multisig enabling the LivenessModule.
 To mitigate this, we will require monitoring that will alert us of unexpected liveness challenges.
 
 ### Blocking Minority is Below Quorum
-If a multisig uses a simple majority or less as quorum this module will be ineffective for the
-stated scenario, because malicious actors will reach quorum before the reach a blocking minority.
+In order for the module to be effective for the stated scenario, a multisig must require at least a simple majority (defined as `threshold > (total + 1 )/ 2`)  as quorum, otherwise malicious actors will reach quorum before reaching a blocking minority.
 
 Instead of implementing a module that would address this scenario, we accept that this module
 is intended for multisigs where the quorum is greater than a simple majority.
--- a/safe-modules/liveness-module-v2.md
+++ b/safe-modules/liveness-module-v2.md
@@ -47,7 +47,8 @@ this module, as well as any 3rd parties who rely on the security properties of t
 
 - Coinbase (potentially impacted 3rd party)
 - Uniswap (`LivenessModule` user)
-- Optimism Foundation (`LivenessModule` user)
+- Security Council (`LivenessModule` user)
+- Optimism Foundation
 
 ## Requirements and Constraints
 
@@ -72,18 +73,20 @@ mechanism as an alternative to the existing poke mechanism inside of `LivenessGu
 constructor parameters. Any Safe on any network can choose to configure the module and then enable
 the module within the Safe.
 
+If possible, this module should be integrated with all other safe modules and guards in a single
+contract that is enabled and configured in all Optimism Safes.
+
 ### Configuration
 
-A Safe configures the module by setting a challenge period, a challenge bond, and a fallback owner.
+A Safe configures the module by setting a challenge period and a fallback owner.
 
 ### Challenge Protocol
 
-Any user may trigger an Account Liveness Challenge by targeting a specific Safe address. In
-response to this challenge, a threshold of owners (same threshold as the Safe itself) must sign a
-message that confirms the liveness of the safe (e.g., a message over the Safe address and the
-current timestamp). If the Safe fails to reach this threshold, all owners are removed from the
-account and ownership is transferred to the fallback account. If the Safe successfully reaches the
-threshold, the challenge fails and the Safe receives the bond as a reward.
+The fallback owner may trigger an Account Liveness Challenge by targeting a specific Safe address.
+In response to this challenge, a threshold of owners (same threshold as the Safe itself) must 
+execute a regular transaction that confirms the liveness of the safe. If the Safe fails to execute
+this transaction, all owners are removed from the Safe and ownership is transferred to the fallback
+account. If the Safe successfully executes the transaction, the challenge fails and nothing happens.
 
 The Account Liveness Challenge addresses both of the ways in which a Safe can fail to be live. If
 sufficiently many honest users cannot sign that the account cannot reach a threshold, the account
@@ -118,6 +121,19 @@ prefer to simply remove an owner that is unable to sign.
 
 <!-- Explain any risks and uncertainties that this design includes. Highlight aspects of the design
 that remain unclear and any potential issues we may face down the line. -->
+### Malicious Fallback
+While the fallback owner is chosen as a trusted actor, it can always happen that they are
+eventually compromised. In that case, they might issue a liveness challenge and if unnoticed
+gain control of the multisig enabling the LivenessModule.
+
+To mitigate this, we will require monitoring that will alert us of unexpected liveness challenges.
+
+### Blocking Minority is Below Quorum
+If a multisig uses a simple majority or less as quorum this module will be ineffective for the
+stated scenario, because malicious actors will reach quorum before the reach a blocking minority.
+
+Instead of implementing a module that would address this scenario, we accept that this module
+is intended for multisigs where the quorum is greater than a simple majority.
 
 ### Fallback Ownership
 
--- a/safe-modules/liveness-module-v2.md
+++ b/safe-modules/liveness-module-v2.md
@@ -1,0 +1,139 @@
+
+# LivenessModule V2
+
+## Context and Problem Statement
+
+<!-- The Context and Problem Statement section is one of the most critical parts of the design
+document process. Use this section to clearly highlight the background context for the problem,
+the specific issues being faced by customers, and any constraints on a solution as defined either
+by customers or by technical limitations. Context and Problem Statement is an opportunity to tell
+the story that helps to motivate the rest of the design document. -->
+
+[Safe Accounts](https://safe.global) are configured as a set of owner addresses where some
+threshold of addresses can sign and ultimately execute a transaction through the account. We must
+solve for two safety cases, first that a threshold of signers do not maliciously execute a
+transaction (safety failure) and second that a blocking minority of signers
+(`total - threshold + 1`) don't become unable to sign (liveness failure). Either case can be
+equally catastrophic but we tend to focus on the safety failure case. Here we attempt to produce a
+solution for the liveness failure case that fits within certain constraints.
+
+We define multisig liveness as the ability for the multisig to reach consensus and execute a
+transaction within a bounded amount of time. Note that a multisig does not need to reach consensus
+over every possible transaction (because the members have the ability to reject transactions as
+needed). We instead say that a multisig is live as long as it is able to reach consensus over a
+transaction that states "this multisig is live".
+
+At a high level, a multisig fails to be live if at least a blocking minority of signers prevents
+the multisig from being able to reach consensus. This can happen if signers are honest but unable
+to sign (e.g., keys are lost) or if signers are malicious and refuse to sign. We must address both
+problems if we want to avoid multisig liveness failures. The existing `LivenessGuard` is
+insufficient in this regard because it does not handle malicious signers.
+
+Any system to address the above must also scale sufficiently such that a signer is not
+unnecessarily burdened by the liveness mechanism (this is a product constraint). The
+`LivenessGuard` is also insufficient here because it does not scale across multiple chains and it
+demands signatures on a regular basis (as opposed to only demanding signatures as needed).
+
+## Customer Description
+
+<!-- Provide a brief summary of the customers for this design document. -->
+
+The customers for this design doc are any participants in the multisig accounts that would utilize
+this module, as well as any 3rd parties who rely on the security properties of these accounts.
+
+### Customer Reviewers
+
+<!-- Identify at least one customer who should be involved in the review of this document. -->
+
+- Coinbase (potentially impacted 3rd party)
+- Uniswap (`LivenessModule` user)
+- Optimism Foundation (`LivenessModule` user)
+
+## Requirements and Constraints
+
+<!-- Identify the solution requirements and any additional design constraints from the Context and
+Problem Statement section in a bulleted list. -->
+
+- Solve both the owner liveness and account liveness problems *securely*.
+- Keep code as minimal as possible to avoid complex logic in Safe modules.
+- Reduce mental overhead for Safe owners and multisig leads.
+
+## Proposed Solution
+
+<!-- Explain the solution that you believe best addresses the problem described above. -->
+
+We propose the introduction of a new module, `LivenessModuleV2` (name is unimportant) to replace
+the existing `LivenessGuard`/`LivenessModule`. `LivenessModuleV2` would employ a challenge/response
+mechanism as an alternative to the existing poke mechanism inside of `LivenessGuard`.
+
+### Singleton
+
+`LivenessModuleV2` is a singleton that exists at a known address on all chains and has no
+constructor parameters. Any Safe on any network can choose to configure the module and then enable
+the module within the Safe.
+
+### Configuration
+
+A Safe configures the module by setting a challenge period, a challenge bond, and a fallback owner.
+
+### Challenge Protocol
+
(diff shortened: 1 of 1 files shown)


safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Alternatives Considered
+
+<!-- Describe any alternatives that were considered during the development of this design. Explain
+why the alternative designs were ultimately not chosen and where they failed to meet the product
+requirements. -->
+
+### Owner Liveness Challenge
+
+We previously considered a model that would allow users to challenge either the entire Safe (as
```
alcueca, line 101: We don't describe the alternative model where the entire Safe is challenged, and it is confusing because in this design doc we are proposing a module where the entire Safe is challenged.

Was the Account Liveness Challenge like the Owner Liveness Challenge, but in a loop so that it challenges all owners in turn?

In any case, I would just mention the Owner Liveness Challenge here. I think that's enough.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+One open question is what the Security Council should set as a fallback. We'd likely want this to
+be another multisig operated by an independent and well-trusted entity.
+
+### Blunt Instrument
+
+It should be noted that the updated `LivenessModule` is more of a blunt instrument than the
+existing module. We no longer have the ability to remove specific owners, we instead trigger a
+fallback immediately if a threshold cannot be reached. This is less surgical, but we believe it is
+equally effective and ultimately safer/simpler than the existing module because users don't need to
+account for as many parameters.
```
alcueca, line 139: You might want to note that if at some point removing individual inactive signers becomes too much of a burden for the active signers, it would be possible to implement a module that challenges an individual signer to show liveness or be removed. Or you can also add it to the out of scope section.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+A Safe configures the module by setting a challenge period, a challenge bond, and a fallback owner.
+
+### Challenge Protocol
+
+Any user may trigger an Account Liveness Challenge by targeting a specific Safe address. In
+response to this challenge, a threshold of owners (same threshold as the Safe itself) must sign a
+message that confirms the liveness of the safe (e.g., a message over the Safe address and the
+current timestamp). If the Safe fails to reach this threshold, all owners are removed from the
+account and ownership is transferred to the fallback account. If the Safe successfully reaches the
+threshold, the challenge fails and the Safe receives the bond as a reward.
```
alcueca, line 86: I would simplify even further and instead of talking about a bond, I would consider it as a payment. The challenger always has to pay the amount, and it always be credited to the Safe.

In all likelihood, the challenge would be done by someone that trusts the fallback owner, which then can arrange for the payment to be reimbursed if necessary. Not implementing a path in the module where assets leave the Safe is much safer.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+At a high level, a multisig fails to be live if at least a blocking minority of signers prevents
+the multisig from being able to reach consensus. This can happen if signers are honest but unable
+to sign (e.g., keys are lost) or if signers are malicious and refuse to sign. We must address both
+problems if we want to avoid multisig liveness failures. The existing `LivenessGuard` is
+insufficient in this regard because it does not handle malicious signers.
+
+Any system to address the above must also scale sufficiently such that a signer is not
+unnecessarily burdened by the liveness mechanism (this is a product constraint). The
+`LivenessGuard` is also insufficient here because it does not scale across multiple chains and it
+demands signatures on a regular basis (as opposed to only demanding signatures as needed).
```
alcueca, line 35: I think that we should describe the current Liveness Guard and the problems it causes here as well. Namely harassment of both signers and chainops for no good reason.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+- Coinbase (potentially impacted 3rd party)
+- Uniswap (`LivenessModule` user)
+- Optimism Foundation (`LivenessModule` user)
+
+## Requirements and Constraints
+
+<!-- Identify the solution requirements and any additional design constraints from the Context and
+Problem Statement section in a bulleted list. -->
+
+- Solve both the owner liveness and account liveness problems *securely*.
```
alcueca, line 57: Maybe expand a bit on this one on what *securely* means. Which is not allowing malicious users to use the solution to maintain or increase their control of the multisig.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+  percentage which creates more surface area for potential issues.
+
+Additionally, in most cases, the Owner Liveness Challenge would be unused because the Safe would
+prefer to simply remove an owner that is unable to sign.
+
+## Risks and Uncertainties
+
+<!-- Explain any risks and uncertainties that this design includes. Highlight aspects of the design
+that remain unclear and any potential issues we may face down the line. -->
+
```
alcueca, line 121: There is the risk that a multisig is challenged and no one notices. We will need monitoring to mitigate that.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+The customers for this design doc are any participants in the multisig accounts that would utilize
+this module, as well as any 3rd parties who rely on the security properties of these accounts.
+
+### Customer Reviewers
+
+<!-- Identify at least one customer who should be involved in the review of this document. -->
+
+- Coinbase (potentially impacted 3rd party)
+- Uniswap (`LivenessModule` user)
+- Optimism Foundation (`LivenessModule` user)
```
alcueca, line 50: Not the Security Council?

maurelian: in fact the current liveness system is on the SC not the Foundation, but I think the point is that it should be on both.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+
+The Owner Liveness Challenge faces a number of problems:
+
+- It creates a second challenge path.
+- It only solves the honest user case, whereas the Account Liveness Challenge solves both cases.
+- It requires that the Safe configure various new parameters like minimum owner count and threshold
+  percentage which creates more surface area for potential issues.
+
+Additionally, in most cases, the Owner Liveness Challenge would be unused because the Safe would
+prefer to simply remove an owner that is unable to sign.
```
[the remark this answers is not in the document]
alcueca: Recorded this comment in the threat model, so that the mitigation is later tracked.

image

safe-modules/liveness-module-v2.md
smartcontracts: This is something we wanted to avoid (this would be the Owner Liveness Challenge). I don't think it would be too burdensome, ideally we shouldn't need to develop this. A single transaction to swap an owner shouldn't be too much effort for signers.

safe-modules/liveness-module-v2.md
smartcontracts: Hm, interesting. I like it. Note that the funds wouldn't leave the module until a challenge is successful, but payment is fine, it works fine for the cases we are expecting.

safe-modules/liveness-module-v2.md
DanielVF: I'm a fan of the blunt approach. It's much easy to understand, and easier to not run into unanticipated situations in the real world across many safes.

safe-modules/liveness-module-v2.md
DanielVF: I like payment as well - removes a lot of math and tracking.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+### Challenge Protocol
+
+Any user may trigger an Account Liveness Challenge by targeting a specific Safe address. In
+response to this challenge, a threshold of owners (same threshold as the Safe itself) must sign a
+message that confirms the liveness of the safe (e.g., a message over the Safe address and the
+current timestamp). If the Safe fails to reach this threshold, all owners are removed from the
+account and ownership is transferred to the fallback account. If the Safe successfully reaches the
+threshold, the challenge fails and the Safe receives the bond as a reward.
+
+The Account Liveness Challenge addresses both of the ways in which a Safe can fail to be live. If
```
DanielVF, line 88: I don't think this addresses the malicious angle.

Let's say a malicious minority refuses to sign needed transactions. The safe is in practice dead. Then someone proposes a liveness check. The malicious minority would then sign the liveness check, and the safe is still as it was, unable to do anything real.

Perhaps another approach would be to have the liveness ping come from the safe itself. In this case the honest majority can bundle together both an honest proposal to kick the malicious actor, and the liveness ping. If the malicious minority does not sign, power goes to the fallback. If they do sign, power is taken away from them. Win Win.

safe-modules/liveness-module-v2.md
DanielVF: It is critical that it is monitored, yes.

safe-modules/liveness-module-v2.md
DanielVF: Certain multi-sig size combos do not suffer from any either kind of minority problems. A 5/8 multisig for example, does not allow a minority to break quorum or interfere in any way with a majority.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+
+- It creates a second challenge path.
+- It only solves the honest user case, whereas the Account Liveness Challenge solves both cases.
+- It requires that the Safe configure various new parameters like minimum owner count and threshold
+  percentage which creates more surface area for potential issues.
+
+Additionally, in most cases, the Owner Liveness Challenge would be unused because the Safe would
+prefer to simply remove an owner that is unable to sign.
+
+## Risks and Uncertainties
```
alcueca, line 117: Another potential risk is this module interacting with a potential Timelock implementation. Reusing the default safe execution flow would mean less logic, but potential race conditions or logic clashes.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+
+Both the existing `LivenessModule` and this updated iteration rely on fallback ownership. In the
+case that the challenge fails, ownership of the account must be transferred to some fallback
+address. We believe this is acceptable and reasonable. Most multisig accounts can simply name the
+Optimism Foundation multisig as a fallback, and the Optimism Foundation multisig can name the
+Security Council as a fallback.
+
+One open question is what the Security Council should set as a fallback. We'd likely want this to
+be another multisig operated by an independent and well-trusted entity.
+
```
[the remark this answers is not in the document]
alcueca: Discuss whether the Foundation can be the fallback for the SC, and viceversa, at the same time.

safe-modules/liveness-module-v2.md
alcueca: `blocking_minority` is defined in the first paragraph of the design doc.

safe-modules/liveness-module-v2.md
wildmolasses: @alcueca i think your comment here is a good idea, will it be included?

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,139 @@
...
+We previously considered a model that would allow users to challenge either the entire Safe (as
+happens in the Account Liveness Challenge) or an individual owner (i.e., an Owner Liveness
+Challenge). An Owner Liveness Challenge would be more similar to the behavior of the existing
+`LivenessModule`, where a user could target a specific owner and require that owner respond or be
+removed from the Safe.
+
+The Owner Liveness Challenge faces a number of problems:
+
+- It creates a second challenge path.
```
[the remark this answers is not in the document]
maurelian: I think this is referring to a module that would have two challenge paths: 
1. Account Liveness Challenge
2. Owner Liveness Challenge

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,155 @@
...
+
+We define multisig liveness as the ability for the multisig to reach consensus and execute a
+transaction within a bounded amount of time. Note that a multisig does not need to reach consensus
+over every possible transaction (because the members have the ability to reject transactions as
+needed). We instead say that a multisig is live as long as it is able to reach consensus over a
+transaction that states "this multisig is live".
+
+At a high level, a multisig fails to be live if at least a blocking minority of signers prevents
+the multisig from being able to reach consensus. This can happen if signers are honest but unable
+to sign (e.g., keys are lost) or if signers are malicious and refuse to sign. We must address both
```
maurelian, line 28: What I don't understand is how the proposed solution detects the presence of malicious signers? ie. If Mallory controls a quorum blocking number of signers, then she has invested a lot into gaining control of the Safe, and is likely motivated to either:
- profit from some form of griefing by preventing an upgrade
- gaining control to yet more keys

So why would she burn all that hard work by refusing to sign a harmless liveness proof?

maurelian: I am satisfied by [this comment](https://github.com/ethereum-optimism/design-docs/pull/320/files#r2283404448), which is included in the document on Line 93.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,155 @@
...
+### Malicious Fallback
+While the fallback owner is chosen as a trusted actor, it can always happen that they are
+eventually compromised. In that case, they might issue a liveness challenge and if unnoticed
+gain control of the multisig enabling the LivenessModule.
+
+To mitigate this, we will require monitoring that will alert us of unexpected liveness challenges.
+
+### Blocking Minority is Below Quorum
+If a multisig uses a simple majority or less as quorum this module will be ineffective for the
+stated scenario, because malicious actors will reach quorum before the reach a blocking minority.
```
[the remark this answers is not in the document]
alcueca: Mmm, the situation presents itself with a simple majority for multisigs with an uneven number of signers, or half of the signers for multisigs with an even number of them.

safe-modules/liveness-module-v2.md
```diff
@@ -0,0 +1,155 @@
...
+
+At a high level, a multisig fails to be live if at least a blocking minority of signers prevents
+the multisig from being able to reach consensus. This can happen if signers are honest but unable
+to sign (e.g., keys are lost) or if signers are malicious and refuse to sign. We must address both
+problems if we want to avoid multisig liveness failures. The existing `LivenessGuard` is
+insufficient in this regard because it does not handle malicious signers.
+
+Any system to address the above must also scale sufficiently such that a signer is not
+unnecessarily burdened by the liveness mechanism (this is a product constraint). The
+`LivenessGuard` is also insufficient here because it does not scale across multiple chains and it
```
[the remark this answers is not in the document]
maurelian: > We need a liveness module that scales with the number of chains.

more chains doesn't necessarily equate to more Safes though. 

Regardless, now I understand what this is saying, so I don't think we need to really nitpick about the flaws of the current module. Resolving.

safe-modules/liveness-module-v2.md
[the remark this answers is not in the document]
maurelian: Good point. I do think it's important to document when the module is worth using or not. Appreciate if you can validate my logic here:

Given the definition of a blocking minority:

```python
blocking_minority = total - threshold + 1
```

And given that the module is ineffective if 'malicious actors will reach the threshold before they reach a blocking minority', 
then the following statement tells us when the module is effective:

```python
blocking_minority < threshold
```

Thus we derive the following test to determine if the module is effective:

```python
  blocking_minority < threshold
  total - threshold + 1 < threshold
  total < 2*threshold - 1 (OR threshold > (total + 1 )/ 2)
```

Test cases: 

| threshold | total | total < 2*threshold - 1 | effective? |
| --------- | ----- | ----------------------- | ---------- |
| 4         | 8     | 8 < 7                   | false      |
| 5         | 8     | 8 < 9                   | true       |
| 4         | 7     | 7 < 7                   | false      |
| 5         | 7     | 7 < 9                   | true       |

If this is correct, then it's not quite the case that the module doesn't work for a simple majority (ie. it works for 5 of 8). 

Then we can rewrite the statement more precisely as: 

```suggestion
In order for the module to be effective for the stated scenario, a multisig must require at least a simple majority (ie. `threshold > (total + 1 )/ 2`)  as quorum, otherwise malicious actors will reach quorum before reaching a blocking minority.
```

alcueca:
@mds1, @smartcontracts, @maurelian, we are ready to merge this PR. Could I please get your approval?

PR #353, opened 2025-11-06
feat: OPCM v2 design doc

smartcontracts:
This PR adds the design document for OPCM v2.

### DIFF
--- a/protocol/opcm-v2.md
+++ b/protocol/opcm-v2.md
@@ -14,6 +14,8 @@ The Proofs team is most heavily impacted by OPCM v1's limitations. When protocol
 
 4. **High knowledge burden**: Developers need deep understanding of how OPCM integrates with op-deployer (the Go deployment tool), including the glue code that connects Solidity contracts to Go tooling. This prevents developers from focusing on their specific domain without understanding the entire system.
 
+5. **Input changes are hard**: Changes to the inputs of OPCM cascades into many different changes across the repository, and the places in which those changes should be made are not always obvious.
+
 The cumulative effect of these issues, combined with our improved understanding of the right architectural direction, makes this the appropriate time to redesign OPCM. OPCM v2 aims to simplify the interface, eliminate hardcoded assumptions, unify deployment and upgrade paths, and reduce the knowledge burden on developers while maintaining full backwards compatibility with existing functionality.
 
 **Key constraints:**
@@ -80,8 +82,8 @@ OPContractsManager v2 redesigns the architecture around three core principles: *
 ### Architecture Overview
 
 OPCM v2 consolidates the 5 separate functions of v1 into 2 primary functions:
-- `deploy(Config memory config)`: Deploys a new OP Stack chain
-- `upgrade(Config memory config)`: Upgrades an existing chain
+- `deploy(DeployConfig memory deployConfig)`: Deploys a new OP Stack chain
+- `upgrade(UpgradeConfig memory upgradeConfig)`: Upgrades an existing chain
 
 Both functions share the same initialization path for contracts, eliminating the special upgrade functions that previously required maintenance between releases.
 
@@ -91,15 +93,23 @@ Both functions share the same initialization path for contracts, eliminating the
 
 Replace hardcoded assumptions about game types with dynamic arrays:
 ```solidity
-struct GameConfig {
+struct DisputeGameConfig {
+    bool enabled;
+    uint256 initBond;
     GameType gameType;
-    address implementation;
-    bytes32 prestate;
+    bytes gameArgs; // ABI-encoded struct specific to this game type
 }
 
-struct Config {
+struct DeployConfig {
     // ... other config
-    GameConfig[] games;  // Support arbitrary number of game types
+    DisputeGameConfig[] disputeGameConfig;  // Support arbitrary number of game types
+}
+
+struct UpgradeConfig {
+    ISystemConfig systemConfig;
+    DIsputeGameConfig[] disputeGameConfig;
+    ConfigOverrides[] configOverrides; // Sometimes we're required to set a new config variable
+    string[] permittedProxyDeployments; // Sometimes we need to permit new proxies to be deployed
 }
 ```
 
@@ -142,7 +152,9 @@ After assembling the chain world, both functions gather the configuration inputs
 - **Deployments**: All configuration input comes directly from the deploy input parameter
 - **Upgrades**: Only a subset of configuration is provided in the upgrade input; the remaining configuration is gathered automatically from the existing deployed contracts
 
-This asymmetry allows upgrades to specify only what changes, while deployments must provide complete configuration.
+This asymmetry allows upgrades to specify only what changes, while deployments must provide complete configuration. When upgrading, if there is a net new parameter that doesn't already exist on chain, the user will need to provide that parameter as a configuration override.
+
+**NOTE**: Configuration overrides also mean that the `upgrade` function can be used to manage configuration of onchain values.
 
 **Phase 3: Common Transformation**
 


protocol/opcm-v2.md
```diff
@@ -0,0 +1,234 @@
...
+
+Both functions feed into a shared transformation function that takes the proxies and configuration as input and updates/initializes those proxies appropriately. This common transformation path ensures consistent behavior between initial deployments and subsequent upgrades, eliminating the need for deployment-specific vs. upgrade-specific initialization logic.
+
+### Requirements Mapping
+
+- **Deploy new chains**: `deploy()` function
+- **Upgrade existing chains**: `upgrade()` function  
+- **Dynamic game types**: GameConfig array in Config struct
+- **Add game types**: Implicit in `upgrade()` with modified games array
+- **Update prestates**: Implicit in `upgrade()` with modified games array
```
mds1: What do you think about preserving `addGameType` and `updatePrestates` as convenience functions? We often want to do just those actions (e.g. changing prestate for a hard fork) and it feels error prone and hard to review to pass a full `Config` into upgrade for that. These convenience functions would just gather the other inputs before calling `upgrade`

protocol/opcm-v2.md
```diff
@@ -0,0 +1,234 @@
...
+The load-or-deploy pattern attempts to load each contract address from an expected source (e.g., a known deployment address for upgrades, or a predictable location). If the contract cannot be found at the expected location, it is deployed instead. This unified pattern means the same chain world loading function serves both deploy and upgrade operations.
+
+**Phase 2: Input Gathering**
+
+After assembling the chain world, both functions gather the configuration inputs needed to properly configure the proxy contracts:
+
+- **Deployments**: All configuration input comes directly from the deploy input parameter
+- **Upgrades**: Only a subset of configuration is provided in the upgrade input; the remaining configuration is gathered automatically from the existing deployed contracts
+
+This asymmetry allows upgrades to specify only what changes, while deployments must provide complete configuration.
```
ajsutton: How is this expressed in code? How do you know what you need to provide when calling upgrade?  The ABIs above pass the same Config object to both deploy and upgrade so it seems unclear how to know if an upgrade requires a new setting be specified or not.

protocol/opcm-v2.md
```diff
@@ -0,0 +1,234 @@
...
+- Developers only need to add new CLI arguments to op-deployer when introducing net-new inputs
+- Reduces knowledge burden—developers don't need to understand integration layer
+
+### Implementation Flow
+
+Both `deploy()` and `upgrade()` follow a unified three-phase flow:
+
+**Phase 1: Chain World Assembly (Load-or-Deploy Pattern)**
+
+Both functions begin by gathering the complete set of contracts for the chain—the "chain world"—which includes SystemConfig, OptimismPortal, AnchorStateRegistry, and all other chain contracts.
```
mslipper: Does this include the Superchain contracts? How do we upgrade those?

protocol/opcm-v2.md
```diff
@@ -0,0 +1,234 @@
...
+### Key Architectural Changes
+
+**1. Dynamic Dispute Game Interface**
+
+Replace hardcoded assumptions about game types with dynamic arrays:
+```solidity
+struct GameConfig {
+    GameType gameType;
+    address implementation;
+    bytes32 prestate;
```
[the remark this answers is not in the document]
ajsutton: From discussion: OPCM still only supports a fixed set of game types and knows how to convert from the particular config for that game type (which can be customised for it) to the game args needed (mixing in addresses for things like DelayedWETH etc).

protocol/opcm-v2.md
```diff
@@ -0,0 +1,234 @@
...
+**Phase 3: Common Transformation**
+
+Both functions feed into a shared transformation function that takes the proxies and configuration as input and updates/initializes those proxies appropriately. This common transformation path ensures consistent behavior between initial deployments and subsequent upgrades, eliminating the need for deployment-specific vs. upgrade-specific initialization logic.
+
+### Requirements Mapping
+
+- **Deploy new chains**: `deploy()` function
+- **Upgrade existing chains**: `upgrade()` function  
+- **Dynamic game types**: GameConfig array in Config struct
+- **Add game types**: Implicit in `upgrade()` with modified games array
```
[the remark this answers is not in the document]
ajsutton: From discussion: For game types, we have a fixed list that is supported so they can just be removed upfront and then re-add what is specified to make this work.

For the generic case we should ensure that initialisers only set simple fields and don't write to maps so they are safe to call repeatedly.

protocol/opcm-v2.md
maurelian: good callout. The doc mentions 5 functions  (`deploy()`, `upgrade()`, `addGameType()`, `updatePrestate()`, `interopMigrate()`), but there is now also `upgradeSuperchainConfig()`.

protocol/opcm-v2.md
```diff
@@ -0,0 +1,234 @@
...
+- Defensive coding practices throughout implementation
+- State diff review for the first OPCM v2 upgrade to verify correctness
+- Multiple rounds of internal security review before audit
+
+### Functionality Coverage Risk: Missing OPCM v1 Capabilities
+
+**Risk**: We might inadvertently miss some existing functionality that OPCM v1 provides when transitioning to the new 2-function interface.
+
+**Mitigation**:
+- Ensure all existing OPCM v1 tests pass with v2 implementation
```
maurelian: Are we planning to use the '[strangler pattern](https://learn.microsoft.com/en-us/azure/architecture/patterns/strangler-fig#solution)' here, so that both OPCMs exist and are tested in parallel during development of V2? This was used successfully when migrating the deploy scripts to the new pattern without the I/O contracts.

protocol/opcm-v2.md
```diff
@@ -0,0 +1,234 @@
...
+- **Backwards compatibility**: Support all existing OPCM v1 operations and use cases
+- **Gas efficiency**: Single deployment transaction must fit within Fusaka per-tx gas limit
+- **Security**: Must be auditable and maintain security properties of v1
+- **Maintainability**: Unified code paths for deploy and upgrade operations
+- **Developer experience**: Minimize knowledge burden through auto-generated glue code
+- **Release readiness**: Always in deployable state without post-upgrade cleanup
+
+### Constraints
+
+- Must integrate with existing op-deployer Go tooling
```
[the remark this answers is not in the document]
smartcontracts: We are aiming to maintain the existing interface for now, but then deprecate that interface in favor of a simpler one

protocol/opcm-v2.md
ajsutton: I think what's confusing me is lines 83 and 84 have `deploy(Config)` and `upgrade(Config)` which makes it look like they take the same inputs. In which case, how do you know which parts of that `Config` are required to be specified for the upgrade being done?

https://github.com/ethereum-optimism/design-docs/pull/353/files/1acc722cdf458c0744e166c0c37443a4a8b3f3b4#diff-62c8771657b68fe38ae394ef3d255c7928c0b7853b0077ca70ff62007de8beeeR83

But maybe I'm looking at the wrong place for the ABI you updated.

protocol/opcm-v2.md
```diff
@@ -0,0 +1,246 @@
...
+}
+
+struct DeployConfig {
+    // ... other config
+    DisputeGameConfig[] disputeGameConfig;  // Support arbitrary number of game types
+}
+
+struct UpgradeConfig {
+    ISystemConfig systemConfig;
+    DIsputeGameConfig[] disputeGameConfig;
```
0xiamflux: nit: typo on `DIsputeGameConfig`, should be `DisputeGameConfig`

PR #354, opened 2025-11-07
design doc: add fusaka on L2

joshklop:


protocol/fusaka-on-l2.md
```diff
@@ -0,0 +1,101 @@
...
+As a rule of thumb, including code snippets (except for defining an external API)
+is likely too low level. -->
+
+Assuming the division of relevant and irrelevant EIPs is correct, we need to answer the adoption questions.
+
+### 7825 Tx Gas Limit Cap: What should the cap be on L2? Should it be configurable?
+
+We propose adopting a constant cap that matches L1: 16,777,216 gas (2^24).
+This is the simplest option that maximizes Ethereum equivalence.
+If there is a strong need for a different cap or flexibility in setting the cap, that can be considered in future hardforks.
```
sebastianst: I agree. There's a lot of value in maximizing Eth equivalence.

The one exception I propose to make here is for deposit transactions. These should be excluded from the gas limit. This makes it easier to any future crazy upgrade txs to not need to work around such a limit, like the L2CM migration.

joshklop: I feel this is also somewhat related to https://github.com/ethereum-optimism/design-docs/pull/351#discussion_r2498337596

Can we add a way for the EL to check that a deposit tx is an upgrade transaction, and not subtract the upgrade transaction's gas used from the block gas limit? That way inclusion is guaranteed.

niran: We agree too. We've already been [enforcing a 25M cap at the mempool level](https://docs.base.org/base-chain/network-information/block-building#per-transaction-gas-maximum) since September 17, and we want to further reduce that to 16.7M along with the rest of the Superchain when Fusaka's EIPs make it into OP Stack. This lower limit improves flashblock performance and mitigates the degradation than underpriced opcodes can cause (e.g. XEN transactions).

PR #361, opened 2026-01-15
design-doc: initial SDM design doc draft

nonsense:
**Description**

This is a design doc for **SDM**.

**Background**

First, we implemented SDM PoC 1 in `op-geth`, which was an attempt to build the ideas summarized at:
- https://www.notion.so/oplabs/Fee-Rebates-2c6f153ee162804ab80bfc48ea85095d and
- https://www.notion.so/oplabs/external-draft-A-New-Gas-Model-for-the-OP-Stack-2c8f153ee16280ca9b23dbb4195c2bbb

Then, PoC 1 was superseded by PoC 2, implemented in `op-reth`, which is the **current design** we plan on going to production with. In preparation for the design review, please read the following documents:

- SDM PoC2 Review document - https://www.notion.so/oplabs/SDM-PoC-2-Review-330f153ee162801e8fc8e5e79c010e40
- SDM PoC2 Benchmarking document - https://www.notion.so/oplabs/SDM-Benchmarking-for-PoC-2-Block-Level-Warming-330f153ee16280d4b2effdf0b3b53eb7
- https://www.notion.so/uniswaplabs/Block-Level-Warming-2aac52b2548b8029b9dcfc4f51dc0a34

---

Fixes: https://github.com/ethereum-optimism/design-docs/issues/360

### DIFF
--- a/protocol/sdm.md
+++ b/protocol/sdm.md
@@ -19,15 +19,13 @@ The first application of SDM is the **block-level warming rebates** policy: it a
 
 ## Problem Statement + Context
 
-The EVM gas model is a static, in-protocol approximation of resource consumption. It does not capture all execution characteristics that matter operationally to an OP Stack sequencer. In particular, it does not account for the fact that transaction ordering within a block can change the effective cost of later transactions when accounts and storage slots have already been accessed.
+The EVM gas model is a static, in-protocol approximation of resource consumption. It does not capture all execution characteristics that matter operationally to an OP Stack sequencer.
 
-This creates a few problems:
+The sequencer currently cannot reflect execution efficiencies and inefficiencies in fees, even when those are material.
 
-- transactions that benefit from prior accesses in the same block are priced as if they were paying the full cold-access cost;
-- the sequencer cannot reflect block-local execution efficiencies in fees, even when those efficiencies are material;
-- repeated access patterns within a block are not surfaced in a way that can be inspected or analyzed later.
+The goal of SDM is to let the sequencer account for such effects explicitly through subjective sequencer-defined refunds, starting with a narrow and observable policy: **block-level warming rebates** - an implementation of [EIP-7863](https://eips.ethereum.org/EIPS/eip-7863).
 
-The goal of SDM is to let the sequencer account for such effects explicitly through sequencer-defined refunds, starting with a narrow and observable policy: **block-level warming rebates** - an implementation of [EIP-7863](https://eips.ethereum.org/EIPS/eip-7863).
+Currently the EVM gas model does not account for the fact that transaction ordering within a block can change the effective cost of later transactions when accounts and storage slots have already been accessed. As a result, transactions that benefit from prior accesses in the same block are priced as if they were paying the full cold-access cost.
 
 ## Proposed Solution
 
--- a/protocol/sdm.md
+++ b/protocol/sdm.md
@@ -1,0 +1,157 @@
+# Sequencer-Defined Metering (SDM) - an offchain and out-of-protocol mechanism for gas refunds by the sequencer
+
+| Author             | _Anton Evangelatov, Josh Klopfenstein_ |
+| ------------------ | -------------------------------------- |
+| Created at         | _2026-01-13_                           |
+| Initial Reviewers  | _TBD_                                  |
+| Need Approval From | _TBD_                                  |
+| Status             | _Draft_                                |
+
+## Purpose
+
+Allow the sequencer of OP Stack chains to influence the fee paid on a per-transaction basis, aligning fees with the actual performance characteristics of each transaction, rather than rely solely on the static, in-protocol EVM gas model.
+
+## Summary
+
+Introduce a subjective sequencer-defined gas refund alongside `gas used` and `gas limit`. SDM is introducing a new system post-execution transaction, with type `0x7D`, which contains post-execution data injected by the sequencer. The motivation for SDM and the first application of SDM is the **block-level warming rebates** policy: it allows transactions to benefit from already warmed accounts and storage slots from earlier in the same block by a sequencer-defined rebate, as described in [EIP-7863](https://eips.ethereum.org/EIPS/eip-7863). This is the initial go-to-market policy; future SDM policies may issue different kinds of refunds. Users still sign transactions in the usual way for the worst case, while the sequencer may charge less when execution benefits from block-local access reuse.
+
+## Problem Statement + Context
+
+The EVM gas model is a static, in-protocol approximation of resource consumption. It does not capture all execution characteristics that matter operationally to an OP Stack sequencer. In particular, it does not account for the fact that transaction ordering within a block can change the effective cost of later transactions when accounts and storage slots have already been accessed.
+
+This creates a few problems:
+
+- transactions that benefit from prior accesses in the same block are priced as if they were paying the full cold-access cost;
+- the sequencer cannot reflect block-local execution efficiencies in fees, even when those efficiencies are material;
+- repeated access patterns within a block are not surfaced in a way that can be inspected or analyzed later.
+
+The goal of SDM is to let the sequencer account for such effects explicitly, starting with a narrow and observable policy: **block-level warming rebates** - an implementation of [EIP-7863](https://eips.ethereum.org/EIPS/eip-7863).
+
+## Proposed Solution
+
+The proposed solution is to introduce a sequencer-defined refund alongside **gas used** (EVM gas used) and **gas limit**, at the expense of putting more trust in the sequencer of a given OP Stack chain. In the initial policy, the sequencer computes block-warming rebates off-chain and communicates them in-block.
+
+Under this policy:
+
+- normal EVM execution remains unchanged;
+- users continue to sign standard transactions with a normal gas limit;
+- the sequencer computes a per-transaction rebate when a transaction benefits from accounts or storage slots already warmed earlier in the block;
+- the sequencer includes a new system post-execution transaction, with type `0x7D`, into each block. This transaction contains all rebates for all transactions in the block;
+- the rebate is exposed as metadata and projected onto receipts for each transaction.
+
+### Core Mechanism
+
+1. **Track warming across the block:** During block execution, the sequencer tracks which accounts and storage slots have already been warmed by earlier transactions.
+2. **Compute per-transaction rebate:** For each later non-deposit transaction, if an accessed account or storage slot was already warmed earlier in the block, the sequencer accumulates a rebate.
+3. **Serialize the rebate data in-block:** At the end of block building, the sequencer emits a system post-execution transaction, with type `0x7D`, containing the refund entries.
+4. **Project rebates onto receipts:** Nodes decode the post-execution transaction and expose the corresponding per-transaction rebate in RPC receipts.
+
+This gas model uses an additional refund mechanism alongside the canonical EVM gas used, where the sequencer can account for execution-local effects that the static gas schedule does not express. In the current PoC, this is realized as a refund-only model:
+
+$$
+\text{effective gas}=\text{gas used}-\text{sequencer-defined refund}
+$$
+
+There is also a block-level cap on the **total permissible gas used before rebates**. This is the hard limit on the sum of the raw `gas used` values of transactions that the sequencer is allowed to include in a block, before subtracting any SDM rebates. In production, this limit will be set to `MAX_GAS_LIMIT` from `L1.SystemConfig`.
+
+Given the **core mechanism**, we get the following effects:
+
+1. **Existing UX preserved:** Users still sign standard transactions specifying the maximum they are willing to pay.
+2. **Minimal execution changes:** The EVM gas schedule remains unchanged; SDM is carried as additional sequencer-produced accounting.
+3. **Inspectable accounting:** The rebate data is explicit in-block and can be projected onto receipts and replayed later.
+
+### Design Principles
+
+The presented design has been thought out based on the following principles:
+
+- Preserve existing UX with respect to transaction signing.
+- Stick with a minimal set of changes.
+- Keep the first policy narrow and inspectable.
+- Make the sequencer’s accounting externally replayable.
+
+### Architecture and PoC
+
+The current PoC implements SDM as **block-level warming** in `op-reth`. The PoC is implemented in branch `nonsense/opgas-reth-poc2-tracer`, and you can find an overview document specifically for it at: [SDM PoC 2 Review document](https://www.notion.so/oplabs/SDM-PoC-2-Review-330f153ee162801e8fc8e5e79c010e40)
+
+Within the current PoC the refunds are not applied to `gas used`, so that blocks are replayable and don't modify consensus fields, such as the `state root` during benchmarking with real-world chain data.
+
+In the production implementation of SDM `gas used` will include the refund and will not be equal to the canonical EVM gas used.
+
+#### Refund accumulation
+
+During block execution, the executor maintains persistent warming state across transactions in the block. When a later user transaction accesses state that was already warmed earlier in the block, a rebate is accumulated.
+
+The current PoC uses rebate values that come directly from [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929):
+
+- **warm account:** 2500 gas refund
+- **warm storage read:** 2000 gas refund
+- **warm storage write:** 2100 gas refund
+
+Deposits may warm state for later transactions, but do not themselves receive rebates.
+
+These values match the cold -> warm access deltas used by tx-level warming.
+
+#### Propagating calculated refunds to block assembly
+
+After execution, per-transaction rebate entries are accumulated as `(tx_index, gas_refund)` pairs. During payload building, these entries are encoded into a system OP transaction of type `0x7D`, namely `TxPostExec`.
+
+The post-execution transaction payload is RLP encoded and currently contains:
+
+- a `nonce` to ensure uniqueness,
+- a `version`, and
+- a list of refund `entries`, where each entry contains:
+  - `index`: the original transaction index in the block
+  - `gas_refund`: the sequencer-computed rebate
+
+#### Receipt / RPC projection
+
+Nodes inspect the block for the system post-execution transaction, decode its payload, and project the refund for each transaction into the RPC receipt as **`opGasRefund`**.
+
+#### Replay and inspection
+
+The PoC also includes replay and inspection tooling that can:
+
+- counterfactually replay a historical block with SDM accounting,
+- synthesize the expected SDM payload,
+- compare replay results against the embedded payload and receipt-level refund projection,
+- attribute each refund to the earlier transaction that first warmed the account or storage slot.
+
+#### Benchmarking: How effective the PoC is
+
+Benchmarking document for PoC 2 can be found at: [SDM Benchmarking for PoC 2](https://www.notion.so/oplabs/SDM-Benchmarking-for-PoC-2-Block-Level-Warming-330f153ee16280d4b2effdf0b3b53eb7)
+
+#### Integration with Flashblocks
+
+A concrete design option for integration of SDM with Flashblocks is to preserve a single final post-execution transaction in the sealed block.
+
+The core idea is:
+
+- `op-rbuilder` computes and streams **refund-aware prefixes** on every flashblock
+- each prefix includes the **current candidate `TxSDM`** for the whole prefix
+- `rollup-boost` no longer blindly concatenates all transactions, but it treats `TxPostExec` as a **reserved mutable slot** and always keeps only the **latest** `TxPostExec`
+
+This results in:
+
+- exactly one `TxPostExec` in the final block
+- per-prefix cumulative fields (`gas_used`, `state_root`, `receipts_root`, `block_hash`) already computed correctly by the builder
+- no need for `rollup-boost` to execute or recompute roots
+
+This is a viable design, but it is **not append-only** in the current strict sense. It introduces a new rule:
+
+> All user transactions are append-only across prefixes, but the reserved `TxPostExec` position is mutable and may be replaced by later prefixes.
+
+## Note on Sequencer Trust
+
+The current PoC makes the SDM accounting explicit in-block and externally replayable.
+
+## Security Considerations
+
+Considering that the functionality to issue refunds by the sequencer increases trust in the OP chain sequencer, it should be possible to encode and program a stop-gap such that, in the event of a chain incident, the L2 chain operator could turn off the refund mechanism on-demand and revert back to plain EVM gas accounting.
+
+## Alternative Policies Considered
+
+### Wall-clock execution time
+
+An alternative SDM policy is to price transactions based on wall-clock execution time. This would let the sequencer align fees more directly with observed execution cost, rather than with specific execution effects such as block-level warming.
+
+This may ultimately address a broader class of mispricing issues, but it also introduces more design complexity around measurement, calibration, reproducibility, and trust. The current proposal starts with block-warming because it is narrower, easier to inspect, and easier to replay.
--- a/protocol/sdm.md
+++ b/protocol/sdm.md
@@ -15,7 +15,7 @@ Allow the sequencer of OP Stack chains to influence the fee paid on a per-transa
 
 Introduce SDM as a general mechanism for subjective sequencer-defined gas refunds alongside `gas used` and `gas limit`. SDM adds a new system post-execution transaction, with type `0x7D`, which contains post-execution refund data injected by the sequencer.
 
-The first application of SDM is the **block-level warming rebates** policy: it allows transactions to benefit from already warmed accounts and storage slots from earlier in the same block by a sequencer-defined rebate, as described in [EIP-7863](https://eips.ethereum.org/EIPS/eip-7863). This is the initial go-to-market policy; future SDM policies may issue different kinds of refunds. Users still sign transactions in the usual way for the worst case, while the sequencer may charge less when a given SDM policy determines that a refund is appropriate.
+The first application of SDM is the **block-level warming rebates** policy: it allows transactions to benefit from already warmed accounts and storage slots from earlier in the same block by a sequencer-defined rebate, as described in [EIP-7863](https://eips.ethereum.org/EIPS/eip-7863). This is the initial go-to-market policy; future SDM policies may issue different kinds of refunds. Users still sign transactions in the usual way, while the sequencer rebates according to its own off-chain SDM policies.
 
 ## Problem Statement + Context
 
--- a/protocol/sdm.md
+++ b/protocol/sdm.md
@@ -42,10 +42,15 @@ Under this policy:
 
 ### Core Mechanism
 
-1. **Track warming across the block:** During block execution, the sequencer tracks which accounts and storage slots have already been warmed by earlier transactions.
-2. **Compute per-transaction rebate:** For each later non-deposit transaction, if an accessed account or storage slot was already warmed earlier in the block, the sequencer accumulates a rebate.
-3. **Serialize the rebate data in-block:** At the end of block building, the sequencer emits a system post-execution transaction, with type `0x7D`, containing the refund entries.
-4. **Project rebates onto receipts:** Nodes decode the post-execution transaction and expose the corresponding per-transaction rebate in RPC receipts.
+SDM is the **general mechanism** by which the sequencer can attach a subjective, per-transaction refund to normal transaction execution. It should be thought of as a framework for sequencer-defined refunds, not as a synonym for any one refund policy.
+
+At a high level, SDM works as follows:
+
+1. **Execute transactions normally:** User and deposit transactions execute under the standard EVM gas schedule.
+2. **Compute policy-defined refunds off-chain:** The sequencer applies one or more SDM policies to determine whether any transaction should receive a subjective refund, and in what amount.
+3. **Serialize refund metadata in-block:** At the end of block building, the sequencer emits a system post-execution transaction, with type `0x7D`, containing the refund entries.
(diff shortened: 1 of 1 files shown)


protocol/opgas.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Core Mechanism
+
+1. **Set the block gas limit to the worst-case EVM gas-per-millisecond throughput $G$:** Calculate the gas per millisecond that the chain can safely handle assuming every transaction is maximally expensive. This becomes the block gas limit.
+2. **Execute transactions and measure wall-clock time:** The sequencer runs each transaction and records how long it actually takes to execute. For a transaction that executes with gas used $g$ and time $t$, adjust its gas used proportional to the worst-case throughput $G$, capped at the signed transaction gas limit.
+3. **Rebate proportional to performance:** If a transaction executes faster than the worst-case assumption, the sequencer rebates gas proportionally. A transaction that runs in 10% of the expected time gets 90% of its gas rebated.
```
ajsutton: This likely leads to blocks that are unprovable because the execution of blocks in a fault proof or ZK VM will exhibit very different performance. Especially with precompiles, the sequencer may be able to execute something hundreds or thousands of times faster than a proofs/ZK VM.

If the gas _cost_ is refunded proofs will be fine, but if opgas allows the overall block to use more gas than it would have otherwise, it can be a significant problem for provability.

ajsutton (commented):
Having an out-of-consensus method to reduce gas costs makes sense and allows much faster iteration of approaches, but I'd say that wall clock time is far too simplistic and opens up a number of potentially dangerous attack scenarios both for nodes and for proofs systems.

protocol/opgas.md
```diff
@@ -0,0 +1,124 @@
...
+
+- Preserve existing UX with respect to transaction signing.
+- Stick with a minimal set of changes.
+- EIP-1559 continues to function.
+
+### Architecture and PoC
+
+#### Refund accumulation
+
+During transaction execution, `op-geth` already contains code for *Refund Accumulation*. When a storage slot is cleared (set from non-zero to zero), refunds are added to the refund counter in core/vm/gas_table.go. At the end of transaction execution, `op-geth` already contains code for *Refund Application* at `core/state_transition.go`.
```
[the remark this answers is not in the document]
nonsense: Good callout. Refunds are capped at 20% at the moment according to https://eips.ethereum.org/EIPS/eip-3529 and https://github.com/ethereum-optimism/op-geth/blob/optimism/core/state_transition.go#L776-L793

protocol/sdm.md
```diff
@@ -0,0 +1,157 @@
...
+In the production implementation of SDM `gas used` will include the refund and will not be equal to the canonical EVM gas used.
+
+#### Refund accumulation
+
+During block execution, the executor maintains persistent warming state across transactions in the block. When a later non-deposit, non-SDM transaction accesses state that was already warmed earlier in the block, a rebate is accumulated.
+
+The current PoC uses the following rebate values:
+- **warm account:** 2500 gas
+- **warm storage read:** 2000 gas
+- **warm storage write:** 2100 gas
```
joshklop: How does this relate to [EIP-7863](https://eips.ethereum.org/EIPS/eip-7863)?

protocol/sdm.md
```diff
@@ -0,0 +1,157 @@
...
+
+#### Propagating calculated refunds to block assembly
+
+After execution, per-transaction rebate entries are accumulated as `(tx_index, gas_refund)` pairs. During payload building, these entries are encoded into a synthetic OP transaction of type `0x7d` - namely `TxSDM`.
+
+The SDM payload currently contains:
+- a `version`, and
+- a list of refund `entries`, where each entry contains:
+  - `index`: the original transaction index in the block
+  - `gas_refund`: the sequencer-computed rebate
```
joshklop: Should it have some kind of nonce to ensure `TxSDM` hashes are unique? This is important for syncing the unsafe chain and the existing RPC APIs.

nonsense: Good catch. We do need some kind of uniqueness for the TxSDM, we could use the Deposit transaction `source_hash` as part of it.

protocol/sdm.md
```diff
@@ -0,0 +1,157 @@
...
+| Need Approval From | *TBD* |
+| Status | *Draft* |
+
+## Purpose
+
+Allow the sequencer of OP Stack chains to influence the fee paid on a per-transaction basis, aligning fees with the actual performance characteristics of each transaction, rather than rely solely on the static, in-protocol EVM gas model.
+
+## Summary
+
+Introduce a subjective sequencer-defined gas refund alongside `gas used` and `gas limit`. In the initial policy, SDM is implemented as **block-warming**: transactions that benefit from accounts or storage slots already warmed earlier in the same block receive a sequencer-defined rebate. This is the initial go-to-market policy only; future SDM policies may issue different kinds of refunds. Users still sign transactions in the usual way for the worst case, while the sequencer may charge less when execution benefits from block-local access reuse.
```
sebastianst: I'd add a 1-2 sentence summary of the implementation design to the Summary section (new post-exec tx type, pre-rebate total gas-used limit). And maybe rephrase block-warming as the motivation for implementing this feature and its first application.

protocol/sdm.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Proposed Solution
+
+The proposed solution is to introduce a sequencer-defined refund alongside **gas used** (EVM gas used) and **gas limit**, at the expense of putting more trust in the sequencer of a given OP Stack chain. In the initial policy, the sequencer computes block-warming rebates off-chain and communicates them in-block.
+
+Under this policy:
+- normal EVM execution remains unchanged;
+- users continue to sign standard transactions with a normal gas limit;
+- the sequencer computes a per-transaction rebate when a transaction benefits from accounts or storage slots already warmed earlier in the block;
+- the rebate is exposed as SDM metadata and projected onto receipts as `opGasRefund`.
```
sebastianst: But the rebates are still stored in a new `PostExecData` tx type? And the new receipt field is only informational?

I'd mention here that the rebates are stored in a new post exec tx type rather than the receipts (which doesn't work because they're not part of exec payloads so are not available for block execution). I was confused at first reading this and the post exec tx is more important than the out-of-protocol receipt field addition, which just happens on the rpc layer.

protocol/sdm.md
nonsense: I added a bit more explanation, but the initial policy is literally an implementation of EIP-7863.

protocol/sdm.md
```diff
@@ -0,0 +1,163 @@
...
+- Preserve existing UX with respect to transaction signing.
+- Stick with a minimal set of changes.
+- Keep the first policy narrow and inspectable.
+- Make the sequencer’s accounting externally replayable.
+
+### Architecture and PoC
+
+The current PoC implements SDM as **block-level warming** in `op-reth`. The PoC is implemented in branch `nonsense/opgas-reth-poc2-tracer`, and you can find an overview document specifically for it at: [SDM PoC 2 Review document](https://www.notion.so/oplabs/SDM-PoC-2-Review-330f153ee162801e8fc8e5e79c010e40)
+
+Within the current PoC the refunds are not applied to `gas used`, so that blocks are replayable and don't modify consensus fields, such as the `state root` during benchmarking with real-world chain data.
```
sebastianst, line 84: Why was this excluded from the PoC? It would have been interesting to inform the required lift of this from the PoC.

protocol/sdm.md
```diff
@@ -0,0 +1,163 @@
...
+
+During block execution, the executor maintains persistent warming state across transactions in the block. When a later non-deposit, non-SDM transaction accesses state that was already warmed earlier in the block, a rebate is accumulated.
+
+The current PoC uses rebate values that come directly from [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929):
+
+- **warm account:** 2500 gas refund
+- **warm storage read:** 2000 gas refund
+- **warm storage write:** 2100 gas refund
+
+Deposits may warm state for later transactions, but do not themselves receive rebates.
```
sebastianst, line 98: that makes sense 👍🏻

protocol/sdm.md
```diff
@@ -0,0 +1,163 @@
...
+
+These values match the cold -> warm access deltas used by tx-level warming.
+
+#### Propagating calculated refunds to block assembly
+
+After execution, per-transaction rebate entries are accumulated as `(tx_index, gas_refund)` pairs. During payload building, these entries are encoded into a synthetic OP transaction of type `0x7d` - namely `TxSDM`.
+
+The SDM payload currently contains:
+
+- a `hash` to ensure uniqueness,
```
sebastianst: a tx itself always has a hash, so this hash is superfluous.

protocol/sdm.md
```diff
@@ -0,0 +1,163 @@
...
+
+Deposits may warm state for later transactions, but do not themselves receive rebates.
+
+These values match the cold -> warm access deltas used by tx-level warming.
+
+#### Propagating calculated refunds to block assembly
+
+After execution, per-transaction rebate entries are accumulated as `(tx_index, gas_refund)` pairs. During payload building, these entries are encoded into a synthetic OP transaction of type `0x7d` - namely `TxSDM`.
+
+The SDM payload currently contains:
```
sebastianst: Is it RLP-encoded? In any case, would be good to mention the encoding format here.

protocol/sdm.md
```diff
@@ -0,0 +1,163 @@
...
+
+- a `hash` to ensure uniqueness,
+- a `version`, and
+- a list of refund `entries`, where each entry contains:
+  - `index`: the original transaction index in the block
+  - `gas_refund`: the sequencer-computed rebate
+
+#### Receipt / RPC projection
+
+Nodes inspect the block for the synthetic SDM transaction, decode its payload, and project the refund for each transaction into the RPC receipt as **`opGasRefund`**.
```
sebastianst: I think this should be the only place where the new receipt field is mentioned. It's just informational and out-of-protocol anyways and its references confused me in the earlier sections, as if they're made consensus receipt fields.

protocol/sdm.md
```diff
@@ -0,0 +1,163 @@
...
+
+> All user transactions are append-only across prefixes, but the reserved `TxSDM` position is mutable and may be replaced by later prefixes.
+
+## Note on Sequencer Trust
+
+The current PoC makes the sequencer’s SDM accounting explicit in-block and externally replayable.
+
+## Security Considerations
+
+Considering that the functionality to issue refunds by the sequencer increases trust in the OP chain sequencer, it should be possible to encode and program a stop-gap such that, in the event of a chain incident, the L2 chain operator could turn off the refund mechanism on-demand and revert back to plain EVM gas accounting.
```
sebastianst, line 157: But isn't the chain operator the same actor?
Could we say, the stop gap is to just not include the final `TxSDM`, so we keep it as optional, and if it's absent, rebates are simply not applied.

protocol/sdm.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Alternatives Considered
+
+### Wall-clock execution time
+
+An alternative SDM policy is to price transactions based on wall-clock execution time. This would let the sequencer align fees more directly with observed execution cost, rather than with specific execution effects such as block-level warming.
+
+This may ultimately address a broader class of mispricing issues, but it also introduces more design complexity around measurement, calibration, reproducibility, and trust. The current proposal starts with block-warming because it is narrower, easier to inspect, and easier to replay.
```
sebastianst: I'd rename this to "Other applications". Wall-clock based rebates are just another application of the same SDM feature.

protocol/sdm.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Transaction Timeout with Reversion
+
+Allow the sequencer to revert transactions that exceed their resource budget. This accomplishes similar goals but grants additional power to the sequencer (arbitrary reversion capability), which may raise concerns. The rebate approach is more conservative in terms of sequencer authority.
```
sebastianst: Also, nobody would benefit. The resources have already been consumed, but now the tx is still reverted. Loose-loose.

protocol/sdm.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Open questions
+
+1. Should future SDM policies remain refund-only, or support direct repricing?
+2. How should SDM interact with the EIP-3529 gas-refund cap? Today, gas refunds are capped at 20%, but this is not visible in the current PoC because it only calculates refunds without actually applying the cap. Should SDM preserve the cap, or disable it? For an initial SDM implementation, that safesty option is to keep EIP-3529 enabled, and cap refunds.
```
sebastianst: I still believe we can make rebates work with the other EIPs mentioned here. E.g. if there's no rebate, apply the other limits as usual, but if there is, allow the rebate to get larger.
We still need to define a total per-block gas-used-before-rebates limit, e.g. `gas_limit * 2`, so a sequencer can't shove an unlimited amount of txs into a block. This becomes a new effective EVM-gas block gas limit. This is an important security feature that should be called out.

protocol/sdm.md
```diff
@@ -0,0 +1,163 @@
...
+
+This is a viable design, but it is **not append-only** in the current strict sense. It introduces a new rule:
+
+> All user transactions are append-only across prefixes, but the reserved `TxSDM` position is mutable and may be replaced by later prefixes.
+
+## Note on Sequencer Trust
+
+The current PoC makes the sequencer’s SDM accounting explicit in-block and externally replayable.
+
+## Security Considerations
```
sebastianst, line 155: Another concern is that the sequencer can potentially allow many more txs into a block than before. This can probably be mitigated by still limiting the total pre-rebate gas-used, see other comment.

protocol/sdm.md
```diff
@@ -0,0 +1,163 @@
...
+- **warm storage read:** 2000 gas refund
+- **warm storage write:** 2100 gas refund
+
+Deposits may warm state for later transactions, but do not themselves receive rebates.
+
+These values match the cold -> warm access deltas used by tx-level warming.
+
+#### Propagating calculated refunds to block assembly
+
+After execution, per-transaction rebate entries are accumulated as `(tx_index, gas_refund)` pairs. During payload building, these entries are encoded into a synthetic OP transaction of type `0x7d` - namely `TxSDM`.
```
sebastianst: Since we add versioning to this new tx type's payload, I'd prefer we call this something more generic in the prod implementation, like `PostExecDataTx`.

sebastianst (commented):
I would find it clearer to read if SDM is first described generally and then block warming is only described in a later section. It's just one application of the general feature.

protocol/sdm.md
nonsense: you're right about the `tx hash`, i should rename this - i meant it for uniqueness. we do need a uniqueness seed within TxSDM.

protocol/sdm.md
nonsense: It is easier to replay historical blocks and get benchmarks if you don't modify the state of every block. I already have a branch on top of the PoC which implements the actual refund application onto `gas used`.

protocol/sdm.md
nonsense: >But isn't the chain operator the same actor?

Not sure what you mean with this.

>Could we say, the stop gap is to just not include the final TxSDM, so we keep it as optional, and if it's absent, rebates are simply not applied.

That's exactly how I plan on implementing this.

protocol/sdm.md
nonsense: Agree

protocol/sdm.md
nonsense: I removed this altogether, to keep the document as short and on point as possible.

protocol/sdm.md
nonsense: I think this is the case for every transaction, and what one would expect in an Ethereum / OP node, no?

protocol/sdm.md
nonsense: Included another line which explicitly says that rebates are stored in a new system tx.

protocol/sdm.md
joshklop: We can use [this](https://github.com/ethereum-optimism/optimism/blob/7fd04488d9e9558ff0dc78fb0ba03460301f78db/packages/contracts-bedrock/src/L1/SystemConfig.sol#L101-L104).

protocol/opgas.md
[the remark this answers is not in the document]
nonsense: We will track the existing EVM gas refunds separately to SDM refunds, so that we don't have a cap on SDM refunds.

protocol/sdm.md
sebastianst: It could also be tightly packed, or SSZ or whatever. So explicitly mentioning the encoding type always makes sense.

protocol/sdm.md
```diff
@@ -0,0 +1,160 @@
...
+
+These values match the cold -> warm access deltas used by tx-level warming.
+
+#### Propagating calculated refunds to block assembly
+
+After execution, per-transaction rebate entries are accumulated as `(tx_index, gas_refund)` pairs. During payload building, these entries are encoded into a system OP transaction of type `0x7d` - namely `TxSDM`.
+
+The post-execution transaction payload currently contains:
+
+- a `nonce` to ensure uniqueness,
```
joshklop, line 108: How is the `nonce` determined? Can it just be the L2 block number?

protocol/sdm.md
```diff
@@ -0,0 +1,160 @@
...
+- **warm storage read:** 2000 gas refund
+- **warm storage write:** 2100 gas refund
+
+Deposits may warm state for later transactions, but do not themselves receive rebates.
+
+These values match the cold -> warm access deltas used by tx-level warming.
+
+#### Propagating calculated refunds to block assembly
+
+After execution, per-transaction rebate entries are accumulated as `(tx_index, gas_refund)` pairs. During payload building, these entries are encoded into a system OP transaction of type `0x7d` - namely `TxSDM`.
```
trianglesphere: I suspect this is low risk, but are there any other chains that use 0x7d as the EIP 2718 tx type.

protocol/sdm.md
sebastianst: I see, ~so that rebates can't be replayed~ edit: it's so that the tx hash is unique. I'd rather call it a `nonce` or so then. There's also the argument that the block hash commits to the transaction root hash, which commits also to this tx's hash. But still may be better to include a nonce. Need to think if it's critical that the tx hash is unique...

protocol/sdm.md
sebastianst: Good idea

protocol/sdm.md
```diff
@@ -0,0 +1,160 @@
...
+- counterfactually replay a historical block with SDM accounting,
+- synthesize the expected SDM payload,
+- compare replay results against the embedded payload and receipt-level `opGasRefund`,
+- attribute each refund to the earlier transaction that first warmed the account or storage slot.
+
+#### Benchmarking: How effective the PoC is
+
+Benchmarking document for PoC 2 can be found at: [SDM Benchmarking for PoC 2](https://www.notion.so/oplabs/SDM-Benchmarking-for-PoC-2-Block-Level-Warming-330f153ee16280d4b2effdf0b3b53eb7)
+
+#### Integration with Flashblocks
```
trianglesphere, line 131: After discussing, we may want to move the refund data to receipt field or into a metadata field and then only include the `TxSDM` at the end when the sequencer requests the full block from op-node

protocol/sdm.md
sebastianst: We choose this as `DepositTxType - 1`, I think that makes semantically sense since it's the second special OP transaction type, and we don't want to count upwards because `0x7F` is already the last one allowed.

protocol/sdm.md
nonsense: do we have example somewhere where a payload in a tx is not RLP encoded? My understanding is that RLP is used virtually everywhere on the execution layer, and SSZ is used on the beacon chain.

protocol/sdm.md
sebastianst: That could also work. I'm still wondering if we really need uniqueness of the `PostExec` tx. It's committed to via the tx root hash, so maybe it doesn't matter and we could just not have any nonce/seed.

protocol/sdm.md
sebastianst: OK let's go with this limit for now. If we discover a compelling reason to tie the limit to the original block gas limit via a multiple or so, we can revisit this decision. We should talk about this during the failure mode analysis for this feature.

protocol/sdm.md
sebastianst: So the data field of regular txs for example doesn't enforce any encoding theoretically, it can be anything, just by convention it's most of the time rlp encoded. And in our other protocol specs we have a lot of bespoke encodings that are more efficient than RLP, like singular or span batches, or the data we pack into the `extraData`. It wouldn't be a bad idea to use a bespoke encoding for this maybe, to conserve space.

protocol/sdm.md
sebastianst: I meant that the sequencer and chain operator are the same actor, so it doesn't increase security for the chain operator to add this feature. May have misunderstood your sentence. In any case, I agree the stopgap makes sense, I'm in favor of allowing the new tx to be omitted. Another case for this is that there will be many chains who don't wanna do fee rebates at all, so for them this feature is totally transparent because there's no new tx.

protocol/sdm.md
joshklop: It will need to be query-able by hash.

protocol/sdm.md
nonsense: OK, adding a note that the post-execution transaction payload is RLP encoded for now, we can decide if we want to change it during the implementation phase.

protocol/sdm.md
nonsense: I added a note around the total permissible gas used before rebates.

protocol/sdm.md
nonsense: I think it is important for any tx to be uniquely identifiable, so we should definitely enforce uniqueness.

protocol/sdm.md
```diff
@@ -0,0 +1,157 @@
...
+## Proposed Solution
+
+The proposed solution is to introduce a sequencer-defined refund alongside **gas used** (EVM gas used) and **gas limit**, at the expense of putting more trust in the sequencer of a given OP Stack chain. In the initial policy, the sequencer computes block-warming rebates off-chain and communicates them in-block.
+
+Under this policy:
+
+- normal EVM execution remains unchanged;
+- users continue to sign standard transactions with a normal gas limit;
+- the sequencer computes a per-transaction rebate when a transaction benefits from accounts or storage slots already warmed earlier in the block;
+- the sequencer includes a new system post-execution transaction, with type `0x7D`, into each block. This transaction contains all rebates for all transactions in the block;
```
sebastianst: between those two points it should be mentioned that the rebate is actually applied during the state transition and the user is refunded that rebate.

in the specs we also need to formalize how those rebates interplay with other refunds from the evm exec.

protocol/sdm.md
```diff
@@ -0,0 +1,157 @@
...
+- the rebate is exposed as metadata and projected onto receipts for each transaction.
+
+### Core Mechanism
+
+1. **Track warming across the block:** During block execution, the sequencer tracks which accounts and storage slots have already been warmed by earlier transactions.
+2. **Compute per-transaction rebate:** For each later non-deposit transaction, if an accessed account or storage slot was already warmed earlier in the block, the sequencer accumulates a rebate.
+3. **Serialize the rebate data in-block:** At the end of block building, the sequencer emits a system post-execution transaction, with type `0x7D`, containing the refund entries.
+4. **Project rebates onto receipts:** Nodes decode the post-execution transaction and expose the corresponding per-transaction rebate in RPC receipts.
```
sebastianst: I still find this confusing to read. It conflates the general SDM/rebates protocol with the specific application of block level slot warming.

protocol/sdm.md
sebastianst: I'm still not convinced that it is a hard requirement for txs to have unique hashes across blocks. We're only using the PostExec tx as a data container, so if we can remove the nonce from the design, we should do it. I can't see a security-related reason why it must be there. We don't need to solve this issue for this design doc, but I'd like to see a convincing proof that it needs to be unique when we implement this.

> It will need to be query-able by hash.
It still would be. The tx hash would resolve to a unique PostExec tx. That this tx may be present in multiple blocks wouldn't matter. Note that txs are generally not tied to a block.

protocol/sdm.md
nonsense: done!

protocol/sdm.md
nonsense: I updated the "Core Mechanism" section to clearly separate between the SDM protocol/framework and block-level warming policy.

protocol/sdm.md
sebastianst: There actually is an argument for unique tx hashes: The [eth_getTransactionReceipt](https://ethereum.org/developers/docs/apis/json-rpc/#eth_gettransactionreceipt) rpc resolves a tx hash to its receipt, and the receipt is specific about its block.

So if we generate receipts for this new tx type, we need to give it a nonce. In that case I'd suggest we use the block number (less data than hash), so it's tied to the block and we don't require a state lookup to confirm it's unique.

protocol/sdm.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Purpose
+
+Allow the sequencer of OP Stack chains to influence the fee paid on a per-transaction basis, aligning fees with the actual performance characteristics of each transaction, rather than rely solely on the static, in-protocol EVM gas model.
+
+## Summary
+
+Introduce SDM as a general mechanism for subjective sequencer-defined gas refunds alongside `gas used` and `gas limit`. SDM adds a new system post-execution transaction, with type `0x7D`, which contains post-execution refund data injected by the sequencer.
+
+The first application of SDM is the **block-level warming rebates** policy: it allows transactions to benefit from already warmed accounts and storage slots from earlier in the same block by a sequencer-defined rebate, as described in [EIP-7863](https://eips.ethereum.org/EIPS/eip-7863). This is the initial go-to-market policy; future SDM policies may issue different kinds of refunds. Users still sign transactions in the usual way for the worst case, while the sequencer may charge less when a given SDM policy determines that a refund is appropriate.
```
joshklop: nit: edits for clarity.
```suggestion
The first application of SDM is the **block-level warming rebates** policy: it allows transactions to benefit from already warmed accounts and storage slots from earlier in the same block by a sequencer-defined rebate, as described in [EIP-7863](https://eips.ethereum.org/EIPS/eip-7863). This is the initial go-to-market policy; future SDM policies may issue different kinds of refunds. Users still sign transactions in the usual way, while the sequencer rebates according to its own off-chain SDM policies.
```

protocol/sdm.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Problem Statement + Context
+
+The EVM gas model is a static, in-protocol approximation of resource consumption. It does not capture all execution characteristics that matter operationally to an OP Stack sequencer.
+
+The sequencer currently cannot reflect execution efficiencies and inefficiencies in fees, even when those are material.
+
+The goal of SDM is to let the sequencer account for such effects explicitly through subjective sequencer-defined refunds, starting with a narrow and observable policy: **block-level warming rebates** - an implementation of [EIP-7863](https://eips.ethereum.org/EIPS/eip-7863).
```
joshklop, line 26:
```suggestion
The goal of SDM is to let the sequencer account for such effects explicitly through sequencer-defined  (i.e., off-chain) refunds, starting with a narrow and observable policy: **block-level warming rebates** - an implementation of [EIP-7863](https://eips.ethereum.org/EIPS/eip-7863).
```

protocol/sdm.md
```diff
@@ -0,0 +1,165 @@
...
+Under SDM generally:
+
+- normal EVM execution remains unchanged;
+- users continue to sign standard transactions with a normal gas limit;
+- the sequencer computes a per-transaction refund according to one or more SDM policies;
+- the sequencer includes a new system post-execution transaction, with type `0x7D`, into each block. This transaction contains all refunds for all transactions in the block;
+- the refund is applied during the state transition, and the user is refunded that amount by having the `gas used` of their transaction reduced by the sequencer;
+- the refund is exposed as metadata and projected onto receipts for each transaction.
+
+For the initial policy, the sequencer computes **block-level warming rebates** off-chain and communicates them in-block.
```
joshklop, line 43: nit: removing some detail for clarity
```suggestion
For the initial policy, the sequencer computes **block-level warming rebates** off-chain.
```

protocol/sdm.md
joshklop: @nonsense correct me if you meant something different, but I think the idea is that an SDM policy could introduce a dos vector. The chain operator may want to disable rebates in an emergency. Because the sequencer controls rebates, verifiers have no way to disable the feature.

@sebastianst does this align with how you're thinking about it?

joshklop (approved):
LGTM! Thanks Anton. I've left a few non-blocking comments.

protocol/sdm.md
nonsense: I'd rather we keep this sentence -- the sequencer must communicate the rebates, so that verifiers can validate the block.

protocol/sdm.md
nonsense: I think the current version is better, because `subjective` here means `not necessarily deterministic`, and I don't like `off-chain` because the refunds do go onchain.

nonsense:
> @nonsense correct me if you meant something different, but I think the idea is that an SDM policy could introduce a dos vector. The chain operator may want to disable rebates in an emergency. Because the sequencer controls rebates, verifiers have no way to disable the feature.

@joshklop yes, SDM policy could have unintended consequences, not just a DoS vector, therefore the chain operator may want to disable rebates in an emergency. Verifiers have no way of disabling the feature.

PR #369, opened 2026-03-17
feat: fma for zk dispute game

0xOneTony:
## Summary
  - Add FMA document for the ZK Dispute Game covering 8 failure modes
  - Includes risk assessments, mitigations, detection strategies, recovery paths, and audit requirements.
  - Consolidated action items table with 30 items across all failure modes.

  ## References
  - [ZK Dispute Game Design Doc](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/zk-dispute-game.md)
  - [ZK Dispute Game Specs](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk)

### DIFF
--- a/security/fma-zk-dispute-game.md
+++ b/security/fma-zk-dispute-game.md
@@ -48,7 +48,7 @@ Below are references for this project:
 - **Mitigations:**
     1. `IZKVerifier` interface enables verifier swap without redeploying the game.
     2. `prove()` constructs `publicValues` from immutable on-chain state — only `_proofBytes` is caller-supplied.
-    3. Verifier contract should be independently audited by multiple firms.
+    3. Verifier contract will be audited before mainnet deployment.
     4. `DISPUTE_GAME_FINALITY_DELAY_SECONDS` airgap + `DelayedWETH` delay give the Guardian two windows to intervene.
 - **Detection:**
     - Off-chain monitoring that re-executes L2 state transitions and alerts on `DEFENDER_WINS` for incorrect claims.
@@ -59,7 +59,6 @@ Below are references for this project:
     2. OPCM deploys patched verifier and updates `gameArgs`.
     3. For sub-case B: Guardian blacklists affected games (REFUND mode) while corrected verifier is deployed.
 - **Action Item(s):**
-    - [ ]  FM1: Verifier audited by at least two independent firms before mainnet.
     - [ ]  FM1: Off-chain monitoring that verifies every game's `rootClaim` against trusted L2 node.
     - [ ]  FM1: Guardian runbook for verifier compromise (pause/blacklist/retire sequence).
     - [ ]  FM1: Fuzz test `IZKVerifier.verify()` with malformed inputs to confirm it always reverts.
@@ -292,10 +291,9 @@ Below is a consolidated list of all action items from the failure modes above.
 
 | Action Item | Description | Source |
 | --- | --- | --- |
-| FM1-1 | Verifier audited by at least two independent firms before mainnet. | [FM1](#fm1-zk-verifier-soundness-or-completeness-break) |
-| FM1-2 | Off-chain monitoring that verifies every game's `rootClaim` against trusted L2 node. | [FM1](#fm1-zk-verifier-soundness-or-completeness-break) |
-| FM1-3 | Guardian runbook for verifier compromise (pause/blacklist/retire sequence). | [FM1](#fm1-zk-verifier-soundness-or-completeness-break) |
-| FM1-4 | Fuzz test `IZKVerifier.verify()` with malformed inputs to confirm it always reverts. | [FM1](#fm1-zk-verifier-soundness-or-completeness-break) |
+| FM1-1 | Off-chain monitoring that verifies every game's `rootClaim` against trusted L2 node. | [FM1](#fm1-zk-verifier-soundness-or-completeness-break) |
+| FM1-2 | Guardian runbook for verifier compromise (pause/blacklist/retire sequence). | [FM1](#fm1-zk-verifier-soundness-or-completeness-break) |
+| FM1-3 | Fuzz test `IZKVerifier.verify()` with malformed inputs to confirm it always reverts. | [FM1](#fm1-zk-verifier-soundness-or-completeness-break) |
 | FM2-1 | Implement monitoring that alerts when an unchallenged game has an incorrect `rootClaim` with time remaining in the challenge window. | [FM2](#fm2-unchallenged-fraudulent-proposal) |
 | FM2-2 | Ensure `maxChallengeDuration` accounts for L1 congestion/censorship and bond economics incentivize challengers. | [FM2](#fm2-unchallenged-fraudulent-proposal) |
 | FM2-3 | Document the Guardian fallback procedure for games that were not challenged in time. | [FM2](#fm2-unchallenged-fraudulent-proposal) |
--- a/security/fma-zk-dispute-game.md
+++ b/security/fma-zk-dispute-game.md
@@ -114,7 +114,7 @@ Below are references for this project:
     
 - **Risk Assessment:** Critical severity / Low likelihood
 - **Mitigations:**
-    1. Bond distribution is fully specified in the [game mechanics](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/game-mechanics.md#bond-distribution) with a clear table of every scenario.
+    1. Bond distribution is fully specified in the [game mechanics](https://github.com/ethereum-optimism/specs/blob/main/specs/fault-proof/stage-one/zk/game-mechanics.md#bond-distribution) with a clear table of every scenario.
     2. The `DelayedWETH` pattern is shared with `FaultDisputeGame` and has been audited in that context.
     3. REFUND mode returns each bond to its original depositor, which is always balanced by definition.
     4. Bond distribution is deterministic — no external inputs or oracle dependencies.
@@ -153,7 +153,7 @@ Below are references for this project:
 - **Description:** The MCP clone pattern packs 8 fields into `gameArgs` via `abi.encodePacked`. A mismatch between `OPContractsManager._makeGameArgs()` encoding and `ZKDisputeGame` CWIA offset decoding causes the game to read wrong values — wrong `verifier` address, wrong `challengerBond`, wrong durations, etc. Note: `l2SequenceNumber` type changed from `uint256` to `uint64` in `_extraData`, relevant to packing correctness.
 - **Risk Assessment:** Medium severity / Low likelihood
 - **Mitigations:**
-    1. The encoding is defined in the spec ([Game Args Layout](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/zk-dispute-game.md#game-args-layout)) and implemented in `OPContractsManager._makeGameArgs()`.
+    1. The encoding is defined in the spec ([Game Args Layout](https://github.com/ethereum-optimism/specs/blob/main/specs/fault-proof/stage-one/zk/zk-dispute-game.md#game-args-layout)) and implemented in `OPContractsManager._makeGameArgs()`.
     2. Round-trip tests that encode via `_makeGameArgs()` and decode via the game's accessor functions verify consistency.
     3. The CWIA pattern is used extensively in the existing OP Stack (e.g., `FaultDisputeGame`), so there is prior art and tooling for validating it.
 - **Detection:**
--- a/security/fma-zk-dispute-game.md
+++ b/security/fma-zk-dispute-game.md
@@ -30,10 +30,10 @@ This document covers the `ZKDisputeGame`, a single-round ZK-proof-based dispute 
 Below are references for this project:
 
 - [Design Doc](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/zk-dispute-game.md)
-- [ZK Dispute Game Spec](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/zk-dispute-game.md)
-- [Game Mechanics Spec](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/game-mechanics.md)
-- [IZKVerifier Interface Spec](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/zk-interface.md)
-- [ZK Fault Proof VM Spec](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/zk-fault-proof-vm.md)
+- [ZK Dispute Game Spec](https://github.com/ethereum-optimism/specs/blob/main/specs/fault-proof/stage-one/zk/zk-dispute-game.md)
+- [Game Mechanics Spec](https://github.com/ethereum-optimism/specs/blob/main/specs/fault-proof/stage-one/zk/game-mechanics.md)
+- [IZKVerifier Interface Spec](https://github.com/ethereum-optimism/specs/blob/main/specs/fault-proof/stage-one/zk/zk-interface.md)
+- [ZK Fault Proof VM Spec](https://github.com/ethereum-optimism/specs/blob/main/specs/fault-proof/zk-fault-proof-vm.md)
 
 ## Failure Modes and Recovery Paths
 
--- a/security/fma-zk-dispute-game.md
+++ b/security/fma-zk-dispute-game.md
@@ -8,7 +8,7 @@
     - [FM2: Unchallenged Fraudulent Proposal](#fm2-unchallenged-fraudulent-proposal)
     - [FM3: Missed Child Blacklist After Parent Invalidation](#fm3-missed-child-blacklist-after-parent-invalidation)
     - [FM4: Bond Accounting Failure and DelayedWETH Insolvency](#fm4-bond-accounting-failure-and-delayedweth-insolvency)
-    - [FM5: Upgrade Lifecycle Mismatch](#fm5-upgrade-lifecycle-mismatch)
+    - [FM5: Prestate Mismatch](#fm5-prestate-mismatch)
     - [FM6: CWIA Game Args Encoding Mismatch](#fm6-cwia-game-args-encoding-mismatch)
     - [FM7: Bond and Duration Misconfiguration](#fm7-bond-and-duration-misconfiguration)
     - [FM8: Self-Challenge Front-Running to Recover Bonds](#fm8-self-challenge-front-running-to-recover-bonds)
@@ -133,46 +133,23 @@ The spec requires (iZKG-011) that for every game: `sum(distributions) + sum(burn
     scenarios.
     - [ ]  FM4: Fuzz test bond accounting across randomized game lifecycles, including the burn
     path.
-    - [ ]  FM4: Add `DelayedWETH` balance monitoring alert.
 
 ---
 
-### FM5: Upgrade Lifecycle Mismatch
+### FM5: Prestate Mismatch
 
-- **Description:** Covers the full upgrade lifecycle (aZKG-002, [IZKVerifier upgrade path](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/zk-interface.md#verifier-upgrade-path)):
-    
-    **A -- absolutePrestate mismatch:** The `absolutePrestate` is updated but provers still run the old binary (or vice versa). Proofs fail verification; challenged games time out as `CHALLENGER_WINS` and proposers lose bonds. Also applies when governance delays mean the prestate isn't updated before a hard fork activates on L2.
-    
-    **B -- Verifier upgrade race:** The old verifier is deprecated while in-flight games still reference it via immutable CWIA args. Those games become unprovable.
-    
-    **C -- Aggressive retirement timestamp:** `updateRetirementTimestamp()` set too close to present forces in-flight games into REFUND mode unnecessarily.
-    
-    **D -- Cross-prestate parent chaining:** Parent games intentionally don't need the same `absolutePrestate` as children (avoids 7-day re-proofs on every update). But retiring old-prestate games doesn't cover children that chained off them with a different prestate — the Guardian must trace across prestate boundaries.
-    
-- **Risk Assessment:** High severity / Low-to-medium likelihood
+- **Description:** The `absolutePrestate` is updated on-chain but provers still run the old binary (or vice versa) (aZKG-002). Proofs fail verification; challenged games time out as `CHALLENGER_WINS` and proposers lose bonds.
+- **Risk Assessment:** Medium severity / Low likelihood
 - **Mitigations:**
-    1. In-progress games use old configuration immutably (CWIA args fixed at clone creation).
-    2. Upgrade sequencing: update `absolutePrestate` on-chain first, then distribute new binary to provers, then resume proposing. For hard forks, prestate update MUST precede activation.
-    3. Old verifier contracts MUST remain functional until all in-flight games resolve or are blacklisted.
-    4. Retirement timestamps should allow at least `maxChallengeDuration + maxProveDuration + DISPUTE_GAME_FINALITY_DELAY_SECONDS` lead time.
+    1. In-progress games use old configuration immutably (CWIA args fixed at clone creation). Off-chain software automatically selects the correct prestate based on the absolute prestate hash, supporting multiple prestates concurrently.
+    2. Verifier contracts are immutable — old verifiers remain functional indefinitely and cannot be deprecated.
 - **Detection:**
-    - Monitoring for `gameArgs` changes in the `DisputeGameFactory`.
     - Alerts when a challenged game fails to receive a proof within a reasonable time (indicating prover/prestate mismatch).
-    - Pre-deployment verification that the `absolutePrestate` matches the program binary hash.
-    - Monitoring for games referencing verifier addresses that are no longer functional.
-    - Alerts when retirement timestamp is set within the in-flight game window.
+    - Alerts when a `prove()` transaction for a valid state root reverts.
+    - `op-dispute-mon` already detects games that are forecast to or do resolve incorrectly.
 - **Recovery Path(s):**
-    1. If games were created with a mismatched `absolutePrestate`, they cannot be proven. If challenged, the proposer loses `initBond`. Guardian can blacklist these games to trigger REFUND mode instead.
-    2. Correct the `absolutePrestate` via OPCM and resume normal operation.
-    3. If a hard fork activated before the prestate update, Guardian retires the old game type and OPCM deploys with the correct prestate.
-    4. If the old verifier was made non-functional prematurely, Guardian blacklists affected games to trigger REFUND mode, then OPCM re-deploys the old verifier or migrates to a new one.
-- **Action Item(s):**
-    - [ ]  FM5: Document a program upgrade runbook that specifies the exact sequencing of `absolutePrestate` update, prover binary distribution, and proposer resumption.
-    - [ ]  FM5: For hard fork upgrades, the `absolutePrestate` update MUST be included in the governance proposal alongside the L2 activation timestamp, with the on-chain update applied before activation.
-    - [ ]  FM5: Add automated verification in the deployment pipeline that the `absolutePrestate` in `gameArgs` matches the hash of the deployed program binary.
-    - [ ]  FM5: Document the requirement that old verifier contracts must remain functional until all in-flight games resolve, and add monitoring for verifier liveness.
-    - [ ]  FM5: Establish a minimum retirement timestamp lead time policy and enforce it in Guardian tooling.
-    - [ ]  FM5: Document the cross-prestate parent chaining implications for retirement cascades in the Guardian runbook.
+    1. Fix the off-chain software configuration. Proposers lose `initBond` on any challenged games that couldn't be proven, but the system is otherwise unaffected.
+    2. Guardian can blacklist affected games to trigger REFUND mode if needed.
 
 ---
 
@@ -291,13 +268,6 @@ Below is a consolidated list of all action items from the failure modes above.
 | FM2-1 | Ensure `maxChallengeDuration` accounts for L1 congestion/censorship and bond economics incentivize challengers. | [FM2](#fm2-unchallenged-fraudulent-proposal) |
 | FM4-1 | Implement iZKG-011 conservation invariant tests across all bond distribution scenarios. | [FM4](#fm4-bond-accounting-failure-and-delayedweth-insolvency) |
 | FM4-2 | Fuzz test bond accounting across randomized game lifecycles, including the burn path. | [FM4](#fm4-bond-accounting-failure-and-delayedweth-insolvency) |
-| FM4-3 | Add `DelayedWETH` balance monitoring alert. | [FM4](#fm4-bond-accounting-failure-and-delayedweth-insolvency) |
-| FM5-1 | Document a program upgrade runbook that specifies the exact sequencing of `absolutePrestate` update, prover binary distribution, and proposer resumption. | [FM5](#fm5-upgrade-lifecycle-mismatch) |
-| FM5-2 | For hard fork upgrades, the `absolutePrestate` update MUST be included in the governance proposal alongside the L2 activation timestamp, with the on-chain update applied before activation. | [FM5](#fm5-upgrade-lifecycle-mismatch) |
-| FM5-3 | Add automated verification in the deployment pipeline that the `absolutePrestate` in `gameArgs` matches the hash of the deployed program binary. | [FM5](#fm5-upgrade-lifecycle-mismatch) |
-| FM5-4 | Document the requirement that old verifier contracts must remain functional until all in-flight games resolve, and add monitoring for verifier liveness. | [FM5](#fm5-upgrade-lifecycle-mismatch) |
-| FM5-5 | Establish a minimum retirement timestamp lead time policy and enforce it in Guardian tooling. | [FM5](#fm5-upgrade-lifecycle-mismatch) |
-| FM5-6 | Document the cross-prestate parent chaining implications for retirement cascades in the Guardian runbook. | [FM5](#fm5-upgrade-lifecycle-mismatch) |
 | FM6-1 | Implement comprehensive round-trip encoding/decoding tests for all 8 `gameArgs` fields, including edge-case values (zero, max, addresses with leading zeros). | [FM6](#fm6-cwia-game-args-encoding-mismatch) |
 | FM6-2 | Add fuzz tests that verify `_makeGameArgs()` output is correctly decoded by the game's accessor functions. | [FM6](#fm6-cwia-game-args-encoding-mismatch) |
 | FM6-3 | Review the `Duration` type's packed size and ensure it matches the offset calculations in the CWIA decoding logic. | [FM6](#fm6-cwia-game-args-encoding-mismatch) |
--- a/security/fma-zk-dispute-game.md
+++ b/security/fma-zk-dispute-game.md
@@ -55,7 +55,7 @@ Below are references for this project:
     - Fuzz testing verifier with edge-case inputs (empty bytes, zero values, gas-limited calls).
     - Monitoring for challenged games that never receive a valid proof (sub-case B indicator).
 - **Recovery Path(s):**
-    1. Guardian pauses system, blacklists affected games, retires game type if systemic.
+    1. Guardian pauses system, blacklists affected games, retires game type updating the retirement timestamp if systemic.
     2. OPCM deploys patched verifier and updates `gameArgs`.
     3. For sub-case B: Guardian blacklists affected games (REFUND mode) while corrected verifier is deployed.
 - **Action Item(s):**
@@ -122,7 +122,6 @@ Below are references for this project:
     5. If deadlock griefing is systematic, OPCM can increase `challengerBond` to raise the cost.
 - **Action Item(s):**
     - [ ]  FM3: Build tooling that, given a blacklisted game, automatically enumerates all descendant games by querying `DisputeGameFactory` for all `ZKDisputeGame` instances and filtering by `parentIndex`, including cross-prestate ancestry.
-    - [ ]  FM3: Consider adding an on-chain mechanism that automatically propagates blacklist status to descendants, or that checks ancestor blacklist status in `closeGame()`.
     - [ ]  FM3: Document the Guardian runbook for cascading blacklists, including cross-prestate tracing, tooling, and verification steps.
     - [ ]  FM3: Add monitoring that alerts when any game with a blacklisted ancestor approaches the finality delay window.
     - [ ]  FM3: Analyze the expected chain depth under normal operation and the maximum delay an attacker can impose per `challengerBond` spent via resolution ordering deadlock.
@@ -301,10 +300,9 @@ Below is a consolidated list of all action items from the failure modes above.
 | FM2-2 | Ensure `maxChallengeDuration` accounts for L1 congestion/censorship and bond economics incentivize challengers. | [FM2](#fm2-unchallenged-fraudulent-proposal) |
 | FM2-3 | Document the Guardian fallback procedure for games that were not challenged in time. | [FM2](#fm2-unchallenged-fraudulent-proposal) |
 | FM3-1 | Build tooling that, given a blacklisted game, automatically enumerates all descendant games by querying `DisputeGameFactory` for all `ZKDisputeGame` instances and filtering by `parentIndex`, including cross-prestate ancestry. | [FM3](#fm3-parent-chain-invalidation-and-propagation-failure) |
-| FM3-2 | Consider adding an on-chain mechanism that automatically propagates blacklist status to descendants, or that checks ancestor blacklist status in `closeGame()`. | [FM3](#fm3-parent-chain-invalidation-and-propagation-failure) |
-| FM3-3 | Document the Guardian runbook for cascading blacklists, including cross-prestate tracing, tooling, and verification steps. | [FM3](#fm3-parent-chain-invalidation-and-propagation-failure) |
-| FM3-4 | Add monitoring that alerts when any game with a blacklisted ancestor approaches the finality delay window. | [FM3](#fm3-parent-chain-invalidation-and-propagation-failure) |
-| FM3-5 | Analyze the expected chain depth under normal operation and the maximum delay an attacker can impose per `challengerBond` spent via resolution ordering deadlock. | [FM3](#fm3-parent-chain-invalidation-and-propagation-failure) |
+| FM3-2 | Document the Guardian runbook for cascading blacklists, including cross-prestate tracing, tooling, and verification steps. | [FM3](#fm3-parent-chain-invalidation-and-propagation-failure) |
+| FM3-3 | Add monitoring that alerts when any game with a blacklisted ancestor approaches the finality delay window. | [FM3](#fm3-parent-chain-invalidation-and-propagation-failure) |
+| FM3-4 | Analyze the expected chain depth under normal operation and the maximum delay an attacker can impose per `challengerBond` spent via resolution ordering deadlock. | [FM3](#fm3-parent-chain-invalidation-and-propagation-failure) |
 | FM4-1 | Implement iZKG-011 conservation invariant tests across all bond distribution scenarios. | [FM4](#fm4-bond-accounting-failure-and-delayedweth-insolvency) |
 | FM4-2 | Fuzz test bond accounting across randomized game lifecycles, including the burn path. | [FM4](#fm4-bond-accounting-failure-and-delayedweth-insolvency) |
 | FM4-3 | Add `DelayedWETH` balance monitoring alert. | [FM4](#fm4-bond-accounting-failure-and-delayedweth-insolvency) |
--- a/security/fma-zk-dispute-game.md
+++ b/security/fma-zk-dispute-game.md
@@ -183,13 +183,13 @@ The spec requires (iZKG-011) that for every game: `sum(distributions) + sum(burn
     - **Bonds too low:** Spam proposals and frivolous challenges are cheap. Challenging costs only `challengerBond`; defending requires expensive ZK proof generation. If proving cost > `challengerBond`, defending is economically irrational.
     - **Bonds too high:** Honest proposers and challengers priced out, undermining permissionless participation.
     - **Durations too short:** Provers may not generate the proof in time. L1 censorship of `prove()` also becomes feasible (cost proportional to `maxProveDuration`). Short `maxChallengeDuration` compounds FM2 risk.
-    - **Durations too long:** Withdrawal finality delayed, eliminating the latency benefit of ZK proofs.
+    - **Durations too long:** Withdrawal finality delayed on the unchallenged path. The latency benefit of ZK proofs is only realized if someone eagerly proves, which has its own cost.
     - **Block range selection:** The proposer is responsible for creating games with block ranges they can prove within `maxProveDuration`. Proving larger ranges in a single proof is generally cheaper per-block in zkVMs however a proposer who creates an unprovable range loses their own `initBond`.
     - **L1 gas pressure:** ZK proof verification can cost several hundred thousand gas units depending on the backend. During extreme gas spikes, `prove()` inclusion becomes expensive, not technically unfeasible though given the margin against the block gas limit.
 - **Risk Assessment:** Medium severity / Medium likelihood
 - **Mitigations:**
     1. All parameters are in `gameArgs` and tunable per chain by OPCM without redeploying.
-    2. `challengerBond` should exceed 2x expected proving cost so defending is always profitable.
+    2. `challengerBond` should exceed expected proving cost so defending is always profitable.
     3. Third-party provers can earn `challengerBond`, creating a market incentive for proving services.
     4. `maxProveDuration` must be long enough that L1 censorship of `prove()` is economically infeasible (aZKG-007).
     5. Rational proposers won't create unnecessarily large block ranges (aZKG-006).
@@ -197,10 +197,10 @@ The spec requires (iZKG-011) that for every game: `sum(distributions) + sum(burn
     - Monitoring challenge rates, proving costs vs `challengerBond`, and `prove()` inclusion failures.
     - Alerts when challenged games approach `maxProveDuration` without proof submission.
 - **Recovery Path(s):**
-    1. OPCM adjusts `challengerBond` and/or `maxProveDuration`. "Game type changed mid-play" ensures in-flight games get REFUND treatment.
-    2. Guardian can blacklist exploitative games or retire the game type.
+    1. OPCM adjusts `challengerBond` and/or `maxProveDuration` for new games. Existing in-flight games continue to play out with their existing rules.
+    2. Guardian can blacklist exploitative games, update the retirement timestamp to invalidate all in-flight games, or change the respected game type.
 - **Action Item(s):**
-    - [ ]  FM7: Establish a minimum `challengerBond` policy such that `challengerBond > 2x expected proving cost` to ensure defending is always profitable.
+    - [ ]  FM7: Establish a minimum `challengerBond` policy such that `challengerBond > expected proving cost` to ensure defending is always profitable.
     - [ ]  FM7: Calibrate `maxProveDuration` per aZKG-007 with analysis of L1 censorship costs and proof generation time on reference hardware.
     - [ ]  FM7: Monitor challenge rates in production and have a runbook for adjusting bond and duration parameters if griefing is detected.
     - [ ]  FM7: Add gas consumption regression tests for the verifier.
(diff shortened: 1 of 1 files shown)


security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,326 @@
...
+
+This document covers the `ZKDisputeGame`, a single-round ZK-proof-based dispute game for the OP Stack. It replaces the multi-round bisection protocol with a single cryptographic proof that verifies an L2 state transition from a parent output root to a claimed output root. The system integrates into the existing OP Stack dispute infrastructure: `DisputeGameFactory`, `AnchorStateRegistry`, `DelayedWETH`, and `OPContractsManager`.
+
+Below are references for this project:
+
+- [Design Doc](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/zk-dispute-game.md)
+- [ZK Dispute Game Spec](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/zk-dispute-game.md)
+- [Game Mechanics Spec](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/game-mechanics.md)
+- [IZKVerifier Interface Spec](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/zk-interface.md)
+- [ZK Fault Proof VM Spec](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/zk-fault-proof-vm.md)
```
0xOneTony: todo: update links once specs get merged

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+
+## Failure Modes and Recovery Paths
+
+### FM1: ZK Verifier Soundness or Completeness Break
+
+- **Description:** Two sub-cases of `IZKVerifier` malfunction (aZKG-001):
+    
+    **A -- Soundness break:** A bug in the verifier or the `IZKVerifier` adapter causes `verify()` to return without reverting for an incorrect state transition. Since `prove()` treats any non-reverting return as proof acceptance, the game transitions to `DEFENDER_WINS` with a fraudulent `rootClaim`, enabling withdrawal of funds that don't exist on L2. This covers both cryptographic flaws (e.g., broken field arithmetic, pairing checks) and input validation gaps (e.g., `verify()` silently returns on empty proof bytes, zero `programId`, or gas-limited subcalls instead of reverting).
```
0xChin: instead of depending on the verifier reverting (which it may or may not do). why don’t we require the verifier to do something similar to the `onERC721Received` callback, which requires the interface to return a specific selector to ensure that it complied with the interface? https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/IERC721Receiver.sol#L11

another approach, require the interface to return `bool isValidProof_` and then `if (!isValidProof) revert();`

wwared: A second soundness scenario should be included for the case where the ZK program has a bug in it that enables a prover to generate a valid proof for invalid data by manipulating private values. This is a soundness issue in the ZK program. See also https://github.com/ethereum-optimism/specs/pull/896#discussion_r2956256839)

It is a critical failure mode for the ZK game architecture that would enable invalid proofs to be generated and incorrect state roots to resolve to `DEFENDER_WINS`.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    
+    **B -- Completeness break:** The verifier rejects valid proofs. All challenged games resolve as `CHALLENGER_WINS`, proposers lose bonds, and withdrawals stall.
+    
+- **Risk Assessment:** Critical severity / Low likelihood
+- **Mitigations:**
+    1. `IZKVerifier` interface enables verifier swap without redeploying the game.
+    2. `prove()` constructs `publicValues` from immutable on-chain state — only `_proofBytes` is caller-supplied.
```
wwared, line 53: I would add an item along the lines of:
* The ZK program validates that the private prover inputs are derived directly from or cryptographically linked to the public inputs.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+- **Detection:**
+    - Off-chain monitoring that re-executes L2 state transitions and alerts on `DEFENDER_WINS` for incorrect claims.
+    - Fuzz testing verifier with edge-case inputs (empty bytes, zero values, gas-limited calls).
+    - Monitoring for challenged games that never receive a valid proof (sub-case B indicator).
+- **Recovery Path(s):**
+    1. Guardian pauses system, blacklists affected games, retires game type updating the retirement timestamp if systemic.
+    2. OPCM deploys patched verifier and updates `gameArgs`.
+    3. For sub-case B: Guardian blacklists affected games (REFUND mode) while corrected verifier is deployed.
+- **Action Item(s):**
+    - [ ]  FM1: Verifier audited by at least two independent firms before mainnet.
```
wwared: An additional item to add:
* FM1: ZK Program audited by at least two independent firms before mainnet.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    
+- **Risk Assessment:** High severity / Low-to-medium likelihood
+- **Mitigations:**
+    1. In-progress games use old configuration immutably (CWIA args fixed at clone creation).
+    2. Upgrade sequencing: update `absolutePrestate` on-chain first, then distribute new binary to provers, then resume proposing. For hard forks, prestate update MUST precede activation.
+    3. Old verifier contracts MUST remain functional until all in-flight games resolve or are blacklisted.
+    4. Retirement timestamps should allow at least `maxChallengeDuration + maxProveDuration + DISPUTE_GAME_FINALITY_DELAY_SECONDS` lead time.
+- **Detection:**
+    - Monitoring for `gameArgs` changes in the `DisputeGameFactory`.
+    - Alerts when a challenged game fails to receive a proof within a reasonable time (indicating prover/prestate mismatch).
```
wwared, line 142: Might be good to also add something along the lines of:
- Alerts when a `prove()` transaction for a valid state root call reverts

That scenario can also indicate a prover/prestate mismatch

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+
+---
+
+### FM7: Bond and Duration Misconfiguration
+
+- **Description:** Economic parameter calibration risks (aZKG-004, aZKG-007). All sub-cases are fundamentally about calibrating `initBond`, `challengerBond`, `maxChallengeDuration`, and `maxProveDuration`:
+    - **Bonds too low:** Spam proposals and frivolous challenges are cheap. Challenging costs only `challengerBond`; defending requires expensive ZK proof generation. If proving cost > `challengerBond`, defending is economically irrational.
+    - **Bonds too high:** Honest proposers and challengers priced out, undermining permissionless participation.
+    - **Durations too short:** Provers may not generate the proof in time. L1 censorship of `prove()` also becomes feasible (cost proportional to `maxProveDuration`). Short `maxChallengeDuration` compounds FM2 risk.
+    - **Durations too long:** Withdrawal finality delayed, eliminating the latency benefit of ZK proofs.
```
wwared: Technically/pedantically speaking, the duration being too long causes the withdrawal finality to delay only in the optimistic path (i.e. no proof was supplied and no challenge was made). In the ZK proof case the finality is still much closer to "immediate" as soon as the proof is verified and the game resolves (so there are still latency benefits to the ZK proofs). However this implies a proving cost that needs to be paid by the prover/proposer that wants to keep the finality short, which could lead to economic incentive issues (i.e. if the duration is too long either the withdrawal finality is delayed OR some honest prover needs to irrationally lose money by eagerly generating ZK proofs for all states).

I'm not sure if this item needs to be changed, since it is correct as-is, this is a nitpick on the second sentence here (i.e. reaping the latency benefit of ZK proofs is orthogonal to the challenge duration being too long, and can only be realized by the ZK proofs actually being verified on chain).

security/fma-zk-dispute-game.md
ajsutton: We don't need to get two independent reviews of either the program or the ZKVM. Our security model is that dispute games can and will resolve incorrectly - the air gap allows us to intervene to prevent damage if/when they do so.
We don't _want_ that to happen so an audit would make sense, but these will be very expensive audits so I don't think we can justify doing two.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    
+    **A -- Soundness break:** A bug in the verifier or the `IZKVerifier` adapter causes `verify()` to return without reverting for an incorrect state transition. Since `prove()` treats any non-reverting return as proof acceptance, the game transitions to `DEFENDER_WINS` with a fraudulent `rootClaim`, enabling withdrawal of funds that don't exist on L2. This covers both cryptographic flaws (e.g., broken field arithmetic, pairing checks) and input validation gaps (e.g., `verify()` silently returns on empty proof bytes, zero `programId`, or gas-limited subcalls instead of reverting).
+    
+    **B -- Completeness break:** The verifier rejects valid proofs. All challenged games resolve as `CHALLENGER_WINS`, proposers lose bonds, and withdrawals stall.
+    
+- **Risk Assessment:** Critical severity / Low likelihood
+- **Mitigations:**
+    1. `IZKVerifier` interface enables verifier swap without redeploying the game.
+    2. `prove()` constructs `publicValues` from immutable on-chain state — only `_proofBytes` is caller-supplied.
+    3. Verifier contract should be independently audited by multiple firms.
```
ajsutton: As noted below, I don't imagine we will do multiple audits on the verifier given any faults are contained within the air gap.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    4. `DISPUTE_GAME_FINALITY_DELAY_SECONDS` airgap + `DelayedWETH` delay give the Guardian two windows to intervene.
+- **Detection:**
+    - Off-chain monitoring that re-executes L2 state transitions and alerts on `DEFENDER_WINS` for incorrect claims.
+    - Fuzz testing verifier with edge-case inputs (empty bytes, zero values, gas-limited calls).
+    - Monitoring for challenged games that never receive a valid proof (sub-case B indicator).
```
ajsutton: For the two monitoring cases here, we monitor for any dispute game that is forecast to or does resolve incorrectly.  That picks up both cases where a game is challenged but not proven and where an invalid game is proven (or just not challenged).  This is existing behaviour in op-dispute-mon.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+- **Recovery Path(s):**
+    1. Guardian pauses system, blacklists affected games, retires game type updating the retirement timestamp if systemic.
+    2. OPCM deploys patched verifier and updates `gameArgs`.
+    3. For sub-case B: Guardian blacklists affected games (REFUND mode) while corrected verifier is deployed.
+- **Action Item(s):**
+    - [ ]  FM1: Verifier audited by at least two independent firms before mainnet.
+    - [ ]  FM1: Off-chain monitoring that verifies every game's `rootClaim` against trusted L2 node.
+    - [ ]  FM1: Guardian runbook for verifier compromise (pause/blacklist/retire sequence).
```
ajsutton: We already have these as part of our existing dispute game monitoring.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - [ ]  FM1: Fuzz test `IZKVerifier.verify()` with malformed inputs to confirm it always reverts.
+
+---
+
+### FM2: Unchallenged Fraudulent Proposal
+
+- **Description:** If nobody challenges a fraudulent output root within `maxChallengeDuration`, the game resolves as `DEFENDER_WINS` by default. The fraudulent root becomes a Valid Claim eligible to finalize withdrawals. The system's safety depends on at least one honest, well-funded, always-online challenger. The ZK game requires only a single `challenge()` call (simpler than multi-round bisection), but every proposal must still be independently re-executed to verify its `rootClaim`.
```
ajsutton: We don't actually independently re-execute each claim - just compare the proposal to the canonical output root from a trusted node. Admittedly the trusted node is executing derivation, but we don't want it to sound like we need to do a specific derivation task for each game separately or it would open up significant DoS risks.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+- **Risk Assessment:** Critical severity / Low likelihood
+- **Mitigations:**
+    1. Bond economics incentivizes challengers, a successful challenge can earn `initBond + challengerBond`, so `initBond` must be high enough to justify the cost of running a challenger infrastructure. At the same time, `challengerBond` must remain above proving cost but low enough to not discourage participation.
+    2. `maxChallengeDuration` provides a configurable time window. It should be long enough to account for L1 congestion and censorship scenarios.
+    3. The Guardian can blacklist fraudulent games during the `DISPUTE_GAME_FINALITY_DELAY_SECONDS` airgap, even if the challenge window was missed.
+    4. `DelayedWETH` provides an additional freeze window after `closeGame()`.
+    5. Multiple independent challengers can run concurrently for redundancy (though only one challenge per game is accepted).
+- **Detection:**
+    - Off-chain challenger software (`op-challenger` or equivalent) that monitors all new games, re-derives the correct output root, and challenges any incorrect claims.
```
ajsutton:
```suggestion
    - Off-chain challenger software (`op-challenger` or equivalent) that monitors all new games and challenges any incorrect claims.
```

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - Independent monitoring that alerts when a game's `rootClaim` does not match the correct output root and no challenge has been submitted.
+    - Alerts when `maxChallengeDuration` is approaching for an unchallenged game with an incorrect claim.
+- **Recovery Path(s):**
+    1. If the challenge window is missed, the Guardian blacklists the game before `closeGame()`.
+    2. If `closeGame()` was already called, the Guardian pauses the system to prevent withdrawal finalization.
+    3. As a last resort, governance can upgrade contracts to recover.
+- **Action Item(s):**
+    - [ ]  FM2: Implement monitoring that alerts when an unchallenged game has an incorrect `rootClaim` with time remaining in the challenge window.
```
ajsutton: We already have this monitoring via op-dispute-mon's incorrect forecast alerting.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - [ ]  FM2: Ensure `maxChallengeDuration` accounts for L1 congestion/censorship and bond economics incentivize challengers.
+    - [ ]  FM2: Document the Guardian fallback procedure for games that were not challenged in time.
```
ajsutton: We have this via the existing runbooks which cover any kind of dispute game resolving incorrectly.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    
+    **Safeguards:** iZKG-005 propagates `CHALLENGER_WINS` automatically. iZKG-009 prevents children from resolving before parents. `isGameRespected`/`wasRespectedGameTypeWhenCreated` prevents retired game types from finalizing withdrawals. But none of these help when a parent is individually blacklisted after resolving as `DEFENDER_WINS`.
+    
+    **Cross-prestate amplification:** Parents don't need the same `absolutePrestate` as children (intentional — avoids 7-day re-proofs). This means retirement of old-prestate games doesn't cover children with different prestates.
+    
+    **Resolution ordering deadlock:** iZKG-009 requires the parent to be resolved before any child can resolve. Challenging a root game blocks all descendants until the root's proof is submitted or `maxProveDuration` expires. In practice the delay equals the proof generation time (the prover can submit immediately after challenge), not the full `maxProveDuration` — the worst case only applies if proving infrastructure is unavailable. Cost to the attacker is `challengerBond` per challenge (forfeited to the prover). Proposers mitigate by maintaining parallel chains from the anchor state.
```
ajsutton: I'm not sure I see a deadlock here or a failure of any kind. The proposer simply proves their proposals and resolves games in order.

Proposers don't need to maintain parallel chains and doing so would just increase the amount they need to cover bonds and the amount of proofs they'd potentially need to generate.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    
+- **Risk Assessment:** High severity / Medium likelihood
+- **Mitigations:**
+    1. `resolve()` propagates `CHALLENGER_WINS` from parent to child (iZKG-005) — automatic for resolution-based invalidity.
+    2. iZKG-009 enforces child-waits-for-parent ordering, giving the Guardian time to act on parents first.
+    3. `isGameClaimValid()` checks blacklist/retirement status before allowing withdrawal finalization.
+    4. `DISPUTE_GAME_FINALITY_DELAY_SECONDS` + `DelayedWETH` delay provide Guardian intervention windows.
+    5. Proposers can maintain parallel chains to mitigate resolution ordering deadlocks.
```
ajsutton: This doesn't seem like a good option for proposers. They should just maintain a single, valid chain of proposals.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+- **Detection:**
+    - Off-chain tooling that monitors blacklist events and automatically enumerates all descendant games of a blacklisted game by traversing `parentIndex` references, including across prestate boundaries.
+    - Alerts when a game whose ancestor is blacklisted reaches a state where `closeGame()` could be called.
```
ajsutton: This shouldn't be required. Monitoring can detect any children that should have been blacklisted by checking if they have an invalid proposal but are forecast to or do resolve as DEFENDER_WINS. If the proposed output root is valid, it doesn't actually matter if it has an invalid chain getting to it (though honest proposers wouldn't use it as a parent unless it had resolved as DEFENDER_WINS already).

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - Monitoring for games that are challenged but remain unresolved close to `maxProveDuration`.
+    - Alerts when descendant games are blocked on parent resolution for extended periods.
```
ajsutton: We already have alerting for games that remain unresolved after they should be resolvable. This would pick up any parent that isn't resolved in a timely manner and then unblock the remaining chain.

If the parent was challenged but not proven, the alerting for a game forecast to resolve incorrectly would pick it up.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - Tracking the depth and structure of game chains.
```
ajsutton: What exactly is doing this tracking?

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+- **Recovery Path(s):**
+    1. Guardian blacklists all identified descendant games individually, tracing across prestate boundaries.
+    2. If some descendants were missed and `closeGame()` was called, Guardian pauses the system to prevent further damage.
+    3. Guardian retires the entire game type as a last resort to put all games into REFUND mode.
```
ajsutton: The failure case here is that the guardian blacklisted the parent but missed the child. So the recovery process would be to retire all games. Once we get it wrong once, we don't want to muck about trying again - just retire everything and we know we're safe (retirement uses the current timestamp so it guarantees all games are invalidated).

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    4. For resolution ordering deadlocks: provers submit proofs for challenged games to unblock the chain. Proposers create new parallel chains from the anchor state to bypass blocked chains.
+    5. If deadlock griefing is systematic, OPCM can increase `challengerBond` to raise the cost.
```
ajsutton: I think having a malicious actor grief by challenging every proposal should be a separate failure mode.  It would force the proposer to prove the full chain in the worst case so is low severity (although it would likely actually speed up withdrawals since you no longer wait for the challenge delay to elapse). Mitigation is likely do nothing and profit or increase the challenger bond if it isn't covering proving costs.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+- **Action Item(s):**
+    - [ ]  FM3: Build tooling that, given a blacklisted game, automatically enumerates all descendant games by querying `DisputeGameFactory` for all `ZKDisputeGame` instances and filtering by `parentIndex`, including cross-prestate ancestry.
+    - [ ]  FM3: Document the Guardian runbook for cascading blacklists, including cross-prestate tracing, tooling, and verification steps.
+    - [ ]  FM3: Add monitoring that alerts when any game with a blacklisted ancestor approaches the finality delay window.
```
ajsutton: We likely won't do this - any invalid proposal would already be picked up by existing monitoring and if it's a valid proposal we don't care that the parent was blacklisted.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - [ ]  FM3: Analyze the expected chain depth under normal operation and the maximum delay an attacker can impose per `challengerBond` spent via resolution ordering deadlock.
```
ajsutton: Challenging a proposal would likely speed up its resolution as the proposer would then prove it instead of it having to wait out the challenge duration.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+- **Recovery Path(s):**
+    1. Guardian pauses the system and blacklists affected games (REFUND mode).
+    2. If `DelayedWETH` is insolvent, governance must deploy a replacement — the contract is
+    immutable.
+- **Action Item(s):**
+    - [ ]  FM4: Implement iZKG-011 conservation invariant tests across all bond distribution
+    scenarios.
+    - [ ]  FM4: Fuzz test bond accounting across randomized game lifecycles, including the burn
+    path.
+    - [ ]  FM4: Add `DelayedWETH` balance monitoring alert.
```
ajsutton: We have this already in op-dispute-mon.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+
+---
+
+### FM5: Upgrade Lifecycle Mismatch
+
+- **Description:** Covers the full upgrade lifecycle (aZKG-002, [IZKVerifier upgrade path](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/zk-interface.md#verifier-upgrade-path)):
+    
+    **A -- absolutePrestate mismatch:** The `absolutePrestate` is updated but provers still run the old binary (or vice versa). Proofs fail verification; challenged games time out as `CHALLENGER_WINS` and proposers lose bonds. Also applies when governance delays mean the prestate isn't updated before a hard fork activates on L2.
+    
+    **B -- Verifier upgrade race:** The old verifier is deprecated while in-flight games still reference it via immutable CWIA args. Those games become unprovable.
```
ajsutton: Why would those games become unprovable? They should remain provable using the old verifier and prestate.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    
+    **C -- Aggressive retirement timestamp:** `updateRetirementTimestamp()` set too close to present forces in-flight games into REFUND mode unnecessarily.
```
ajsutton: Retirement always uses the current timestamp. This is a deliberate feature to ensure we invalidate everything. There's no `setRetirementTimestamp`, only `updateRetirementTimestamp`.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    
+    **D -- Cross-prestate parent chaining:** Parent games intentionally don't need the same `absolutePrestate` as children (avoids 7-day re-proofs on every update). But retiring old-prestate games doesn't cover children that chained off them with a different prestate — the Guardian must trace across prestate boundaries.
```
ajsutton: I'm not sure how the different prestate matters here. You can't retire old-prostate games specifically. You can only retire all games so new games would have to build from the anchor state again.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    path.
+    - [ ]  FM4: Add `DelayedWETH` balance monitoring alert.
+
+---
+
+### FM5: Upgrade Lifecycle Mismatch
+
+- **Description:** Covers the full upgrade lifecycle (aZKG-002, [IZKVerifier upgrade path](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/zk-interface.md#verifier-upgrade-path)):
+    
+    **A -- absolutePrestate mismatch:** The `absolutePrestate` is updated but provers still run the old binary (or vice versa). Proofs fail verification; challenged games time out as `CHALLENGER_WINS` and proposers lose bonds. Also applies when governance delays mean the prestate isn't updated before a hard fork activates on L2.
```
ajsutton: Running the wrong prestate and failing to update the prestate prior to a L2 hard fork seem like very different failure cases. For running the wrong  prestate the off-chain software just needs to have its configuration fixed, for the other we'd need to update the prestate on chain or fallback to permissioned games.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    **B -- Verifier upgrade race:** The old verifier is deprecated while in-flight games still reference it via immutable CWIA args. Those games become unprovable.
+    
+    **C -- Aggressive retirement timestamp:** `updateRetirementTimestamp()` set too close to present forces in-flight games into REFUND mode unnecessarily.
+    
+    **D -- Cross-prestate parent chaining:** Parent games intentionally don't need the same `absolutePrestate` as children (avoids 7-day re-proofs on every update). But retiring old-prestate games doesn't cover children that chained off them with a different prestate — the Guardian must trace across prestate boundaries.
+    
+- **Risk Assessment:** High severity / Low-to-medium likelihood
+- **Mitigations:**
+    1. In-progress games use old configuration immutably (CWIA args fixed at clone creation).
+    2. Upgrade sequencing: update `absolutePrestate` on-chain first, then distribute new binary to provers, then resume proposing. For hard forks, prestate update MUST precede activation.
```
ajsutton: There will be active games using multiple different prestates at once.  Off chain software will be designed to automatically select the correct prestate to use based on the absolute prestate hash (op-challenger does this today though it may need to be added to the succinct proposer which does proving).

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    3. Old verifier contracts MUST remain functional until all in-flight games resolve or are blacklisted.
```
ajsutton: We should be using immutable verifier contracts so they can never stop working.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    4. Retirement timestamps should allow at least `maxChallengeDuration + maxProveDuration + DISPUTE_GAME_FINALITY_DELAY_SECONDS` lead time.
```
ajsutton: I'm not sure what this is referring to. When the retirement timestamp is updated all existing games are invalidated. There shouldn't be a need to factor in lead time here.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+- **Detection:**
+    - Monitoring for `gameArgs` changes in the `DisputeGameFactory`.
```
ajsutton: We have monitoring for unexpected transactions from privileged wallets. I'm not sure why we'd need monitoring specifically for gameArgs changes.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - Alerts when a challenged game fails to receive a proof within a reasonable time (indicating prover/prestate mismatch).
+    - Pre-deployment verification that the `absolutePrestate` matches the program binary hash.
+    - Monitoring for games referencing verifier addresses that are no longer functional.
```
ajsutton: Immutable verifier address should make it impossible for a verifier address to become non-functional.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - Alerts when retirement timestamp is set within the in-flight game window.
```
ajsutton: I'm not sure what this alert is for given the updated retirement timestamp would invalid all games so they can then all be ignored.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+- **Recovery Path(s):**
+    1. If games were created with a mismatched `absolutePrestate`, they cannot be proven. If challenged, the proposer loses `initBond`. Guardian can blacklist these games to trigger REFUND mode instead.
+    2. Correct the `absolutePrestate` via OPCM and resume normal operation.
+    3. If a hard fork activated before the prestate update, Guardian retires the old game type and OPCM deploys with the correct prestate.
+    4. If the old verifier was made non-functional prematurely, Guardian blacklists affected games to trigger REFUND mode, then OPCM re-deploys the old verifier or migrates to a new one.
+- **Action Item(s):**
+    - [ ]  FM5: Document a program upgrade runbook that specifies the exact sequencing of `absolutePrestate` update, prover binary distribution, and proposer resumption.
+    - [ ]  FM5: For hard fork upgrades, the `absolutePrestate` update MUST be included in the governance proposal alongside the L2 activation timestamp, with the on-chain update applied before activation.
```
ajsutton: This isn't really an action item - there's nothing to do for it until there is a future upgrade proposal being made.
And we do already have this as part of our standard hard fork process.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - Pre-deployment verification that the `absolutePrestate` matches the program binary hash.
+    - Monitoring for games referencing verifier addresses that are no longer functional.
+    - Alerts when retirement timestamp is set within the in-flight game window.
+- **Recovery Path(s):**
+    1. If games were created with a mismatched `absolutePrestate`, they cannot be proven. If challenged, the proposer loses `initBond`. Guardian can blacklist these games to trigger REFUND mode instead.
+    2. Correct the `absolutePrestate` via OPCM and resume normal operation.
+    3. If a hard fork activated before the prestate update, Guardian retires the old game type and OPCM deploys with the correct prestate.
+    4. If the old verifier was made non-functional prematurely, Guardian blacklists affected games to trigger REFUND mode, then OPCM re-deploys the old verifier or migrates to a new one.
+- **Action Item(s):**
+    - [ ]  FM5: Document a program upgrade runbook that specifies the exact sequencing of `absolutePrestate` update, prover binary distribution, and proposer resumption.
```
ajsutton: We shouldn't need to pause and resume the proposer as part of deploying a new prestate. Existing procedures already cover making new prestates available prior to their deployment.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - [ ]  FM5: For hard fork upgrades, the `absolutePrestate` update MUST be included in the governance proposal alongside the L2 activation timestamp, with the on-chain update applied before activation.
+    - [ ]  FM5: Add automated verification in the deployment pipeline that the `absolutePrestate` in `gameArgs` matches the hash of the deployed program binary.
```
ajsutton: I believe superchain-ops has this validation (ensuring the prestate is from the standard-prestates list) but the main validation is actually the governance process reviewing the proposal. I don't think we need to take any action here.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - [ ]  FM5: Document the requirement that old verifier contracts must remain functional until all in-flight games resolve, and add monitoring for verifier liveness.
```
ajsutton: The requirement would just be that verifier contracts are immutable.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - [ ]  FM5: Establish a minimum retirement timestamp lead time policy and enforce it in Guardian tooling.
+    - [ ]  FM5: Document the cross-prestate parent chaining implications for retirement cascades in the Guardian runbook.
```
ajsutton: I think these are just based on misunderstanding how retirement timestamps work and we shouldn't need to take any action here.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    - **Durations too short:** Provers may not generate the proof in time. L1 censorship of `prove()` also becomes feasible (cost proportional to `maxProveDuration`). Short `maxChallengeDuration` compounds FM2 risk.
+    - **Durations too long:** Withdrawal finality delayed, eliminating the latency benefit of ZK proofs.
+    - **Block range selection:** The proposer is responsible for creating games with block ranges they can prove within `maxProveDuration`. Proving larger ranges in a single proof is generally cheaper per-block in zkVMs however a proposer who creates an unprovable range loses their own `initBond`.
+    - **L1 gas pressure:** ZK proof verification can cost several hundred thousand gas units depending on the backend. During extreme gas spikes, `prove()` inclusion becomes expensive, not technically unfeasible though given the margin against the block gas limit.
+- **Risk Assessment:** Medium severity / Medium likelihood
+- **Mitigations:**
+    1. All parameters are in `gameArgs` and tunable per chain by OPCM without redeploying.
+    2. `challengerBond` should exceed 2x expected proving cost so defending is always profitable.
```
ajsutton: I don't think we want to explicitly specify 2x here. The higher the challengerBond the more you are likely to want to have on hand to fund challenges which reduces capital efficiency.  There is already very strong incentive to challenge invalid proposals for any chain operator since that's what keeps the chain secure.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    3. Third-party provers can earn `challengerBond`, creating a market incentive for proving services.
+    4. `maxProveDuration` must be long enough that L1 censorship of `prove()` is economically infeasible (aZKG-007).
+    5. Rational proposers won't create unnecessarily large block ranges (aZKG-006).
+- **Detection:**
+    - Monitoring challenge rates, proving costs vs `challengerBond`, and `prove()` inclusion failures.
+    - Alerts when challenged games approach `maxProveDuration` without proof submission.
+- **Recovery Path(s):**
+    1. OPCM adjusts `challengerBond` and/or `maxProveDuration`. "Game type changed mid-play" ensures in-flight games get REFUND treatment.
```
ajsutton: Existing games don't enter refund mode - they continue to play out using their existing rules.  We _could_ blacklist to trigger refund mode if really needed but are unlikely to since the cost of proving the entire chain for even a week or two is not unreasonable even if challenge bonds don't cover it.

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    2. Guardian can blacklist exploitative games or retire the game type.
```
ajsutton: You can't retire a game type. You can retire all games or change the respected game type (ie fallback to permissioned games).

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+- **Action Item(s):**
+    - [ ]  FM7: Establish a minimum `challengerBond` policy such that `challengerBond > 2x expected proving cost` to ensure defending is always profitable.
```
ajsutton:
```suggestion
    - [ ]  FM7: Establish a minimum `challengerBond` policy such that `challengerBond > expected proving cost` to ensure defending is always profitable.
```

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,324 @@
...
+    2. The fraud still doesn't succeed — the claim resolves as `CHALLENGER_WINS` and cannot finalize withdrawals.
+    3. The proposer still pays gas for both `create()` and `challenge()`, plus the capital lockup for `initBond + challengerBond` during the game lifecycle.
+- **Detection:**
+    - Monitoring for games where the challenger address is linked to the proposer (same EOA, same deployer, funded from the same source).
+    - Monitoring for repeated fraudulent proposals from the same proposer that are always self-challenged.
+- **Recovery Path(s):**
+    1. If systematic self-challenge front-running is detected, increase `initBond` to raise the capital cost of the attack.
+- **Action Item(s):**
+    - [ ]  FM8: Analyze the economic viability of self-challenge front-running given gas costs and capital lockup.
+    - [ ]  FM8: Evaluate protocol mitigations: multiple challenges or challenger priority mechanisms.
```
ajsutton: I wouldn't expect that we will take any action for this. The malicious proposer is still guaranteed to lose money through gas and may still lose their bonds through either failing to front run or having the guardian intervene and manually reallocate the bonds. And the system is still operating fine with every invalid proposal being challenged.

ajsutton (commented):
I think we've identified months of work to address all these action items when they often aren't providing real value.  In particular we generally don't need to monitor for very specific conditions and can catch the vast majority of risks with just:
* games are forecast to or do resolve incorrectly (including forecasting based on the ancestor's resolution status)
* DelayedWETH balance matches outstanding bonds from games

That won't cover everything but we should review the action items and see if we can use simpler, higher level monitoring to get the same level of coverage. Complex monitors that try to detect very specific scenarios tend to be far less reliable.

security/fma-zk-dispute-game.md
wwared: That's fine, this suggestion was based on having symmetry with the smart contract audit action item above. I agree that two independent audits for the program/zkvm and verifiers would be impractically expensive since they're much larger codebases with way higher surface area, given that the air gap is an important safety feature of the protocol

In particular, I don't think an independent audit of the entire zkVM is necessary or makes too much sense at all - Succinct and Risc0 have already plenty of independent audits for their production level zkVMs which we can rely on, and there is way more value in auditing our verifier contracts and ZK program than the zkVM

I think mentioning that the ZK program itself should be audited (which is not in the FMA doc), so I'll keep this thread unresolved for that to be added

security/fma-zk-dispute-game.md
wwared: The way for the malicious/griefing challenger to delay resolution is to wait right before the challenge duration is about to expire, and then challenge the proposal. However the mitigation for this case is to simply generate the proof and adjust the parameters if necessary (challenging costs and challenging window), and the overall impact would be low (the prover still ends up profiting by simply collecting all the challenger bonds and the additional delay is the cost of generating one proof for the proposal).

security/fma-zk-dispute-game.md
Joxess: You're right here, the downside here is that bridge finality becomes slower by default if the games are not proven in advance

security/fma-zk-dispute-game.md
0xOneTony: Good call, will add

security/fma-zk-dispute-game.md
0xOneTony: Good to know this logic exists, will probably need to be supported for the zk game too right?

security/fma-zk-dispute-game.md
0xOneTony: added as part of https://github.com/ethereum-optimism/design-docs/pull/369/commits/dd7114795de315960c0c496f081a184493598096

security/fma-zk-dispute-game.md
0xOneTony: added in mitigations as part of https://github.com/ethereum-optimism/design-docs/pull/369/commits/dd7114795de315960c0c496f081a184493598096

security/fma-zk-dispute-game.md
0xOneTony: @wwared added as part of the `Audit Requirements` section at https://github.com/ethereum-optimism/design-docs/pull/369/commits/d5efe7ffac76859ec51bd21a8531aae62a70b937

security/fma-zk-dispute-game.md
0xOneTony: changed here https://github.com/ethereum-optimism/design-docs/pull/369/commits/b153d5070b491c91c4ff7afdcaecd75a7fcc3eb4

security/fma-zk-dispute-game.md
0xOneTony: updated at https://github.com/ethereum-optimism/design-docs/pull/369/changes/d5efe7ffac76859ec51bd21a8531aae62a70b937

security/fma-zk-dispute-game.md
0xOneTony: removed here https://github.com/ethereum-optimism/design-docs/pull/369/changes/d5efe7ffac76859ec51bd21a8531aae62a70b937

security/fma-zk-dispute-game.md
0xOneTony: thank you for clarifying, fixed here https://github.com/ethereum-optimism/design-docs/pull/369/changes/d5efe7ffac76859ec51bd21a8531aae62a70b937

security/fma-zk-dispute-game.md
0xOneTony: removed here https://github.com/ethereum-optimism/design-docs/pull/369/changes/d5efe7ffac76859ec51bd21a8531aae62a70b937

security/fma-zk-dispute-game.md
0xOneTony: removed here https://github.com/ethereum-optimism/design-docs/pull/369/changes/d5efe7ffac76859ec51bd21a8531aae62a70b937

security/fma-zk-dispute-game.md
0xOneTony: Agree, removed here https://github.com/ethereum-optimism/design-docs/pull/369/changes/57def0967ced7e0d777702edefe3b1ccdf0e3761

security/fma-zk-dispute-game.md
0xOneTony: removed https://github.com/ethereum-optimism/design-docs/pull/369/changes/57def0967ced7e0d777702edefe3b1ccdf0e3761

security/fma-zk-dispute-game.md
0xOneTony: updated all the monitoring related mitigations for fm3 as part of this https://github.com/ethereum-optimism/design-docs/pull/369/changes/57def0967ced7e0d777702edefe3b1ccdf0e3761

security/fma-zk-dispute-game.md
0xOneTony: The idea was to have some monitor tooling to track thedepth and status of child games in case the parent get blacklisted, but my understanding is op-dispute-mon already has monitoring that flags invalid games so this is not needed, removed here https://github.com/ethereum-optimism/design-docs/pull/369/changes/57def0967ced7e0d777702edefe3b1ccdf0e3761

security/fma-zk-dispute-game.md
0xOneTony: refactored here https://github.com/ethereum-optimism/design-docs/pull/369/changes/57def0967ced7e0d777702edefe3b1ccdf0e3761

security/fma-zk-dispute-game.md
0xOneTony: added as fm9 here https://github.com/ethereum-optimism/design-docs/pull/369/changes/57def0967ced7e0d777702edefe3b1ccdf0e3761

security/fma-zk-dispute-game.md
0xOneTony: fixed here https://github.com/ethereum-optimism/design-docs/pull/369/changes/57def0967ced7e0d777702edefe3b1ccdf0e3761

security/fma-zk-dispute-game.md
0xOneTony: updated as part of https://github.com/ethereum-optimism/design-docs/pull/369/changes/57def0967ced7e0d777702edefe3b1ccdf0e3761

security/fma-zk-dispute-game.md
0xOneTony: sicne verifiers will be immutable and new verifies get deployed, yes this doesnt make sense, will remove

security/fma-zk-dispute-game.md
0xOneTony: good call! refactored fm5 but still added the alert https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: removed here https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: removed here, in a refactor of fm5 https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: removed, https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: removed, was a miss conception https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: repurposed fm5 around running the wrong prestate https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: Good information to know updated the mitigation to reflect this https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: Agree! fixed https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: removed here https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: fixed here https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: was part of the missunderstanding of the update timestamp, removed here https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: removed https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: removed https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: fixed https://github.com/ethereum-optimism/design-docs/pull/369/commits/369ef91b4818c57001b8ffbe36df72d1ad093104

security/fma-zk-dispute-game.md
0xOneTony: kept and improved the wording a bit https://github.com/ethereum-optimism/design-docs/pull/369/changes/18881630e655ee81ec1b03d49849fdac8c0c269c

security/fma-zk-dispute-game.md
0xOneTony: agree, fixed here https://github.com/ethereum-optimism/design-docs/pull/369/changes/18881630e655ee81ec1b03d49849fdac8c0c269c

security/fma-zk-dispute-game.md
0xOneTony: changed https://github.com/ethereum-optimism/design-docs/pull/369/changes/18881630e655ee81ec1b03d49849fdac8c0c269c

security/fma-zk-dispute-game.md
0xOneTony: yea wording was off, updated this like the rest of the cases aswell https://github.com/ethereum-optimism/design-docs/pull/369/changes/18881630e655ee81ec1b03d49849fdac8c0c269c

security/fma-zk-dispute-game.md
0xOneTony: removed both action items https://github.com/ethereum-optimism/design-docs/pull/369/changes/18881630e655ee81ec1b03d49849fdac8c0c269c

security/fma-zk-dispute-game.md
```diff
@@ -38,163 +39,112 @@ Below are references for this project:
...
-    4. Retirement timestamps should allow at least `maxChallengeDuration + maxProveDuration + DISPUTE_GAME_FINALITY_DELAY_SECONDS` lead time.
+    1. In-progress games use old configuration immutably (CWIA args fixed at clone creation). Off-chain software automatically selects the correct prestate based on the absolute prestate hash, supporting multiple prestates concurrently.
+    2. Verifier contracts are immutable — old verifiers remain functional indefinitely and cannot be deprecated.
 - **Detection:**
-    - Monitoring for `gameArgs` changes in the `DisputeGameFactory`.
     - Alerts when a challenged game fails to receive a proof within a reasonable time (indicating prover/prestate mismatch).
-    - Pre-deployment verification that the `absolutePrestate` matches the program binary hash.
-    - Monitoring for games referencing verifier addresses that are no longer functional.
-    - Alerts when retirement timestamp is set within the in-flight game window.
+    - Alerts when a `prove()` transaction for a valid state root reverts.
```
ajsutton: Minor tweak to this - we don't want to alert on any prove() failure for a game with a valid state root. We don't want random users to be able to spam our alerting by posting invalid prove() calls.  But we would track the number of prove failures our proposer/prover encounters.

So something like:
 - Alerts when the proposer generates a proof that is found to be invalid on chain.

ajsutton (approved):
LGTM.

security/fma-zk-dispute-game.md
0xOneTony: good call, fixed here https://github.com/ethereum-optimism/design-docs/pull/369/commits/68d6e4896533fa31025c1f7a2e70d6433ee89c8b

wwared (approved):
LGTM

security/fma-zk-dispute-game.md
0xOneTony: updated as of https://github.com/ethereum-optimism/design-docs/pull/369/commits/a272b8681a578219eb6e79182b5b6da45c1a1c7d

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,270 @@
...
+
+### FM4: Bond Accounting Failure and DelayedWETH Insolvency
+
+- **Description:** Multiple `ZKDisputeGame` instances share the same per-chain `DelayedWETH`. The spec requires (iZKG-011) that for every game: `sum(distributions) + sum(burns) == initBond + challengerBond`. An accounting bug in `closeGame()` that violates this invariant — distributing more than deposited, double-crediting, or mishandling the burn path — could drain `DelayedWETH` or permanently lock funds.
+    
+    The burn path deserves attention: when a parent resolves as `CHALLENGER_WINS` and the child was unchallenged, the child's `initBond` is sent to `address(0)`. A bug here could create unbacked withdrawals or lock residual funds. The bond distribution table has 10 NORMAL + 3 REFUND scenarios, each with distinct logic paths.
+    
+- **Risk Assessment:** Critical severity / Low likelihood
+- **Mitigations:**
+    1. Bond distribution is fully specified in the [game mechanics](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/game-mechanics.md#bond-distribution) with a clear table of every scenario.
```
ajsutton: One more spec link

security/fma-zk-dispute-game.md
```diff
@@ -0,0 +1,270 @@
...
+    2. Guardian can blacklist affected games to trigger REFUND mode if needed.
+
+---
+
+### FM6: CWIA Game Args Encoding Mismatch
+
+- **Description:** The MCP clone pattern packs 8 fields into `gameArgs` via `abi.encodePacked`. A mismatch between `OPContractsManager._makeGameArgs()` encoding and `ZKDisputeGame` CWIA offset decoding causes the game to read wrong values — wrong `verifier` address, wrong `challengerBond`, wrong durations, etc. Note: `l2SequenceNumber` type changed from `uint256` to `uint64` in `_extraData`, relevant to packing correctness.
+- **Risk Assessment:** Medium severity / Low likelihood
+- **Mitigations:**
+    1. The encoding is defined in the spec ([Game Args Layout](https://github.com/defi-wonderland/specs/blob/feat/op-zk-proofs/specs/fault-proof/stage-one/zk/zk-dispute-game.md#game-args-layout)) and implemented in `OPContractsManager._makeGameArgs()`.
```
ajsutton: Two more actually.

security/fma-zk-dispute-game.md
0xOneTony: nice catch, fixed here https://github.com/ethereum-optimism/design-docs/pull/369/commits/e66826db61018179aa6a32efe8eec96abedf49a6

security/fma-zk-dispute-game.md
0xOneTony: fixed here https://github.com/ethereum-optimism/design-docs/pull/369/commits/e66826db61018179aa6a32efe8eec96abedf49a6

PR #370, opened 2026-03-19
feat: design doc for the shared dispute game migration

### LINKED ISSUE #19651: Design doc for shared dispute game
https://github.com/ethereum-optimism/optimism/pull/19253 provides the upgrade to SuperDisputeGame but each chain has it's own. Next we need the ability to merge chains into a shared SuperDisputeGame. This issue is for creating a design doc.

Let's keep this lightweight. `OPCM.migrate()` was already implemented, we can just AI-generate a design doc from this as a starting point and use it to discuss the changes for OPCM v2, and any tradeoffs like (hopefully not) resetting withdrawals, and the need for pre-approved hardcoded super roots.

stevennevins:
**Description**

**Tests**

**Additional context**

**Metadata**

Closes https://github.com/ethereum-optimism/design-docs/pull/370

### DIFF
--- a/protocol/proofs/shared-dispute-game-migration.md
+++ b/protocol/proofs/shared-dispute-game-migration.md
@@ -130,7 +130,7 @@ This step runs ONCE, atomically, for ALL chains in the migration set.
 | Game types | CANNON, PERMISSIONED_CANNON, CANNON_KONA | SUPER_CANNON_KONA, SUPER_PERMISSIONED_CANNON |
 | Proof method | `rootClaim()` (single chain) | `rootClaimByChainId(chainId)` (per-chain from super root) |
 
-**All prior withdrawal proofs are invalidated.** The ASR reference changes — old proofs reference games from the old DGF, which the new ASR does not track. Users must re-prove (`OPContractsManagerMigrator.sol:62-64`).
+**All prior withdrawal proofs are invalidated.** The ASR reference changes — old proofs reference games from the old DGF, which the new ASR does not track. Users must re-prove (`OPContractsManagerMigrator.sol:62-64`). Additionally, the new ASR sets `retirementTimestamp = block.timestamp` on initialization, which retires all games created before or at that block as an extra safety measure (`AnchorStateRegistry.sol:106-113`).
 
 **Irreversible without full contract upgrade.** No `migrateBack()` exists.
 
--- a/protocol/proofs/shared-dispute-game-migration.md
+++ b/protocol/proofs/shared-dispute-game-migration.md
@@ -187,9 +187,6 @@ Rejected. Creates a permanent fork where CGT chains can never participate in int
 
 - Is explicit blocking of CGT chains from joining non-CGT interop sets necessary?
 
-- The migrate function does not upgrade each chain's portal. This should either happen within `migrate()` itself or in a prior upgrade step.
-  - `migrate()` assumes the portal already has `migrateToSuperRoots()` — casts to `IOptimismPortalInterop` and calls it directly (`OPContractsManagerMigrator.sol:244,280`). The current design relies on a prior `OPContractsManagerV2.upgrade()` call to swap the portal impl (`OPContractsManagerV2.sol:744-752`). If the portal hasn't been upgraded, the call reverts.
-
 ### Trust assumptions
 
 - **ProxyAdmin owner coordination**: All chains must trust the same ProxyAdmin owner to execute the migration correctly. A malicious or compromised owner can corrupt all chains simultaneously.
--- a/protocol/proofs/shared-dispute-game-migration.md
+++ b/protocol/proofs/shared-dispute-game-migration.md
@@ -90,8 +90,6 @@ This step runs once per chain BEFORE the atomic migration. It runs for **every**
 - **Legacy game types disabled.** Prevents split-brain where both legacy and super games exist on the same DGF.
 - **OPCM version sequence.** Cannot skip versions.
 
-**Transitional state:** After this step, chains function independently with super game types on their per-chain DGF. This is functional but NOT the target state — shared infrastructure comes in Step 2.
-
 ### 4. Step 2: Atomic Multi-Chain Migrate (`OPContractsManagerMigrator.migrate()`)
 
 This step runs ONCE, atomically, for ALL chains in the migration set.
--- a/protocol/proofs/shared-dispute-game-migration.md
+++ b/protocol/proofs/shared-dispute-game-migration.md
@@ -18,7 +18,7 @@ The migration consolidates N independent chains into a shared set of these contr
 
 Key outcomes:
 - **Portal unification**: `OptimismPortalInterop` features are merged into `OptimismPortal2`, making OP2 the single portal contract. OPInterop is deleted.
-- **Shared infrastructure**: All migrated chains share one `DisputeGameFactory`, one `AnchorStateRegistry`, one `ETHLockbox`, and one `DelayedWETH`.
+- **Shared infrastructure**: All migrated chains share one `DisputeGameFactory`, one `AnchorStateRegistry`, and one `ETHLockbox`. Reuses existing DelayedWETH (`OPContractsManagerMigrator.sol:158-163`).
 - **Two-phase execution**: (1) per-chain OPCM upgrade prepares each chain, then (2) an atomic multi-chain `migrate()` call deploys shared infra and wires everything together.
 - **Proof routing via game type**: `_isSuperGameType()` in OP2 (`OptimismPortal2.sol:642-646`) determines whether to use `rootClaimByChainId()` or `rootClaim()` based on the game type enum. No `superRootsActive` storage flag needed.
 - **CGT chain compatibility**: Not in v1 scope. Attempting to migrate a CGT chain will revert — this prevents CGT chains from being incorrectly pooled into the shared ETHLockbox. `OptimismPortal2` retains CGT support via the `Features` library so CGT chains can be added in a future iteration.
@@ -96,7 +96,7 @@ This step runs ONCE, atomically, for ALL chains in the migration set.
 
 **What happens:**
 
-1. **Deploy shared infrastructure** (`OPContractsManagerMigrator.sol:125-156`): new ETHLockbox, DisputeGameFactory, and AnchorStateRegistry proxies. Reuses existing DelayedWETH.
+1. **Deploy shared infrastructure** (`OPContractsManagerMigrator.sol:125-156`): new ETHLockbox, DisputeGameFactory, and AnchorStateRegistry proxies. Reuses existing DelayedWETH (`OPContractsManagerMigrator.sol:158-163`).
 
 2. **Initialize shared contracts** (`OPContractsManagerMigrator.sol:166-205`): ETHLockbox with first chain's SystemConfig, DGF with ProxyAdmin owner, ASR with starting anchor root and respected game type. The first chain's SystemConfig is used as the canonical governance reference because all chains are validated to share the same ProxyAdmin owner and SuperchainConfig (`OPContractsManagerMigrator.sol:171-174`).
 
--- a/protocol/proofs/shared-dispute-game-migration.md
+++ b/protocol/proofs/shared-dispute-game-migration.md
@@ -195,7 +195,7 @@ Rejected. Creates a permanent fork where CGT chains can never participate in int
 
 - **No rollback mechanism**: If the migration succeeds but a bug is discovered in the shared infrastructure, there is no `undo()`. Fixing requires deploying new shared contracts and migrating again (which the current migrator does not support — see re-migration constraint).
 - **Orphaned in-progress games**: Games created on old per-chain DGFs before migration can still resolve, but their results are not tracked by the new shared ASR. Users who proved withdrawals against these games must re-prove against the new DGF.
-- **No partial migration support**: There is no way to migrate N-1 chains and add the Nth later.
+- **No partial migration support (v1 constraint)**: There is no way to migrate N-1 chains and add the Nth later. Incremental chain addition is planned for a future iteration.
 
 ## Key Reference Files
 
--- a/protocol/proofs/shared-dispute-game-migration.md
+++ b/protocol/proofs/shared-dispute-game-migration.md
@@ -21,7 +21,7 @@ Key outcomes:
 - **Shared infrastructure**: All migrated chains share one `DisputeGameFactory`, one `AnchorStateRegistry`, and one `ETHLockbox`.
 - **Two-phase execution**: (1) per-chain OPCM upgrade prepares each chain, then (2) an atomic multi-chain `migrate()` call deploys shared infra and wires everything together.
 - **Proof routing via game type**: `_isSuperGameType()` in OP2 (`OptimismPortal2.sol:642-646`) determines whether to use `rootClaimByChainId()` or `rootClaim()` based on the game type enum. No `superRootsActive` storage flag needed.
-- **CGT chain compatibility**: Custom Gas Token chains are fully supported through the existing `Features` library in OP2 for single chain interop sets.
+- **CGT chain compatibility**: Not in v1 scope. `OptimismPortal2` retains CGT support via the `Features` library so CGT chains can be added in a future iteration.
 
 ## Problem Statement + Context
 
@@ -187,11 +187,13 @@ Rejected. Creates a permanent fork where CGT chains can never participate in int
 
 ## Risks & Uncertainties
 
+### Resolved
+
+- **CGT chains and interop**: CGT is not required for the first release — no CGT chains are immediately lined up for interop. Future CGT support requires splitting lockbox migration from the shared dispute game migration so CGT chains can join without ETH pooling (they must not deploy ETH bridging contracts like `SuperchainETHBridge`/`ETHLiquidity`).
+
 ### Open questions
 
-- Can CGT chains participate in shared dispute games without pooling ETH, or must all chains share the same native asset, with CGT chains explicitly blocked from joining?
-
-- Is explicit blocking of CGT chains from joining non-CGT interop sets necessary?
+- **Withdrawal proof invalidation**: All prior withdrawal proofs are invalidated during migration. Is this accepted for v1, or do we need to design a way to preserve in-flight proofs? May have a different answer for initial release vs future chain additions.
 
 ### Trust assumptions
 
--- a/protocol/proofs/shared-dispute-game-migration.md
+++ b/protocol/proofs/shared-dispute-game-migration.md
@@ -68,7 +68,7 @@ The migrator currently casts portals to `IOptimismPortalInterop` (`OPContractsMa
 
 Before any migration step, several invariants must hold.
 
-**Governance alignment.** All chains must share the same ProxyAdmin owner (`OPContractsManagerMigrator.sol:96`) and reference the exact same SuperchainConfig address (`OPContractsManagerMigrator.sol:102`). Shared infrastructure is initialized with a single ProxyAdmin and SuperchainConfig — the ETHLockbox enforces this at authorization time (`ETHLockbox.sol:220-224`).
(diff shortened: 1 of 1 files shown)


protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,216 @@
...
+
+Rejected. Creates a permanent fork where CGT chains can never participate in interop.
+
+## Risks & Uncertainties
+
+### Open questions
+
+- Can CGT chains participate in shared dispute games without pooling ETH, or must all chains share the same native asset, with CGT chains explicitly blocked from joining?
+
+- Is explicit blocking of CGT chains from joining non-CGT interop sets necessary?
```
Joxess: I believe it is possible for CGT chains join the interop set and shared dispute game, we might want to skip the lockbox migration step somehow for those

More importantly (not related to this pr), it requires some workaround to enable the right predeploys in the interop fork for CGT chains, especifically not deploying/enabling the current `SuperchainETHBridge`/`ETHLiquidity` contracts

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Trust assumptions
+
+- **ProxyAdmin owner coordination**: All chains must trust the same ProxyAdmin owner to execute the migration correctly. A malicious or compromised owner can corrupt all chains simultaneously.
+
+### Operational gaps
+
+- **No rollback mechanism**: If the migration succeeds but a bug is discovered in the shared infrastructure, there is no `undo()`. Fixing requires deploying new shared contracts and migrating again (which the current migrator does not support — see re-migration constraint).
+- **Orphaned in-progress games**: Games created on old per-chain DGFs before migration can still resolve, but their results are not tracked by the new shared ASR. Users who proved withdrawals against these games must re-prove against the new DGF.
+- **No partial migration support**: There is no way to migrate N-1 chains and add the Nth later.
```
Joxess: This is more like a future feature to consider, right? And not a hard restriction

stevennevins: yeah it's a constraint for what would be the first release of the feature and then we iterate to remove these constraints later

protocol/proofs/shared-dispute-game-migration.md
stevennevins: ah that's a good call out

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,216 @@
...
+
+**Transitional state:** After this step, chains function independently with super game types on their per-chain DGF. This is functional but NOT the target state — shared infrastructure comes in Step 2.
+
+### 4. Step 2: Atomic Multi-Chain Migrate (`OPContractsManagerMigrator.migrate()`)
+
+This step runs ONCE, atomically, for ALL chains in the migration set.
+
+**What happens:**
+
+1. **Deploy shared infrastructure** (`OPContractsManagerMigrator.sol:125-156`): new ETHLockbox, DisputeGameFactory, and AnchorStateRegistry proxies. Reuses existing DelayedWETH.
```
Joxess: I think it requires a new DelayedWETH as well

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,216 @@
...
+Finally, the current codebase has two portal contracts: `OptimismPortal2` (live, supports CGT chains) and `OptimismPortalInterop` (not live, has migration functions but lacks CGT support). Shipping interop through OPInterop would permanently fork the portal — CGT chains could never participate, and every security fix would need to land in both contracts. Merging OPInterop's migration features into OP2 eliminates this split.
+
+## Proposed Solution
+
+### 1. Portal Unification (Merge OPInterop into OP2)
+
+**What changes in `OptimismPortal2`:**
+
+Add from `OptimismPortalInterop`:
+- `migrateToSuperRoots(IETHLockbox _newLockbox, IAnchorStateRegistry _newAnchorStateRegistry)` — swaps lockbox and ASR references, emits `PortalMigrated` event (from `OptimismPortalInterop.sol:381-414`)
```
ajsutton: Should we rename these to be `migrateToSharedDisputeGame`? Individual chains should all be on super roots and don't need to swap lockboxes/ASR as part of that process.

protocol/proofs/shared-dispute-game-migration.md
ajsutton: Could go either way.  We will wind up with a single DelayedWETH for all chains but it's fine if it comes from one of the chains being merged or if its new.

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,216 @@
...
+| Component | Before | After |
+|---|---|---|
+| DisputeGameFactory | Per-chain | Shared across all chains |
+| OptimismPortal | Points to per-chain DGF/ASR/Lockbox | Points to shared DGF/ASR/Lockbox |
+| AnchorStateRegistry | Per-chain | Shared across all chains |
+| ETHLockbox | Per-chain (or none) | Shared across all chains |
+| Game types | CANNON, PERMISSIONED_CANNON, CANNON_KONA | SUPER_CANNON_KONA, SUPER_PERMISSIONED_CANNON |
+| Proof method | `rootClaim()` (single chain) | `rootClaimByChainId(chainId)` (per-chain from super root) |
+
+**All prior withdrawal proofs are invalidated.** The ASR reference changes — old proofs reference games from the old DGF, which the new ASR does not track. Users must re-prove (`OPContractsManagerMigrator.sol:62-64`).
```
ajsutton: Are we accepting that or do we need to design a way to avoid invalidating withdrawals? @karlfloersch 

We can have a different answer for this initial release and for adding future chains (probably - have to work out the details of that still)

protocol/proofs/shared-dispute-game-migration.md
ajsutton: @karlfloersch What's the planned scope for interop + CGT for this first release?

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Risks & Uncertainties
+
+### Open questions
+
+- Can CGT chains participate in shared dispute games without pooling ETH, or must all chains share the same native asset, with CGT chains explicitly blocked from joining?
+
+- Is explicit blocking of CGT chains from joining non-CGT interop sets necessary?
+
+- The migrate function does not upgrade each chain's portal. This should either happen within `migrate()` itself or in a prior upgrade step.
```
ajsutton: Isn't the upgrade in a prior upgrade step?

protocol/proofs/shared-dispute-game-migration.md
karlfloersch: For the first release, CGT is not a requirement at all and we also don't have any CGT chains which are immediately lined up for interop. So I'm not super stressed about any decision we make here as long as when we do need to add CGT we can add it.

That said... in theory CGT & interop should be compatible as long as the CGT chain doesn't deploy the ETH bridging contracts & we don't perform a lockbox migration. So it would definitely be nice to split lockbox migration from super dispute game.

Lmk if that answers your q @ajsutton

pauldowman: We have https://github.com/ethereum-optimism/optimism/issues/19711 but it's scope is just to think about CGT-related issues and put them on the backlog.

ajsutton: Ok, my suggestion is that if you try to migrate a CGT chain to shared dispute games that it reverts for now.  That way we can't break stuff by moving CGT into shared lockbox.

We will need to ensure that the ETH bridging contracts don't get deployed on CGT chains - that happens on L2 automatically when the fork activates and would be very difficult to undo. 19711 has a comment about that now (thanks Paul) and I think that's sufficient for tracking this.

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Summary
+
+Today, each OP Stack chain runs its own dispute infrastructure: a per-chain `DisputeGameFactory`, `AnchorStateRegistry`, `OptimismPortal2`, and `ETHLockbox`. Each game validates a single chain's output root in isolation.
+
+The migration consolidates N independent chains into a shared set of these contracts, backed by **super dispute games** whose `rootClaim` is a `SuperRootProof` containing output roots for ALL participating chains. This enables cross-chain state commitment at the dispute layer.
+
+Key outcomes:
+- **Portal unification**: `OptimismPortalInterop` features are merged into `OptimismPortal2`, making OP2 the single portal contract. OPInterop is deleted.
+- **Shared infrastructure**: All migrated chains share one `DisputeGameFactory`, one `AnchorStateRegistry`, and one `ETHLockbox`.
```
ajsutton: Let's mention DelayedWETH as shared here too.

protocol/proofs/shared-dispute-game-migration.md
ajsutton: @stevennevins We should probably explicitly state that the retirement timestamp will be updated which invalidates all games as well. The new ASR automatically sets the retirement timestamp to the current block but it's an extra safety feature.

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,221 @@
...
+| Component | Impact if compromised |
+|---|---|
+| Shared DGF | No chain can create new dispute games. In-progress games still resolve. |
+| Shared ASR | Game validity checks (`isGameProper`, `isGameRespected`, `isGameClaimValid`) fail across all chains. |
+| Shared ETHLockbox | All ETH across all chains at risk. Mitigated by authorized portals check and pause mechanism. |
+| SuperchainConfig | Already shared pre-migration. Post-migration, pausing also blocks dispute resolution. |
+
+## Failure Mode Analysis
+
+See the existing FMA document for comprehensive failure mode analysis.
```
ajsutton: Do we have an FMA? Probably worth linking to it here or just remove this if we're going to use threat modelling instead of FMA.

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,221 @@
...
+
+**Two-step failure gap:** Between Step 1 and Step 2, chains run super game types against per-chain infrastructure. If Step 2 is delayed, chains are stuck in a transitional state with no rollback function.
```
ajsutton: We expect most chains to stay in this state using super roots but not in a shared game (or dependency set).  We should probably just not refer to it as a transitional state.

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Impact on Developer Experience
+
+**Users must re-prove withdrawals.** SDKs and bridge UIs need to detect invalidated proofs and guide users through re-proving.
+
+**New game types in tooling.** Off-chain tools must recognize `SUPER_CANNON_KONA` and `SUPER_PERMISSIONED_CANNON`. The `rootClaim` format changes to `SuperRootProof`.
```
ajsutton: Most importantly tools would need to change to use `outputRootForChain` (or whatever its called) rather than just `rootClaim` when checking if an output root is valid for them to use and creating the proof.

ajsutton (commented):
This is looking good.

protocol/proofs/shared-dispute-game-migration.md
stevennevins: Hmm isn't there a transitional state in between before migrate?

stevennevins: or do you mean because it would be atomic?

ajsutton: We will run OPCM upgrade on all chains which moves them from FaultDisputeGame to SuperDisputeGame.  For most chains, that's the end of the story.  For op-mainnet and unchain (initially) we will then, in a separate transaction, run migrate to move them to a shared dispute game.

So both a non-shared SuperDisputeGame and a shared SuperDisputeGame are valid "end states" for a chain to be left in and there's no actual transitional state.

stevennevins: Oh I gotcha

I meant this as the upgrade for the portal to get the migration functions before they can call opcm.migrate

ajsutton: But the migrate would come just from the normal upgrade which is likely the same one that upgrades them to SuperDisputeGame.  Even if its separate, its fine to have a portal that supports migrate but choose not to call it.

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,221 @@
...
+
+**Governance alignment.** All chains must share the same ProxyAdmin owner (`OPContractsManagerMigrator.sol:96`) and reference the exact same SuperchainConfig address (`OPContractsManagerMigrator.sol:102`). Shared infrastructure is initialized with a single ProxyAdmin and SuperchainConfig — the ETHLockbox enforces this at authorization time (`ETHLockbox.sol:220-224`).
+
+**Portal version.** The portal MUST already be upgraded to the version with migration functions. If not, `migrateToSharedDisputeGame()` reverts (`OPContractsManagerMigrator.sol:277-279`).
+
+**Feature flags.** The `OPTIMISM_PORTAL_INTEROP` dev feature flag must be enabled on the OPCM container (`OPContractsManagerMigrator.sol:80`).
+
+### 3. Step 1: Per-Chain OPCM Upgrade (`OPContractsManagerV2.upgrade()`)
+
+This step runs once per chain BEFORE the atomic migration.
```
wwared: A question worth figuring out: this step is going to run for _every_ chain, or only those that effectively migrate to the SharedDisputeGame? Specifically this step upgrades the portal to a version that contains the migration functions, and those functions are only used by the migration.

I think the better option is going to be to only run this step for the chains being migrated to SharedDisputeGame, which would make more sense then for this upgrade to be collapsed into the `migrate()` function call so it all happens at once.

Unless there is some reason why we would want to unconditionally upgrade every portal to include the migration functions. One reason might be to simplify the number of different portal implementations we need to consider

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,221 @@
...
+The migrator currently casts portals to `IOptimismPortalInterop` (`OPContractsManagerMigrator.sol:244`). After unification, it casts to `IOptimismPortal2` instead, since the unified interface lives on OP2.
+
+### 2. Pre-Migration Constraints
+
+Before any migration step, several invariants must hold.
+
+**Governance alignment.** All chains must share the same ProxyAdmin owner (`OPContractsManagerMigrator.sol:96`) and reference the exact same SuperchainConfig address (`OPContractsManagerMigrator.sol:102`). Shared infrastructure is initialized with a single ProxyAdmin and SuperchainConfig — the ETHLockbox enforces this at authorization time (`ETHLockbox.sol:220-224`).
+
+**Portal version.** The portal MUST already be upgraded to the version with migration functions. If not, `migrateToSharedDisputeGame()` reverts (`OPContractsManagerMigrator.sol:277-279`).
+
```
wwared: Might be helpful to clear up some of the operational steps/concerns, for example:

1. The upgrade can be run after the ProxyAdmin owner migration by the new consolidated owner, or before the owner migration by the previous ProxyAdmin. Which one do we want? Should we support both cases?
2. What's the process for the initial setup of the shared contracts, e.g. what needs to be deployed, what needs to be reconfigured, and what gets used as-is? What is different from the migration from the perspective of the first chain setting up the shared infrastructure vs the later ones that simply start using the existing shared contracts?

protocol/proofs/shared-dispute-game-migration.md
wwared: I think it's better to be explicit on whether or not we will migrate every chain to a portal that supports migrate, or if we only migrate the portal for chains that get migrated.

There's a trade-off to consider: migrating unconditionally means we have less contract implementations to support and consider for future upgrades, migrating conditionally means more implementations to consider but the on-chain surface area becomes more minimal (i.e. chains that don't want shared dispute gate functionality don't have the migration functions available in their portals)

protocol/proofs/shared-dispute-game-migration.md
ajsutton: This step runs for _every_ chain. We don't want to introduce different versions of the portal for different chains or it will make future upgrades and contract releases much harder. It's totally fine to have a function available that chain operators choose not to call.

protocol/proofs/shared-dispute-game-migration.md
stevennevins: Linked to the existing one for now and we can make it current if necessary

protocol/proofs/shared-dispute-game-migration.md
stevennevins: Yeah i think moving the functions and having one Portal is the way to go


I think the detail i was assuming here is that. In the next upgrade, we will upgrade the dispute games to the SUPER_* variants.

Then, in the following upgrade, which is the one sequencing-wise I'm covering in this doc. The OptimismPortal2 upgrade happens to give chains the migrate-related functions, and `opcm.migrate` occurs.

I think we could do the portal upgrades inside of migrate, but right now that logic doesn't exist there or we upgrade the portals along with the SUPER* game migration upgrade

protocol/proofs/shared-dispute-game-migration.md
ajsutton: The ProxyAdmin must be the same for all chains. So the ProxyAdmin is the same for all systems (unchanged) and the shared components.

protocol/proofs/shared-dispute-game-migration.md
ajsutton: Ah ok, we want to upgrade all chains to the new portal implementation that has `migrate`. And then some chains will call `migrate` to move to a shared dispute game. So we don't want to do the portal upgrades inside migrate because we want them to apply to every chain and not every chain will want to migrate to shared dispute games.

In terms of what release things are in, it's unclear if there'll be a separate release for switching to SUPER_ games or not but the concept is still the same. We apply the upgrades that all chains get via opcm.upgrade so they all wind up on SUPER_ games with a portal that has a migrate function and then only some chains actually perform the migration to shared dispute games.

protocol/proofs/shared-dispute-game-migration.md
ajsutton: Definitely every one is migrating to SUPER_ games and to portals that support migration. Then only some chains actually migrate to a shared dispute game.

ajsutton (approved):

protocol/proofs/shared-dispute-game-migration.md
Joxess: That makes sense. I'll gladly follow up when necessary

protocol/proofs/shared-dispute-game-migration.md
```diff
@@ -0,0 +1,217 @@
...
+The migration consolidates N independent chains into a shared set of these contracts, backed by **super dispute games** whose `rootClaim` is a `SuperRootProof` containing output roots for ALL participating chains. This enables cross-chain state commitment at the dispute layer.
+
+Key outcomes:
+- **Portal unification**: `OptimismPortalInterop` features are merged into `OptimismPortal2`, making OP2 the single portal contract. OPInterop is deleted.
+- **Shared infrastructure**: All migrated chains share one `DisputeGameFactory`, one `AnchorStateRegistry`, and one `ETHLockbox`. Reuses existing DelayedWETH (`OPContractsManagerMigrator.sol:158-163`).
+- **Two-phase execution**: (1) per-chain OPCM upgrade prepares each chain, then (2) an atomic multi-chain `migrate()` call deploys shared infra and wires everything together.
+- **Proof routing via game type**: `_isSuperGameType()` in OP2 (`OptimismPortal2.sol:642-646`) determines whether to use `rootClaimByChainId()` or `rootClaim()` based on the game type enum. No `superRootsActive` storage flag needed.
```
Joxess, line 23: When a future super game type is added (e.g., `SUPER_ZK_DISPUTE_GAME`), does it need its own enum value and an update to `_isSuperGameType()`, or would it reuse an existing super game type ID with a different implementation on the factory?

Joxess (approved):

protocol/proofs/shared-dispute-game-migration.md
ajsutton: Depends on exactly what the change to the game is, but for SUPER ZK games it would be a new game type and we'd update `_isSuperGameType()`.  Most likely once we ship the migration to super games, we'd invert this check and list the set of known output root games here and assume everything else is super root (or remove this because all reasonably recent games now support the rootForChainId method).

wwared (approved):

PR #371, opened 2026-03-25
protocol: drop overlapping span batches

sebastianst:
## Summary

Explores dropping overlapping span batches as a Karst hardfork derivation rule change and concludes that **the change is not feasible** due to unresolvable implications for fault proofs, node restarts, snap sync, interop, and operational tools.

- Dropping overlapping span batches introduces an implicit requirement that the safe head must be at a span batch boundary — this cannot be guaranteed in practice
- Nine approaches to resolving the fault proof interaction were explored, with concrete attack scenarios for each — none are satisfactory
- **Conclusion:** Overlap checks will be retained in the non-IO derivation pipeline refactor using a deferred L2 block fetching pattern

Related: https://github.com/ethereum-optimism/optimism/pull/19303

🤖 *Generated with [Claude Code](https://claude.com/claude-code)*

### DIFF
--- a/protocol/drop-span-batch-overlap.md
+++ b/protocol/drop-span-batch-overlap.md
@@ -6,47 +6,45 @@
 | Created at         | 2026-03-18                                         |
 | Initial Reviewers  | @ajsutton, @theochap, @inphi                       |
 | Need Approval From | @ajsutton                                          |
-| Status             | Draft                                              |
+| Status             | Final (not implementing)                           |
 
 ## Purpose
 
-This design doc proposes dropping overlapping span batches at the Karst hardfork and explores how to
-maintain fault proof soundness under this new rule.
+This design doc explores dropping overlapping span batches as a derivation rule change and concludes
+that **the change is not feasible** due to unresolvable implications for fault proofs, node restarts,
+and interop. Overlap checks will instead be retained and implemented in any non-IO derivation
+pipeline refactor.
 
-The motivations are:
+The original motivation was:
 
 1. **Simplify the derivation pipeline.** The current span batch overlap verification requires the
    `SafeBlockFetcher` interface to look up historical L2 blocks during batch validation. This
    complicates building a pure derivation pipeline where all L1 inputs are fetched in advance and IO
-   is moved outside the pipeline. Removing the overlap check simplifies the derivation pipeline
+   is moved outside the pipeline. Removing the overlap check would simplify the derivation pipeline
    refactor drafted in [PR #19303](https://github.com/ethereum-optimism/optimism/pull/19303).
-   Alternatively, the overlap check could be retained with a two-pass approach where the pipeline
-   returns the required L2 blocks and the caller provides them (see approach 3 below), but this
-   carries the overlap checking complexity into the new implementation.
 
 2. **Improve fault proof performance.** The current overlap verification in fault proofs requires
    K+1 preimage oracle reads (one parent block reference plus K full block payloads) and O(K×T)
    byte-by-byte transaction comparison, where K is the number of overlapping blocks and T is the
-   average number of transactions per block. Any solution that eliminates the overlap check also
-   eliminates this cost, reducing on-chain dispute cost, off-chain FP VM execution, and ZK proving
-   costs.
+   average number of transactions per block. Eliminating the overlap check would also eliminate this
+   cost, reducing on-chain dispute cost, off-chain FP VM execution, and ZK proving costs.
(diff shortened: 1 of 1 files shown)


protocol/drop-span-batch-overlap.md
```diff
@@ -0,0 +1,310 @@
...
+   - Non-deposit transaction count (must match)
+   - Each transaction byte-for-byte (must match)
+   - L1 origin number (must match)
+
+The `SafeBlockFetcher` interface provides two methods: `L2BlockRefByNumber` and `PayloadByNumber`.
+This interface is threaded through the batch checking call chain from the derivation pipeline.
+
+### Why the overlap check complicates the pure derivation pipeline
+
+The derivation pipeline should ideally depend only on L1 data and the current safe head, with all IO
```
ajsutton, line 72: I think one thing here is that the pipeline would depend on L1 data, the current safe head and _the implicit requirement that the safe head is from a span batch boundary_. We might be able to meet that requirement, but it's an important extra requirement to note and is basically the source of the problem with the fault dispute system.

Even outside of the fault proof system, I'm nervous about making this a strict requirement for derivation to work.

protocol/drop-span-batch-overlap.md
```diff
@@ -0,0 +1,310 @@
...
+Simply dropping overlapping span batches in the fault proof program does not work. The dispute game
+bisects to arbitrary L2 blocks above the `SPLIT_DEPTH` (bisecting over output roots). The agreed
+prestate can place the safe head at any block, including in the middle of a span batch.
+
+If the safe head is mid-span-batch and the overlapping span batch is dropped, the batch that
+originally produced the safe head's blocks would be dropped — and derivation cannot continue.
+
+The fault proof's agreed output root (`OutputV0`) captures L2 execution state (state root, block
+hash) but **not** derivation state (which batch was in progress, the position within it). This gap
+between execution state and derivation state is the core challenge.
```
ajsutton: This actually gets even more difficult with interop.  We'd be bisecting over super roots which are at a particular timestamp and so there's no way to guarantee that each individual chain is at a span batch boundary at that point (or at any point).

protocol/drop-span-batch-overlap.md
```diff
@@ -0,0 +1,310 @@
...
+#### 4. Re-derive from a span batch boundary
+
+Ensure derivation always starts from a span batch boundary, either by re-deriving from the game's
+anchor state or by constraining the dispute game to only bisect to span batch boundaries.
+
+**Problem — too expensive.** Span batches can contain thousands of blocks, and fault proof block
+execution must be over a single block. Re-deriving from a span batch boundary requires executing all
+intermediate blocks up to the disputed block. Constraining the dispute game's bisection to span batch
+boundaries does not help either — bisection above the split depth creates intermediate starting
+points at arbitrary L2 blocks regardless of the anchor state.
```
ajsutton, line 187: The other problem is how do you find out what the span batch boundary actually is without checking every historic span batch?

ajsutton (commented):
Three cases this hits:
* Fault proofs
* Node crashes while applying a span batch and safe head is left part way through it (mitigated by moving to only update safe head once all data from the batch is applied)
* Rewinding the chain with setDebugHead
* Node startup process - where is the L1 block to start from and what safe L2 block should we use? Theoretically we should use the current safe L2 block but op-node often rewinds it for some reason.
* Snap sync might sync to a node in the middle of a span batch.

Fundamentally the difficult question here is how to know which safe block is actually ok to start derivation from. If you take an L2 block and iterate L1 to find the span batch it came from you then have to ask, did that span batch overlap with a previous one?
So you find the previous span batch and check it overlaps or not. But first you need to know if the previous span batch was valid, so you have to find its parent. You wind up having to read the entire batch history in order to validate your starting point and with blobs that starting point won't be available.

protocol/drop-span-batch-overlap.md
sebastianst: updated the doc to conclude we're not implementing this change

protocol/drop-span-batch-overlap.md
sebastianst: updated the doc to conclude we're not implementing this change

protocol/drop-span-batch-overlap.md
sebastianst: updated the doc to conclude we're not implementing this change

ajsutton (approved):

PR #377, opened 2026-04-23
design-doc: enable CircleCI for fork PRs

ajsutton:
## Summary
- Proposes enabling automatic CircleCI feedback on **all** fork PRs to `ethereum-optimism/optimism`, without exposing secrets to fork-authored code.
- Two-phase plan: (1) serialize `mise install` into a single upstream job so GitHub API pressure does not scale with pipeline fanout; (2) flip "Build forked pull requests" ON while hardening contexts, auditing OIDC trust, and adding per-job `when: not is_fork_pr` hygiene gates.
- Documents the exact CircleCI platform guarantees we rely on — secret hiding, OIDC non-issuance, cache isolation, fail-closed restricted contexts — with citations to CircleCI's docs for every security-relevant claim.
- Bailiff is retained as the escape hatch for running the full secret-requiring suite before merge.

## Test plan
- [ ] Design review: confirm the platform-level guarantees cited are the ones reviewers expect to rely on.
- [ ] Confirm CircleCI plan tier (Scale plan required for Config Policies, used as defence in depth).
- [ ] Confirm final list of `when: not is_fork_pr`-gated jobs with CI owners.
- [ ] Decide whether `ops/book/src/ci/pr-authorization.md` updates ship with Phase 2 or separately.

_Draft: looking for reviewer assignment before moving to In Review._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/fork-pr-ci-access.md
+++ b/fork-pr-ci-access.md
@@ -1,0 +1,178 @@
+# Fork PR CI Access: Design Doc
+
+|                    |                                                    |
+| ------------------ | -------------------------------------------------- |
+| Author             | Adrian Sutton                                      |
+| Created at         | 2026-04-24                                         |
+| Initial Reviewers  | _TBD_                                              |
+| Need Approval From | _TBD_                                              |
+| Status             | Draft                                              |
+
+## Purpose
+
+External contributors to `ethereum-optimism/optimism` get no automatic CircleCI feedback on their PRs. They learn whether their change compiles, lints, or passes tests only after a member of the `engineering` team authorizes the PR via the `bailiff` bot, or if they belong to a small number of allowlisted partner orgs.
+
+This is a poor contributor experience and a bottleneck on a small group of maintainers. Every fork PR should automatically get the broadest practical subset of CircleCI feedback, without exposing any secret or privileged resource to fork-authored code.
+
+## Summary
+
+Enable automatic CircleCI on all fork PRs by relying on CircleCI's platform-level secret isolation, not on trust in the PR author. Two separable pieces:
+
+1. **Serialize `mise install`** into one upstream job so GitHub API rate-limit pressure does not scale with pipeline fanout. Independently valuable for trusted CI; prerequisite for fork CI to be reliable.
+2. **Enable fork PRs to run the full `main` workflow** with per-job `when:` gates hiding jobs that are meaningless without secrets, and with context expression restrictions as defence in depth.
+
+Bailiff is retained as the escape hatch for running the full secret-requiring suite before merge. The merge gate continues to require the bailiff-authorized run.
+
+## Problem Statement + Context
+
+### Current behaviour
+
+- **CircleCI project settings.** "Build forked pull requests" is **OFF**. Per the [CircleCI OSS docs][circleci-oss], "By default, CircleCI does not build PRs from forked repositories." CircleCI receives the fork PR webhook, ignores it, and no pipeline is triggered. "Pass secrets to builds from forked pull requests" is also OFF but moot today because no fork pipelines exist.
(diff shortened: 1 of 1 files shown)


fork-pr-ci-access.md
```diff
@@ -0,0 +1,207 @@
...
+
+GHA is deliberately restricted to a small set of security-reviewed workflows. Expanding its surface is out of scope.
+
+### The rate-limit problem
+
+`mise install` runs at the start of most jobs. Each run bursts GitHub API calls to list releases and download binaries. The current config fans out many jobs in parallel that race to populate the cache on a cold branch, each consuming the shared 60 req/hr anonymous budget.
+
+On trusted branches a warm cache hides this. Fork PRs live in their own cache namespace ([cache isolation][circleci-oss]), so they start cold relative to the main repo and the first pipeline per fork would blow the anonymous budget.
+
+### What actually fails without contexts — experiment results
```
ajsutton, line 87: I'm somewhat sceptical that the lack of changes in the branch affected what ran - though it did change circleci config which should trigger everything to rebuild, but I really expected contract upgrade checks and op-deployer tests to fail and they didn't.  That could be because they were skipped or because they get the CI node URL from environment variables rather than context - external PRs still wouldn't get those environment vars so there would be more failures I think.

fork-pr-ci-access.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Purpose
+
+External contributors to `ethereum-optimism/optimism` get no automatic CircleCI feedback on their PRs. They learn whether their change compiles, lints, or passes tests only after a member of the `engineering` team authorizes the PR via the `bailiff` bot, or if they belong to a small number of allowlisted partner orgs.
+
+This is a poor contributor experience and a bottleneck on a small group of maintainers. Every fork PR should automatically get the broadest practical subset of CircleCI feedback, without exposing any secret or privileged resource to fork-authored code.
+
+## Summary
+
+Enable automatic CircleCI on all fork PRs by relying on CircleCI's platform-level secret isolation, not on trust in the PR author. An experiment ([#20301](https://github.com/ethereum-optimism/optimism/pull/20301)) that stripped every `context:` from the config found that only 3 of ~140 jobs actually fail without their contexts, grounding the rest of this proposal in observed data.
```
raffaele-oplabs, line 19: On top of context, optimism repo also uses env variables, I count 17 of them, so at the moment probably many jobs do not fails because they use those variables instead. (We should delete these variable and move them to context)

In order to have a reliable number we would need to first move all these variables to a contex.

fork-pr-ci-access.md
raffaele-oplabs: Env variable as far as I see: https://github.com/ethereum-optimism/design-docs/pull/377/changes#r3136107057

fork-pr-ci-access.md
```diff
@@ -0,0 +1,207 @@
...
+
+| Job | Reason | Context required |
+| --- | --- | --- |
+| `ci-gate` | `ERROR: CIRCLE_API_TOKEN is not set` | `circleci-api-token` |
+| `generate-flaky-tests-report` | `CIRCLE_API_TOKEN length: 0` | `circleci-api-token` |
+| `required-contracts-ci` | Same ci-gate logic | `circleci-api-token` |
+
+(Two further failures — `main` and `contracts-feature-tests` — were workflow-level aggregations of the above.)
+
+Every job referencing `circleci-repo-readonly-authenticated-github-token` passed — cache hits masked any rate-limit concern, and the jobs' GitHub API usage (where any) stayed inside the anonymous budget for a single pipeline. Every job referencing `slack` or `discord` passed too: notification steps are fire-and-forget and do not fail the job when the webhook environment is missing.
```
raffaele-oplabs, line 99: there is a github token in env var that is used instead of the one in circleci-repo-readonly-authenticated-github-token

ajsutton:
I don't have time to push this forward but anyone else is very welcome to pick it up - either reopen the PR or copy it to your own, whatever works.

PR #378, opened 2026-04-27
Add EIP-8130 OP Stack design docs

chunter-cb:
## Summary
- Add a protocol design doc for adopting EIP-8130 on the OP Stack.
- Add a companion security FMA for the EIP-8130 adoption.

## Test plan
- Documentation-only change; no tests run.

### DIFF
--- a/protocol/adopt-eip-8130-on-op-stack.md
+++ b/protocol/adopt-eip-8130-on-op-stack.md
@@ -1,0 +1,404 @@
+# Adopt EIP-8130 on the OP Stack
+
+|                    |                                                    |
+| ------------------ | -------------------------------------------------- |
+| Author             | Chris Hunter                                       |
+| Created at         | 2026-04-13                                         |
+| Initial Reviewers  | TBD                                                |
+| Need Approval From | TBD                                                |
+| Status             | Draft                                              |
+
+## Purpose
+
+This document proposes adopting [EIP-8130](https://eips.ethereum.org/EIPS/eip-8130) on OP Stack
+L2s and identifies the protocol, client, and tooling work required to ship it safely.
+
+## Summary
+
+EIP-8130 introduces a new typed transaction and an Account Configuration system that enables
+native account abstraction with custom authentication, batching, gas sponsorship, and portable
+accounts and owner management.
+
+We propose that the OP Stack adopt 8130 as an L2 protocol feature in a hardfork. The
+implementation should target semantic compatibility with the upstream EIP wherever possible and
+should not require any OP Stack-specific divergence, as no EVM changes / opcodes are needed.
+
+This document is an adoption proposal, not a replacement for the upstream 8130 spec. The EIP
+should remain the normative source for exact transaction encoding and execution behavior.
+
+## Problem Statement + Context
+
+Existing account abstraction solutions such as ERC-4337 address many UX problems but do so outside
+of the base transaction model. That approach requires specialized mempools, bundlers, and
+validation of wallet behavior before inclusion. This approach adds latency, cost, and integration burden.
+
+EIP-8130 offers a native account abstraction standard. Authentication is separated from wallet logic
+through verifier contracts and account configuration storage. Transactions explicitly identify the
+sender, payer, and authentication path. This gives sequencers and nodes better visibility into
+validation cost and opens a path to native support for passkeys, sponsorship, multisig, and new
+signature schemes without forcing every validator to simulate arbitrary wallet code during
+transaction acceptance.
+
+Other account abstraction approaches can mitigate some of the same UX problems, but they often do
+so with more complex mempool rules and a less uniform model across account types. 8130 instead
+provides a standard transaction surface for EOAs, existing smart accounts, and new smart accounts
+while keeping validation explicit and comparatively simple.
+
+That simplicity matters for both nodes and builders. In proposals where validity depends on broader
+wallet execution or more dynamic state, ingress may require simulation and tracing just to
+determine whether a transaction is admissible, and builders take on more wasted compute and packing
+risk when invalidation is harder to track. 8130 aims to directly address these issues while still providing 
+flexibility. It also gives wallets a clearer standard to target for cross-chain support, and directly 
+includes features we care about such as 2D nonces, key rotation, and transaction expiration.
+
+## Proposed Solution
+
+We propose to adopt EIP-8130 on the OP Stack via a scheduled hardfork and to treat the upstream
+EIP as the source of truth for execution semantics.
+
+### Scope of Adoption
+
+The OP Stack implementation should include the core 8130 feature set:
+
+- the new AA transaction type
+- the Account Configuration system contract
+- the Nonce Manager precompile (if offering 2D or nonceless tx)
+- the Transaction Context precompile 
+- account creation, config changes, and delegation entries in `account_changes`
+- payer authorization and sponsored transactions
+- call phases and receipt extensions
+- protocol-injected logs and RPC extensions described by the EIP
+
+The goal should be to ship 8130 as an additive feature. Existing EOAs, smart contracts, and
+ERC-4337 flows should continue to work. 8130 should provide a native path for wallets that want
+protocol-level account abstraction, not require the ecosystem to abandon existing wallet patterns.
+
+### OP Stack Integration
+
+Adopting 8130 on the OP Stack requires coordinated work across the protocol and the surrounding
+stack:
+
+- `op-reth` must implement the new transaction type, intrinsic gas accounting, validation flow,
+  state transition logic, precompile behavior, receipt fields, and protocol-injected logs.
+- `op-node`, sequencer transaction ingestion, and batch submission paths must support the new typed
+  transaction end to end.
+- fixed addresses for the Account Configuration contract, Nonce Manager, Transaction Context,
+  initial verifier set, and default account implementation must be standardized for chains
+  that opt into the feature. These addresses are TBD until the upstream spec is final.
+- SDKs, wallets, explorers, and indexers must be updated to understand the new transaction type,
+  receipt shape, and RPC extensions.
+
+The required contract surface should remain limited. Once the final 8130 spec is agreed upon, OP
+Stack chains should include the Account Configuration contract and the initial verifier set at
+standard addresses. The remaining required functionality is expected to be precompile-based.
+
+### Spec Dependencies to Pin
+
+The OP Stack adoption should be conditional on pinning the final upstream EIP values and artifacts
+before activation:
+
+- `AA_TX_TYPE` and payer domain-separation byte
+- fixed addresses for the Account Configuration contract, Nonce Manager precompile, Transaction
+  Context precompile, initial verifier set, and default account implementation
+- Account Configuration contract bytecode and storage layout
+- canonical verifier bytecode and deployment process
+- default account implementation bytecode
+- precompile interfaces and gas behavior
+- receipt fields and RPC extension shapes
+
+### Activation, Derivation, and System Transactions
+
+8130 should activate in a named OP Stack hardfork. The fork name and activation schedule are TBD.
+Before activation, the new transaction type must be rejected. After activation, it should be
+accepted as a standard [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) typed transaction.
+
+Because 8130 is an EIP-2718 transaction type, no special derivation rule changes are expected.
+Batches should continue to carry opaque transaction bytes, and execution validity should be enforced
+by the execution client after activation. `op-node` and the batcher still need compatibility work so
+that the new transaction type is accepted, propagated, included, and decoded correctly throughout the
+stack.
+
+Deposit transactions and OP Stack system transactions should not have special 8130 behavior in the
+initial adoption. They should continue to use their existing transaction paths. Future upgrades may
+choose to leverage 8130 for deposits or system transactions, but that is not part of this proposal.
+
+### OP Stack Policy
+
+8130 allows nodes to maintain a verifier allowlist, and the OP Stack should help define a canonical
+verifier set as part of the standard. This verifier set should be agreed upon by the client and
+chain teams implementing 8130.
+
+OP Stack chains should agree on which verifiers belong in this canonical set, and chains that claim
+conformance with the standard must accept every verifier in the set. This gives wallets and SDKs a
+single standard baseline to target across chains.
+
+Chains may choose to enable additional verifiers beyond the canonical set, but that behavior should
+be treated as "out of spec". Wallets and applications should not assume those extensions are
+available on other OP Stack chains.
+
+The OP Stack will adopt the exact 8130 canonical verifier set agreed upon by the teams implementing
+the standard. For an initial rollout, the canonical set is planned to be conservative:
+
+- Secp256k1
+- P256
+- P256 WebAuthn / passkey
+- Delegate, enabling a sub-account model
+
+No other validation paths are considered part of the OP Stack 8130 spec at this time. Additional
+canonical verifiers would need to be agreed upon by the teams implementing 8130 and, ideally,
+standardized upstream.
+
+Protocol-native verifier paths are acceptable when their behavior and standard gas costs are defined
+at the fork. Delegate should charge an additional SLOAD cost for resolving the delegated owner, plus
+the underlying wrapped signature verification cost. Verifier contracts themselves must not be
+delegated accounts.
+
+### Consensus and Mempool Validation Boundary
+
+The consensus requirements for an included 8130 transaction should be lighter and more mechanical
+than the full mempool policy. At minimum, a valid included 8130 transaction means:
+
+- `sender_auth` was valid at inclusion time
+- `payer_auth`, when present, was valid at inclusion time
+- the payer had enough ETH to cover the transaction
+- the nonce was valid under the chain's supported nonce mode
+- `account_changes` were valid
+- sender authentication was validated again at the end of the flow
+
+Other limits should be treated as mempool policy and agreed upon by client teams. Examples include
+limits on owner changes, payer authorization complexity, and other transaction shape constraints
+that prevent expensive or pathological validation before inclusion.
+
+The OP Stack may choose whether to support nonceless 8130 transactions. An initial deployment could
+also defer nonceless mode if that produces a simpler and safer first fork.
+
+### Account Semantics Adopted from 8130
+
+The OP Stack should explicitly adopt the following EIP semantics:
+
+- a code-empty EOA that sends its first 8130 transaction is auto-delegated to
+  `DEFAULT_ACCOUNT_ADDRESS` unless the transaction creates the account or supplies its own
+  delegation entry
+- calls carry no protocol-level ETH value; ETH transfers must be initiated by wallet bytecode
+- each call executes with `msg.sender = from`, `tx.origin = from`, and `msg.value = 0`
+- calls execute in ordered phases; each phase is atomic, completed phases persist, and later phases
+  are skipped after a phase failure
+- owner scope checks and account lock behavior are consensus-critical
+- config changes with `chain_id = 0` are intentionally portable across chains and should be treated
+  as an audit focus
+
+Auto-delegation is a significant UX and safety surface. The default account implementation must be
+pinned and audited before activation, and wallet UX should make clear when a user's account will be
+delegated by sending an 8130 transaction.
+
+### Delegation and Existing OP Stack Semantics
+
+8130 reuses the 7702 delegation-indicator model. This means adoption must be checked against existing
+OP Stack logic that treats EOAs and contracts differently.
+
+At minimum, implementation work should explicitly review:
+
+- address aliasing behavior for delegated accounts
+- bridge convenience methods and any EOA-only checks
+- assumptions about `tx.origin`, `msg.sender`, and code presence
+- tooling that currently treats "has code" and "is a contract" as equivalent
+
+The expected direction is that OP Stack behavior should remain internally consistent with the
+mental model already established for delegated EOAs: a delegated account may have code, but it is
+not equivalent to a general smart contract for every protocol decision.
+
+### Rollout
+
+Because 8130 touches consensus logic, node policy, and user-facing RPC behavior, it should be
+rolled out in stages:
+
+1. ratify the adoption design and complete a companion FMA
+2. choose constants and fixed addresses in coordination with the upstream EIP
+3. prototype the feature in `op-reth`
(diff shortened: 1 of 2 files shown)


protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+protocol-level account abstraction, not require the ecosystem to abandon existing wallet patterns.
+
+### OP Stack Integration
+
+Adopting 8130 on the OP Stack requires coordinated work across the protocol and the surrounding
+stack:
+
+- `op-reth` must implement the new transaction type, intrinsic gas accounting, validation flow,
+  state transition logic, precompile behavior, receipt fields, and protocol-injected logs.
+- `op-node`, sequencer transaction ingestion, and batch submission paths must support the new typed
```
joshklop, line 83: > `op-node`

would this be kona-only or also `op-node` cc @geoknee? Should be light changes on the CL either way iiuc.

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+
+1. ratify the adoption design and complete a companion FMA
+2. choose constants and fixed addresses in coordination with the upstream EIP
+3. prototype the feature in `op-reth`
+4. validate mempool behavior, tooling compatibility, and verifier policy in devnets
+5. complete multiple audit cycles across the EIP, OP Stack implementation, verifier contracts, and
+   wallet integration flows
+6. ship in a named OP Stack hardfork after testnet coverage and async review
+
+The devnet checklist should include:
```
joshklop, line 223: Can (most of) these be written as acceptance tests to catch bugs in CI prior to full devnets?

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+- auto-delegation to the default account implementation and explicit delegation override behavior
+- owner scope checks, account lock behavior, and portable `chain_id = 0` config changes
+- phased call execution, including `msg.value = 0`, persisted completed phases, and skipped later
+  phases after failure
+- wallet, SDK, explorer, and indexer integration tests owned by Base for the initial rollout
+
+Implementation ownership should be made explicit before approval:
+
+- `op-reth` implementation and consensus tests
+- `op-node`, sequencer ingestion, and batcher compatibility
```
joshklop, line 242: similar to the above: we should clarify what CLs are in scope (op-node, kona, or both)

security/adopt-eip-8130-on-op-stack-fma.md
```diff
@@ -0,0 +1,189 @@
...
+  intrinsic gas, verifier behavior, account changes, nonce handling, receipt fields,
+  protocol-injected logs, or precompile behavior. This can produce invalid block acceptance or an
+  execution rule that must be fixed by client release or hardfork.
+- **Risk Assessment:** High severity, possible likelihood.
+- **Mitigations:**
+  - Pin all upstream EIP constants, bytecode, storage layouts, fixed addresses, precompile behavior,
+    and RPC/receipt shapes before activation.
+  - Create positive and negative consensus vectors for every transaction path.
+  - Run devnets with the final `op-reth` implementation.
+  - Complete multiple audit cycles across spec, `op-reth`, contracts, and wallet flows.
```
joshklop, line 54: > Complete multiple audit cycles across... `op-reth`

Highlighting this for discussion. Do we want op-reth changes to be in scope this time?

protocol/adopt-eip-8130-on-op-stack.md
geoknee: Yes I think this should call out both CL implementations.

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+- Account Configuration contract bytecode and storage layout
+- canonical verifier bytecode and deployment process
+- default account implementation bytecode
+- precompile interfaces and gas behavior
+- receipt fields and RPC extension shapes
+
+### Activation, Derivation, and System Transactions
+
+8130 should activate in a named OP Stack hardfork. The fork name and activation schedule are TBD.
+Before activation, the new transaction type must be rejected. After activation, it should be
```
geoknee, line 112: Calling out some prior art, very similar to how we dealt with EIP-7702 transactions in the Isthmus fork https://github.com/ethereum-optimism/optimism/blob/09ef0299a1a9a9d81553291942fb9e169cbbdfb6/op-e2e/actions/proofs/isthmus_setcode_tx_test.go#L148.

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+
+- positive and negative transaction validity vectors for the new EIP-2718 transaction type
+- verifier acceptance and rejection behavior for the canonical verifier set
+- payer authorization, insufficient-balance, nonce, expiry, and `account_changes` failure cases
+- receipt fields, protocol-injected logs, and RPC compatibility
+- `eth_getAcceptedVerifiers`, extended `eth_getTransactionCount` with `nonceKey`, receipt `payer`,
+  and receipt `phaseStatuses`
+- mempool policy limits for transaction shape, owner changes, payer authorization, and validation
+  cost
+- auto-delegation to the default account implementation and explicit delegation override behavior
```
karlfloersch, line 233: Is the idea that we default delegate to a particular pre-deploy and thereby enable a default sig scheme that the chain maintainer can update over time?

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+- The final transaction type, reserved addresses, default wallet bytecode, and canonical verifier
+  set bytecode and addresses are not yet settled.
+- If chains or applications come to depend on verifiers outside the canonical verifier set, wallet
+  interoperability may still fragment.
+- Delegation semantics must be audited carefully anywhere the OP Stack currently distinguishes EOAs
+  from contracts.
+- The change spans consensus logic, node policy, RPC, tooling, and wallet infrastructure, so
+  rollout coordination is non-trivial.
+- Initial deployments may need tighter verifier and nonceless-mode limits than the upstream EIP
+  theoretically permits.
```
karlfloersch, line 404: I have discussed designs where we are able to use EIP 8130 to achieve backwards compatible signatures by writing a verifier which authorizes an EIP 8130 transaction with equivalent semantic meaning as the original transaction format. That does mean; however, that it does not use the passed hash at all and does most via TX_CONTEXT. 

Is this usage intended/encouraged?

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+For example, the OP Stack could adopt verifier and account-configuration concepts without adopting
+the transaction type. This would shrink the initial scope, but it would also leave much of the UX
+benefit on the table and complicate the overall design.
+
+## Open Decisions Before Approval
+
+The following decisions should be resolved, or explicitly accepted as TBD, before this proposal is
+ready for approval:
+
+- fork name and activation schedule
```
geoknee, line 378: I think we wouldn't commit to this here; we tend to run a "turn up and go" approach to forks/upgrades. So when the work is done/merged, we can switch on some feature toggles such that it activates with the next upgrade that has an open scope for features. 

So predicting that from where we stand now comes down to estimating how long it takes to do the work (plus tests and auxiliary tasks like governance posts, documentation, threat modelling etc) and get it merged.

geoknee: More information on the process end to end is here https://github.com/ethereum-optimism/pm/blob/main/src/sdlc.md.

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+- exact canonical verifier bytecode and deployment process
+- whether the first OP Stack deployment supports nonceless transactions or defers `NONCE_KEY_MAX`
+  while supporting standard and user-defined nonce keys
+- default account implementation bytecode and wallet UX requirements for auto-delegation
+- agreed mempool policy limits for validation cost, owner changes, payer authorization, and
+  transaction shape
+- final RPC extension shapes, including accepted-verifier discovery, 2D nonce queries, and receipt
+  extensions
+- final `op-reth` implementation scope and required consensus test coverage
+- audit owners, audit scope, and signoff criteria for each audit cycle
```
geoknee, line 390: References for this https://github.com/ethereum-optimism/pm/blob/main/src/audits.md and https://gov.optimism.io/t/op-labs-audit-framework-when-to-get-external-security-review-and-how-to-prepare-for-it/6864

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+
+## Risks & Uncertainties
+
+- 8130 is still draft and may continue to change upstream.
```
geoknee, line 394: When might we be able to consider 8130 "frozen"?

protocol/adopt-eip-8130-on-op-stack.md
chunter-cb: The wallet validation logic and the wallet code itself is decoupled. Verifiers supply validation logic and wallets add/configure those via the account configuration contract.

The default code would be very simple and just have executeBatch and receiver hooks. Its not expected to be upgraded often, but I agree this should just be a chain owned precompile. The interface the wallet implements should be a minimal agreed standard.

protocol/adopt-eip-8130-on-op-stack.md
chunter-cb: Goal is as soon as possible but working on getting feedback from external teams.

EIP will move out of draft once we have agreed upon the spec.

protocol/adopt-eip-8130-on-op-stack.md
chunter-cb: Though 8130 technically does allow for any logic to go into a verifier the 8130 spec will require chains to not allow these custom verifiers on the AA txn path for validation (if they do its considered "out of spec"). The goal is for wallets to have a highly portable set of validation methods that they are sure are available everywhere. 

We believe the number of validation methods will be small (k1, p256, webauth, ...) and should just be agreed upon via the EIP itself or a companion ERC that gets updated (how consensus works here is still up for debate but if there is a good reason to add a validation method, it shouldnt be difficult to get support as adding a new one just means permissionless deployment + a mempool rule change.)

security/adopt-eip-8130-on-op-stack-fma.md
chunter-cb: Base has a fork of op-reth, we discussed we may be able to include op-reth in the audit itself given minimal changes. Will continue tracking this

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+
+We propose to adopt EIP-8130 on the OP Stack via a scheduled hardfork and to treat the upstream
+EIP as the source of truth for execution semantics.
+
+### Scope of Adoption
+
+The OP Stack implementation should include the core 8130 feature set:
+
+- the new AA transaction type
+- the Account Configuration system contract
```
geoknee, line 64: Is this a precompile?

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+
+8130 should activate in a named OP Stack hardfork. The fork name and activation schedule are TBD.
+Before activation, the new transaction type must be rejected. After activation, it should be
+accepted as a standard [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) typed transaction.
+
+Because 8130 is an EIP-2718 transaction type, no special derivation rule changes are expected.
+Batches should continue to carry opaque transaction bytes, and execution validity should be enforced
+by the execution client after activation. `op-node` and the batcher still need compatibility work so
+that the new transaction type is accepted, propagated, included, and decoded correctly throughout the
+stack.
```
geoknee, line 119: Again just calling out some prior art to guide the implementation https://github.com/ethereum-optimism/optimism/pull/14197

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+
+Deposit transactions and OP Stack system transactions should not have special 8130 behavior in the
+initial adoption. They should continue to use their existing transaction paths. Future upgrades may
+choose to leverage 8130 for deposits or system transactions, but that is not part of this proposal.
+
+### OP Stack Policy
+
+8130 allows nodes to maintain a verifier allowlist, and the OP Stack should help define a canonical
+verifier set as part of the standard. This verifier set should be agreed upon by the client and
```
geoknee, line 128: Any thoughts on how that canonical verifier set should be maintained or expressed? I would suggest hardcoding it into the node software as a default and allowing it to be overridden by command line flags.

geoknee: Just to clarify my understanding also that the allowlist is essentially a sequencer policy, not baked into the protocol? This would allow it to be extended without the need for hardforks.

geoknee: Here's another place where the standard allowlist could be expressed https://specs.optimism.io/protocol/configurability.html?highlight=standard#op-stack-configurability.

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+
+8130 reuses the 7702 delegation-indicator model. This means adoption must be checked against existing
+OP Stack logic that treats EOAs and contracts differently.
+
+At minimum, implementation work should explicitly review:
+
+- address aliasing behavior for delegated accounts
+- bridge convenience methods and any EOA-only checks
+- assumptions about `tx.origin`, `msg.sender`, and code presence
+- tooling that currently treats "has code" and "is a contract" as equivalent
```
geoknee, line 204: Prior art here https://github.com/ethereum-optimism/optimism/pull/14506

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+not equivalent to a general smart contract for every protocol decision.
+
+### Rollout
+
+Because 8130 touches consensus logic, node policy, and user-facing RPC behavior, it should be
+rolled out in stages:
+
+1. ratify the adoption design and complete a companion FMA
+2. choose constants and fixed addresses in coordination with the upstream EIP
+3. prototype the feature in `op-reth`
```
geoknee, line 217: We would typically want to merge this to our `develop` branch incrementally, but behind a feature flag (https://github.com/ethereum-optimism/design-docs/blob/main/protocol/decoupled-features.md). This allows us freedom to activate the feature in e2e tests, and to easily schedule it into a hardfork when ready.

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+data, which increases L2 data footprint.
+
+These tradeoffs are acceptable if the OP Stack keeps the initial verifier profile conservative and
+standardizes enough node policy to avoid pathological validation costs.
+
+### Single Client and Consensus Considerations
+
+New OP Stack feature development, including the next Karst hardfork, is expected to happen on
+`op-reth` only. `op-geth` remains supported for security patches and critical bug fixes through May
+31, 2026, after which support ends. Because 8130 targets new hardfork functionality, the adoption
```
geoknee, line 268: Agreed!

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+
+## Open Decisions Before Approval
+
+The following decisions should be resolved, or explicitly accepted as TBD, before this proposal is
+ready for approval:
+
+- fork name and activation schedule
+- fixed addresses for the Account Configuration contract, precompiles, verifier set, and default
+  account implementation
+- exact canonical verifier bytecode and deployment process
```
geoknee, line 381: I am wondering if this could follow our new L2CM (L2 Contracts Manager) process which will go live soon https://github.com/ethereum-optimism/design-docs/pull/351

geoknee:
I had a question about the relation between 8130 and [EIP 8141](https://eips.ethereum.org/EIPS/eip-8141). If I understand correctly, the two EIPs attempt to save the same problem, but there is a concern with 8141 that it would allow users to DoS the sequencer and cause unbounded unpaid EVM compute. 8130 solves this via an allowlist of verifier contracts. 

Aligning with L1 is a desirable thing in general. Often there are L2 specific concerns which justify a divergence; but I wonder what a minimal diff might look like? Namely, EIP-8141 plus an allowlist of verifier contracts. Is such a thing feasible? What would the tradeoffs be?

protocol/adopt-eip-8130-on-op-stack.md
chunter-cb: It is not baked into the protocol correct. 

It would look like a mempool rule which would just default to what we define in the EIP.

K1, P256, Webauth, Delegate is likely first 4 (will decide this for EIP finalization).

Extending this list will either be with modifications to the EIP itself or by a companion ERC.

protocol/adopt-eip-8130-on-op-stack.md
chunter-cb: It has a solidity version for non 8130 chains. 

It can be implemented as a precompile on 8130 chains though, these both share the same interface. 

We could keep it as solidity as well, and the protocol just reads/updates storage slots directly

protocol/adopt-eip-8130-on-op-stack.md
chunter-cb: Thanks

protocol/adopt-eip-8130-on-op-stack.md
chunter-cb: That should work!

chunter-cb:
> I had a question about the relation between 8130 and [EIP 8141](https://eips.ethereum.org/EIPS/eip-8141). If I understand correctly, the two EIPs attempt to save the same problem, but there is a concern with 8141 that it would allow users to DoS the sequencer and cause unbounded unpaid EVM compute. 8130 solves this via an allowlist of verifier contracts.
> 
> Aligning with L1 is a desirable thing in general. Often there are L2 specific concerns which justify a divergence; but I wonder what a minimal diff might look like? Namely, EIP-8141 plus an allowlist of verifier contracts. Is such a thing feasible? What would the tradeoffs be?

8141 is under development, and spec is changing but can answer this as it is now:
8141 requires the account to approve the transaction and therefore there is no way to run an "approved verifier allowlist". In 8141 the validation method is not known in the transaction type, where as 8130 has the validation method explicit in the tx top level (via the verifier). 8141 tx must be ran in order to know it breaks the rules (typically erc 7562 rules). 

8130 accounts can work in 8141 system as well (but account needs to be upgraded to just call account config).  

There are quite a few differences but 8130 is designed to be more performant and more opinionated where as 8141 leans in to being fully generic and let's follow up specifications try to define logic. Note that 8141 is not generic enough to enable what 8130 provides however (it does not allow approval delegation to external contracts).

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+The goal should be to ship 8130 as an additive feature. Existing EOAs, smart contracts, and
+ERC-4337 flows should continue to work. 8130 should provide a native path for wallets that want
+protocol-level account abstraction, not require the ecosystem to abandon existing wallet patterns.
+
+### OP Stack Integration
+
+Adopting 8130 on the OP Stack requires coordinated work across the protocol and the surrounding
+stack:
+
+- `op-reth` must implement the new transaction type, intrinsic gas accounting, validation flow,
```
pochenai, line 81: Should the design explicitly call out a separate mempool? Validation and mempool are different layers: validation defines admissibility, while the pool handles pending storage, ordering, replacement, eviction, and DoS limits.

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+  state transition logic, precompile behavior, receipt fields, and protocol-injected logs.
+- `op-node`, sequencer transaction ingestion, and batch submission paths must support the new typed
+  transaction end to end.
+- fixed addresses for the Account Configuration contract, Nonce Manager, Transaction Context,
+  initial verifier set, and default account implementation must be standardized for chains
+  that opt into the feature. These addresses are TBD until the upstream spec is final.
+- SDKs, wallets, explorers, and indexers must be updated to understand the new transaction type,
```
pochenai, line 88: It may be worth explicitly requiring op-up support/tests for this tx type.

protocol/adopt-eip-8130-on-op-stack.md
```diff
@@ -0,0 +1,404 @@
...
+be treated as "out of spec". Wallets and applications should not assume those extensions are
+available on other OP Stack chains.
+
+The OP Stack will adopt the exact 8130 canonical verifier set agreed upon by the teams implementing
+the standard. For an initial rollout, the canonical set is planned to be conservative:
+
+- Secp256k1
+- P256
+- P256 WebAuthn / passkey
+- Delegate, enabling a sub-account model
```
pochenai, line 145: Suggest adopting a [Tempo-style access key](https://docs.tempo.xyz/protocol/transactions#access-keys) model instead of full-authority delegation. A delegated/sub-account key should support scoped permissions, e.g. spend limits, allowed targets/selectors/assets, validity windows, and revocation semantics. Full delegation makes compromise of the delegate key equivalent to compromise of the primary account, which seems too strong for the intended sub-account/session-key use case.

chunter-cb: The delegate key type is actually meant to power the sub account model as outlined in this blog post - https://blog.base.dev/subaccounts . So its meant to be that a parent key has full control over its sub accounts, not to be used as a session key on some parent account.

chunter-cb: We are exploring account key policies for session key model here:

https://github.com/ethereum/EIPs/pull/11648/changes 

Key is added and can restrict calls to certain policy contracts (similar to https://github.com/base/account-policies) which will perform any sort of extra validity, spend limits, allowed access rules checks but executed within the EVM. If compliant with the policy the contract can call the wallet to execute the calls. Note this leaves it up to wallet designers to build their policy system. 

Let me know what you think on this + delegate system and if there are any other use cases we might be missing or any other thoughts on the design!

PR #381, opened 2026-05-19
feat: Add FMA for Super ZK Dispute Game

### LINKED ISSUE #20745: [SuperZKDG] Failure Modes Analysis (FMA)
## Scope

Identify and document failure modes across the super-root ZKDG, OPCM integration, and migration paths. Output feeds into the tech design, the implementation work, and the internal review.

---

## Coverage

- **Super-root proof verification edge cases**: invalid proofs, malformed `SuperRootProof` preimages, mismatched `rootClaim`, verifier failures.
- **`extraData` misparse / CWIA misconfiguration**: incorrect byte counts, off-by-one in dynamic offsets, calldata length mismatches, malformed `outputRoots[]`.
- **Path C shared-infrastructure partial upgrade scenarios**: one chain in the interop set upgraded while another isn't, ASR reinitializer race conditions, ETHLockbox state during transition.

---

## Definition of Done

- [ ] Failure modes documented with severity and mitigation
- [ ] Findings reviewed by Tech Lead and Architect
- [ ] Mitigations tracked as follow-up tasks where needed

---

## Weekly Updates

**[2026-05-19]**  [PR is open](https://github.com/ethereum-optimism/design-docs/pull/381), pending review and spec merge to update paths 

**[2026-05-29]**  [PR is open](https://github.com/ethereum-optimism/design-docs/pull/381), all comments were addressed for the first review, pending 2nd review and approval

0xarktos:
Closes: https://github.com/ethereum-optimism/optimism/issues/20745

## Summary

Updates the ZK Dispute Game failure mode analysis for the interop superchain version of ZKDisputeGame, where proposals commit to a super root over all chains rather than a single-chain output root.

### Changes

- Updated the introduction and FM framing for the superchain context: super root commitments, multi-chain proving, multi-chain challenger coverage, and shared parameters across the superchain.
- Broadened FM2 from the no-challenger case into Faulty Challenger Coverage Allows Fraudulent Proposal, covering both missing challengers and incomplete challenger coverage over the chain set.
- Expanded FM1 to clarify the blast radius of verifier, adapter, and ZK program soundness/completeness failures across every committed chain.
- Expanded FM5 to cover both off-chain prestate/config lag and ZK program drift from chain additions, hardforks, or STF changes.
- Reworked FM6 from a generic CWIA encoding mismatch into a full CWIA Game Args Layout and extraData Offset Mismatch failure mode, covering variable-length extraData, runtime offset derivation, and encode/decode/validation consistency.
- Folded challenge griefing into FM7 as part of bond and duration misconfiguration, instead of keeping it as a standalone failure mode.
- Removed obsolete standalone sections for self-challenge front-running and challenge griefing.
- Added FM8: Shared Parameter Drift as Superchain Scales, covering recurring recalibration needs as the superchain evolves.
- Updated audit requirements to include variable-length extraData decoding, the ZK program’s superchain-wide scope, and SuperRootProof decoding.
- Updated the consolidated Action Items table to match the final FM1–FM8 structure.

- TODO: Update spec references once #384 gets merged.


security/fma-zk-dispute-game.md
```diff
@@ -133,85 +136,105 @@ Below are references for this project:
...
+
+    **A -- Off-chain config lag:** `absolutePrestate` is updated on-chain but provers still run the old binary (or vice versa) (aZKG-002). All challenged games time out as `CHALLENGER_WINS` and proposers across the entire superchain lose bonds.
+
+    **B -- New chain added without program update:** When a new chain joins the superchain, the ZK program must be updated to include that chain's state transition function (STF) before games referencing that `chainId` are created. If not, the program either rejects all proofs for those games (completeness break) or silently validates fraudulent state transitions for the new chain.
+
+    **C -- Existing chain changes its STF without program update:** A hardfork or EVM config change on any superchain member without a corresponding ZK program and `absolutePrestate` update causes proof failures for that chain. There is no on-chain enforcement of this coordination, so an unmanaged STF change is indistinguishable from a fraudulent claim. Chains under a self-managed plan that modify core code without coordinating a program update could break proving for the entire superchain.
```
Inphi: These two are the same problem of `derivation rules differ from the superchain on L2`. There are a myriad of ways that the rollup config used by the fault proof can deivate from L2. So I suggest making a single bullet point to capture this issue class

security/fma-zk-dispute-game.md
```diff
@@ -133,85 +136,105 @@ Below are references for this project:
...
+
+- **Risk Assessment:** High severity / Medium likelihood
 - **Mitigations:**
-    1. In-progress games use old configuration immutably (CWIA args fixed at clone creation). Off-chain software automatically selects the correct prestate based on the absolute prestate hash, supporting multiple prestates concurrently.
-    2. Verifier contracts are immutable — old verifiers remain functional indefinitely and cannot be deprecated.
+    1. In-progress games use old configuration immutably (CWIA args fixed at clone creation). Off-chain software selects the correct prestate by hash, supporting multiple prestates concurrently.
+    2. Verifier contracts are immutable, so old verifiers remain functional indefinitely.
```
Inphi: I don't think the verifier contract is related to the issue of mismatched prestates. Does the verifier contract change if the vkey, for example, is constructed from a new sp1 zk circuit? Let's remove this mitigation to avoid confusion if that's not the case.
```suggestion
```

security/fma-zk-dispute-game.md
```diff
@@ -133,85 +136,105 @@ Below are references for this project:
...
+    3. Off-chain software updates for all chains must be coordinated simultaneously when the ZK program changes.
```
Inphi: Changes are only needed if there's a hardfork-induced ZK program change. And it's only the L2 RPCs that need to change (i.e. be updated with the latest op-stack software). But these aren't changes that are specific to the ZK program.
I think, as with the FPVM dispute game, the challenger agent will automatically use the appropriate prestate preimage of a `ZKDisputeGame`'s prestate hash. So operationally, no change is required for the off-chain challenger components.

security/fma-zk-dispute-game.md
```diff
@@ -133,85 +136,105 @@ Below are references for this project:
...
+    - `op-dispute-mon` detects games that are forecast to or do resolve incorrectly.
+    - Alert when a new `chainId` appears in a super root not covered by the current `absolutePrestate`'s known chain set.
+    - Alert when a chain's known hardfork block is reached without a corresponding `absolutePrestate` update.
 - **Recovery Path(s):**
-    1. Fix the off-chain software configuration. Proposers lose `initBond` on any challenged games that couldn't be proven, but the system is otherwise unaffected.
-    2. Guardian can blacklist affected games to trigger REFUND mode if needed.
+    1. For sub-case A: fix the off-chain software configuration. Proposers lose `initBond` on challenged games that couldn't be proven. Guardian can blacklist affected games for REFUND mode.
+    2. For sub-case B: block game creation for super roots containing the new `chainId` until the ZK program is updated and a new `absolutePrestate` is deployed via OPCM.
+    3. For sub-case C: Guardian blacklists in-flight games for the affected chain. OPCM deploys updated ZK program and `absolutePrestate`. In-flight games that cannot be proven resolve via REFUND mode.
```
Inphi: We can simplify these two. In both cases, we would either pause or blacklist invalid games.

security/fma-zk-dispute-game.md
```diff
@@ -133,85 +136,105 @@ Below are references for this project:
...
+- **Action Item(s):**
+    - [ ]  FM5: Add monitoring alert for a new `chainId` in a super root not covered by the current `absolutePrestate` known chain set.
+    - [ ]  FM5: Enforce in the chain governance process that any hardfork or STF change triggers a ZK program upgrade and `absolutePrestate` update before the hardfork activates.
```
Inphi: Until we have a better sense of the offchain infra, it's too early to suggest action items at this time. Let's omit this section.

security/fma-zk-dispute-game.md
```diff
@@ -133,85 +136,105 @@ Below are references for this project:
...
+    1. If detected before deployment, fix the encoding, offset helpers, and validation then redeploy.
+    2. If detected after deployment, OPCM deploys a new implementation with corrected logic and updates the factory. In-progress games with incorrect decoding or missing validation are blacklisted.
 - **Action Item(s):**
-    - [ ]  FM6: Implement comprehensive round-trip encoding/decoding tests for all 8 `gameArgs` fields, including edge-case values (zero, max, addresses with leading zeros).
-    - [ ]  FM6: Add fuzz tests that verify `_makeGameArgs()` output is correctly decoded by the game's accessor functions.
-    - [ ]  FM6: Review the `Duration` type's packed size and ensure it matches the offset calculations in the CWIA decoding logic.
-    - [ ]  FM6: Add specific tests for `l2SequenceNumber` `uint64` encoding in `_extraData` to verify correct packing and offset alignment.
+    - [ ]  FM6: Port `_preExtraDataByteCount()` and `_extraDataByteCount()` from `SuperFaultDisputeGame` without reimplementation, and document any intentional deviations.
```
Inphi, line 176: I don't suggest this. The `SuperFaultDisputeGame` game args parsing is more complicated than it needs otherwise because it needs to also support permissioned game args.

security/fma-zk-dispute-game.md
```diff
@@ -133,85 +136,105 @@ Below are references for this project:
...
 - **Mitigations:**
-    1. The encoding is defined in the spec ([Game Args Layout](https://github.com/ethereum-optimism/specs/blob/main/specs/fault-proof/stage-one/zk/zk-dispute-game.md#game-args-layout)) and implemented in `OPContractsManager._makeGameArgs()`.
-    2. Round-trip tests that encode via `_makeGameArgs()` and decode via the game's accessor functions verify consistency.
-    3. The CWIA pattern is used extensively in the existing OP Stack (e.g., `FaultDisputeGame`), so there is prior art and tooling for validating it.
+    1. Port `_preExtraDataByteCount()` and `_extraDataByteCount()` directly from `SuperFaultDisputeGame` rather than reimplementing them. Any deviation from the established pattern must be explicitly justified.
+    2. The encoding is defined in the spec ([Game Args Layout](https://github.com/ethereum-optimism/specs/blob/main/specs/fault-proof/stage-one/zk/zk-dispute-game.md#game-args-layout)) and implemented in `OPContractsManager._makeGameArgs()`.
+    3. Round-trip tests that encode via `_makeGameArgs()` and decode via the game's accessor functions run at n=1, n=2, and n=max supported chain counts.
 - **Detection:**
```
Inphi, line 169: The best detection is to simulate an OPCM upgrade or gameargs update and check the resulting game args before execution. We have superchain-ops task templates that do this and signers will check the sim per superchain ops SOC.

security/fma-zk-dispute-game.md
```diff
@@ -133,85 +136,105 @@ Below are references for this project:
...
-    - [ ]  FM7: Establish a minimum `challengerBond` policy such that `challengerBond > expected proving cost` to ensure defending is always profitable.
-    - [ ]  FM7: Calibrate `maxProveDuration` per aZKG-007 with analysis of L1 censorship costs and proof generation time on reference hardware.
+    - [ ]  FM7: Establish a minimum `challengerBond` policy such that `challengerBond > expected full superchain proving cost` to ensure defending is always profitable.
+    - [ ]  FM7: Calibrate `maxProveDuration` per aZKG-007 with analysis of L1 censorship costs and proof generation time on reference hardware for the full chain set.
     - [ ]  FM7: Monitor challenge rates in production and have a runbook for adjusting bond and duration parameters if griefing is detected.
     - [ ]  FM7: Add gas consumption regression tests for the verifier.
 
 ---
 
 ### FM8: Self-Challenge Front-Running to Recover Bonds
```
Inphi: I'd delete this section. Or make it much shorter to 2-3 sentences. Low sev, low likelihood issues are not worth calling out in the fma.

security/fma-zk-dispute-game.md
```diff
@@ -223,17 +246,78 @@ Below are references for this project:
 
 ### FM9: Challenge Griefing
```
Inphi: This looks like a sub-issue of FM7 (Bond misconfiguration). Can we briefly combine FM7 and FM9? Other than via bond misconfiguration, the likelihood of this occurring is low.

security/fma-zk-dispute-game.md
```diff
@@ -223,17 +246,78 @@ Below are references for this project:
...
+    - No specific detection required. Games resolve correctly, and the proposer simply proves and collects bonds.
 - **Recovery Path(s):**
-    1. No action required if `challengerBond` covers proving costs — the proposer profits from the griefing.
-    2. If `challengerBond` does not cover proving costs, OPCM increases `challengerBond` to restore profitability.
+    1. No action required if `challengerBond` covers superchain proving costs, since the proposer profits from the griefing.
+    2. If `challengerBond` does not cover proving costs, OPCM increases `challengerBond` to restore profitability. If superchain size grows and proving cost rises, OPCM must recalibrate `challengerBond` upward proactively.
+
+---
+
+### FM10: Anchor State Type Mismatch on First Super Game
```
Inphi: This is another instance of https://github.com/ethereum-optimism/design-docs/blob/main/security/fma-interop-portal.md#fm1-incorrect-state-reporting-by-anchorstateregistry. Given that this issue isn't specific to the `ZKDisputeGame`, I don't think we need to specially call it out here.
The appropriate mitigation steps are covered in that other fma as well.

security/fma-zk-dispute-game.md
```diff
@@ -223,17 +246,78 @@ Below are references for this project:
...
+    - The `AnchorStateRegistry` and the game cannot detect this on `initialize`. After migration, perform an on-chain sanity check that the root returned by the `AnchorStateRegistry` is super root compatible (check the version byte in the decoded `SuperRootProof`).
+- **Recovery Path(s):**
+    1. Roll back `respectedGameType` to the previous game type.
+    2. Re-run OPCM migration with a correctly formatted super root anchor.
+- **Action Item(s):**
+    - [ ]  FM10: Verify OPCM migration script sets a valid super root anchor atomically before or in the same transaction as `setRespectedGameType(ZKDisputeGame)`.
+
+---
+
+### FM11: Partial Chain Monitoring Allows Fraudulent Output Root to Stand
```
Inphi: FM2 is quite similar to this as the detection and mitigation are identical. Let's broaden FM2 to cover _any_ instance of faulty challengers, not just the no challenger case and delete this section.

Inphi (commented):
The FMA can be more succinct. It's important to focus on the salient failure modes that are specific to the `ZKDisputeGame`. The goal is to get a better understanding of new failure modes that are not captured in any document. Please review the [interop fma](https://github.com/ethereum-optimism/design-docs/blob/main/security/fma-interop-portal.md) and other similar documents so that we're not repeating failure modes that are covered elsewhere. If an already documented failure mode changes due to the introduction of the `ZKDisputeGame`, then that should be called out here.

security/fma-zk-dispute-game.md
0xarktos: Good point. Addressed here [875c9da](https://github.com/ethereum-optimism/design-docs/pull/381/commits/875c9dad5949bd328af5751190e17620f9a49cdc) along with the other FM5 changes

security/fma-zk-dispute-game.md
0xarktos: Removed here: [875c9da](https://github.com/ethereum-optimism/design-docs/pull/381/commits/875c9dad5949bd328af5751190e17620f9a49cdc)

security/fma-zk-dispute-game.md
0xarktos: Opted to merge 1. and 3 together here [875c9da](https://github.com/ethereum-optimism/design-docs/pull/381/commits/875c9dad5949bd328af5751190e17620f9a49cdc). Given some functionality will be needed on the off-chain side i think it's good to mention it here.

security/fma-zk-dispute-game.md
0xarktos: Great point. Fixed here [875c9da](https://github.com/ethereum-optimism/design-docs/pull/381/commits/875c9dad5949bd328af5751190e17620f9a49cdc)

security/fma-zk-dispute-game.md
0xarktos: Added a TODO here [875c9da](https://github.com/ethereum-optimism/design-docs/pull/381/commits/875c9dad5949bd328af5751190e17620f9a49cdc) . Let me know if this is not what you want to see and we can omit it completely.

security/fma-zk-dispute-game.md
0xarktos: Can you expand a bit more on this? The idea for the `SuperZKDisputeGame` was to follow a similar pattern to `SuperFaultDisputeGame` when it comes to the CWIA data layout . Obviously offsets are adjusted for the different parameters needed but what were the alternatives considered given the game implementation arguments (the ones stored in the factory) also need to be permissioned in the ZK game?

security/fma-zk-dispute-game.md
0xarktos: Merged the detection points together and added your suggestion here: [e49a98c](https://github.com/ethereum-optimism/design-docs/pull/381/commits/e49a98c6b524cf16afa9c404c84f9fa7bf668aa6)

security/fma-zk-dispute-game.md
0xarktos: Agree on this. This was brought up in the internal review and dismissed as a non-issue well as it has no actual impact on honest participants. Removed here: [2d2dc76](https://github.com/ethereum-optimism/design-docs/pull/381/commits/2d2dc766215cdbfa1cf928458f7fba552b492211)

security/fma-zk-dispute-game.md
0xarktos: Yeah the biggest impact of challenging every game if we assume well-configured bonds could be the time delay if a challenge comes just before the game expires so it makes sense to merge them. Made sure to mention the core concerns in FM7 here [293b3e2](https://github.com/ethereum-optimism/design-docs/pull/381/commits/293b3e205d7bdf9b780544866bdb9d0396dfd4fe)

security/fma-zk-dispute-game.md
0xarktos: It is indeed not specific to the `ZKDisputeGame` but rather to the migration to Super Games altogether. I think that since the update of the Starting Super Root is planned but critical to get right before super games are enabled , it would be a good idea to include a mention of it in the Interop-Portal FMA you linked.

Agree that we can remove it here though. Done here [9151618](https://github.com/ethereum-optimism/design-docs/pull/381/commits/91516188ba1b1f24f8859e5eee9bbd2d2673ea67)

security/fma-zk-dispute-game.md
0xarktos: Agree that merging makes sense here. Done in [67bdb8a](https://github.com/ethereum-optimism/design-docs/pull/381/commits/67bdb8a788135b2ec7cbedcc89d925108e1a0f4b)

0xarktos:
> The FMA can be more succinct. It's important to focus on the salient failure modes that are specific to the `ZKDisputeGame`. The goal is to get a better understanding of new failure modes that are not captured in any document. Please review the [interop fma](https://github.com/ethereum-optimism/design-docs/blob/main/security/fma-interop-portal.md) and other similar documents so that we're not repeating failure modes that are covered elsewhere. If an already documented failure mode changes due to the introduction of the `ZKDisputeGame`, then that should be called out here.

Hey, thanks for the review! The existing ZK FMA is indeed extensive so my rationale, since this is a security document, was to try to include new potential issues as a result of the migration and bring them up in the `ZKDisputeGame` context. I notice the Interop FMA is both shorter and each mode more broad (covers more cases). Is this typically the format that you like to see? 

Addressed all comments so let me know if any further fixes are needed.

security/fma-zk-dispute-game.md
```diff
@@ -133,107 +134,115 @@ Below are references for this project:
...
-    - **Block range selection:** The proposer is responsible for creating games with block ranges they can prove within `maxProveDuration`. Proving larger ranges in a single proof is generally cheaper per-block in zkVMs however a proposer who creates an unprovable range loses their own `initBond`.
-    - **L1 gas pressure:** ZK proof verification can cost several hundred thousand gas units depending on the backend. During extreme gas spikes, `prove()` inclusion becomes expensive, not technically unfeasible though given the margin against the block gas limit.
+    - **Super root timestamp selection:** The proposer is responsible for selecting a super root timestamp that all included chains can prove within `maxProveDuration`. Proving cost scales with chain count, not a single block range.
+    - **L1 gas pressure:** ZK proof verification can cost several hundred thousand gas units depending on the backend. During extreme gas spikes, `prove()` inclusion becomes expensive, not technically unfeasible given the margin against the block gas limit.
```
Inphi: The gas cost of verification is basically bounded (about 200-300K), well below the gas limit, so one can always verify the proof so long as they can afford the fee. Is this highlighting the possibly high cost of verification or something else?

Inphi (approved):
LGTM. I have just one comment on the point about L1 gas pressure. I'm wondering if it's a significant enough issue worth calling out unless I've misunderstood the issue.

security/fma-zk-dispute-game.md
Inphi: That's good to me. Thanks.

security/fma-zk-dispute-game.md
Inphi: Sorry, I forgot to respond to this comment. The issue with the current arg parsing is that it cannot assume a fixed extra args length because a super permissioned dispute game (before its simplification) may append extra args at the end. So rather than use fixed offsets to read game arguments, it's done with clunky arithmetic:
```js
function _extraDataByteCount() internal pure returns (uint256) {
    uint256 immutableArgsLength = msg.data.length - _getImmutableArgsOffset() - 2;
    return immutableArgsLength - _preExtraDataByteCount() - gameImplArgsByteCount();
}

function absolutePrestate() public pure returns (Claim absolutePrestate_) {
    absolutePrestate_ = Claim.wrap(_getArgBytes32(_preExtraDataByteCount() + _extraDataByteCount()));
}
```

Whereas, the absolutePrestate, for example, can be simplified for the ZKDisputeGame:
```js
function absolutePrestate() public pure returns (Claim absolutePrestate_) {
    absolutePrestate_ = Claim.wrap(_getArgBytes32(88));
}
```
During game initialization, we ensure the extra arg length is sane, but not, to guard against attackers manipulating the length to poison the game arg. But with the simplified approach, such guards could be strengthened by asserting against a fixed number of bytes.

security/fma-zk-dispute-game.md
ashitakah: Good question, the wording was ambiguous. The failure mode here is economic, not technical: during extreme L1 fee spikes, the cost of prove() submission (gas fee × ~200-300K) may exceed the prover's expected reward (the challenger bond they stand to win).

ashitakah: [d5f842d](https://github.com/ethereum-optimism/design-docs/pull/381/commits/d5f842deccdb16943de6279ba009912f3723eadb)

security/fma-zk-dispute-game.md
```diff
@@ -133,107 +134,115 @@ Below are references for this project:
...
+    - Alert when any chain in the superchain set updates gas limit, block time, or EVM version, flagging for immediate parameter review.
+    - Periodic review of proof time benchmarks and aggregate TVL as superchain composition changes.
 - **Recovery Path(s):**
-    1. No action required if `challengerBond` covers proving costs — the proposer profits from the griefing.
-    2. If `challengerBond` does not cover proving costs, OPCM increases `challengerBond` to restore profitability.
+    1. OPCM upgrade with recalibrated `maxProveDuration`, `maxChallengeDuration`, and bond values. Existing in-flight games continue under their original values.
+- **Action Item(s):**
+    - [ ]  FM9: Establish a parameter review process triggered by any chain addition or per-chain config change (gas limit, block time, EVM version).
+    - [ ]  FM9: Calibrate `maxProveDuration` and `maxChallengeDuration` to the worst-case chain in the current superchain set and document the required margin (≥20%).
+    - [ ]  FM9: Calibrate `initBond` to aggregate superchain TVL and `challengerBond` to full superchain proving cost on reference hardware, and document the methodology.
```
wwared: nit: Mislabelled

```suggestion
    - [ ]  FM8: Establish a parameter review process triggered by any chain addition or per-chain config change (gas limit, block time, EVM version).
    - [ ]  FM8: Calibrate `maxProveDuration` and `maxChallengeDuration` to the worst-case chain in the current superchain set and document the required margin (≥20%).
    - [ ]  FM8: Calibrate `initBond` to aggregate superchain TVL and `challengerBond` to full superchain proving cost on reference hardware, and document the methodology.
```

security/fma-zk-dispute-game.md
```diff
@@ -133,107 +134,115 @@ Below are references for this project:
...
+- **Description:** Each game's Clone-With-Immutable-Args (CWIA) payload contains a fixed pre-extraData section followed by a variable-length `extraData` of `9 + n×64 bytes` (1 version byte, 8-byte super root timestamp, n `(chainId:32, outputRoot:32)` pairs), with the implementation argument region appended after. All implementation argument getters (`verifier`, `absolutePrestate`, `anchorStateRegistry`, `weth`, bond amounts, durations) derive their byte offsets at runtime via `_preExtraDataByteCount() + _extraDataByteCount()`. `OPContractsManager._makeGameArgs()` produces the encoded payload, and `initialize()` enforces two invariants on the decoded values: `l2ChainId == 0`, and `l2SequenceNumber` (uint64) is the super root timestamp. A mismatch anywhere in this encode / decode / validate chain — off-by-one offsets, stale hardcoded constants, wrong field order, incorrect type sizes, or missing validation — corrupts every implementation argument simultaneously for all games created from that implementation, or silently accepts malformed games.
```
wwared: The `l2ChainId == 0` invariant is no longer needed since we removed that field.

wwared: This mentions "a variable-length `extraData` of `9 + n×64 bytes` (1 version byte, 8-byte super root timestamp, n `(chainId:32, outputRoot:32)` pairs)", which does not include the `parentIndex` `uint32` value that's present in the spec document before the super root proof bytes.

We should instead mention that the `extraData` section is composed of the `parentIndex` (4 bytes) plus the `superRootProof` variable-length bytes that follow the format described here.

Additionally, we need to make sure that both this and the specs use the correct and aligned format for the super root proof bytes encoding. The specs simply mention it's "ABI-encoded" but this FMA doc goes into more specific detail.

security/fma-zk-dispute-game.md
```diff
@@ -108,23 +109,23 @@ Below are references for this project:
...
 
-- **Description:** Multiple `ZKDisputeGame` instances share the same per-chain `DelayedWETH`. The spec requires (iZKG-011) that for every game: `sum(distributions) + sum(burns) == initBond + challengerBond`. An accounting bug in `closeGame()` that violates this invariant — distributing more than deposited, double-crediting, or mishandling the burn path — could drain `DelayedWETH` or permanently lock funds.
-    
+- **Description:** Multiple `ZKDisputeGame` instances share the same per-chain `DelayedWETH`. The spec requires (iZKG-011) that for every game: `sum(distributions) + sum(burns) == initBond + challengerBond`. An accounting bug in `closeGame()` that violates this invariant by distributing more than deposited, double-crediting, or mishandling the burn path could drain `DelayedWETH` or permanently lock funds.
+
     The burn path deserves attention: when a parent resolves as `CHALLENGER_WINS` and the child was unchallenged, the child's `initBond` is sent to `address(0)`. A bug here could create unbacked withdrawals or lock residual funds. The bond distribution table has 10 NORMAL + 3 REFUND scenarios, each with distinct logic paths.
```
wwared: I think the updated specs PR has a different number of bond distribution scenarios in the table? (correct me if I'm wrong)

```suggestion
    The burn path deserves attention: when a parent resolves as `CHALLENGER_WINS` and the child was unchallenged, the child's `initBond` is sent to `address(0)`. A bug here could create unbacked withdrawals or lock residual funds. The bond distribution table has 7 NORMAL + 2 REFUND scenarios, each with distinct logic paths.
```

security/fma-zk-dispute-game.md
```diff
@@ -133,107 +134,115 @@ Below are references for this project:
...
+    2. The encoding is defined in the spec ([Game Args Layout](https://github.com/ethereum-optimism/specs/blob/main/specs/fault-proof/stage-one/zk/zk-dispute-game.md#game-args-layout)) and implemented in `OPContractsManager._makeGameArgs()`.
+    3. Round-trip tests that encode via `_makeGameArgs()` and decode via the game's accessor functions run at n=1, n=2, and n=max supported chain counts.
 - **Detection:**
-    - Unit tests that round-trip all 8 fields through encode/decode.
-    - Integration tests that create a game via the factory and verify all accessor functions return expected values.
-    - Invariant/fuzz tests that verify accessor output matches encoding input for random values.
-    - Specific tests for the `l2SequenceNumber` `uint64` encoding in `_extraData`.
+    - Simulate the OPCM upgrade or `gameArgs` update using superchain-ops task templates and verify the resulting game argument before execution. Signers will to check the simulation as part of the standard operating procedure.
```
wwared: nit:

```suggestion
    - Simulate the OPCM upgrade or `gameArgs` update using superchain-ops task templates and verify the resulting game argument before execution. Signers will check the simulation as part of the standard operating procedure.
```

security/fma-zk-dispute-game.md
```diff
@@ -133,107 +134,115 @@ Below are references for this project:
...
+    - **Challenge griefing:** A sustained challenger forcing the proposer to prove every game is normally harmless. The attacker forfeits `challengerBond` per attempt and the proposer collects it on each successful proof. Worst case adds up to one full proof generation time of delay per game if the challenge occurs just before the game closes. Because `ZKDisputeGame` proves over all chains simultaneously, per-challenge proving cost is materially higher than in a single-chain game, making the `challengerBond > full superchain proving cost` invariant more critical to maintain. If bonds are set too low, griefing by challening every proposal can become economically rational for attackers.
```
wwared: nit:

```suggestion
    - **Challenge griefing:** A sustained challenger forcing the proposer to prove every game is normally harmless. The attacker forfeits `challengerBond` per attempt and the proposer collects it on each successful proof. Worst case adds up to one full proof generation time of delay per game if the challenge occurs just before the game closes. Because `ZKDisputeGame` proves over all chains simultaneously, per-challenge proving cost is materially higher than in a single-chain game, making the `challengerBond > full superchain proving cost` invariant more critical to maintain. If bonds are set too low, griefing by challenging every proposal can become economically rational for attackers.
```

security/fma-zk-dispute-game.md
```diff
@@ -133,107 +134,115 @@ Below are references for this project:
...
-    - `op-dispute-mon` already detects games that are forecast to or do resolve incorrectly.
+    - Alerts when a challenged game fails to receive a proof within a reasonable time.
+    - Alerts when the proposer generates a proof found invalid on-chain.
+    - `op-dispute-mon` detects games that are forecast to or do resolve incorrectly.
+    - Alert when a new `chainId` appears in a super root not covered by the current `absolutePrestate`'s known chain set.
+    - Alert when a chain's known hardfork block is reached without a corresponding `absolutePrestate` update.
 - **Recovery Path(s):**
-    1. Fix the off-chain software configuration. Proposers lose `initBond` on any challenged games that couldn't be proven, but the system is otherwise unaffected.
-    2. Guardian can blacklist affected games to trigger REFUND mode if needed.
+    - Guardian pauses the system or blacklists invalid games. Unprovable games resolve via REFUND mode.
```
wwared: nit: This changed from a numbered list to a non-numbered list, but I think it should still be a numbered list (to match the other sections)

```suggestion
    1. Guardian pauses the system or blacklists invalid games. Unprovable games resolve via REFUND mode.
```

wwared (approved):
LGTM, a few comments below, most non-blocking nits. The main important thing to update is that we should align the description of the encoding of the super root proof bytes in the CWIA section with the specs and implementation.

security/fma-zk-dispute-game.md
ashitakah: Fixed, thanks, corrected the labels

security/fma-zk-dispute-game.md
ashitakah: Right, that field is gone. Removed the invariant from the description and updated the test action item.

ashitakah: Good catch. Checked against the implementation: extraData is the 4-byte parentIndex followed by the SuperRootProof bytes, which are tightly packed (version 0x01 || 8-byte timestamp || n×(chainId:32, outputRoot:32)) and parsed bytewise by Encoding.decodeSuperRootProof. Updated the description accordingly, and added an action item to fix the spec's "ABI-encoded" wording so both match the packed layout.

security/fma-zk-dispute-game.md
ashitakah: 👍

security/fma-zk-dispute-game.md
ashitakah: 👍

security/fma-zk-dispute-game.md
ashitakah: 👍

security/fma-zk-dispute-game.md
ashitakah: Updated to 7 NORMAL + 2 REFUND.

security/fma-zk-dispute-game.md
ashitakah: [49a8ce2](https://github.com/ethereum-optimism/design-docs/pull/381/commits/49a8ce2bf75e48a06c8249c40353b904b685821a)

security/fma-zk-dispute-game.md
ashitakah: [49a8ce2](https://github.com/ethereum-optimism/design-docs/pull/381/commits/49a8ce2bf75e48a06c8249c40353b904b685821a)

security/fma-zk-dispute-game.md
ashitakah: [49a8ce2](https://github.com/ethereum-optimism/design-docs/pull/381/commits/49a8ce2bf75e48a06c8249c40353b904b685821a)

security/fma-zk-dispute-game.md
ashitakah: [49a8ce2](https://github.com/ethereum-optimism/design-docs/pull/381/commits/49a8ce2bf75e48a06c8249c40353b904b685821a)

security/fma-zk-dispute-game.md
ashitakah: [49a8ce2](https://github.com/ethereum-optimism/design-docs/pull/381/commits/49a8ce2bf75e48a06c8249c40353b904b685821a)

ashitakah: [49a8ce2](https://github.com/ethereum-optimism/design-docs/pull/381/commits/49a8ce2bf75e48a06c8249c40353b904b685821a)

security/fma-zk-dispute-game.md
ashitakah: [49a8ce2](https://github.com/ethereum-optimism/design-docs/pull/381/commits/49a8ce2bf75e48a06c8249c40353b904b685821a)

ashitakah:
> LGTM, a few comments below, most non-blocking nits. The main important thing to update is that we should align the description of the encoding of the super root proof bytes in the CWIA section with the specs and implementation.

@wwared All comments fixed!

PR #388, opened 2026-06-24
sdm: add failure mode analysis

nonsense:
Adds a public failure mode analysis for SDM and PostExec transactions, covering consensus, accounting, fault proofs, activation, and RPC compatibility.

Closes https://github.com/ethereum-optimism/protocol-team/issues/102

### DIFF
--- a/security/fma-sdm-postexec.md
+++ b/security/fma-sdm-postexec.md
@@ -1,0 +1,662 @@
+# Sequencer Defined Metering (SDM) & PostExec Transactions: Failure Modes and Recovery Path Analysis
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Introduction](#introduction)
+- [Failure Modes and Recovery Paths](#failure-modes-and-recovery-paths)
+  - [A. Consensus & State-Root Divergence](#a-consensus--state-root-divergence)
+    - [FM1: Divergence in applying or validating sequencer-set refunds → chain split or unprovable output root](#fm1-divergence-in-applying-or-validating-sequencer-set-refunds--chain-split-or-unprovable-output-root)
+    - [FM2: Refund mis-metering — over-refund, fee-vault drain, value non-conservation](#fm2-refund-mis-metering--over-refund-fee-vault-drain-value-non-conservation)
+    - [FM3: Refunds hide real compute → block-gas rate-limit bypass (no aggregate cap — superseded by pre-refund enforcement)](#fm3-refunds-hide-real-compute--block-gas-rate-limit-bypass-no-aggregate-cap--superseded-by-pre-refund-enforcement)
+    - [FM4: Malformed / malicious PostExec payload accepted (structural validation gaps)](#fm4-malformed--malicious-postexec-payload-accepted-structural-validation-gaps)
+  - [B. Block Building & Derivation](#b-block-building--derivation)
+    - [FM5: Flashblock materialization diverges from sealed canonical block](#fm5-flashblock-materialization-diverges-from-sealed-canonical-block)
+    - [FM6: Span-batch derivation has no PostExec structural gate](#fm6-span-batch-derivation-has-no-postexec-structural-gate)
+    - [FM7: Disabled / flaky flashblocks SDM test leaves the divergence detector dark](#fm7-disabled--flaky-flashblocks-sdm-test-leaves-the-divergence-detector-dark)
+  - [C. Fault Proofs](#c-fault-proofs)
+    - [FM8: SDM unproven through the FPVM/Cannon path; no non-zero refund fixture](#fm8-sdm-unproven-through-the-fpvmcannon-path-no-non-zero-refund-fixture)
+  - [D. Activation & Configuration](#d-activation--configuration)
+    - [FM9: Inconsistent Lagoon activation across chains/components → fragmentation & unprovable blocks](#fm9-inconsistent-lagoon-activation-across-chainscomponents--fragmentation--unprovable-blocks)
+    - [FM10: Latent SDM↔Interop coupling (decoupling rename not applied)](#fm10-latent-sdminterop-coupling-decoupling-rename-not-applied)
+    - [FM11: Operator opt-in flag in wrong / split state (non-persistent, multi-endpoint)](#fm11-operator-opt-in-flag-in-wrong--split-state-non-persistent-multi-endpoint)
+  - [E. User-Facing & Operational](#e-user-facing--operational)
+    - [FM12: Synthetic 0x7D tx breaks standard user-facing RPC consumers (expose, don't filter)](#fm12-synthetic-0x7d-tx-breaks-standard-user-facing-rpc-consumers-expose-dont-filter)
+    - [FM13: Receipt refund mis-attribution (index anchoring)](#fm13-receipt-refund-mis-attribution-index-anchoring)
+    - [FM14: Admin RPC abuse / exposure](#fm14-admin-rpc-abuse--exposure)
+    - [FM15: Policy-abstraction misconfiguration (silent economic error)](#fm15-policy-abstraction-misconfiguration-silent-economic-error)
+  - [Generic items we need to take into account](#generic-items-we-need-to-take-into-account)
+- [Action Items](#action-items)
+- [Audit Requirements](#audit-requirements)
+- [Appendix](#appendix)
+  - [Appendix A: Verification environment & method](#appendix-a-verification-environment--method)
+  - [Appendix B: Doc-vs-code discrepancies found while writing this FMA](#appendix-b-doc-vs-code-discrepancies-found-while-writing-this-fma)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+|                    |                       |
+| ------------------ | --------------------- |
+| Author             | Anton Evangelatov     |
+| Created at         | 2026-06-16            |
+| Initial Reviewers  | _TBD_                 |
+| Need Approval From | _TBD (Security team)_ |
+| Status             | Draft                 |
+
+> [!NOTE]
+> First draft. Anything I couldn't confirm in code is flagged unverified.
+
+## Introduction
+
+This document covers **Sequencer Defined Metering (SDM)** and the **PostExec transaction** it's built on.
+
+**What it is.** SDM lets the sequencer reduce the gas a user is metered for — chiefly by refunding the EIP-2929
+"cold→warm" access surcharge the protocol charges but the sequencer chooses not to pass on. The block builder appends
+**exactly one synthetic trailing transaction of type `0x7D` (`TxPostExec`)** to each L2 block. That tx carries a
+`PostExecPayload`: a list of per-transaction refund entries `SDMGasEntry{index, gas_refund}`. Execution clients
+re-execute the block in PostExec-aware mode, apply the refunds (adjusting `gasUsed`, receipts, fee-vault settlement, and
+the state root), and verifiers re-check the embedded payload rather than recomputing it.
+
+**Where the code lives (scope of this analysis):**
+
+- **EVM / refund math:** `rust/alloy-op-evm/src/block/mod.rs`, `rust/alloy-op-evm/src/inspector.rs`, `rust/op-revm/src/spec.rs`.
+- **Block building:** op-reth payload builder; op-rbuilder flashblocks (`rust/op-rbuilder/.../builders/flashblocks/payload.rs`); rollup-boost (`crates/rollup-boost/src/flashblocks/service.rs`).
+- **Consensus / derivation:** op-node batch derivation (`op-node/rollup/derive/batches.go`), span-batch codec; `op-chain-ops/pkg/sdm/`.
+- **Fault proofs:** kona proof executor (`rust/kona/crates/proof/executor/...`), FPVM/Cannon path, absolute prestates.
+- **Activation & config:** `op-node/rollup/toggles.go`, rollup-config Lagoon time, EL operator opt-in (`admin_setSdmPostExecOptIn`).
+- **User-facing:** op-reth eth RPC receipts/blocks; flashblocks-rpc.
+
+**Activation vehicle (verified).** SDM rides the **Lagoon** hardfork (the renamed Interop fork; rollup-config field
+`LagoonTime`, JSON `lagoon_time`, serde-aliased from `interop_time`). In all three clients `is_sdm_active` currently
+delegates through Interop → Lagoon (op-node `toggles.go:28`; op-reth `evm/src/lib.rs:138`; kona `rollup.rs:308`). The
+revm spec for the SDM era is `OpSpecId::INTEROP` (eth base `OSAKA`, `op-revm/src/spec.rs:32-46`). Producing `0x7D` also
+requires a per-node **operator opt-in** flag (`admin_setSdmPostExecOptIn`, in-memory, default OFF). **SDM is not active
+on any production chain or default devnet today** — it runs only under test overrides and forced-activation acceptance
+tests.
+
+**References:**
+
+- FMA template & process: `design-docs/assets/fma-template.md`, [FMAs in the SDLC](https://github.com/ethereum-optimism/pm/blob/main/src/fmas.md).
+- Generic failure modes to inherit: `design-docs/security/fma-generic-hardfork.md`, `fma-generic-contracts.md`.
+- SDM working notes: `~/sdm-docs/` — esp. `README.md`, `sdm-fault-proofs.md`, `sdm-infinite-mint-guards-and-intrinsic-warm-bug.md`, `max-gas-limit-enforcement.md`, `sdm-refund-cap-plan.md`, `flashblocks-sdm-postexec-divergence.md`, `postexec-derivation-gating-plan.md`, `sdm-hardfork-estimate.md`, `2026-06-15-sdm-toggle-lagoon-rename-plan.md`, `sdm-admin-rpc-plan.md`, `sdm-policy-abstraction-plan.md`.
+- Key PRs/issues: #20738 (replay-cost fix), #21034, #908 (post-exec), #21105 (Lagoon rename), #21110 (explorer support), #21324 (proof test coverage), #21354 (Osaka flashblocks divergence).
+
+## Failure Modes and Recovery Paths
+
+### A. Consensus & State-Root Divergence
+
+#### FM1: Divergence in applying or validating sequencer-set refunds → chain split or unprovable output root
+
+- **Description:** Refunds are **sequencer-defined**: the block producer computes them and embeds the values in the
+  `0x7D` payload. Verifiers, replicas, and the fault-proof program **do not recompute refunds** — they take the claimed
+  values as authoritative and only (i) check structural validity plus the per-tx bound `refund ≤ evm_gas_used`,
+  (ii) apply the post-exec settlement and set `gas_used = evm_gas_used − refund`, and (iii) require the resulting
+  receipts and state root to match. So the split surface is **not** "different clients compute different refunds" but the
+  steps every client must perform **identically on the same payload**: computing `evm_gas_used`, applying the settlement,
+  and enforcing the structural + per-tx-bound checks. If two verifiers apply the same payload to different state roots,
+  or one accepts what another rejects, that's a **chain split**; the same divergence between sequencer and proof program
+  is an **unprovable (or wrongly-provable) output root** that stalls withdrawals. A _wrong refund value_ is a separate,
+  non-split failure (FM2) — every verifier faithfully applies the same wrong value. The Osaka entry-count incident
+  (#21354) shows the surface is live even though that specific case was a test artifact (FM7).
+- **Risk Assessment:** **High impact, low likelihood.** Because verifiers apply rather than recompute, the whole
+  "recompute-and-disagree" class is out of scope by design, and the apply/validate path is a single shared Rust crate
+  (`alloy-op-evm` Verify mode) used by both op-reth and the kona proof program. **op-geth is end-of-life and won't
+  support SDM**, so there's no second independent EVM re-implementation. The residual differential surface is narrow:
+  drift between op-reth Verify mode and the kona proof program (incl. MIPS-vs-native, FM8), and between the Rust
+  validators and the **Go-side** structural/derivation checks (op-node, `op-chain-ops/pkg/sdm`), which independently
+  reimplement the _validation_ (not the refund math).
+- **Mitigations:**
+  1. _Reduce chance:_ keep one shared apply/validate path (`block/mod.rs` Verify). The consensus-critical invariants —
+     the per-tx bound `refund ≤ evm_gas_used` (`:554,933`), the structural checks (`:130,528-556,1080-1087`), and the
+     conservation/underflow guards (`:634,663`) — must be identical across **every verifier and the proof program**.
+     kona enforces the same bound and structural checks and doesn't recompute warming (`sdm-fault-proofs.md`).
+  2. _Add:_ a **canonical spec for the application & validation semantics** — settlement math,
+     `gas_used = evm_gas_used − refund`, receipt adjustment, the per-tx bound, and the structural rules (one `0x7D`,
+     last-in-block, valid indices, block-number anchor) — so the Go derivation/decode checks (op-node,
+     `op-chain-ops/pkg/sdm`) and the kona proof program can't drift from op-reth Verify mode. Add a **differential test**
+     asserting op-reth Verify mode and the kona proof program apply the same `0x7D` payload to the same block hash /
+     state root, and that the Go structural checks accept/reject exactly the payloads the Rust verifier does. (The
+     warming **exclusion set** and refund constants at `inspector.rs:19-21` are _sequencer policy_, not consensus — their
+     cross-builder consistency is a builder/economic concern, FM2/FM5.)
+- **Detection:** `debug_replaySDMBlock` (synthesized-vs-embedded payload equality); kona native + FPVM output-root match
+  vs op-reth; acceptance tests `TestSDMEnabledPayloadAndReplayMatch`, `TestSDMPostExecBlockDerivesAndChainProgresses`;
+  in production, replica state-root disagreement or op-challenger unable to counter a claim.
+- **Recovery Path(s):** A live post-activation split needs an **emergency coordinated hardfork** to pin the canonical
+  math. If the divergence is only in the proof path (kona), a kona-client hotfix plus a **new absolute prestate release**
+  (governance-approved swap in superchain-registry + op-challenger-runner config). Both high-effort.
+- **Tests covering this FM (this monorepo):**
+  - _Unit — shared apply/validate path (`rust/alloy-op-evm/src/block/tests.rs`):_ `test_post_exec_producer_verifier_roundtrip`, `test_post_exec_settlement_deltas_conserve_value`, `test_post_exec_settlement_deltas_skip_non_refunding_txs`, `test_settlement_state_account_preserves_original_info`, `test_post_exec_settlement_conserves_total_eth_supply`, `test_post_exec_settlement_underflow_is_rejected`.
+  - _Unit — kona proof-program parity (`rust/kona/crates/proof/executor/src/builder/core/tests.rs`):_ `post_exec_valid_empty_payload_executes_without_state_or_gas_change` plus the reject-matrix (FM4), all through `StatelessL2Builder`.
+  - _Unit — Go-side structural/derivation reimplementation (`op-node/rollup/derive/`):_ `TestDeriveSpanBatchSDMGate`, `TestSpanBatchReadPostExecTxData`; the `batches_test.go` cases `"postExec tx included pre-SDM"` / `"postExec tx included at SDM"`; cross-encode round-trips `TestPostExecTxMarshalBinaryDifferential`, `TestUnmarshalPostExecTxRoundTrip` (`op-core/types/post_exec_tx_test.go`).
+  - _Acceptance:_ `TestSDMEnabledPayloadAndReplayMatch`, `TestSDMPostExecBlockDerivesAndChainProgresses` (span + singular), `TestSDMPostExecBlockDerivesOnIsolatedVerifier` (tests/sdm/block_test.go); `TestInteropSingleChainFaultProofsWithSDM` (tests/interop/proofs-singlechain/).
+  - _Gap (BLOCKING):_ no **differential test** asserting op-reth Verify mode and the kona proof program apply the _same non-empty_ `0x7D` payload to the same block hash, nor that the Go structural checks accept/reject exactly the same payloads as the Rust verifier — the cross-implementation drift surface this FM is about is untested.
+
+#### FM2: Refund mis-metering — over-refund, fee-vault drain, value non-conservation
+
+- **Description:** SDM moves real value — refunds reduce what users pay into the fee vaults. Two historical/latent bug
+  classes: (a) the **intrinsic-warm over-refund** — granting the 2500 warm rebate for a tx's own intrinsically-warm
+  accounts (sender/`to`/created-address) and for fee-vault settlement touches, which over-credits senders and **drains
+  OP fee vaults** (Immunefi #81847); and (b) **value non-conservation** — settlement deltas that don't net to zero, in
+  the limit minting ETH. The intrinsic-warm fix has landed on this branch, but the regression surface is large: any
+  future change to `collect_intrinsic_warmth` / `note_account_touch` could silently reintroduce it (the buggy behavior
+  was previously locked in by green tests).
+- **Risk Assessment:** **Medium impact, low likelihood** now the fix has landed. It was _systematic_ when present (fires
+  on sequential same-EOA txs and popular `to` contracts), so a regression would be high-impact again.
+- **Mitigations:**
+  1. _Reduce chance:_ three-layer defense — per-tx bound `refund ≤ evm_gas_used` (`block/mod.rs:554,933`); settlement
+     **underflow guard** `checked_sub` → `PostExecSettlementUnderflow` (`:634`); value conservation via
+     `post_exec_settlement_deltas` (`:663`). Intrinsic-warm fix at `inspector.rs:208,259,261,415`. Regression tests:
+     `test_intrinsic_warm_{to,sender,created}_address_is_not_rebated`, `test_fee_recipient_settlement_touch_is_not_rebated`,
+     `test_post_exec_settlement_{conserves_total_eth_supply,underflow_is_rejected}`.
+  2. _Add:_ a spec/reference cross-check of the exclusion set (FM1); a test asserting saturating arithmetic
+     (`inspector.rs:134`, `block/mod.rs:688-708`) can't silently truncate a settlement delta under max realistic inputs
+     (today saturation is defense-in-depth behind the underflow guard, but nothing pins that it can't mask an accounting
+     bug); re-baseline the downstream Go acceptance test (`TestSDMStorageRefundBreakdown`) on devstack CI.
+- **Detection:** fee-vault balance monitor; acceptance tests asserting zero refund for plain transfers;
+  conservation/underflow unit tests; replay parity.
+- **Recovery Path(s):** A code fix (already done for the known bug). **Note:** since all clients compute the same wrong
+  value, an over-refund that shipped would _not_ be a chain split — it's value misrouting needing treasury
+  reconciliation / governance, not a fork. A true ETH-creation bug would be critical and need an emergency upgrade.
+- **Tests covering this FM (this monorepo):**
+  - _Unit — intrinsic-warm exclusion regression set (`block/tests.rs`):_ `test_intrinsic_warm_{to,sender,created,beneficiary,precompile}_..._is_not_rebated`, `test_intrinsic_warm_7702_authority_is_not_rebated`, `test_fee_recipient_settlement_touch_is_not_rebated` (the Immunefi #81847 fee-vault-drain class).
+  - _Unit — value conservation / underflow (`block/tests.rs`):_ `test_post_exec_settlement_conserves_total_eth_supply`, `test_post_exec_settlement_underflow_is_rejected`, `test_post_exec_settlement_deltas_conserve_value`.
+  - _Unit — inspector accounting (`rust/alloy-op-evm/src/post_exec/inspector/tests.rs`):_ `repeated_account_touch_refunds_once`, `repeated_storage_refunds_without_account_double_count`, `deposit_warms_but_does_not_claim`, `settlement_note_warms_without_claiming_but_opcode_access_still_rebates`, `post_exec_tx_never_claims_refunds`, `intrinsic_access_list_warmth_does_not_claim`.
+  - _Acceptance:_ `TestSDMStorageRefundBreakdown`, `TestSDMMixedWorkloadSmoke`, `TestSDMDisabledNoRefunds` (tests/sdm/); `TestFlashblocksSDMPhantomWarmingDivergence` (tests/flashblocks/).
+  - _Gap (NON-BLOCKING):_ no test pins that **saturating** settlement arithmetic can't silently truncate a delta under max realistic inputs; the `TestSDMStorageRefundBreakdown` devstack-CI re-baseline is still pending.
+
+#### FM3: Refunds hide real compute → block-gas rate-limit bypass (no aggregate cap — superseded by pre-refund enforcement)
+
+- **Description:** Refunds lower canonical `gasUsed`. With only a _per-tx_ bound, a sequencer could refund ~100% of every
+  tx, driving canonical `gasUsed → 0` while the real pre-refund EVM work stays large — so the block-gas-limit rate limit
+  no longer bounds real compute, and replicas/challengers fall behind the tip, threatening the fault-proof
+  liveness/Stage-1 guarantee. **Design decision (2026-06-19): no percentage-based aggregate refund cap. The cap is
+  `evm_gas_used`** — the per-tx bound `refund ≤ evm_gas_used`, combined with pre-refund gas-limit enforcement:
+  - **Pre-refund gas-limit enforcement** (the `evm_gas_used` accumulator + admission check, `block/mod.rs:350,847-848,1025`)
+    — **landed**; bounds _real compute_ per block (`sum(evm_gas_used) ≤ block_gas_limit`), the Stage-1 concern. This is
+    the layer that neutralizes the compute-blowup attack.
+  - **Per-tx bound `refund ≤ evm_gas_used`** (`block/mod.rs:554,933`) — **landed**; the only refund cap, by design.
+  - **Aggregate refund cap** (`MAX_REFUND_BPS`, designed in `sdm-refund-cap-plan.md`) — **will not be implemented**; that
+    plan is closed (it was never implemented — only a stray comment at `tests.rs:1066`).
+- **Risk Assessment:** **Low residual (post-decision).** Pre-refund enforcement already bounds real compute, so the
+  Stage-1 violation the cap was meant to prevent is handled without it. Accepted residual: a sequencer may still refund
+  up to 100% of every tx, so canonical `gasUsed` can legitimately approach 0 while real compute stays bounded — and the
+  magnitude of any value misrouting is bounded only per-tx by `evm_gas_used`, not by an aggregate percentage. That's an
+  economic/policy property, **not** a consensus or Stage-1 concern, and is accepted by the decision above.
+- **Mitigations:**
+  1. _Reduce chance:_ pre-refund enforcement bounds `sum(evm_gas_used) ≤ block_gas_limit` at the executor level (op-reth
+     + kona both route through `OpBlockExecutor`). Tests: `test_pre_refund_gas_limit_counts_sdm_refunded_gas`,
+     `test_verifier_rejects_malicious_payload_whose_refunds_hide_pre_refund_overuse`.
+  2. _DECIDED — no aggregate cap:_ the per-tx bound plus pre-refund enforcement supersede `MAX_REFUND_BPS`; no shared
+     constant / verifier `finish()` cap / producer clamp / kona cap-parity work is needed. Still open: confirm the
+     **`finish()` pre-refund debug-assert** the `max-gas-limit` notes describe — **absent** today (`block/mod.rs:1077-1106`
+     only checks unconsumed entries); add an e2e malicious-sequencer test and a **cross-flashblock** cumulative-overrun
+     test (the per-flashblock executor accumulator resets; the cross-flashblock bound relies on the producer's persistent
+     `ExecutionInfo.cumulative_evm_gas_used`).
+- **Detection:** verifier rejects with `TransactionGasLimitMoreThanAvailableBlockGas`; a per-block
+  `refund / refundable_evm_gas` ratio metric (planned in the cap plan §6.5, **not yet added**); an explorer showing a
+  full block with `gasUsed ≈ 0`.
+- **Recovery Path(s):** A consensus-rule addition — trivial if it lands before activation, a coordinated
+  upgrade/hardfork if SDM is already live.
+- **Tests covering this FM (this monorepo):**
+  - _Unit — pre-refund enforcement (`block/tests.rs`):_ `test_pre_refund_gas_limit_counts_sdm_refunded_gas` (tx fitting canonical `gas_used` but exceeding pre-refund `evm_gas_used` is rejected); `test_pre_refund_gas_limit_never_binds_with_sdm_off`; `test_evm_gas_used_tracks_pre_refund_gas_under_sdm` (`evm_gas_used − gas_used == total refund`); `test_verifier_rejects_malicious_payload_whose_refunds_hide_pre_refund_overuse`; `test_verifier_accepts_payload_when_pre_refund_stays_below_limit`.
+  - _Decided — not a gap:_ the **aggregate refund cap** won't be implemented, so its absence needs no test. _Still-open test gaps:_ no **cross-flashblock cumulative pre-refund-overrun** test; no **e2e malicious-sequencer** test; the `finish()` pre-refund debug-assert is absent (`block/mod.rs:1077-1106`).
+
+#### FM4: Malformed / malicious PostExec payload accepted (structural validation gaps)
+
+- **Description:** The embedded `0x7D` payload is attacker-controllable by a malicious sequencer. Incomplete structural
+  validation could let a payload target a deposit tx or the `0x7D` tx itself, reference an out-of-range/duplicate index,
+  leave entries unconsumed, or place a second/non-final PostExec tx — any of which mis-applies refunds and diverges
+  state. A related risk is **Go/Rust validation drift**: the Go decoder
+  `op-chain-ops/pkg/sdm/payload.go::DecodePayload` validates **only `Version`** (`:41`) — it doesn't bound `refund ≤ gas`,
+  reject duplicate/out-of-range indexes, or enforce the one-per-block / last-in-block / block-number-anchor invariants.
+  Those are Rust-side only.
+- **Risk Assessment:** **High impact, low likelihood.** The Rust verifier checks look complete and sit on the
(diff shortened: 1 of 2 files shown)


security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+
+The refund **mechanism** is consensus-critical, but the refund **policy** is sequencer-defined. Verifiers enforce
+payload structure, `refund <= evmGasUsed`, and settlement validity; they do not recompute whether a refund was
+justified by the sequencer's intended policy. The version-1 production policy is block-level warming: it rebates
+an EIP-2929 cold-to-warm surcharge only when the transaction actually paid that surcharge. This policy boundary
+is an explicit sequencer trust assumption.
+
+SDM activates with the Lagoon network upgrade. Producing refunds additionally requires an operator-controlled
+execution-layer opt-in. Verification does not depend on this opt-in: after activation, every verifier must accept
+and apply a valid SDM payload whether or not that verifier would produce one.
```
geoknee:
```suggestion
and apply a valid SDM payload if the sequencer did.
```

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+  validity or state-root disagreement between clients causes a chain split.
+- **Risk Assessment:** High impact, low likelihood. PostExec affects consensus-visible gas accounting, receipts,
+  balances, and state roots. The specification, shared execution code, and execution-layer validation reduce the
+  likelihood, but producer lifecycle bugs or an independent implementation can still diverge.
+- **Existing Mitigations:**
+  1. The specifications require at most one final `0x7D`, the correct block-number anchor, an active schema, and
+     non-empty, non-zero entries ordered by strictly increasing transaction index.
+  2. Entries may target only standard Ethereum transactions. Execution enforces `refund <= evmGasUsed`, rejects
+     settlement underflow, and consumes every entry exactly once.
+  3. Incremental streams use replace-not-append semantics for the cumulative PostExec payload; the final streamed
```
geoknee: We use "streams" quite a bit, and I guess we mean the subblocks stream. Better to just use that concrete identifier IMO.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+     bytes are compared with the sealed transaction, and sealing aborts if PostExec finalization fails.
+  4. Derivation enforces activation while the execution layer remains authoritative for post-activation payload
```
geoknee: In what sense does "derivation enforce activation:?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+     validity and deposits-only recovery.
+  5. Existing unit and acceptance tests cover structural rejection, producer/verifier round trips, and
+     streamed-to-sealed payload equality.
+- **Possible Mitigations:**
+  1. Add a non-empty fixture that compares execution-client and fault-proof-program block hash, receipts root,
```
geoknee: We can probably just say "check for consensus between" rather than listing all of these fields. If the block hash agrees, this implies the other fields do too.

What is the "non empty fixture"? Would this be the action tests that we decided not to do in the end?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+     gas used, and state root.
+  2. Add deposits-only replacement metrics and cross-client malformed-payload recovery coverage.
```
geoknee: I believe we will only get a deposit-only replacement once an invalid input is batched to L1 and then derived. We would most likely want to (and maybe already would with our current alerts) learn of a problem much sooner, when verifiers try to insert the invalid unsafe payload.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+an EIP-2929 cold-to-warm surcharge only when the transaction actually paid that surcharge. This policy boundary
+is an explicit sequencer trust assumption.
+
+SDM activates with the Lagoon network upgrade. Producing refunds additionally requires an operator-controlled
+execution-layer opt-in. Verification does not depend on this opt-in: after activation, every verifier must accept
+and apply a valid SDM payload whether or not that verifier would produce one.
+
+The normative specifications are the source of truth:
+
+- [Post-Execution Transactions](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/lagoon/post-exec.md)
```
geoknee, line 46: Do we have specs covering the behaviour of SDM as it relates to subblocks?

This document implies there are specific choices made about how it should work, so we ought to have a spec for it.

security/fma-sdm-postexec.md
geoknee: So we will see an unsafe head stall, and if it can be attributed to an invalid unsafe payload caused by SDM, I would suggest the runbook would have us:
* stop the batcher to (maybe, if done quickly enough) prevent the invalid data being batched to L1
* hit the kill switch to disable SDM (we now suspect a bug in the sequencer or verifier code or both)
* rewind the sequencer to the block before the bad one

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+- [Sequencer-Defined Metering](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/lagoon/sdm.md)
+- [Lagoon Network Upgrade](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/lagoon/overview.md)
+- [SDM design document](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/sdm.md)
+
+## Failure Modes and Recovery Paths
+
+### FM1: PostExec construction or application diverges, causing invalid blocks or chain splits
```
geoknee: Might be worth splitting invalid blocks and chain splits into an FM1a and FM1b, since the recovery path may be quite different in either case.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+  from an erroneous one because refund policy is not part of consensus. Historically, this class included
+  rebates for intrinsically warm or otherwise unmetered accesses: the transaction received a cold-to-warm rebate
+  even though the EVM never charged the cold surcharge.
+- **Risk Assessment:** Medium impact, low likelihood. The likely outcome is bounded economic misallocation rather
+  than a chain split. A settlement conservation bug would have higher impact, but is constrained by consensus
+  checks and conservation tests.
+- **Existing Mitigations:**
+  1. Consensus limits each refund to the transaction's `evmGasUsed` and rejects settlement underflow.
+  2. Settlement tests assert that sender credits equal recipient debits across transaction types, fee settings,
+     and arithmetic boundaries.
```
geoknee: This feels like a mitigation for an infinite ETH mint bug or similar, rather than economic loss for the operator. I suggest that should be it's own failure mode (very high impact, very low probability).

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+  policy-aware checker.
+- **Recovery Path(s):** Disable refund production, correct the policy, and redeploy. Already finalized refunds are
+  consensus-valid and should not be rolled back; material losses require accounting and, if appropriate,
+  treasury reconciliation. A true value-conservation failure requires an emergency upgrade.
+
+### FM3: Refunds bypass block resource limits
+
+- **Description:** Canonical gas decreases after refunds. If block admission used only post-refund gas, a producer
+  could include substantially more EVM work than the block gas limit intends, causing replicas and challengers to
+  fall behind even though the header's `gasUsed` remains low.
```
geoknee: This is essentially a Denial Of Service vector?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+  reproduce it. Causes include different transaction decoding, settlement behavior, configuration, or behavior
+  that works in native execution but fails inside the production FPVM. SDM also activates with Lagoon, whose
+  dependency-set and proof-program requirements must be satisfied by the target chain.
+- **Risk Assessment:** High impact, medium likelihood until production-FPVM coverage exists. An unprovable output
+  stalls permissionless withdrawals and may require replacing the fault-proof program under governance.
+- **Existing Mitigations:**
+  1. Fault-proof execution uses the same PostExec parser and refund application logic as the execution client.
+  2. Native proof tests exercise non-empty SDM blocks.
+- **Possible Mitigations:**
+  1. Prove a non-empty SDM block in the production FPVM rather than only running the program natively.
```
geoknee: We should check in with a proofs expert about what our standards are on this kind of thing. IIUC this can take a long time to run through (playing a dispute game down to a single step and executing that on chain). 

My assumption is that we usually rely on testing equivalence between the client code and the fault proof _program_, and take the virtual machine part and test that separately.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+- **Risk Assessment:** High impact, low likelihood with an activation preflight. Cross-component activation
+  disagreement can halt or split the chain. An opt-in-only mismatch has lower impact because verification is
+  independent of opt-in and missing refunds remain valid.
+- **Existing Mitigations:**
+  1. Every component derives SDM activation from the Lagoon schedule.
+  2. Refund production is enabled only when both Lagoon is active and the operator opt-in is set.
+  3. Verification ignores the local production opt-in.
+  4. The operator opt-in defaults off, supports boot-time configuration, and exposes effective state through the
+     admin status method.
+- **Possible Mitigations:**
```
geoknee, line 199: We should probably mention the superchain registry here, that is the main tool we use to help ensure all nodes have the correct timestamp for a hardfork. 

This would also be a good opportunity to refresh and reference https://github.com/ethereum-optimism/design-docs/blob/main/security/fma-generic-hardfork.md since it is a general concern and not special to SDM.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+  accounting risk, not a consensus risk.
+- **Existing Mitigations:**
+  1. The PostExec specification defines the minimal RPC transaction shape, receipt behavior, and
+     `opGasRefund` semantics.
+  2. Payload validation guarantees ordered, unique, in-range indices before receipt projection.
+  3. Unit and acceptance tests cover PostExec serialization and receipt refund projection.
+- **Possible Mitigations:**
+  1. Add end-to-end coverage for `eth_getBlockByNumber`, `eth_getTransactionByHash`, transaction/receipt
+     cross-links, and representative generic Ethereum decoders.
+  2. Publish the RPC schema and coordinate with explorer and infrastructure providers before activation.
```
geoknee, line 216: We could add here, as discussed, exposing a public SDM-enabled devnet for ecosystem integrators to build against. Base do something similar and call it a "vibe net". Or simply pointing integrators to a tool like kupcake so they can run a devnet like that themselves locally.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Action Items
+
+The following are candidate follow-ups for Security and engineering review. They are **not commitments or launch
+blockers** until an owner accepts them and a tracking issue or PR is created.
+
+| ID | Candidate Follow-up | Failure Modes | Owner | Status |
+| --- | --- | --- | --- | --- |
+| A1 | Run a production-FPVM proof over a block with a non-empty SDM payload. | FM1, FM4 | _TBD_ | Proposed |
+| A2 | Add a post-Isthmus, Lagoon-active, non-empty fixture that compares execution-client and proof-program block hash and state root. | FM1, FM4 | _TBD_ | Proposed |
```
geoknee:
```suggestion
| A2 | Add an action test that compares execution-client and proof-program block hash and state root. | FM1, FM4 | _TBD_ | Proposed |
```
Here's the kicker: this reads like overly-verbose, AI generated material. Let's be more terse.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+| A3 | Document the proof-program and dependency-set requirements for every SDM-active chain. | FM4, FM5 | _TBD_ | Proposed |
```
geoknee: Aren't dependency-set requirements being documented a concern for the interop team / interop FMA?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+
+The audit should cover:
+
+- PostExec decoding and structural validation;
+- SDM canonical-gas calculation and fee settlement, including value conservation and arithmetic boundaries;
+- pre-refund block resource-limit enforcement;
+- transaction rollback and policy snapshot behavior;
+- producer, streamed-subblock, and sealed-block consistency;
+- production refund-policy correctness against the EIP-2929 metered-access rules; and
+- execution-client versus fault-proof-program parity, including the production FPVM.
```
geoknee: Again I'm not sure if worrying about the FPVM here is really necessary. Let's discuss.

security/fma-sdm-postexec.md
einar-oplabs:
```suggestion
and apply a valid SDM payload.
```

security/fma-sdm-postexec.md
geoknee: There's also "draining the fee vaults". Not as bad as infinite ETH mint bug, but worse than simply "lost revenue".

geoknee (commented):
Good start, left a bunch of comments where we can tighten it up.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+- [Sequencer-Defined Metering](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/lagoon/sdm.md)
+- [Lagoon Network Upgrade](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/lagoon/overview.md)
+- [SDM design document](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/sdm.md)
+
+## Failure Modes and Recovery Paths
+
+### FM1: PostExec construction or application diverges, causing invalid blocks or chain splits
+
+- **Description:** The producer's streamed, derived, or sealed block views may contain different PostExec
+  payloads; a sealed payload may be malformed; or consensus consumers may apply the same valid payload
```
einar-oplabs: what exactly is a 'block view'. And in which sense 'may it contain **different** PostExec payloads'.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+     bytes are compared with the sealed transaction, and sealing aborts if PostExec finalization fails.
+  4. Derivation enforces activation while the execution layer remains authoritative for post-activation payload
+     validity and deposits-only recovery.
+  5. Existing unit and acceptance tests cover structural rejection, producer/verifier round trips, and
+     streamed-to-sealed payload equality.
+- **Possible Mitigations:**
+  1. Add a non-empty fixture that compares execution-client and fault-proof-program block hash, receipts root,
+     gas used, and state root.
+  2. Add deposits-only replacement metrics and cross-client malformed-payload recovery coverage.
+  3. Add consumer-side checks and alerts for missing or inconsistent cumulative streamed payloads.
```
einar-oplabs: I think we have logging of most errors, but what about high error rates - like if 90% of blocks fail to validate  - do we get notified somehow?

security/fma-sdm-postexec.md
einar-oplabs: This FM could be **structurally invalid blocks**.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+- **Existing Mitigations:**
+  1. Consensus limits each refund to the transaction's `evmGasUsed` and rejects settlement underflow.
+  2. Settlement tests assert that sender credits equal recipient debits across transaction types, fee settings,
+     and arithmetic boundaries.
+  3. Policy tests cover EIP-2929 metered accesses and known intrinsically warm or protocol-generated accesses that
+     must not receive a rebate.
+  4. Stateful policy data is snapshotted and restored when a candidate transaction fails execution or is not
+     committed to the block.
+  5. Operators can disable refund production without changing verification behavior.
+- **Possible Mitigations:**
```
einar-oplabs, line 92: What if we had policy-combinators, e.g.

    `CeilPolicy(FaultyBlockWarmingPolicy, PercentagePolicy(50))`

This way we could bound the error and risk.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,291 @@
...
+  1. Every component derives SDM activation from the Lagoon schedule.
+  2. Refund production is enabled only when both Lagoon is active and the operator opt-in is set.
+  3. Verification ignores the local production opt-in.
+  4. The operator opt-in defaults off, supports boot-time configuration, and exposes effective state through the
+     admin status method.
+- **Possible Mitigations:**
+  1. Add a cross-component activation preflight covering timestamps, legacy keys, execution-client support,
+     proof-program selection, and dependency-set configuration.
+  2. Add restart health checks that verify the intended effective opt-in state.
+  3. Document and test that mutating admin methods are not exposed on public HTTP or WebSocket endpoints.
```
einar-oplabs: Do we have a mechanism like a 24hr ahead-of-launch NOOP log statement, where nodes state their intent to activate in 24 hrs in a way that we can verify?

einar-oplabs (approved):

security/fma-sdm-postexec.md
einar-oplabs: **Opus 5 (high)**: No — today nothing counts a rejection, so there is no series to alert on.

The SDM dashboard (`optimism-premium/observability/sdm/dashboard.json`) has post-exec panels, but they're built on `reth_sync_execution_post_execution_histogram_count` and `reth_sync_block_validation_post_execution_validation_duration_count` — validations *performed* and how long they took — plus `reth_payloads_failed_jobs` on the build side. There is no `..._rejected_total` / `..._invalid_total`, and no alert rule in `k8s/grafana-cloud/terraform-rules/` mentions sdm, postexec, or refund (only `subblocks.yaml`, 4 alerts).

What would actually catch the 90% case, and only as a second-order symptom:

- **uniform rejection** → chain stops advancing → head-lag and `sequencing-window-expiry-risk` alerts in `replicas.yaml`. Loud, but says nothing about the cause.
- **partial rejection / client split** → `healthcheck_is_currently_diverged` (P1, 5m) and `replica-target-height-low-match`.
- **90% failing while the chain still advances** — e.g. the stream-only mismatches that self-heal per this FM's recovery path — → nothing. Logs only, and nobody pages on log volume.

The precedent to copy is right next door: subblocks shipped with counter-rate alerts (`increase(reth_subblocks_broadcast_success_total[5m]) < 1`, dropped-subblock counters summed over 5m, p5-vs-target cadence). So I'd turn the two "Possible" bullets into a concrete ask: `postexec_validation_rejected_total` labeled by reason (structural / entry / settlement), `deposits_only_replacement_total`, an alert on rejected-over-total exceeding a threshold for 5m, and a page on any nonzero deposits-only replacement — as a pre-mainnet-activation requirement rather than an optional mitigation.

Worth also rewording **Existing Detection**: "execution-engine `INVALID` responses" reads like alerting when it's really just logs.

security/fma-sdm-postexec.md
einar-oplabs: **Opus 5 (high)**: There's no NOOP-log or attestation mechanism, and I don't think we need to invent one — the intent is already queryable per node, ahead of time, in three places:

- **`admin_sdmStatus(query_timestamp)`** — on develop in both op-reth (`rust/op-reth/crates/rpc/src/sdm_admin.rs`) and op-rbuilder. It returns `operatorSdmOptIn`, `protocolActive` evaluated at *any* timestamp you pass, `effective`, and `activationTime`. Passing `activationTime + 1` is literally "do you intend to be SDM-active then?", 24h out or 24 days.
- **`eth_config`** (EIP-7910), op-geth `internal/ethapi/api.go:1422` — `current`/`next`/`last`, each with `activationTime` and `forkId`. OP fork timestamps are `*Time` fields on `ChainConfig` and `gatherForks` collects them reflectively, so the SDM gate is inside that `forkId`.
- **`optimism_rollupConfig`** (`op-node/node/api.go:175`) for the CL-side schedule, with the superchain registry as the expected value.

So the two real gaps, which is what I'd add to the doc:

1. **Nothing scrapes or compares any of it.** Verification is manual. Possible Mitigation 1 should say the preflight is a *scheduled* check that diffs each node's `next.activationTime`/`forkId` and `admin_sdmStatus(T+1)` against the registry starting N hours out, and pages on disagreement. That's the checkable form of the 24h intent statement.
2. **`admin_sdmStatus` can't be the fleet-wide signal.** It's admin-namespace, and mitigation 3 on this very line says keep admin off public endpoints — so it's unavailable for third-party nodes and isn't Prometheus-scraped. Cheapest fix: expose the same three values as gauges (`sdm_protocol_active`, `sdm_operator_opt_in`, `sdm_activation_time`). Metrics already flow fleet-wide through existing scrape config, which makes a cross-node disagreement alert a one-liner *and* makes it work before activation.

One asymmetry worth noting: op-node has no equivalent status method on develop (the `SdmStatus` doc comment claims the shape matches op-node, but nothing there implements it), so CL-side intent is only inferable from `optimism_rollupConfig`. At the protocol layer, EIP-2124 forkid does carry the next scheduled fork in the EL handshake, so ELs can in principle detect future-fork disagreement pairwise at peering time — but that covers neither the CL nor the operator opt-in.

security/fma-sdm-postexec.md
nonsense: Yes, streams here is referencing subblocks.

security/fma-sdm-postexec.md
nonsense: If 90% of blocks fail to validate for whatever reason, this is out of scope for SDM, we should have alerts on ERROR rates and second-order effects for that.

security/fma-sdm-postexec.md
nonsense: Yes, you're right. This was an issue when starting a network with SDM, as SDM is ran on Lagoon, but really this is out of scope for SDM.

security/fma-sdm-postexec.md
nonsense: > exposing a public SDM-enabled devnet for ecosystem integrators to build against

I'd rather we don't do any of these, until at least one ecosystem integrator has requested anything like this.

Ecosystem integrators just need to be aware of the fact that 0x7d will exist, and show its contents, nothing more is required of them - there is no need for any special visualization or parsing, unless we want first-class support for showing tuples (tx-id, refund).

We don't do anything special for the Deposit tx either -- here is an example: https://optimistic.etherscan.io/tx/0x9d2928961ac70c8c195200cf0d74a985b376c1b2300312bbb5c0c0442f94553d

security/fma-sdm-postexec.md
nonsense: Agreed. “Non-empty” means a block containing at least one SDM refund.

security/fma-sdm-postexec.md
nonsense: Not currently, the subblocks behavior is described only in the SDM design doc. I am not sure there is another way it could work though, given the requirement that every subblock should be a valid block.

security/fma-sdm-postexec.md
nonsense: Agreed.

security/fma-sdm-postexec.md
nonsense: Agreed.

security/fma-sdm-postexec.md
nonsense: Yes, I'll make that explicit.

security/fma-sdm-postexec.md
nonsense: Agreed. I’ll add the Superchain Registry as the expected activation schedule. The generic hardfork FMA is already referenced.

security/fma-sdm-postexec.md
nonsense: Agreed that execution-client versus proof-program parity is the important SDM-specific property.

security/fma-sdm-postexec.md
nonsense: The ambiguity has been removed.

security/fma-sdm-postexec.md
nonsense: Agreed. This is now FM1a: “Structurally invalid or inconsistent PostExec blocks.”

security/fma-sdm-postexec.md
nonsense: This would add configuration and audit complexity, so we'd rather not do it for now.

security/fma-sdm-postexec.md
nonsense: Added as a proposed item, so we can discuss and make a decision.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
...
+  multiple `0x7D` transactions, an incorrect position or block anchor, and invalid transaction indices. A
+  subblock-stream-only mismatch breaks pending views and preconfirmations. Uniform rejection of a malformed sealed
+  block causes a deposits-only replacement and user-transaction reorg.
+- **Risk Assessment:** High impact, low likelihood. Producer lifecycle bugs can invalidate preconfirmations or omit
+  user transactions from the derived chain. The specification, execution-layer validation, and lifecycle tests
+  reduce the likelihood.
+- **Existing Mitigations:**
+  1. The specifications require at most one final `0x7D`, the correct block-number anchor, an active schema, and
+     non-empty, non-zero entries ordered by strictly increasing transaction index.
+  2. Incremental subblock streams use replace-not-append semantics for the cumulative PostExec payload; the final
```
pcw109550: Can you please elaborate more on this mitigation?

> use replace-not-append semantics

What does replace-not-append mean? 

> final subblock-stream payload is compared with the sealed transaction

Does "sealed transaction" mean the postExec tx? If yes, shouldn't it be compared with every subblock stream(0th subblock to Nth subblock)

> sealing aborts if PostExec finalization fails.

so.. what happens if abort happens? We get a canonical block which does not have postExec Tx? We cannot build a canonical block at all? We only include deposit txs in the block?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
...
+- [Post-Execution Transactions](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/lagoon/post-exec.md)
+- [Sequencer-Defined Metering](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/lagoon/sdm.md)
+- [Lagoon Network Upgrade](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/lagoon/overview.md)
+- [SDM design document](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/sdm.md)
+
+## Failure Modes and Recovery Paths
+
+### FM1a: Structurally invalid or inconsistent PostExec blocks
+
+- **Description:** A producer may emit a stale cumulative PostExec payload in a subblock stream, produce a final
```
pcw109550: Should generalize this description to also capture blocks which has subblocks disabled. SDM can be enabled even if we disable subblocks.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
...
+     non-empty, non-zero entries ordered by strictly increasing transaction index.
+  2. Incremental subblock streams use replace-not-append semantics for the cumulative PostExec payload; the final
+     subblock-stream payload is compared with the sealed transaction, and sealing aborts if PostExec finalization fails.
+  3. Before Lagoon activation, blocks containing PostExec transactions are invalid. After activation, the
+     execution layer validates PostExec payloads, while derivation handles deposits-only recovery when a payload
+     is invalid.
+  4. Existing unit and acceptance tests cover structural rejection and subblock-stream-to-sealed payload equality.
+- **Possible Mitigations:**
+  1. Add deposits-only replacement metrics and cross-client malformed-payload recovery coverage.
+- **Existing Detection:** Execution-engine `INVALID` responses and differential-test failures.
```
pcw109550, line 65: Does exec engine throw `INVALID` when it tries to sanity check the structure enforced by https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/lagoon/post-exec.md#block-level-structural-rules ?

If yes, can we also emit metrics at the exec engine to detect the reason of `INVALID` happened because of postExec structure check?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
...
+- **Possible Detection:** Alerts for deposits-only replacements.
+- **Recovery Path(s):** A subblock-stream-only mismatch self-heals when the canonical block arrives but invalidates
+  the affected preconfirmation. Uniform rejection of a sealed payload results in a deposits-only block and
+  requires users to resubmit omitted transactions.
+
+### FM1b: PostExec application diverges across clients
```
pcw109550, line 69: My understanding is that this application happens at execution client, and we only maintain op-reth. So my ask is, which clients are we comparing? op-reth vs kona-executor ?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
...
+  from an erroneous one because refund policy is not part of consensus. Historically, this class included
+  rebates for intrinsically warm or otherwise unmetered accesses: the transaction received a cold-to-warm rebate
+  even though the EVM never charged the cold surcharge.
+- **Risk Assessment:** Medium impact, low likelihood. The economic misallocation is bounded per transaction by its
+  `evmGasUsed` and does not cause a chain split.
+- **Existing Mitigations:**
+  1. Consensus limits each refund to the transaction's `evmGasUsed`.
+  2. Policy tests cover EIP-2929 metered accesses and known intrinsically warm or protocol-generated accesses that
+     must not receive a rebate.
+  3. Stateful policy data is snapshotted and restored when a candidate transaction fails execution or is not
```
pcw109550: Is this part of our spec? Also

> candidate transaction fails execution

this happen? This can happen because of incorrect policy refunds?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
...
+- **Existing Mitigations:**
+  1. Settlement requires sender credits to equal the sum of recipient debits and rejects debit underflow.
+  2. Settlement tests cover transaction types, fee settings, and arithmetic boundaries.
+  3. Operators can disable refund production without changing verification behavior.
+- **Possible Mitigations:**
+  1. Independently audit settlement and add invariant or property tests for value conservation.
+  2. Reconcile sender credits and fee-recipient debits from sampled canonical blocks.
+- **Existing Detection:** Conservation and differential tests detect known settlement discrepancies before
+  deployment.
+- **Possible Detection:** Alert on reconciliation failures or anomalous fee-vault and beneficiary balance changes.
```
pcw109550, line 111: How will we detect this "reconciliation failures"?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
...
+- **Recovery Path(s):** Disable refund production, correct the policy, and redeploy. Already finalized refunds are
+  consensus-valid and should not be rolled back; material losses require accounting and, if appropriate,
+  treasury reconciliation.
+
+### FM2b: Settlement violates value conservation
+
+- **Description:** A settlement implementation bug may credit the sender without applying equal debits, debit the
+  wrong fee recipient, or otherwise violate value conservation. If deployed consistently across clients, this
+  could mint ETH or drain fee-vault or beneficiary balances without causing an immediate chain split.
+- **Risk Assessment:** Very high impact, very low likelihood. A value-conservation failure can corrupt canonical
```
pcw109550: We may maintain a external monitor like monitorism / hexagate to detect that the invariant holds.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
...
+- **Existing Detection:** Conservation and differential tests detect known settlement discrepancies before
+  deployment.
+- **Possible Detection:** Alert on reconciliation failures or anomalous fee-vault and beneficiary balance changes.
+- **Recovery Path(s):** Disable refund production immediately and deploy corrected clients. Canonical balance
+  corruption may require an emergency upgrade and the established incident-response process.
+
+### FM3: Refunds bypass block resource limits
+
+- **Description:** Canonical gas decreases after refunds. If block admission used only post-refund gas, a producer
+  could include substantially more EVM work than the block gas limit intends, creating a resource-exhaustion
```
pcw109550: Is this only possible when the postExec tx was unintentionally inserted between user txs?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
...
+- **Possible Detection:** A production-FPVM regression test that detects native-versus-FPVM trace divergence.
+- **Recovery Path(s):** Deploy a corrected proof program and publish a new absolute prestate through the approved
+  governance and registry process. Use the established fault-proof recovery process while the corrected program
+  is prepared.
+
+### FM5: Activation or operator configuration is inconsistent
+
+- **Description:** SDM uses the Lagoon activation timestamp across the consensus client, execution client, and
+  fault-proof stack. A timestamp mismatch can make one component produce or accept `0x7D` while another considers
+  it inactive. Legacy configuration keys may be silently ignored. An execution client without PostExec support
```
pcw109550: What is "legacy configuration key" ? Searched in the specs but could not find.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
...
+  2. Add a post-Isthmus, Lagoon-active fixture that compares execution-client and proof-program block hash and
+     state root.
+- **Existing Detection:** Proof-runner failures, challenger disagreement, and output-root disagreement in native
+  acceptance tests.
+- **Possible Detection:** A production-FPVM regression test that detects native-versus-FPVM trace divergence.
+- **Recovery Path(s):** Deploy a corrected proof program and publish a new absolute prestate through the approved
+  governance and registry process. Use the established fault-proof recovery process while the corrected program
+  is prepared.
+
+### FM5: Activation or operator configuration is inconsistent
```
pcw109550: I would decouple this FM5 to activation inconsistency(which applies for general consensus touching Hardfork), and SDM operation config drift.

The chain will still progress if multiple sequencers use different SDM policy assuming the HF is all correctly activated to all components

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
...
+  1. The Superchain Registry defines the expected Lagoon activation schedule, and every component derives SDM
+     activation from that schedule.
+  2. Refund production is enabled only when both Lagoon is active and the operator opt-in is set.
+  3. Verification ignores the local production opt-in.
+  4. The operator opt-in defaults off, supports boot-time configuration, and exposes effective state through the
+     admin status method.
+- **Possible Mitigations:**
+  1. Add a cross-component activation preflight covering timestamps, legacy keys, execution-client support, and
+     proof-program selection.
+  2. Add restart health checks that verify the intended effective opt-in state.
```
pcw109550: Lets say there were three sequencers, and all lagoon hf activated. For some reason, single sequencer was opt-out, not producing postExec blocks, and other two were opt-in. We should detect this drift and consolidate the behavior.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,319 @@
+# Sequencer-Defined Metering (SDM) and Post-Execution Transactions: Failure Modes and Recovery Path Analysis
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Introduction](#introduction)
+- [Failure Modes and Recovery Paths](#failure-modes-and-recovery-paths)
```
pcw109550, line 7: Should we include batcher-related failure modes? Batcher needs to batch postExec txs, so this should be also covered?

security/fma-sdm-postexec.md
nonsense: Replace-not-append means every subblock carries the latest PostExec candidate separately from its transaction delta. Earlier candidates do not contain all refunds, since refunds accumulate with every new subblock.

security/fma-sdm-postexec.md
nonsense: Agree. I'll update the doc to say that malformed sealed PostExec blocks are a general risk.

security/fma-sdm-postexec.md
nonsense: Yes, the relevant comparison today is op-reth’s native execution path versus kona-executor’s fault-proof path. They share core execution code, but their parsing, configuration, and integration paths can still diverge. I’ll update the wording to make this concrete.

security/fma-sdm-postexec.md
nonsense: Fixed.

security/fma-sdm-postexec.md
nonsense: Agreed. I’ll split this into activation inconsistency, which can halt or split the chain, and sequencer production configuration drift, which remains consensus-valid but can cause inconsistent refund behavior. Different valid SDM policies do not prevent chain progress and are better covered by FM2a.

security/fma-sdm-postexec.md
nonsense: Agreed. We should compare the effective SDM opt-in status across all active sequencers and alert on differences, especially after restarts. Monitoring PostExec production can supplement this, although absence of `0x7D` alone is not conclusive because a block may have no eligible refunds. I’ll add this detection and consolidation step to FM5b.

security/fma-sdm-postexec.md
nonsense: Agreed. Although the L1 batcher transaction format is unchanged, singular- and span-batch encoding must preserve the PostExec transaction and its ordering. Omitting it can produce a valid but different derived block without refunds, while corruption or reordering can make the derived payload invalid. I’ll incorporate this into FM1a and mention the existing singular/span derivation coverage.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,340 @@
...
+
+### FM1a: Structurally invalid or inconsistent PostExec blocks
+
+- **Description:** A producer may seal a malformed PostExec block whether it uses standard block building or
+  incremental subblocks. Examples include multiple `0x7D` transactions, an incorrect position or block anchor, and
+  invalid transaction indices. A batcher may omit, corrupt, or reorder `0x7D`; omission can produce a valid derived
+  block without refunds that differs from the unsafe block, while corruption or reordering can make the derived
+  payload invalid. With subblock streaming enabled, a producer may additionally emit a stale cumulative PostExec
+  payload or a final subblock-stream payload that differs from the sealed transaction. A subblock-stream-only mismatch
+  breaks pending views and preconfirmations. Uniform rejection of a malformed sealed block causes a deposits-only
```
op-will: > With subblock streaming enabled, a producer may additionally emit a stale cumulative PostExec payload or a final subblock-stream payload that differs from the sealed transaction. A subblock-stream-only mismatch breaks pending views and preconfirmations.

Is there a reason why the `0x7D` must be included in subblocks? It introduces this necessity for replace-not-append, and adds some failure modes, but I'm unsure what it actually does for the consumer of a subblocks stream. If my understanding is correct, the contract for subblocks from a consumer's standpoint is that any observed transactions are guaranteed to be in that order in the sealed block, but the subblocks tx stream is not guaranteed to be complete (i.e. there may be more txs in the finalized block _after_ the txs that the consumer has read from the websocket). 

Is there a world in which it could be omitted from subblocks entirely and just appended to the tx list when the block is built and returned to the CL?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,340 @@
...
+
+### FM1b: PostExec application diverges across clients
+
+- **Description:** Native op-reth execution and kona-executor's fault-proof execution path may apply the same valid
+  PostExec payload differently. Examples include applying a refund twice or calculating canonical gas, receipts, or
+  settlement differently. A validity, block-hash, or state-root disagreement between the two paths causes a chain
+  split or an invalid fault-proof result.
+- **Risk Assessment:** High impact, low likelihood. PostExec affects consensus-visible gas accounting, receipts,
+  balances, and state roots. The two paths share core execution code, reducing the likelihood, but their parsing,
+  configuration, and integration paths can still diverge.
```
op-will: Re: low likelihood, applying a refund twice was already a bug, right? Just confirming the mitigations that have been added since then to move this to low likelihood.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,340 @@
...
+- **Existing Mitigations:**
+  1. Entries may target only standard Ethereum transactions. Execution enforces `refund <= evmGasUsed`, rejects
+     settlement underflow, and consumes every entry exactly once.
+  2. Existing tests cover producer/verifier round trips, settlement, and structural validity.
+- **Possible Mitigations:**
+  1. Add op-reth and kona-executor parity coverage for a block containing an SDM refund.
+  2. Add differential coverage across native and fault-proof execution for valid and malformed PostExec payloads.
```
op-will, line 78: These are desirable additions, as the "Existing Mitigations" don't seem like they are actually mitigations in that they help define validity, but the FM is that a valid PostExec payload is _applied_ differently.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,340 @@
...
+- [Lagoon Network Upgrade](https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/lagoon/overview.md)
+- [SDM design document](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/sdm.md)
+
+## Failure Modes and Recovery Paths
+
+### FM1a: Structurally invalid or inconsistent PostExec blocks
+
+- **Description:** A producer may seal a malformed PostExec block whether it uses standard block building or
+  incremental subblocks. Examples include multiple `0x7D` transactions, an incorrect position or block anchor, and
+  invalid transaction indices. A batcher may omit, corrupt, or reorder `0x7D`; omission can produce a valid derived
```
op-will: > A batcher may omit, corrupt, or reorder `0x7D`

Does it make sense for a batcher to decode (ideally using the exact same lib as derivation) the batch that it just encoded and verify that it matches as a verification step?

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,340 @@
...
+- **Risk Assessment:** Medium impact, low likelihood. The economic misallocation is bounded per transaction by its
+  `evmGasUsed` and does not cause a chain split.
+- **Existing Mitigations:**
+  1. Consensus limits each refund to the transaction's `evmGasUsed`.
+  2. Policy tests cover EIP-2929 metered accesses and known intrinsically warm or protocol-generated accesses that
+     must not receive a rebate.
+  3. Stateful policy data is snapshotted and restored when a candidate transaction fails execution or is not
+     committed to the block.
+  4. Operators can disable refund production without changing verification behavior.
+- **Possible Mitigations:**
```
op-will, line 92: Could we add auditing to live code as a measure of verification? For the block warming example, something like:
- Keep a map of slot -> access count
- Calculate the total expected rebate for the block as `(access_count-1) × 2000` for each slot
- Verify that the cumulative 0x7D refund matches the separately-calculated value

Note: may need to be more granular than just "access" to differentiate between SLOAD, SSTORE, etc. and change `2000` to be the price of the op in question. This may become more difficult, but the general question of auditing via a separate but valid calc holds.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,340 @@
...
+### FM2b: Settlement violates value conservation
+
+- **Description:** A settlement implementation bug may credit the sender without applying equal debits, debit the
+  wrong fee recipient, or otherwise violate value conservation. If deployed consistently across clients, this
+  could mint ETH or drain fee-vault or beneficiary balances without causing an immediate chain split.
+- **Risk Assessment:** Very high impact, very low likelihood. A value-conservation failure can corrupt canonical
+  balances, but the settlement equations, underflow checks, and conservation tests substantially reduce the
+  likelihood.
+- **Existing Mitigations:**
+  1. Settlement requires sender credits to equal the sum of recipient debits and rejects debit underflow.
```
op-will, line 104: How is this enforced? Getting into implementation-specific details, is this something that the typestate pattern in Rust could mitigate and be easily unit tested?

security/fma-sdm-postexec.md
nonsense: I agree we need such alerts. We don't have alerts right now. See https://github.com/ethereum-optimism/protocol-team/issues/237

security/fma-sdm-postexec.md
nonsense: I am not sure I follow. PostExec tx intentionally (or unintentionally) inserted between user tx would result into an invalid block.

security/fma-sdm-postexec.md
nonsense: > Is there a reason why the 0x7D must be included in subblocks?

Every `subblock` must produce a valid block, so yes.

security/fma-sdm-postexec.md
nonsense: Within a specific policy - yes, not within SDM.

security/fma-sdm-postexec.md
nonsense: Agree, we should add tests for these, they are in progress.

security/fma-sdm-postexec.md
nonsense: I have thought about the same, but I think that'd explode the complexity in the batcher.

security/fma-sdm-postexec.md
nonsense: We have that in the `replay` crate. It really doesn't belong on the verifier side though, as we are building SDM to support any kind of policy, not just deterministic ones.

security/fma-sdm-postexec.md
nonsense: Search for `test_post_exec_settlement_*` - on a unit test level.

security/fma-sdm-postexec.md
nonsense: @geoknee on a second thought - I think this is really an implementation detail, and not something we should have specs for, but I am open to discuss it further.

Invariants are:
- every subblock must produce a valid block
- PostExec tx is accumulating refunds

Then clearly we have a `replace-not-append` semantics.

security/fma-sdm-postexec.md
nonsense: No, we can't fail a tx because of incorrect refund.

security/fma-sdm-postexec.md
op-will: Note: meant to this on 2b. Leaving here since there is discussion already.

op-will: > we are building SDM to support any kind of policy, not just deterministic ones

Note: I think the same logic holds—we just wouldn't have deterministic verification like this for policies that are non-deterministic.

security/fma-sdm-postexec.md
op-will: > If my understanding is correct, the contract for subblocks from a consumer's standpoint is that any observed transactions are guaranteed to be in that order in the sealed block, but the subblocks tx stream is not guaranteed to be complete

Did you see this point? There are no actual guarantees on subblock transmission or reception, so I think the only contract we have is that what we send you, and the order in which we send it, must hold.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,264 @@
...
+  2. Alert on unusually high raw EVM gas relative to canonical gas.
+- **Existing Detection:** Block execution time and replica lag are observable through standard node monitoring.
+- **Possible Detection:** Expose raw EVM gas alongside canonical gas and alert on an unusually high ratio.
+- **Recovery Path(s):** Disable SDM production or reduce sequencing load while patching the affected builder or verifier. If deployed clients disagree on the resource-limit rule, recovery follows the chain-split path in FM1b.
+
+### FM4: Fault-proof programs cannot prove an SDM block
+
+- **Description:** An execution client may validate an SDM block while the deployed fault-proof program fails to reproduce it. Causes include different transaction decoding, settlement behavior, configuration, or behavior that works in native execution but fails inside the production FPVM. The deployed proof program and prestate must support Lagoon and SDM.
+- **Risk Assessment:** High impact, medium likelihood until production-FPVM coverage exists. An unprovable output stalls permissionless withdrawals and may require replacing the fault-proof program under governance.
```
op-will: > medium likelihood until production-FPVM coverage exists

This reads like a measured estimate, but no native/FPVM divergence has been found or ruled out, right? PM1/PM2 wouldn't refine the likelihood so much as resolve it completely.

Suggestion: reclassify Action Item 2 as a pre-activation gate, like Action Item 3 for FM2c. The argument is asymmetry, not probability — PM1 is cheap, recovery is a governance-gated prestate replacement while withdrawals are stalled. Would need to be clear about which Action Item 2 route (client/program parity, FPVM execution, or an end-to-end dispute game).

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,264 @@
...
+- **Risk Assessment:** Medium impact, low likelihood with configuration monitoring. Drift does not invalidate blocks or halt the chain, but it can create inconsistent user charges, revenue, and operator-policy compliance.
+- **Existing Mitigations:**
+  1. Refund production is enabled only when both Lagoon is active and the operator opt-in is set.
+  2. Verification ignores the local production opt-in and does not recompute the sequencer's refund policy.
+  3. The operator opt-in defaults off, supports boot-time configuration, and exposes effective state through the admin status method.
+- **Possible Mitigations:**
+  1. Add restart and cross-sequencer health checks that verify the intended effective opt-in.
+  2. Document and test that mutating admin methods are not exposed on public HTTP or WebSocket endpoints.
```
op-will, line 201: The mutating admin method is not persistent either, right? It would be good to:
1. Persist admin mutation state
2. Remove persisted admin mutation state on boot if it matches boot config
3. Alert when different than boot config state

This combination makes it so the mitigation doesn't resolve alerts until it is addressed completely (i.e. update the config and restart). This could be a good larger-scale operational feature than SDM, though. Might suggest mitigating via clearly-defined runbook and defer this boot config vs run config discrepancy detection, alerting, etc.

security/fma-sdm-postexec.md
op-will: > Ecosystem integrators just need to be aware of the fact that 0x7d will exist, and show its contents

Agreed. Can we solidify and publish the spec? Broadcasting that updated consumers are backward compatible and that they should update early seems like it can only help.

security/fma-sdm-postexec.md
```diff
@@ -0,0 +1,264 @@
...
+| A7 | Decide and document the PostExec receipt's auxiliary L1 fee fields. | FM6 | _TBD_ | Proposed |
+| A8 | Add an operator runbook for opt-in health checks, restarts, and admin-RPC exposure. | FM5b | _TBD_ | Proposed |
+
+- [ ] Resolve all review comments and incorporate accepted decisions into this document (Assignee: document author).
+
+## Audit Requirements
+
+**Recommendation: complete an external audit before production activation.**
+
+The audit should cover:
```
op-will, line 250: Might be implied, but code, test, structural, and procedural resilience to change and error detection is very useful. I'd argue that it's just as if not more important than point-in-time correctness.

security/fma-sdm-postexec.md
op-will: To be clear here (since typestate is not necessarily correct for this case), what I mean is that we can write our types in such a way in that they cannot be instantiated nor mutated in a way that violates this invariant*. This is a stronger guarantee than testing use logic because it necessarily applies to all current and future uses rather than just testing a specific set of logic.

*Typically this is done via a builder as the only possible way to construct, private fields, and locked down mutation functions that assert invariants.

security/fma-sdm-postexec.md
nonsense: I don't think we have any code that persists configs / admin mutations. I am reluctant to introduce this as part of SDM honestly.

security/fma-sdm-postexec.md
nonsense: If you're asking how this is implemented see `execute_transaction_without_commit`.

We first compute deltas at `post_exec_settlement_deltas` for a given tx.

We then apply the delta for the tx at `apply_post_exec_refund_to_state`.

security/fma-sdm-postexec.md
nonsense: > no native/FPVM divergence has been found or ruled out, right?

No, and I wouldn't expect to find one. If we have tests for SDM with native execution in `kona` (and we do), that's good enough, and we don't need any more tests.

We shouldn't have to run the compiled Kona program inside the production Cannon FPVM.

security/fma-sdm-postexec.md
nonsense: Sure, will add metrics for the INVALID reasons.

